PR #20, opened 2019-08-05
Fix the status bar discoloration for claim subscreens

hedvigoscar:
Also refactor it to activity because I want it like that


hedvigoscar:
Self-reviewed

PR #25, opened 2019-08-07
New offer images

hedvigoscar:
Also we get to remove all the crappy pngs
Also its all blobby
Thanks, I love it


hedvigoscar:
Self-reviewed

PR #27, opened 2019-08-07
Fix missing text after selecting cashback

hedvigoscar:
Also convert the view to an activity.
Way better.

Also Fragmentize this query. Much nicer.


hedvigoscar:
Self-reviewed

PR #29, opened 2019-08-08
Fix members seeing Logged In screen instead of Terminated screen when…

hedvigoscar:
… they are terminated


hedvigoscar:
Self-reviewed

PR #30, opened 2019-08-08
Improve RecyclerViews with Glide-loaded images

hedvigoscar:
But unfortunately this does not solve the scroll-yank related to images entirely


hedvigoscar:
Self-reviewed

PR #42, opened 2019-08-21
DIrect debit deep link support

hedvigoscar:


app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -42,54 +31,85 @@ class SplashActivity : BaseActivity() {
...
+                    .subscribe({ loginStatus ->
+                        startDefaultActivity(loginStatus)
+                    }, { Timber.e(it) })
             }
-        }.addOnFailureListener {
-            startActivity(Intent(this, MarketingActivity::class.java))
         }
     }
 
     private fun navigateToActivity(loginStatus: LoginStatus) = when (loginStatus) {
```
FredrikAreschoug: Is this function used?

PR #48, opened 2019-09-06
Handle playback failed

FredrikAreschoug:
From experience there is not much we can do when the MediaPlayer fails :/
https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/4ba4a14827f505f3fbb1ab8bac54e30b?time=last-seven-days&sessionId=5D70F8A000D000010956DED0132BA0F3_DNE_0_v2


hedvigoscar:
This must be the fault of the hardware vendor, right?
I agree with this solution though - we are left without any real options here

PR #51, opened 2019-09-18
Dark Theme

hedvigoscar:
Currently missing:
- Settings icon
- 'Typing'-animation. This is going to have to be an AnimatedVectorDrawable probably


app/src/main/res/layout/referral_header.xml
```diff
@@ -11,7 +11,8 @@
     <com.hedvig.app.feature.referrals.ProgressTankView
         android:id="@+id/discountView"
         android:layout_width="match_parent"
-        android:layout_height="200dp" />
+        android:layout_height="200dp"
```
FredrikAreschoug: when we are in the process of cleaning up move this do dimen

app/src/main/java/com/hedvig/app/feature/welcome/WelcomeDialog.kt
```diff
@@ -16,7 +16,7 @@ class WelcomeDialog : DismissablePager() {
 
     override val tracker: WelcomeTracker by inject()
     override val items: List<DismissablePagerPage> by lazy {
-        arguments!!.getParcelableArrayList<DismissablePagerPage>(ITEMS)
+        arguments!!.getParcelableArrayList<DismissablePagerPage>(ITEMS)!!
```
FredrikAreschoug: Could be nice to do a ?: throw("custom error instead") Kotlin null pointer exception can be hard to decipher :)

PR #63, opened 2019-10-16
Language as a setting

hedvigoscar:
It's not the best thing I've ever seen, but it actually works pretty well.

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/settings/Language.kt
+++ b/app/src/main/java/com/hedvig/app/feature/settings/Language.kt
@@ -1,0 +1,124 @@
+package com.hedvig.app.feature.settings
+
+import android.annotation.TargetApi
+import android.content.Context
+import android.content.res.Configuration
+import android.os.Build
+import android.os.LocaleList
+import androidx.preference.PreferenceManager
+import java.util.Locale
+
+enum class Language {
+    SYSTEM_DEFAULT,
+    SV_SE,
+    EN_SE;
+
+    fun apply(context: Context?): Context? {
+        val locale = into()
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            applyV24(context, locale)
+        } else {
+            applyV21(context, locale)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+            is LocaleWrapper.MultipleLocales -> {
+                val locales = locale.locales
+                LocaleList.setDefault(locales)
+                if (context == null) {
+                    return null
+                }
+                val config = Configuration(context.resources.configuration)
+                config.setLocales(locales)
+                return context.createConfigurationContext(config)
```
FredrikAreschoug: this is basically the same code as line 47-52 maybe you could put in a function, I't also pretty similar to 30-38 so have a look at that as well

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+    }
+
+    private fun into(): LocaleWrapper = when (this) {
+        SV_SE -> LocaleWrapper.SingleLocale(Locale.forLanguageTag("sv-SE"))
+        EN_SE -> LocaleWrapper.SingleLocale(Locale.forLanguageTag("en-SE"))
+        SYSTEM_DEFAULT -> {
+            DefaultLocale.get()
+        }
```
FredrikAreschoug: No need for the `{}`

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    override fun toString() = when (this) {
+        SYSTEM_DEFAULT -> "system_default"
```
FredrikAreschoug: "system_default" "sv-SE" "en-SE" to constants

PR #66, opened 2019-10-23
Ratings Requester

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/ratings/RatingsDialog.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.DialogFragment
+import com.hedvig.app.R
+import com.hedvig.app.util.extensions.view.setHapticClickListener
+import kotlinx.android.synthetic.main.dialog_ratings.*
+import org.koin.android.ext.android.inject
+
+class RatingsDialog : DialogFragment() {
```
FredrikAreschoug: Maybe we should store that this dialog har been dismissed, so If we would want to push this in the future we don't have to show it to user who already seen it

PR #69, opened 2019-10-28
Fix/pick single locale

FredrikAreschoug:


app/src/main/java/com/hedvig/app/HedvigApplication.kt
```diff
@@ -2,6 +2,7 @@ package com.hedvig.app
 
 import android.app.Application
 import android.content.Context
+import android.content.res.Configuration
```
hedvigoscar:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -16,14 +16,14 @@ enum class Language {
...
+            apply(context, locale)
         } else {
-            applyV21(context, locale)
+            applySingleLocale(context, locale)
         }
     }
 
     @SuppressWarnings("deprecation")
-    private fun applyV21(context: Context?, locale: LocaleWrapper): Context? {
+    private fun applySingleLocale(context: Context?, locale: LocaleWrapper): Context? {
```
hedvigoscar: Are we obligated to always use the deprecated API, regardless of API level?

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -40,17 +40,10 @@ enum class Language {
...
             is LocaleWrapper.SingleLocale -> {
-                val unwrappedLocale = locale.locale
-                Locale.setDefault(unwrappedLocale)
-                if (context == null) {
-                    return null
-                }
-                val config = Configuration(context.resources.configuration)
-                config.setLocale(unwrappedLocale)
-                return context.createConfigurationContext(config)
+                return applySingleLocale(context, locale)
```
hedvigoscar: Does this block also need an `updateResources`-call?

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
FredrikAreschoug: yes, my guess is that if the phone uses this single locale implementation we need to do it the deprecated way. But don't ask me why, in my opinion it is strange that phones with higher api:s use this implementation

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
FredrikAreschoug: yes it must call `context.resources.updateConfiguration(config, context.resources.displayMetrics)`

PR #77, opened 2020-01-17
[WIP]Feature/confetti

MathiasOhrstrom:


app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,50 @@
...
+        app:layout_constraintTop_toTopOf="parent"
+        android:layout_marginBottom="200dp"/>
+
+    <TextView
+        android:id="@+id/fromText"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:fontFamily="@font/circular_bold"
+        android:text="From"
+        android:textColor="#0D0066"
```
FredrikAreschoug: Always put colors in `colors.xml`. But in this case you should use style found in `text_styles.xml`. Then you could probably remove `fontFamily`

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,50 @@
...
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        android:layout_marginBottom="200dp"/>
+
+    <TextView
+        android:id="@+id/fromText"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:fontFamily="@font/circular_bold"
+        android:text="From"
```
FredrikAreschoug: Never set text directly in the layout file. Then int won't translate. Whit that said we built som stuff so we sync translation on app build let's talk about that on Monday. Please remind me :)

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,50 @@
...
+        android:layout_height="wrap_content"
+        android:fontFamily="@font/circular_bold"
+        android:text="Happy Birthday"
+        android:textColor="#0D0066"
+        android:textSize="36sp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        android:layout_marginBottom="200dp"/>
```
FredrikAreschoug: Dimensions goes in `dimens.xml`

app/src/main/java/com/hedvig/app/feature/profile/data/ProfileRepository.kt
```diff
@@ -24,6 +25,7 @@ class ProfileRepository(private val apolloClientWrapper: ApolloClientWrapper) {
         return Rx2Apollo
             .from(apolloClientWrapper.apolloClient.query(profileQuery).watcher())
             .map { it.data() }
+
```
FredrikAreschoug:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/profile/data/ProfileRepository.kt
```diff
@@ -13,6 +13,7 @@ import com.hedvig.android.owldroid.graphql.UpdateEmailMutation
 import com.hedvig.android.owldroid.graphql.UpdatePhoneNumberMutation
 import com.hedvig.app.ApolloClientWrapper
 import io.reactivex.Observable
+import java.lang.RuntimeException
```
FredrikAreschoug: Not used

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -61,6 +59,7 @@ class LoggedInActivity : BaseActivity() {
         setContentView(R.layout.activity_logged_in)
         toolbar.updatePadding(end = resources.getDimensionPixelSize(R.dimen.base_margin_double))
 
+
```
FredrikAreschoug:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -43,6 +39,8 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import org.koin.android.ext.android.inject
 import org.koin.android.viewmodel.ext.android.viewModel
+import timber.log.Timber
+
```
FredrikAreschoug: Remove if not used

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+    private lateinit var birthDate: String
+
+    val isBirthdayData: MutableLiveData<Boolean> = MutableLiveData()
+    private val disposables = CompositeDisposable()
+
+
+
+    private fun isBirthday(): Boolean {
+
+        val birthMonth = birthDate.substring(4, birthDate.length - 6).toInt()
```
FredrikAreschoug: Since `birthDate.length` always should be 12 this could be set to 6?

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+        val birthDay = birthDate.substring(6, birthDate.length - 4).toInt()
+
+
+        val currentMonth = Calendar.getInstance().get(Calendar.MONTH) + 1
```
FredrikAreschoug: you should be able to create one Calendar for this. Try to be restrictive with creating resources. Phones are way better now then apilevel 1 but we should still try to do as lean app as possible

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+        val currentDay = Calendar.getInstance().get(Calendar.DAY_OF_MONTH)
+
+        Timber.d("current date: $currentMonth/$currentDay Birth date $birthMonth/$birthDay")
```
FredrikAreschoug: we usually remove debugging logging before merging to master :)

app/src/main/java/com/hedvig/app/feature/birthday/BirthdayActivity.kt
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            override fun onAnimationRepeat(animation: Animator) {
+                Timber.d("repeat")
+            }
+        })
+
+        animateText()
+
+        confetti_animation_view.setMaxFrame(63)
```
FredrikAreschoug: 63 should be moved to a constant variable in that way you can explain what it is whit the name of the variable.

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +103,50 @@ class SplashActivity : BaseActivity() {
...
+
+
+
+        return currentDay == birthDay && currentMonth == birthMonth
+
+    }
+
+
+    private fun hasBenCelebrated(): Boolean {
+        val pref = applicationContext.getSharedPreferences("MyPref", 0) // 0 - for private mode
```
FredrikAreschoug: Look into how we use `SharedPreferences` in the rest of the app

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +103,50 @@ class SplashActivity : BaseActivity() {
...
     }
 
+    private fun startBirthdayActivity(){
+        startActivity(Intent(this, BirthdayActivity::class.java))
+    }
+
+    private fun isBirthday(): Boolean {
+        val socialSecurityNumber = "199701174617"
+
+        val birthMonth = socialSecurityNumber.substring(4, socialSecurityNumber.length - 6).toInt()
```
FredrikAreschoug: You have the same logic in `BirthdayViewModel.kt` if you need to do this in two places please have a look at kotlin extensions and maybe create one.

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -11,24 +14,39 @@ import com.hedvig.app.feature.profile.ui.payment.TrustlyActivity
...
+    private val birthdayViewModel: BirthdayViewModel by viewModel()
+
+    private var isBirthday: Boolean = false
 
     override fun onStart() {
         super.onStart()
 
+        birthdayViewModel.apply {
+            isBirthdayData.observe(lifecycleOwner = this@SplashActivity) { data ->
+                isBirthday = data!!
```
FredrikAreschoug: You should almost never use `!!` please try to find another solution for this.

hedvigoscar:
Nice! I see that you've resolved a lot of the comments but not pushed any code, is the code that resolves the issues in your local branch? If so, please push it. I'd like to review the most recent version of the branch 🙂

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -25,15 +28,8 @@ import com.hedvig.app.feature.welcome.WelcomeDialog
...
-import com.hedvig.app.util.extensions.monthlyCostDeductionIncentive
-import com.hedvig.app.util.extensions.observe
-import com.hedvig.app.util.extensions.setupLargeTitle
-import com.hedvig.app.util.extensions.showShareSheet
-import com.hedvig.app.util.extensions.startClosableChat
-import com.hedvig.app.util.extensions.view.remove
-import com.hedvig.app.util.extensions.view.setHapticClickListener
-import com.hedvig.app.util.extensions.view.show
-import com.hedvig.app.util.extensions.view.updatePadding
+import com.hedvig.app.util.extensions.*
```
hedvigoscar: I'm guessing that your IDE does this automatically.

I'd like you to add `ktlint` to your IDE: https://github.com/pinterest/ktlint

and possibly also add some form of `Save Actions` where you can enable the IDE to perform code formatting on save. This will make sure that the codebase stays tidy over time 🙂

app/src/main/java/com/hedvig/app/feature/birthday/data/BirthdayRepository.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.app.feature.birthday.data
+
+import io.reactivex.Observable
+
+class BirthdayRepository {
+
+
+    fun fetchBirthDate(): Observable<String> {
+        return Observable.just("199701171234")
```
hedvigoscar: This probably needs to be changed later on, but I'm sure you know that already 😉

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        val calendar = Calendar.getInstance()
+
+        val currentMonth = calendar.get(Calendar.MONTH) + 1
+        val currentDay = calendar.get(Calendar.DAY_OF_MONTH)
+
+        return currentDay == birthDay && currentMonth == birthMonth
+    }
+
+    fun getBirthDay() {
```
hedvigoscar: This function only appears to be called once when the ViewModel is created. Since that is the case, we can move the function call to an `init`-block in the ViewModel and remove the explicit call in the `SplashActivity`.

app/src/main/java/com/hedvig/app/feature/birthday/BirthdayActivity.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.hedvig.app.feature.loggedin.ui.LoggedInActivity
+import com.hedvig.app.util.extensions.view.hide
+import com.hedvig.app.util.extensions.view.show
+import kotlinx.android.synthetic.main.activity_birthday.*
+import timber.log.Timber
+
+
+class BirthdayActivity : AppCompatActivity() {
+
+    private var animationDuration = 900
```
hedvigoscar: These can be moved to a `companion object` and be made `const val` 🙂

app/src/main/java/com/hedvig/app/feature/birthday/BirthdayActivity.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.hedvig.app.R
+import com.hedvig.app.feature.loggedin.ui.LoggedInActivity
+import com.hedvig.app.util.extensions.view.hide
+import com.hedvig.app.util.extensions.view.show
+import kotlinx.android.synthetic.main.activity_birthday.*
+import timber.log.Timber
+
+
+class BirthdayActivity : AppCompatActivity() {
```
hedvigoscar: We can use the following new constructor to not have to add `setContentView(...)` in `onCreate()` 🙂 

```suggestion
class BirthdayActivity : AppCompatActivity(R.layout.activity_birthday) {
```

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,48 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/from"
+        android:textSize="@dimen/text_xlarge"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/happyBirthdayText" />
+
+    <com.airbnb.lottie.LottieAnimationView
+        android:id="@+id/confetti_animation_view"
```
hedvigoscar:
```suggestion
        android:id="@+id/confetti_animation"
```

Not super-important, but we probably do not need to suffix the id with `view`, to avoid hungarian notiation and all 🙄

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
hedvigoscar: I think this would be wise 🙂

app/src/main/java/com/hedvig/app/feature/birthday/BirthdayActivity.kt
hedvigoscar: also you can suffix the value with `L` to make it a long in the first place, as it's only used as a long:

```suggestion
    private const val animationDuration = 900L
```

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.reactivex.rxkotlin.plusAssign
+import timber.log.Timber
+import java.util.*
+
+
+class BirthdayViewModel(
+    private val birthdayRepository: BirthdayRepository
+) : ViewModel() {
+
+    private lateinit var birthDate: String
```
hedvigoscar: This value only seems to be used in the `isBirthday`-function. We could perhaps remove it from the ViewModel and make it a parameter for that function, making it a little bit more pure 🙂

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    val isBirthdayData: MutableLiveData<Boolean> = MutableLiveData()
+    private val disposables = CompositeDisposable()
+
+
+
+    private fun isBirthday(): Boolean {
+
+        val birthMonth = birthDate.substring(4, birthDate.length - 6).toInt()
```
hedvigoscar: If we make this function a bit more pure in line with my suggestion around the `birthDate`-parameter, we could move it to a companion object or to its own file, allowing it to be reused a little bit better. 🙂

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="com.hedvig.app.feature.birthday.BirthdayActivity">
+
+    <TextView
+        style="@style/Header"
```
hedvigoscar: If possible, we should use one of the entirely qualified styles and not set any `textSize` directly on the `TextView`. `@style/Header.Large` seems to match the style you're looking for here 🙂

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
```
hedvigoscar: It would be interesting to see a screenshot of this activity. Could you attach one to the PR? 🙂

app/src/main/res/layout/activity_birthday.xml
```diff
@@ -0,0 +1,48 @@
...
+        android:text="@string/happy_birthday"
+        android:textSize="@dimen/text_xxlarge"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        android:layout_marginBottom="200dp"/>
+
+    <TextView
+        style="@style/Header"
```
hedvigoscar: Same as with the previous style, except here `@style/Header.Regular` seems to be the one you're looking for.

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -11,24 +14,43 @@ import com.hedvig.app.feature.profile.ui.payment.TrustlyActivity
...
 class SplashActivity : BaseActivity() {
     private val loggedInService: LoginStatusService by inject()
+    private val birthdayViewModel: BirthdayViewModel by viewModel()
+
+    private var isBirthday: Boolean = false
 
     override fun onStart() {
         super.onStart()
 
+        birthdayViewModel.apply {
```
hedvigoscar: If we move the `getBirthDay()`-call to an `init`-block in the ViewModel, we can skip using `apply` here and simply observe directly on the `isBirthdayData`-property.

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +107,34 @@ class SplashActivity : BaseActivity() {
...
+
+
+    private fun hasBenCelebrated(): Boolean {
+        val pref = applicationContext.getSharedPreferences("MyPref", 0) // 0 - for private mode
+
+        return pref.getBoolean("celebrated",false)
+    }
+
+    private fun setHasBenCelebrated() {
+        val socialSecurityNumber = "199701174617"
```
hedvigoscar: This value probably needs to be loaded from the ViewModel, but I guess this is just some WIP-code 😉

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +107,34 @@ class SplashActivity : BaseActivity() {
...
+    }
+
+
+    private fun hasBenCelebrated(): Boolean {
+        val pref = applicationContext.getSharedPreferences("MyPref", 0) // 0 - for private mode
+
+        return pref.getBoolean("celebrated",false)
+    }
+
+    private fun setHasBenCelebrated() {
```
hedvigoscar: This function does not seem to get called 😞

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +107,34 @@ class SplashActivity : BaseActivity() {
...
+        val socialSecurityNumber = "199701174617"
+
+        val currentDay = Calendar.getInstance().get(Calendar.DAY_OF_MONTH)
```
hedvigoscar: Since we're doing the same calculation in the ViewModel, I think we can re-use that code fairly easily and with decent benefits.

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -85,6 +107,34 @@ class SplashActivity : BaseActivity() {
...
+        val birthDay = socialSecurityNumber.substring(6, socialSecurityNumber.length - 4).toInt()
+
+        val pref = applicationContext.getSharedPreferences("MyPref", 0) // 0 - for private mode
+        val editor: SharedPreferences.Editor = pref.edit()
+
+        if (currentDay == birthDay) {
+            editor.putBoolean("celebrated", true)
```
hedvigoscar: This preference will make sure that the user will only be celebrated on the first birthday they open the app - we probably need to store the year that the user has been celebrated in order to allow subsequent years to receive a celebration as well 🙂

hedvigoscar:
I forgot to mention one thing!

What's the source of the Lottie animations we're using for this feature?

If possible, we could export them as AnimatedVectorDrawables which I think could be preferable to using Lottie animations.

app/src/main/res/layout/activity_birthday.xml
MathiasOhrstrom: image
With [this](https://lottiefiles.com/13982-testing) animation

hedvigoscar: Neat!

hedvigoscar: Neat!

app/src/main/java/com/hedvig/app/feature/birthday/viewmodel/BirthdayViewModel.kt
MathiasOhrstrom: `birthDate: String` is used in both `isBirthday()` and `getBirthday()`

hedvigoscar: That's true! However, `getBirthday()` only uses it to write to the property, and `isBirthday()` then subsequently reads from it. `isBirthday()` could be rewritten to have the signature `isBirthday(birthdate: String)`, and as such we avoid the property in the ViewModel 🙂

PR #80, opened 2020-01-27
Feature/change date

MathiasOhrstrom:


app/src/main/res/layout/feature_bubbles.xml
```diff
@@ -222,19 +222,35 @@
...
     </androidx.constraintlayout.widget.ConstraintLayout>
 
+    <Button
+        android:id="@+id/dateButton"
+        style="@style/HedvigButton.Outlined.Negative"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/base_margin_double"
+        android:layout_marginBottom="@dimen/base_margin"
+        android:textColor="@color/off_white"
```
hedvigoscar: We should probably not have both a style with `.Negative` and a literal `textColor` 🙂

app/src/main/java/com/hedvig/app/feature/offer/ChangeDateBottomSheet.kt
```diff
@@ -117,9 +104,12 @@ class ChangeDateBottomSheet : RoundedBottomSheetDialogFragment() {
...
                 isDatePicked = true
 
-                dateText.text = "$dayOfMonth $month $year"
+                dialog?.dateText?.text = "$dayOfMonth $month $year"
                 animateHintMove()
-                dateText.show()
+                dialog?.dateText?.show()
+
+                dialog?.chooseDateButton?.isEnabled = true
+                dialog?.chooseDateButton?.style(R.style.HedvigButton_Purple)
```
hedvigoscar: This is kind of cool!
Do we need to add a library for it though? I guess it could be sufficient just to update the colors, maybe?
The reason I'm a bit skeptical is that annotation processors are kind of heavy to introduce, and if we can avoid it we might want to.

app/src/main/res/values/styles.xml
```diff
@@ -107,10 +107,20 @@
         <item name="android:textColor">@color/white</item>
     </style>
 
+    <style name="HedvigButton.Purple.inactive">
```
hedvigoscar:
```suggestion
    <style name="HedvigButton.Purple.Disabled">
```

PR #81, opened 2020-01-27
Key Gear: Tab

hedvigoscar:
Missing right now:

- GraphQL model classes (schema is not ready)
- Translations

We'll use this as a base for the rest of the feature implementation.

There are a lot of unfilled TODOs - these will be filled later

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/keygear/KeyGearFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/keygear/KeyGearFragment.kt
@@ -1,0 +1,40 @@
+package com.hedvig.app.feature.keygear
+
+import android.os.Bundle
+import android.view.View
+import com.hedvig.app.R
+import com.hedvig.app.feature.loggedin.ui.BaseTabFragment
+import com.hedvig.app.ui.decoration.GridSpacingItemDecoration
+import com.hedvig.app.util.extensions.dp
+import com.hedvig.app.util.extensions.observe
+import com.hedvig.app.util.extensions.view.remove
+import com.hedvig.app.util.extensions.view.show
+import kotlinx.android.synthetic.main.fragment_key_gear.*
+import kotlinx.android.synthetic.main.loading_spinner.*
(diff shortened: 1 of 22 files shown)


app/src/main/java/com/hedvig/app/util/extensions/DimensionExtensions.kt
```diff
@@ -0,0 +1,6 @@
+package com.hedvig.app.util.extensions
+
+import android.content.res.Resources
+
+val Int.dp: Int
```
FredrikAreschoug: I don't love the name of this extension `toDp`?

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+class KeyGearFragment : BaseTabFragment() {
+    override val layout = R.layout.fragment_key_gear
+
+    private val viewModel: KeyGearViewModel by sharedViewModel()
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        items.adapter = KeyGearItemsAdapter()
+        items.addItemDecoration(GridSpacingItemDecoration(8.dp))
```
FredrikAreschoug: Maybe we should extract this constant to dimensions instead

app/src/main/java/com/hedvig/app/util/extensions/DimensionExtensions.kt
hedvigoscar: I think it's quite a nice name - I stole it from some Jetpack Compose-code that used this type of extension.

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearFragment.kt
hedvigoscar: I see your point - perhaps we could use a kotlin-based `const val`?

PR #82, opened 2020-01-30
Create Key Gear Item-Screen: Add Photos

hedvigoscar:
This implementation lacks backend integration
as there is no backend to integrate to right now.


app/src/main/java/com/hedvig/app/util/SpringUtils.kt
```diff
@@ -0,0 +1,113 @@
...
+    return springAnim
+}
+
+/**
+ * Map from a [DynamicAnimation.ViewProperty] to an `id` suitable to use as a [View] tag.
+ */
+@IdRes
+private fun getKey(property: DynamicAnimation.ViewProperty): Int {
+    return when (property) {
+        SpringAnimation.TRANSLATION_X -> R.id.translation_x
```
FredrikAreschoug: Strange that we need to do it like this feels like this should be supported in a better way from the system

hedvigoscar: This is actually the thing that Nick Butcher created and showed off in his talk: https://www.youtube.com/watch?v=f3Lm8iOr4mE

I had written one that was heavily inspired by that one, but I found the actual implementation in the source code for Plaid and decided to use that one straight away.

It looks a bit weird but it is indeed the way to go to implement this kind of Spring animation extension 🙂

app/src/main/java/com/hedvig/app/Constants.kt
```diff
@@ -1,3 +1,5 @@
 package com.hedvig.app
 
 const val BASE_MARGIN = 8
+const val BASE_MARGIN_HALF = BASE_MARGIN / 2
```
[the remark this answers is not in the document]
hedvigoscar: We can, but I'm leaning towards this solution due to dimens being a little bit unergonomic when used in code.

Also, when moving from XML to code you can do a bit more, for example express all units as function of the base unit, which is just kind of nice

PR #83, opened 2020-01-31
fixed layout issues on smaller phones

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/offer/OfferActivity.kt
```diff
@@ -171,7 +171,7 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
 
     private fun setupScrollListeners() {
         offerScroll.setOnScrollChangeListener { _: NestedScrollView, _, scrollY, _, _ ->
-            parallaxContainer.translationY = scrollY / 1.25f
+            parallaxContainer.translationY = scrollY / 7.0f
```
hedvigoscar: How does this affect the parallax?

app/src/main/res/values/styles.xml
```diff
@@ -112,6 +112,8 @@
     </style>
 
     <style name="HedvigButton.Outlined.White">
+        <item name="android:paddingStart">12dp</item>
```
hedvigoscar: I feel like this might be a little bit unexpected from the perspective of consuming this style: Would the user expect the padding to change by setting a different color on the button?

app/src/main/res/layout/feature_bubbles.xml
```diff
@@ -227,9 +227,10 @@
         style="@style/HedvigButton.Outlined.White"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
+        android:paddingStart="12dp"
```
hedvigoscar: Move to dimens

PR #84, opened 2020-02-03
Edge to edge on marketing and chat screen

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/chat/ui/ChatActivity.kt
```diff
@@ -82,6 +84,44 @@ class ChatActivity : BaseActivity(R.layout.activity_chat) {
...
+        }
+
+        messages.doOnApplyWindowInsets { view, insets, initialState ->
+            view.updatePadding(
+                top = initialState.paddings.top + frameLayout.measuredHeight,
+                bottom = initialState.paddings.bottom + insets.systemWindowInsetBottom + input.textInputContainer.measuredHeight
+            )
+        }
+
+        input.textInputContainer.remove()
```
hedvigoscar: Let's make this procedure a function on the `ChatInputView` - a method that measures its text-wise height and cleans up any show/hides it has to do in the process, then returns the measurements 🙂

PR #89, opened 2020-02-07
Key gear item detail-screen: Photo Carousel and more

hedvigoscar:
This PR got a bit out of hand due to the different blockers and branches that had to be created around them.

Once it's reviewed and merged however I think we're back on track


hedvigoscar:
Closing this because it's history is just totally wack

PR #91, opened 2020-02-10
Key Gear Item Detail: Add/View Receipt

hedvigoscar:
Missing: Translation for the bottom sheet to upload a receipt.
Saving this for later as there is currently no translation available.

### DIFF
--- a/app/src/main/res/layout/activity_key_gear_item_detail.xml
+++ b/app/src/main/res/layout/activity_key_gear_item_detail.xml
@@ -45,7 +45,7 @@
 
             <View
                 android:layout_width="match_parent"
-                android:layout_height="1dp"
+                android:layout_height="@dimen/horizontal_line_height"
                 android:layout_marginTop="@dimen/base_margin_quintuple"
                 android:background="@color/separator" />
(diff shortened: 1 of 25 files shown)


app/src/main/res/layout/activity_key_gear_item_detail.xml
```diff
@@ -40,15 +41,77 @@
                 android:layout_height="wrap_content"
                 android:layout_marginStart="@dimen/base_margin_triple"
                 android:layout_marginTop="@dimen/base_margin_quintuple"
-                android:paddingBottom="1500dp"
                 android:text="@string/KEY_GEAR_ITEM_VIEW_COVERAGE_TABLE_TITLE" />
 
+            <View
+                android:layout_width="match_parent"
+                android:layout_height="1dp"
```
FredrikAreschoug: don't we have an dimen for this?

hedvigoscar: We don't! We have 1dp all over the place it turns out. But I added one, lets try to use that one

PR #92, opened 2020-02-10
Key gear item detail purchase date

MathiasOhrstrom:


app/src/main/res/layout/hedvig_edit_text.xml
```diff
@@ -0,0 +1,46 @@
...
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent" />
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+    <TextView
+        android:id="@+id/textHint"
+        style="@style/Body"
```
hedvigoscar: Avoid using the base `Body`-style, always specify the variant (probably `Body.Regular` in this case.

PR #96, opened 2020-02-13
Concrete implementation of API calls

hedvigoscar:
I love coroutines

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/keygear/ui/createitem/CreateKeyGearItemViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/keygear/ui/createitem/CreateKeyGearItemViewModel.kt
@@ -4,8 +4,12 @@ import android.net.Uri
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/hedvig/app/feature/keygear/ui/createitem/CreateKeyGearItemViewModel.kt
```diff
@@ -75,10 +79,23 @@ abstract class CreateKeyGearItemViewModel : ViewModel() {
...
 
-class CreateKeyGearItemViewModelImpl : CreateKeyGearItemViewModel() {
+class CreateKeyGearItemViewModelImpl(
+    private val keyGearItemsRepository: KeyGearItemsRepository
+) : CreateKeyGearItemViewModel() {
     override val createResult = MutableLiveData<CreateKeyGearItemMutation.Data>()
 
     override fun createItem() {
-        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+        viewModelScope.launch {
```
FredrikAreschoug: This is so nice! when I started with coroutines viewmodel was not a thing and it was always a bit unclear wich scope you should use! This makes everything so much clearer!

hedvigoscar: It is indeed the best thing since sliced bread 😎

PR #97, opened 2020-02-13
Key Gear: Items in Tab, concrete implementation

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/keygear/data/KeyGearItemsRepository.kt
```diff
@@ -1,16 +1,16 @@
...
     private val apolloClientWrapper: ApolloClientWrapper
 ) {
-    fun keyGearItems() = Rx2Apollo.from(
+    fun keyGearItems() =
         apolloClientWrapper
             .apolloClient
             .query(KeyGearItemsQuery())
             .watcher()
-    )
+            .toChannel()
```
FredrikAreschoug: Nice!

app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearViewModel.kt
```diff
@@ -3,12 +3,25 @@ package com.hedvig.app.feature.keygear.ui.tab
...
+class KeyGearViewModelImpl(
+    private val repository: KeyGearItemsRepository
+) : KeyGearViewModel() {
     override val data = MutableLiveData<KeyGearItemsQuery.Data>()
+
+    init {
+        viewModelScope.launch {
+            for (response in repository.keyGearItems()) {
+                data.postValue(response.data())
+            }
```
FredrikAreschoug: with this forloop wont we just end up with the last `response.data()` posted to data? can't we just use `repository.keyGearItems().last()` then? or am I missing something?

hedvigoscar: You're probably not missing something - it's most likely me not being used to using coroutine channels.

What I wanted to express was:

Whenever the channel emits a value, post it to the LiveData.

If this implementation is incorrect, please let me know 🙂

FredrikAreschoug: I don't think what I'm commenting on has anything to do with the channel. I think that part will work like the Rx implementation. From my understanding, you will get the response of `[keyGerItems]` from the graphql query. With this implementation, you will loop through that list and post all of them one by one to the MutableLiveData which will end up with the last being the one posted to ` data: MutableLiveData<KeyGearItemsQuery.Data>`

hedvigoscar: `KeyGearItemsRepository.keyGearItems()` has the return type `Response<KeyGearItemsQuery.Data!>` 🙂 Maybe we should rename it, or add the type signature explicitly.

On the topic of channels, do you think my usage of them is correct? I'm unable to test this properly as backend is not quite ready yet.

hedvigoscar: I think we can try it live.

PR #99, opened 2020-02-14
Feature/key gear valuation mutations

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,23 +1,39 @@
...
 import com.hedvig.app.util.extensions.view.setHapticClickListener
 import kotlinx.android.synthetic.main.activity_key_gear_valuation.*
 import org.koin.android.viewmodel.ext.android.viewModel
+import org.threeten.bp.YearMonth
+import timber.log.Timber
 import java.text.DateFormatSymbols
 
-class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuation) {
+class KeyGearValuationActivity :
+    BaseActivity(R.layout.activity_key_gear_valuation) {
```
hedvigoscar: This formatting looks weird 🤔

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,23 +1,39 @@
...
+    private val model: KeyGearItemDetailViewModel by viewModel()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val id = intent.getStringExtra(ITEM_ID)
+        var date: YearMonth? = null
+
+        continueButton.isEnabled = false
+        continueButton.backgroundTintList =
```
hedvigoscar: I think we should take the Hedvig Button style to the next level and use a Color state list with a disabled-selector:

https://developer.android.com/guide/topics/resources/color-list-resource

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -28,16 +44,51 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
         }
 
         continueButton.setHapticClickListener {
-            //TODO
-            model.submit()
-            onBackPressed()
+            val price = priceInput.getText()
+            date?.let { date ->
```
hedvigoscar: When you need to null-check multiple values, you can use a neat little function called `safeLet`.

You provide `safeLet` with `n` nullable values, and it only executes the provided block of code when all of them are non-null.

Example:

```kotlin
import com.hedvig.app.util.safeLet
// ...
val a: String? = null
val b = 123
safeLet(a, b) { a, b ->
    // This block does not execute because `a` is null
    print(a)
    print(b)
}

val c = false
val d = Pair(1, 2)
safeLet(c, d) { c, d ->
    // This block of code does execute, because c and d are both non-null
    print(c)
    print(d)
}
```

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -28,16 +44,51 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
+                ColorStateList.valueOf(this.compatColor(R.color.purple))
+        } else {
+            continueButton.isEnabled = false
+            continueButton.backgroundTintList =
+                ColorStateList.valueOf(this.compatColor(R.color.semi_light_gray))
+        }
+    }
+
+    companion object {
+
```
hedvigoscar:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/keygear/ui/itemdetail/KeyGearItemDetailViewModel.kt
```diff
@@ -4,24 +4,58 @@ import android.net.Uri
...
 }
 
-class KeyGearItemDetailViewModelImpl : KeyGearItemDetailViewModel() {
+class KeyGearItemDetailViewModelImpl(private val repository: KeyGearItemsRepository) :
+    KeyGearItemDetailViewModel() {
     override val data = MutableLiveData<KeyGearItemQuery.KeyGearItem>()
+
     override val isUploading = LiveEvent<Boolean>()
 
+    override fun choosePurchaseDate(yearMonth: YearMonth) {
```
hedvigoscar: I think this function could be a part of the base abstract class, as it will always be the same whether or not you are mocking network calls 🙂

hedvigoscar: Come to think of it, I think this code should actually be in its own ViewModel - ViewModels are not shared between activities.

app/src/main/java/com/hedvig/app/ui/view/DateInputView.kt
```diff
@@ -46,6 +46,8 @@ class DateInputView @JvmOverloads constructor(
             dateText.text = text
         }
 
+    fun getDateText() = dateText.text.toString()
```
hedvigoscar: Is this function used?

app/src/main/java/com/hedvig/app/ui/view/HedvigEditText.kt
```diff
@@ -54,6 +55,10 @@ class HedvigEditText @JvmOverloads constructor(
         }
     }
 
+    fun getText() = textInput.text.toString()
+
+    fun getEditText(): EditText = textInput
```
hedvigoscar: I think this implementation detail should be hidden by `HedvigEditText` - we could provide a way to add an `onChange`-listener to the parent `HedvigEditText` and pass it down to the wrapped `EditText`.

app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -168,7 +169,7 @@ val viewModelModule = module {
     viewModel { ReferralViewModel(get()) }
     viewModel { WelcomeViewModel(get()) }
     viewModel { LanguageViewModel(get()) }
-    viewModel { KeyGearItemDetailViewModelImpl() }
+    viewModel { KeyGearItemDetailViewModelImpl(get()) }
```
hedvigoscar:
```suggestion
```

app/src/main/res/color/hedvig_button_state.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="@color/semi_light_gray" android:state_enabled="false" />
```
hedvigoscar: Is this also the correct colour for dark theme? 🙂

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,23 +1,35 @@
...
+    private val model: KeyGearItemDetailViewModel by viewModel()
+    private val valuationModel: KeyGearValuationViewModel by viewModel()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val id = intent.getStringExtra(ITEM_ID)
+        var date: YearMonth? = null
+
+        continueButton.isEnabled = false
```
hedvigoscar: Is this attribute modifiable through XML?

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,23 +1,35 @@
...
 
-    private val model: KeyGearValuationViewModelImpl by viewModel()
+    private val model: KeyGearItemDetailViewModel by viewModel()
+    private val valuationModel: KeyGearValuationViewModel by viewModel()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        val id = intent.getStringExtra(ITEM_ID)
+        var date: YearMonth? = null
```
hedvigoscar: Having this variable in this scope might make the code resistant to refactoring. Should we lift it to the Activity-scope? (set it as a private var on the Activity-level)

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -28,16 +40,42 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
-            //TODO
-            model.submit()
-            onBackPressed()
+            val price = priceInput.getText()
+
+            safeLet(date, id) { date, id ->
+                val monetaryValue =
+                    MonetaryAmountV2Input.builder().amount(price).currency("SEK").build()
+                model.updatePurchaseDateAndPrice(id, date, monetaryValue)
+                onBackPressed()
```
hedvigoscar: What action we take here should not be `onBackPressed()` in the future, hence we should add a TODO here explaining that.

I think it's probably also appropriate to add the transitionary state here with the loading button 🙂

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -28,16 +40,42 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
         }
 
-        model.purchaseDate.observe(this) { yearMonth ->
+        valuationModel.purchaseDate.observe(this) { yearMonth ->
+            setButtonState(priceInput.getText().isNotEmpty(), yearMonth != null)
             yearMonth?.let {
+                date = yearMonth
                 dateInput.text =
                     "${DateFormatSymbols().months[yearMonth.month.value - 1]} ${yearMonth.year}"
+                Timber.d(yearMonth.toString())
```
hedvigoscar:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,23 +1,35 @@
...
 import kotlinx.android.synthetic.main.activity_key_gear_valuation.*
 import org.koin.android.viewmodel.ext.android.viewModel
+import org.threeten.bp.YearMonth
+import timber.log.Timber
 import java.text.DateFormatSymbols
 
 class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuation) {
 
-    private val model: KeyGearValuationViewModelImpl by viewModel()
+    private val model: KeyGearItemDetailViewModel by viewModel()
```
hedvigoscar: I think we should not act on this ViewModel - we should move all code specficially related to this screen to the `KeyGearValuationViewModel` 🙂

app/src/main/java/com/hedvig/app/feature/keygear/data/KeyGearItemsRepository.kt
```diff
@@ -23,13 +29,61 @@ class KeyGearItemsRepository(
...
             .query(KeyGearItemsQuery())
             .watcher()
             .toChannel()
 
+    fun keyGearItem(id: String): Channel<Response<KeyGearItemQuery.Data>> {
+        keyGearItemQuery = KeyGearItemQuery(id)
+        return apolloClientWrapper.apolloClient.query(keyGearItemQuery).watcher().toChannel()
+    }
+
+    suspend fun updatePurchasePriceAndDateAsync(
```
hedvigoscar: 👏

PR #100, opened 2020-02-17
Key Gear: Illustrations by category

hedvigoscar:
This had major implications for the shared element transitions, and as such there's a lot of code around it!

Bonus: Update LeakCanary


hedvigoscar:
They seem a bit similar on the surface, but they actually deal with very different topics.

One shows the photos of one Key Gear-Item, and one shows all Key Gear Items that you have previously created, including a static first item that is an 'Create New Key Gear Item'-button.

I think we would lose more by trying to extract some common code from these, so I would argue against it 🙂

PR #101, opened 2020-02-17
Key Gear: Upload receipt, concrete implementation

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/keygear/data/KeyGearItemsRepository.kt
```diff
@@ -106,4 +110,64 @@ class KeyGearItemsRepository(
...
+        context.contentResolver.openInputStream(file)?.into(uploadFile)
+        val uploadResult = apolloClientWrapper
+            .apolloClient
+            .mutate(UploadFileMutation(FileUpload(mimeType, uploadFile)))
+            .toDeferred()
+            .await()
+
+        val uploadData = uploadResult.data()
+        if (uploadData == null) {
+            e { "Failed to upload photo" }
```
FredrikAreschoug: Not sure if I like it, really not a fan of code that doesn't explain what it does.

hedvigoscar: I'm not sure that I follow 🤔

PR #102, opened 2020-02-18
Feature/key gear valuation info

MathiasOhrstrom:


app/src/main/res/values/text_styles.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    <dimen name="text_xxxlarge">38sp</dimen>
```
hedvigoscar: Is this correct?

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationInfoActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+                    "PURCHASE_PRICE" to purchasePrice,
+                    "VALUATION_PRICE" to 1234
+                )
+            )
+        }
+        model.loadItem(id)
+    }
+
+    private fun setPercentage(percentage: Int) {
+        valuationPercentage.setMarkdownText("$percentage%")
```
hedvigoscar: Should this be markdown text? The text does not appear to contain any markdown. A spanned string might be more appropriate.

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationInfoActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        val id = intent.getStringExtra(ITEM_ID)
+
+        model.data.observe(this) { data ->
+            //TODO get all data
+            val category = data?.fragments?.keyGearItemFragment?.category.toString()
+            val purchasePrice = data?.fragments?.keyGearItemFragment?.purchasePrice.toString()
+
+            setPercentage(47)
+            //TODO fix correct markdown from * to **
```
hedvigoscar: Is something amiss with the markdown? 🤔

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -44,31 +82,89 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
+            }
+            start()
+        }
     }
 
     private fun setButtonState(hasPrice: Boolean, hasDate: Boolean) {
-        continueButton.isEnabled = hasPrice && hasDate
+        if (hasPrice && hasDate) {
+            save.isEnabled = true
+            saveContainer.backgroundTintList =
```
hedvigoscar: Should we use the disabled-feature here as well? 🙂

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationInfoActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+    companion object {
+        private const val ITEM_ID = "ITEM_ID"
+
+        fun newInstance(context: Context, id: String) =
+            Intent(context, KeyGearValuationInfoActivity::class.java).apply {
+                putExtra(ITEM_ID, id)
+            }
+    }
+}
+
```
hedvigoscar:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationInfoActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+        close.setHapticClickListener {
+            onBackPressed()
+        }
+
+        model.data.observe(this) { data ->
+            data?.let {
+                val category = data.fragments.keyGearItemFragment.category.toString().toLowerCase()
+                val purchasePrice = data.fragments.keyGearItemFragment.purchasePrice?.amount
+                val valuationPercentage =
+                    (data.fragments.keyGearItemFragment.valuation as KeyGearItemFragment.AsKeyGearItemValuationFixed).ratio
```
hedvigoscar: This cast is highly unsafe. We need to behave differently based on the type here.

Use a `when`-statement:

```kotlin
when (val valuation = data.fragments.keyGearItemFragment.valuation)
    is KeyGearItemFragment.AsKeyGearItemValuationFixed -> { ... }
    is KeyGearItemFragment.AsKeyGearItemValuationMarketPrice -> { ... }
    else -> { ... }
```

hedvigoscar: Come to think of it - this screen has no interaction whatever with the data except for displaying it.

We could actually make it entirely parameterized and avoid using a ViewModel entirely - this would probably be quite desirable actually as it decreases complexity significantly 🙂

app/src/main/res/values/text_styles.xml
```diff
@@ -34,6 +39,11 @@
         <item name="android:textColor">@color/text_emphasized</item>
     </style>
 
+    <style name="Headline.xLarge">
```
hedvigoscar:
```suggestion
    <style name="Headline.XLarge">
```

For the sake of consistency, lets use `XLarge` over `xLarge`, given that we already have `Body.XSmall`

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
MathiasOhrstrom: Dont know if that is possible since i disable the button but change the color of the parent frameLayout

hedvigoscar: Is it possible to set the FrameLayout to be disabled? If not, we can disregard it, but it's interesting to see if we can solve the problem using the existing facilities, so to speak

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -121,6 +130,14 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
         }
     }
 
+    private fun valuationType(item: KeyGearItemQuery.KeyGearItem): ValuationType {
+        return if (item.fragments.keyGearItemFragment.valuation is KeyGearItemFragment.AsKeyGearItemValuationFixed) {
```
hedvigoscar: This is a little dangerous: Consider the scenario when we introduce a third valuation type. This code will incorrectly identify that third type as `MARKET_PRICE`. We need to consider a strategy that does not have the issue.

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationInfoActivity.kt
```diff
@@ -3,65 +3,88 @@ package com.hedvig.app.feature.keygear
...
             Intent(context, KeyGearValuationInfoActivity::class.java).apply {
-                putExtra(ITEM_ID, id)
+                putExtra(CATEGORY, category)
+                putExtra(VALUATION_DATA, valuationData)
             }
     }
 }
 
+@Parcelize
+data class ValuationData(
```
hedvigoscar: Let's move these classes to their own files 🙂

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -111,30 +112,33 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
         }
     }
 
-    private fun valuationType(item: KeyGearItemQuery.KeyGearItem): ValuationType {
-        return if (item.fragments.keyGearItemFragment.valuation is KeyGearItemFragment.AsKeyGearItemValuationFixed) {
-            ValuationType.FIXED
-        } else {
-            ValuationType.MARKET_PRICE
+    private fun valuationType(item: KeyGearItemQuery.KeyGearItem): ValuationType? {
+        return when {
```
hedvigoscar:
```suggestion
        return when (item.fragments.keyGearItemFragment.valuation) {
```
This will allow you to not repeat yourself on the following rows in the when-statement 🙂

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -111,30 +112,33 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
+            item.fragments.keyGearItemFragment.valuation is KeyGearItemFragment.AsKeyGearItemValuationFixed -> ValuationType.FIXED
```
hedvigoscar:
```suggestion
            is KeyGearItemFragment.AsKeyGearItemValuationFixed -> ValuationType.FIXED
```

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -111,30 +112,33 @@ class KeyGearValuationActivity : BaseActivity(R.layout.activity_key_gear_valuati
...
+            item.fragments.keyGearItemFragment.valuation is KeyGearItemFragment.AsKeyGearItemValuationMarketValue -> ValuationType.MARKET_PRICE
```
hedvigoscar:
```suggestion
            is KeyGearItemFragment.AsKeyGearItemValuationMarketValue -> ValuationType.MARKET_PRICE
```

PR #104, opened 2020-02-18
Key Gear: Switch to LocalDate and DatePickerDialog in valuation activity

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationActivity.kt
```diff
@@ -1,35 +1,40 @@
...
+            val calendar = Calendar.getInstance()
+            DatePickerDialog(
+                this,
+                DatePickerDialog.OnDateSetListener { _, year, month, dayOfMonth ->
+                    date = LocalDate.of(year, month, dayOfMonth)
+                },
+                calendar.get(Calendar.YEAR),
+                calendar.get(Calendar.MONTH),
+                calendar.get(Calendar.DAY_OF_MONTH)
+            ).show()
```
MathiasOhrstrom: Can you set a future date with this datePicker?

hedvigoscar: Good catch! I think so, which is not great.

PR #134, opened 2020-02-28
changed some margins

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearFragment.kt
```diff
@@ -79,6 +79,7 @@ class KeyGearFragment : BaseTabFragment() {
             illustration.remove()
             title.remove()
             description.remove()
+            items.updateMargin(top = (resources.getDimension(R.dimen.base_margin_octuple) / resources.displayMetrics.density).toInt())
```
hedvigoscar:
```suggestion
            items.updateMargin(top = resources.getDimensionPixelSize(R.dimen.base_margin_octuple))
```

Or perhaps even

```kotlin
... top = BASE_MARGIN_X
```

hedvigoscar: And let's not forget to reset it in the other case as well 🙂

app/src/main/res/layout/fragment_key_gear.xml
```diff
@@ -66,7 +66,9 @@
             android:id="@+id/items"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:layout_marginStart="@dimen/base_margin_double"
```
hedvigoscar: Does this play correctly with the `ItemDecoration` we're using for this `RecyclerView`? 🙂

PR #141, opened 2020-03-04
Chore: Upgrade apollo, switch to kotlin code generation

hedvigoscar:
This took less time than expected and the outcome is kind of great!﻿

### DIFF
--- a/app/src/main/java/com/hedvig/app/service/push/managers/PaymentNotificationManager.kt
+++ b/app/src/main/java/com/hedvig/app/service/push/managers/PaymentNotificationManager.kt
@@ -6,7 +6,6 @@ import android.content.Intent
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.app.TaskStackBuilder
-import com.google.firebase.messaging.RemoteMessage
 import com.hedvig.app.R
 import com.hedvig.app.feature.loggedin.ui.LoggedInActivity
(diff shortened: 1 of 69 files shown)


app/src/main/java/com/hedvig/app/service/push/managers/PaymentNotificationManager.kt
```diff
@@ -90,38 +89,38 @@ object PaymentNotificationManager {
...
+    //         .build()
 
-        NotificationManagerCompat
-            .from(context)
-            .notify(CLAIM_PAID_NOTIFICATION_ID, notification)
-    }
+    //     NotificationManagerCompat
+    //         .from(context)
+    //         .notify(CLAIM_PAID_NOTIFICATION_ID, notification)
+    // }
```
FredrikAreschoug: remove or keep?

hedvigoscar:
This pull request is _free-for-all_. As it aims to not change any behaviour, and manual testing has so far shown that it does maintain the intended behaviour.

app/src/main/java/com/hedvig/app/service/push/managers/PaymentNotificationManager.kt
hedvigoscar: I wanted to keep it because I know we will implement the functionality later on.

hedvigoscar:
I think a more thorough look-through is not quite necessary. Let's merge it 🙂

PR #142, opened 2020-03-04
Norway: Authentication

hedvigoscar:


hedvigoscar:
Note that we're lacking some design and such for this feature, but it should be in a PoC-ready state

hedvigoscar:
Actually nevermind, there is some UI available. I'm going to implement it before we should review this PR

FredrikAreschoug:
Missed your comment above my approve :|

hedvigoscar:
It's fine, the UI implications are small 🙂

hedvigoscar:
We're still missing some UI, but this is clearly better than what was there before :sligh'

PR #148, opened 2020-03-20
apollo catch

MathiasOhrstrom:


hedvigoscar:
Looks good!

We also need to catch errors in Coroutine Flows, which we do via the first method listed here: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect.html

As such:

```kotlin
exampleFunctionReturningAFlow()
    .onEach { response -> doStuffWithResponse(response) }
    .catch { e -> handleException(e) }
    .collect()
```

It's probably also a good idea to log the error via Slimber for any given error we get:

```kotlin
val result = runCatching { ... }
if (result.isFailure)
    result.exceptionOrNull()?.let { e { it } }
```

This is quick example code and warrants some cleanup so that it feels good 😉

app/src/main/java/com/hedvig/app/feature/keygear/KeyGearValuationViewModel.kt
```diff
@@ -38,19 +40,26 @@ class KeyGearValuationViewModelImpl(
     ) {
         viewModelScope.launch {
             val result = runCatching { repository.updatePurchasePriceAndDateAsync(id, date, price) }
+            if (result.isFailure) {
+                result.exceptionOrNull()?.let { e { it.localizedMessage } }
```
hedvigoscar:
```suggestion
                result.exceptionOrNull()?.let { e { it } }
```

Afaik, Slimber has support for error-logging exceptions without converting them to Strings. I think this is the best way to go here 🙂

app/src/main/java/com/hedvig/app/feature/keygear/ui/itemdetail/KeyGearItemDetailViewModel.kt
```diff
@@ -34,36 +36,49 @@ class KeyGearItemDetailViewModelImpl(
 
     override fun loadItem(id: String) {
         viewModelScope.launch {
-            runCatching {
+            val result = runCatching {
```
hedvigoscar: We do not need to wrap `Flow`s with `runCatching`, as are already safe when we add a `catch`-clause for them 🙂

app/src/main/java/com/hedvig/app/feature/keygear/ui/itemdetail/KeyGearItemDetailViewModel.kt
```diff
@@ -34,36 +36,49 @@ class KeyGearItemDetailViewModelImpl(
...
                 repository
                     .keyGearItem(id)
+                    .catch { e -> e { e.localizedMessage } }
                     .collect { response ->
                         data.postValue(response.data()?.keyGearItem)
                     }
```
hedvigoscar: Should we use the following form instead?

```suggestion
                    .keyGearItem(id)
                    .onEach { response ->
                        data.postValue(response.data()?.keyGearItem)
                    }
                    .catch { e -> e { e.localizedMessage } }
                    .collect()
```

PR #150, opened 2020-03-24
opens Adyen instead of Trustly if on NO market

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/offer/OfferSignDialog.kt
```diff
@@ -119,11 +122,19 @@ class OfferSignDialog : DialogFragment() {
 
     private fun goToDirectDebit() {
         requireContext().storeBoolean(IS_VIEWING_OFFER, false)
+        val pref = PreferenceManager.getDefaultSharedPreferences(context)
+        val market = Market.values()[pref.getInt(Market.MARKET_SHARED_PREF, -1)]
```
hedvigoscar: It's important that we don't default to `-1` as this will always cause a crash in case we're in an incorrect state.

app/src/main/res/values/styles.xml
```diff
@@ -137,5 +140,11 @@
         <item name="android:backgroundDimEnabled">false</item>
     </style>
 
+    <style name="AdyenCheckout.TextInputLayout" parent="Widget.MaterialComponents.TextInputLayout.OutlinedBox">
+        <item name="boxStrokeColor">@color/primaryColor</item>
```
hedvigoscar: We can use theme attributes here
```suggestion
        <item name="boxStrokeColor">?colorPrimary</item>
```

PR #151, opened 2020-03-25
Feature/chat brand pivot

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
```diff
@@ -141,6 +142,18 @@ fun Context.showAlert(
         }
         .show()
         .apply {
+            getButton(AlertDialog.BUTTON_POSITIVE).apply {
```
hedvigoscar: This is probably not a good way to theme an AlertDialog 😞

I think that by default AlertDialogs use the theme provided in the attribute `?alertDialogTheme`, check that out!

app/src/main/res/layout/activity_chat.xml
```diff
@@ -2,6 +2,8 @@
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
+    android:theme="@style/Hedvig.Theme.Chat"
```
hedvigoscar:
```suggestion
```

This should be specified in the `AndroidManifest.xml`.

If not automatically applied in the editor, check if the `tools:context` attribute is set 🙂

app/src/main/res/values/theme.xml
```diff
@@ -63,4 +63,20 @@
     <attr name="colorButton" format="reference" />
     <attr name="textColorButton" format="reference" />
 
+    <style name="Hedvig.Theme.Chat" parent="Hedvig.Theme">
+        <item name="android:colorBackground">@color/colorBackgroundChat</item>
```
hedvigoscar: This seems suspect. Is the background color of the chat inconsistent with the rest of the application?

MathiasOhrstrom: Yes, the dark background on the chat screen is #1B1B1B and the others are #121212. Maybe I should talk to Zak about that

hedvigoscar: Yes, go 👮

app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
```diff
@@ -127,34 +125,20 @@ fun Context.showAlert(
...
-            getButton(AlertDialog.BUTTON_NEGATIVE).apply {
-                background = ColorDrawable(Color.TRANSPARENT)
-                setTextColor(resources.getColor(R.color.purple))
-            }
-            getButton(AlertDialog.BUTTON_NEUTRAL).apply {
-                background = ColorDrawable(Color.TRANSPARENT)
-                setTextColor(resources.getColor(R.color.purple))
-            }
-            window?.setBackgroundDrawable(ColorDrawable(Color.TRANSPARENT))
             getButton(AlertDialog.BUTTON_POSITIVE)?.isAllCaps = false
```
hedvigoscar: While we're at it, let's go all the way with theming and use the theme to control this as well.

PR #158, opened 2020-03-27
Norway: Web-Onboarding WebView

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/marketing/ui/MarketingActivity.kt
```diff
@@ -290,18 +293,37 @@ class MarketingActivity : BaseActivity() {
         login.show()
         getHedvig.show()
 
+        val market = getMarket()
+        if (market == null) {
+            e { "Programmer error: No market set when viewing ${this.javaClass.name}" }
+            return
+        }
```
MathiasOhrstrom: Should we open MarketPickerAcitvity here?

hedvigoscar: Good idea! I think yes

hedvigoscar: Actually, we've actually validated earlier in onCreate that we have a Market. Given that, I think that this is fine.

PR #159, opened 2020-03-30
Norway: Adyen Implementation

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/profile/ui/payment/connect/ConnectPaymentActivity.kt
```diff
@@ -9,34 +10,68 @@ import android.os.Bundle
...
-    @SuppressLint("SetJavaScriptEnabled")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_trustly)
 
+        val market = getMarket()
+        if (market == null) {
+            e { "Programmer error: ${this.javaClass.name} accessed without a Market selected" }
+            return
+        }
```
MathiasOhrstrom: Should we open MarketPicker here also?

hedvigoscar: Probably a good idea - will implement

PR #171, opened 2020-04-03
fixed crash and defaults to SE if no matched market

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/language/LanguageAndMarketViewModel.kt
```diff
@@ -145,10 +146,22 @@ class LanguageAndMarketViewModel(
...
+            val response = runCatching {
+                marketRepository.geoAsync().await()
+            }
+            if (response.isFailure) {
+                response.exceptionOrNull()?.let { e(it) }
+                updateMarket(Market.SE)
+            } else {
+                preselectedMarket.postValue(response.getOrNull()?.data()?.geo?.countryISOCode)
+                response.getOrNull()?.data()?.geo?.let { geo ->
+                    val market: Market = try {
```
hedvigoscar, line 158: We can just skip pre-selecting a market if we can't coerce the Geo-information to a market 🙂 so just return early at this point really

hedvigoscar: also same if we get an error in the network request 🙂

app/src/main/java/com/hedvig/app/feature/language/LanguageAndMarketViewModel.kt
```diff
@@ -151,16 +151,15 @@ class LanguageAndMarketViewModel(
...
-                    }
-                    updateMarket(market)
+                return@launch
+            }
+            preselectedMarket.postValue(response.getOrNull()?.data()?.geo?.countryISOCode)
+            response.getOrNull()?.data()?.geo?.let { geo ->
+                try {
+                    updateMarket(Market.valueOf(geo.countryISOCode))
+                } catch (e: Exception) {
+                    e(e)
```
hedvigoscar, line 161: This is probably just going to be `cant cast this geo to this market`, so we don't need to send it to crashlytics.
```suggestion
```

PR #172, opened 2020-04-03
always opens market picker if LoginStatus is ONBOARDING even if marke…

MathiasOhrstrom:
…t has ben selected before


hedvigoscar:
I think this is intended behaviour? If you have selected your market once you should not have to select it again. You can change it via the settings later in case you made a mistake.

MathiasOhrstrom:
> I think this is intended behaviour? If you have selected your market once you should not have to select it again. You can change it via the settings later in case you made a mistake.

I think this will be better because if i pick the wrong market i can only change it if i go into the chat activity and then go to settings, or if i clear the app data. I think this can be confusing for the enduser.
Now that if think of it, we can maybe add a button in the marketingActivity to go back to marketpicker instead

PR #178, opened 2020-04-07
some theme fixes

MathiasOhrstrom:


app/src/main/res/layout/activity_marketing.xml
```diff
@@ -35,6 +35,7 @@
         android:layout_marginEnd="@dimen/base_margin_double"
         android:layout_marginBottom="@dimen/base_margin"
         android:text="@string/MARKETING_SCREEN_CTA"
+        android:textColor="?textColorButton"
```
hedvigoscar, line 38: This should not be necessary, this color should be applied automatically through `style="?materialButtonContainedLargeStyle"`

app/src/main/res/layout/dialog_change_start_date.xml
```diff
@@ -3,6 +3,7 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tool="http://schemas.android.com/tools"
     android:layout_width="match_parent"
+    android:theme="@style/Hedvig.Theme"
```
hedvigoscar, line 6: Is this the optimal way to apply the theme should be applied? Preferably, our theme would automatically be applied to all bottom sheets.

app/src/main/res/values/button_themes.xml
```diff
@@ -12,7 +12,9 @@
         <item name="shapeAppearance">?shapeAppearanceLargeComponent</item>
     </style>
 
-    <style name="Hedvig.Button.Text" parent="Widget.MaterialComponents.Button.TextButton" />
+    <style name="Hedvig.Button.Text" parent="Widget.MaterialComponents.Button.TextButton">
```
hedvigoscar, line 15: Is this correct? (I don't know the answer - double-check with Zak 🙂)

app/src/main/res/layout/activity_marketing.xml
MathiasOhrstrom: yea, i thought so too but i does not seem to work, but i get a white text when on light theme even if we have `<item name="textColorButton">@color/hedvig_black</item>`

hedvigoscar: This is actually intended, check out the `ThemeOverlay` applied to the root of this Activity. Basically, this screen only has white text no matter what in accordance with a decision from design.

app/src/main/res/layout/dialog_change_start_date.xml
MathiasOhrstrom: dont know why i did that since that is what is does 🤷‍♀

PR #182, opened 2020-04-08
slimba

MathiasOhrstrom:


app/build.gradle
```diff
@@ -161,8 +161,7 @@ dependencies {
 
     implementation 'net.ypresto.timbertreeutils:timbertreeutils:1.0.0'
 
-    // Timber
-    implementation 'com.jakewharton.timber:timber:4.7.1'
```
hedvigoscar, line 165: We're still using `Timber` under the hood (just gaining some performance by calling it via `Slimber`), so lets keep this dependency around

app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -144,7 +144,7 @@ val applicationModule = module {
         if (isDebug()) {
             val logger = HttpLoggingInterceptor(object : HttpLoggingInterceptor.Logger {
                 override fun log(message: String) {
-                    Timber.tag("OkHttp").i(message)
```
hedvigoscar, line 147: These calls are not equal, we can continue using the former 🙂

app/src/main/java/com/hedvig/app/HedvigApplication.kt
```diff
@@ -95,15 +97,15 @@ class HedvigApplication : Application() {
             .from(apolloClientWrapper.apolloClient.mutate(NewSessionMutation()))
             .subscribe({ response ->
                 if (response.hasErrors()) {
-                    Timber.e("Failed to register a hedvig token: %s", response.errors().toString())
+                    e { response.errors().toString() }
```
hedvigoscar, line 100: I think we lost some log text on the way here

app/src/main/java/com/hedvig/app/feature/keygear/ui/ReceiptActivity.kt
```diff
@@ -199,7 +208,7 @@ class ReceiptActivity : BaseActivity(R.layout.activity_receipt) {
                 out.close()
                 file.absolutePath
             } catch (e: Exception) {
-                Timber.e(e, "Error saving image")
+                e { "$e Error saving image" }
```
hedvigoscar, line 211: These calls are not the same. Do the following:

```suggestion
                e(it) { "Error saving image" } 
```

app/build.gradle
MathiasOhrstrom: Timber is included in `com.github.PaulWoitaschek:Slimber:1.0.7`

hedvigoscar: I know, but if we use a dependency directly we should be listing it as a direct dependency of our project. If not, we're relying on an implementation detail of our direct dependency, which might change when they do an update.

PR #189, opened 2020-04-15
Try to prevent job cancellations from killing the app

hedvigoscar:


app/src/main/java/com/hedvig/app/util/extensions/CoroutineScopeExtensions.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.launch
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+val timberLoggingCoroutineExceptionHandler =
+    CoroutineExceptionHandler { _, throwable ->
+        if (throwable !is CancellationException) {
+            e(throwable) { "Caught potentially fatal exception in Coroutine" }
```
FredrikAreschoug, line 15: Just to be sure you will still see the exception and the stack trace when logging whit `e(throwable)`?

hedvigoscar: I believe so, yes. `e(throwable)` is functionally equivalent to `Timber.e(throwable)`, which does log the stack trace and some additional information.

hedvigoscar:
As this PR is superfluous, I'm killing it

PR #197, opened 2020-05-05
Feature/more edge to edge

MathiasOhrstrom:


app/src/main/res/layout/activity_contract_detail.xml
```diff
@@ -8,18 +8,15 @@
...
-        android:layout_height="wrap_content"
-        app:navigationIcon="@drawable/ic_close"
-        app:title="@string/CONTRACT_DETAIL_MAIN_TITLE" />
-
     <androidx.core.widget.NestedScrollView
         android:id="@+id/scrollView"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:layout_marginTop="?actionBarSize">
+        android:layout_marginTop="?actionBarSize"
```
hedvigoscar, line 15: If we're using a `ConstraintLayout` as parent, we do not need to use margins to offset the scrollview from the toolbar, we can use constraints instead

app/src/main/res/layout/activity_my_info.xml
```diff
@@ -20,11 +25,58 @@
...
+                    android:padding="@dimen/sphere_text_padding"
+                    android:textColor="?colorOnPrimary"
+                    app:autoSizeMaxTextSize="@dimen/sphere_text_max_text_size"
+                    app:autoSizeMinTextSize="@dimen/sphere_text_min_text_size"
+                    app:autoSizeStepGranularity="@dimen/sphere_text_step_granularity"
+                    app:autoSizeTextType="uniform"
+                    tools:text="Testina Testsdotteri" />
+            </FrameLayout>
+
+            <!--            <include-->
```
hedvigoscar, line 77: 👮

app/src/main/res/layout/dashboard_header.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
```
hedvigoscar, line 2: This file seems like it can easily be inlined

app/src/main/res/layout/fragment_dashboard.xml
```diff
@@ -4,7 +4,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:paddingBottom="@dimen/bottom_navigation_height"
+    android:clipToPadding="false"
```
hedvigoscar, line 7: Is this attribute entirely necessary in all the places it is in right now? Seems like it's been added in many places.

app/src/main/res/layout/hedvig_toolbar.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<androidx.appcompat.widget.Toolbar xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/hedvigToolbar"
```
hedvigoscar, line 4: This file seems like it could easily be inlined where it's used

app/src/main/res/values/styles.xml
```diff
@@ -6,6 +6,14 @@
         <item name="android:navigationBarColor">@color/translucent_background</item>
     </style>
 
+    <style name="Hedvig.Theme.LoggedIn">
+        <item name="android:navigationBarColor">@color/color_background_85</item>
```
hedvigoscar, line 10: Shouldn't the navigation bar be transparent and the views behind it inset instead?

app/src/main/AndroidManifest.xml
```diff
@@ -47,8 +47,8 @@
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
-            android:theme="@style/Hedvig.Theme"
-            android:windowSoftInputMode="adjustResize" />
+            android:theme="@style/Hedvig.Theme.Chat"
+            android:windowSoftInputMode="adjustPan" />
```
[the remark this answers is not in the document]
MathiasOhrstrom: For some reason when I use `adjustResize` hides `ChatInputView` and the toolbar when opening the keyboard and they are still off screen when the keyboard is dismissed

app/src/main/java/com/hedvig/app/feature/claims/ui/ClaimsFragment.kt
```diff
@@ -38,10 +39,14 @@ class ClaimsFragment : BaseTabFragment() {
     private val requestBuilder: RequestBuilder<PictureDrawable> by lazy { buildRequestBuilder() }
     private val baseMargin: Int by lazy { resources.getDimensionPixelSize(R.dimen.base_margin) }
 
+    private var toolbar: androidx.appcompat.widget.Toolbar? = null
     override val layout = R.layout.fragment_claims
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        toolbar = activity?.findViewById(R.id.hedvigToolbar)
```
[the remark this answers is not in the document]
MathiasOhrstrom: Is it possible to get a view from the parent activity in a fragment? When i try i get `null`

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
hedvigoscar: This is probably caused by an underlying bug in how the `ChatActivity` handles the IME showing, which re-triggers WindowInsets. The issue should be resolved at that level instead.

app/src/main/java/com/hedvig/app/feature/claims/ui/ClaimsFragment.kt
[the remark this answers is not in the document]
hedvigoscar: Generally, I think there are some architectural problems with accessing views in a parent `Activity` from a child `Fragment`: The parent `Activity` should probably have ownership of its own `View`s. Perhaps this could be resolved by using a `ViewModel` that the parent and child shares? That could facilitate the communication between the two units.

app/src/main/java/com/hedvig/app/feature/claims/ui/ClaimsFragment.kt
```diff
@@ -72,6 +77,28 @@ class ClaimsFragment : BaseTabFragment() {
         })
     }
 
+    private fun setupScrollListener() {
+        claimsNestedScrollView.setOnScrollChangeListener { _: NestedScrollView?, _: Int, scrollY: Int, _: Int, oldScrollY: Int ->
+            val dy = oldScrollY - scrollY
```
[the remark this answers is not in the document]
MathiasOhrstrom: When you scroll we get the percentage of `toolbar.height` scrolled with `claimsNestedScrollView` to set the elevation for the toolbar(So if `toolbar.height == 100` and `claimsNestedScrollView.computeVerticalScrollOffset() == 30` we have scrolled 30%) And when you have scrolled too 100% of the toolbar.heaigt the elevation is maxed out so you dont get to big of a shadow under the toolbar

app/src/main/res/layout/dashboard_header.xml
MathiasOhrstrom: This is a header for a `recyclerView` should i create the view programmatically?

app/src/main/res/layout/fragment_dashboard.xml
MathiasOhrstrom: True! I forgot to remove them while trying to figure out what the problem was before 😅

app/src/main/res/layout/dashboard_header.xml
hedvigoscar: Ah! My bad. This implementation is correct

app/src/main/res/values/styles.xml
MathiasOhrstrom: It does not seem possible to have the navigationbar completely transparent, if i use this color #00FFFFFF it increases the alpha as seen in this image bild

app/src/main/java/com/hedvig/app/feature/offer/OfferActivity.kt
```diff
@@ -51,6 +53,23 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
 
+        offerRoot.useEdgeToEdge()
+        offerRoot.setEdgeToEdgeSystemUiFlags(true)
+
+        offerToolbar.doOnApplyWindowInsets { view, insets, initialState ->
+            view.updatePadding(top = initialState.paddings.top + insets.systemWindowInsetTop)
+        }
+
+        offerHeader.doOnApplyWindowInsets { view, _, initialState ->
+            view.updatePadding(top = initialState.paddings.top + offerToolbar.measuredHeight)
```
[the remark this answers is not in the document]
MathiasOhrstrom: It is to add padding so the layout dont start under the transparent toolbar. The only other way i can see is to add manual padding to `offerheader` in XML but that seems like it could cause problems on other screen sizes

app/src/main/res/values/styles.xml
hedvigoscar: I think it should be possible to have the navigation bar transparent, although on 3-button navigation modes it usually gets some color by default (which is fine).

app/src/main/java/com/hedvig/app/feature/offer/OfferActivity.kt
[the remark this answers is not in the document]
hedvigoscar: We should be able to pad the scroll content container with `?actionBarSize` and then use a windowInsetListener to increase that padding by `systemWindowInsetTop`

app/src/main/res/values/styles.xml
MathiasOhrstrom: isn't it nicer looking if the navigationbar has the exact same color as the tab bar? since it is possible

hedvigoscar: Good point. Can you attach screenshots using this color on 3-button + gesture navigation? Also using full transparent on gesture navigation 🙂

MathiasOhrstrom: With `<item name="android:navigationBarColor">@color/color_background_85</item>`
Screenshot_20200507-100634_Hedvig Dev

With `<item name="android:navigationBarColor">@color/color_background_85</item>`
Screenshot_20200507-100714_Hedvig Dev

With `<item name="android:navigationBarColor">@color/transparent</item>`
Screenshot_20200507-100913_Hedvig Dev

hedvigoscar: Alright, that looks dope. Lets go with that

app/src/main/java/com/hedvig/app/feature/offer/OfferActivity.kt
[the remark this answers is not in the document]
MathiasOhrstrom: I now realize that the extra padding from `offerToolbar.measuredHeight` wasn't even needed 😅 Dont know why i added it

app/src/main/java/com/hedvig/app/feature/chat/ui/ChatInputView.kt
```diff
@@ -229,10 +252,12 @@ class ChatInputView : FrameLayout {
                 SingleSelectChoiceType.SELECTION -> sendSingleSelect(value)
                 SingleSelectChoiceType.LINK -> singleSelectLink(value)
             }
-
         }
 
         singleSelectContainer.addView(singleSelectButton)
+
+        singleSelectContainer.measure(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT)
+        chatRecyclerView.updatePadding(bottom = chatRecyclerViewInitialPadding + singleSelectContainer.measuredHeight - measureTextInput())
```
hedvigoscar, line 260: Does this overwrite padding we've added through system window insets?

app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearFragment.kt
```diff
@@ -33,9 +34,13 @@ class KeyGearFragment : BaseTabFragment() {
 
     private var hasSentAutoAddedItems = false
 
+    private var toolbar: androidx.appcompat.widget.Toolbar? = null
```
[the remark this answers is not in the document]
hedvigoscar: This still warrants fixing 😉

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInFragmentViewModel.kt
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.app.feature.loggedin.ui
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+abstract class LoggedInFragmentViewModel : ViewModel() {
```
hedvigoscar, line 6: This entire class can probably be a part of `LoggedInViewModel`. Also, we only need to use the `abstract`/`Impl`-pattern for `ViewModel`s when there is some behaviour that we want to be able to mock out - network calls, for example. This functionality should be the same regardless of if mocking is enabled or not, and can therefore stand on its own and needs no `abstract`/`Impl`-variant.

app/src/main/java/com/hedvig/app/feature/profile/ui/ProfileFragment.kt
```diff
@@ -34,10 +34,37 @@ class ProfileFragment : BaseTabFragment() {
 
     override val layout = R.layout.fragment_profile
 
+    private var toolbar: androidx.appcompat.widget.Toolbar? = null
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
+        toolbar = activity?.findViewById(R.id.hedvigToolbar)
```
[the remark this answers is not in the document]
hedvigoscar: 👮

app/src/main/res/layout/activity_chat.xml
```diff
@@ -9,13 +9,14 @@
     tools:context=".feature.chat.ui.ChatActivity">
 
     <FrameLayout
-        android:id="@+id/frameLayout"
+        android:id="@+id/hedvigToolbar"
         android:layout_width="match_parent"
-        android:layout_height="@dimen/chat_top_bar_height"
-        android:background="?colorSurface"
+        android:layout_height="wrap_content"
+        android:background="@color/color_transparent_chat_85"
```
hedvigoscar, line 15: This is actually potentially dangerous (and perhaps the cause of the issue) - using `ColorStateList`s as background colors is only valid on like API 28+ or something crazy high like that. What you do to remedy this is to use a 'placeholder' background that is just something like a square with white or black or whatever (we have some examples of this in the Key Gear-feature) and then use `backgroundTint` to apply it. Try that out before we go to using static colors in `colors_light`/`colors_dark` 🙂

MathiasOhrstrom: That worked 🎉

app/src/main/java/com/hedvig/app/feature/chat/ui/ChatInputView.kt
MathiasOhrstrom: Yes. But ut works as intended now as far as i can tell, after a minor fix

hedvigoscar: There might be some potential for simplification here, but lets keep it as is for now.

PR #201, opened 2020-05-11
Feature/last edge to edge

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentHistoryActivity.kt
```diff
@@ -19,16 +22,22 @@ class PaymentHistoryActivity : BaseActivity() {
...
+        toolbar.setNavigationOnClickListener {
+            onBackPressed()
+        }
+        paymentHistory.setupToolbarScrollListener(toolbar)
 
         profileViewModel.data.observe(lifecycleOwner = this) { data ->
             data?.chargeHistory?.let { chargeHistory ->
-                paymentHistory.adapter = PaymentHistoryAdapter(wrapCharges(chargeHistory))
+                paymentHistory.adapter =
+                    PaymentHistoryAdapter(listOf(ChargeWrapper.Title) + wrapCharges(chargeHistory))
```
hedvigoscar, line 40: There is some room for improvement here if you feel like it: We could make the `PaymentHistoryAdapter` not take its data in the constructor, in a manner similar to our more recent `RecyclerView.Adapter`s.

Not extremely necessary but it would be neat I suppose

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentHistoryAdapter.kt
```diff
@@ -17,6 +18,7 @@ class PaymentHistoryAdapter(
     override fun getItemViewType(position: Int) = when (items[position]) {
         is ChargeWrapper.Header -> HEADER
         is ChargeWrapper.Item -> ITEM
+        is ChargeWrapper.Title -> TITLE
```
hedvigoscar, line 21: More optional improvements: Make this `RecyclerView.Adapter` follow all our more recent patterns 🙂

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentHistoryAdapter.kt
```diff
@@ -10,21 +10,27 @@ import kotlinx.android.synthetic.main.payment_history_item.view.*
...
+
+    var items: List<ChargeWrapper> = listOf()
+
     override fun getItemCount() = items.size
 
-    override fun getItemViewType(position: Int) = when (items[position]) {
-        is ChargeWrapper.Header -> HEADER
-        is ChargeWrapper.Item -> ITEM
-        is ChargeWrapper.Title -> TITLE
+    override fun getItemViewType(position: Int) = when (position) {
```
hedvigoscar, line 19: I'm not sure that this is correct 🤔 

The previous code was arguably more correct in this case, this might accidentally be correct right now but not later 😉

hedvigoscar: The most recent best practice is to use the Layout ID of the View type for ID, then we do not have to maintain constants in the companion object 🙂

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentHistoryAdapter.kt
```diff
@@ -10,21 +10,27 @@ import kotlinx.android.synthetic.main.payment_history_item.view.*
...
+        0 -> TITLE
+        1 -> HEADER
+        else -> ITEM
     }
 
     override fun onCreateViewHolder(
         parent: ViewGroup,
         viewType: Int
     ): ViewHolder = when (viewType) {
+        TITLE -> ViewHolder.TitleViewHolder(
```
hedvigoscar, line 29: Here we can refactor the `ViewHolder` to have the responsibility of inflating its layout in the constructor, similar to how we do in other more recent `RecyclerView.Adapter`s 🙂

hedvigoscar: Also in the `onBindViewHolder`-method, we can refactor the `bind`-functionality to be a part of the `ViewHolder` as well, giving the `ViewHolder` complete ownership of both its layout and its data binding procedure.

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentHistoryActivity.kt
```diff
@@ -34,10 +34,12 @@ class PaymentHistoryActivity : BaseActivity() {
         }
         paymentHistory.setupToolbarScrollListener(toolbar)
 
+        paymentHistory.adapter = PaymentHistoryAdapter()
```
hedvigoscar, line 37: 100% on point

app/src/main/res/drawable/background_square.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
-    <solid android:color="?colorSurface" />
+    <solid android:color="#ff00ff" />
```
hedvigoscar, line 3: Is this correct?

app/src/main/res/drawable/background_bottom_nav.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/color_background_85" />
```
hedvigoscar, line 3: Nice! Have you tested this on API 28? 😉

app/src/main/res/drawable/background_square.xml
MathiasOhrstrom: Yes! As I intended for this to only be used when we have to use `colorStateList`, so that it is apparent when it is not working

app/src/main/res/drawable/background_bottom_nav.xml
MathiasOhrstrom: Tested on API 24 and 29

app/src/main/res/drawable/background_square.xml
hedvigoscar: Big brain-move right there, I like it!

app/src/main/res/drawable/background_bottom_nav.xml
hedvigoscar: 👏

PR #230, opened 2020-06-24
Forever: Piechart

hedvigoscar:


hedvigoscar:
Yes! I'll take the liberty of removing that code in a different PR though 🙂

PR #231, opened 2020-06-24
Forever: Information screen fixes

hedvigoscar:
- The incentive described in the translation is no longer hardcoded, but instead properly inserted through a placeholder
- The up-button in the toolbar now actually navigates up


hedvigoscar:
I think I'm starting to see that value of the builders, yeah. I'm available at any time 🙂

PR #236, opened 2020-06-26
Shared Test Data Builders

hedvigoscar:


hedvigoscar:
Note that the build is failing due to it using an old command to download the apollo schema, which now has to be done through `./gradlew apollo:downloadApolloSchema` instead.

PR #253, opened 2020-07-17
Forever: Edit code, submitting state

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/referrals/ui/editcode/ReferralsEditCodeActivity.kt
```diff
@@ -72,7 +75,34 @@ class ReferralsEditCodeActivity : BaseActivity(R.layout.activity_referrals_edit_
                 }
             }
         }
-        code.setText(currentCode)
+        code.setOnEditorActionListener { _, actionId, _ ->
+            if (actionId == EditorInfo.IME_ACTION_DONE) {
+                code.dismissKeyboard()
```
hedvigoscar, line 80: If submitting using IME dismisses the keyboard, then shouldn't the `Save`-button do the same?

PR #264, opened 2020-07-28
Embark: Revive branch

hedvigoscar:


hedvigoscar:
Marked as draft because I want to set up some more scaffolding before I'm done with this PR

hedvigoscar:
What I've done in the branch is essentially to refactor the tests for Embark to:

- Use the rules defined during previous development to clear apollo cache/set up an apollo mock server
- Use data builders instead of re-defining the data next to each test

I also encountered some minor bugs on the way which I fixed. For example, the behaviour for default responses was not working according to specification.

hedvigoscar:
Oh, and also:

- Add a more modern development testing screen, where all the different test data items can be tested

PR #272, opened 2020-08-11
App Nav: Home Tab, Active in Future

hedvigoscar:


hedvigoscar:
While this PR is ready for review, its upstream PR is not.

I leave it up to you whether you want to check it out now at all.

hedvigoscar:
The upstream PR is now ready for review. I recommend that you review that one first, and then handle this one.

PR #278, opened 2020-08-19
App Navigation: Home, Active

hedvigoscar:


build.gradle
```diff
@@ -44,3 +45,24 @@ allprojects {
...
     }
 }
+
+apply plugin: "build-time-tracker"
+buildtimetracker {
+    reporters {
+        csv {
+            output "build/times.csv"
+            append true
+            header false
```
MathiasOhrstrom, line 55: This looks cool! How do you the data?

MathiasOhrstrom (approved):
✅

build.gradle
hedvigoscar: It's associated with builds in the terminal, and afaik a file in the build folder as well.

PR #279, opened 2020-08-20
Home: Active + Info Cards

hedvigoscar:


app/src/main/res/values/widget_themes.xml
```diff
@@ -43,4 +43,12 @@
         <item name="colorSurface">@color/colorInfoCardSurface</item>
     </style>
 
+    <style name="ThemeOverlay.Hedvig.WarningCard" parent="">
+        <item name="colorSurface">?colorWarning</item>
+        <item name="colorPrimary">@color/hedvig_black</item>
+        <item name="android:textColorPrimary">#AB121212</item>
+        <item name="android:textColorSecondary">@color/hedvig_black</item>
```
MathiasOhrstrom, line 50: Should this be hardcoded?

MathiasOhrstrom (changes requested):

app/src/main/res/values/widget_themes.xml
hedvigoscar: 😬

PR #290, opened 2020-09-01
Fix/toolbar alpha fix

MathiasOhrstrom:


app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -62,7 +67,15 @@ class HomeFragment : Fragment(R.layout.home_fragment) {
                     }
                 }
             addItemDecoration(HomeItemDecoration())
-            addOnScrollListener(ScrollPositionListener(loggedInViewModel::onScroll))
+            addOnScrollListener(
+                ScrollPositionListener(
+                    loggedInViewModel::onScroll,
+                    viewLifecycleOwner
+                )
```
hedvigoscar, line 74: How about we do this?

```suggestion
                ScrollPositionListener(
                    { scrollPosition ->
                        scroll = scrollPosition
                        loggedInViewModel.onScroll(scrollPosition)
                    }},
                    viewLifecycleOwner
                )
```

That way we don't have to add two scroll listeners 🙂

hedvigoscar (changes requested):
Do this change for all the `ScrollPositionListener`s, when it's done we can merge 😉

PR #312, opened 2020-09-21
Fix: Weird activity recreate-related navigation behaviours

hedvigoscar:
- Opportunistic cleanup
- Add a test to ensure post sign-behaviour
- Remove some unused await-statements
- Remove some unused code
- Remove some purposeless code (?)
- The correct tab is now shown after activity recreation, and the toolbar inset is also correct after such an event


MathiasOhrstrom (approved):
🧹

PR #314, opened 2020-09-22
Feature/tooltip

MathiasOhrstrom:
Item from design: [home_tab.hint]

### Screenshots

Screenshot 1
image

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
@@ -3,6 +3,7 @@ package com.hedvig.app.feature.loggedin.ui
 import android.animation.ArgbEvaluator
 import android.content.Context
 import android.content.Intent
+import android.graphics.Color
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
@@ -206,9 +207,10 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
                         .arrowTargetMargin(-20)
                         .arrowSourceMargin(-20)
                         .position(ViewTooltip.Position.BOTTOM)
-                        .color(binding.toolbar.context.colorAttr(R.attr.colorTooltip))
-                        .textColor(binding.toolbar.context.colorAttr(R.attr.colorPrimary))
+                        .color(colorAttr(R.attr.colorTooltip))
+                        .textColor(colorAttr(R.attr.colorPrimary))
                         .text(R.string.home_tab_chat_hint_text)
+                        .withShadow(false)
                         .onDisplay {
                             setLastOpen(date.time)
                         }
--- a/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
@@ -9,7 +9,6 @@ import android.view.Menu
 import android.view.MenuItem
 import androidx.core.view.doOnLayout
 import androidx.core.view.isEmpty
-import androidx.core.view.marginTop
 import androidx.dynamicanimation.animation.FloatValueHolder
 import androidx.dynamicanimation.animation.SpringAnimation
 import androidx.dynamicanimation.animation.SpringForce
@@ -17,10 +16,7 @@ import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.observe
 import com.github.florent37.viewtooltip.ViewTooltip
 import com.hedvig.android.owldroid.type.Feature
-import com.hedvig.app.BASE_MARGIN
 import com.hedvig.app.BASE_MARGIN_DOUBLE
-import com.hedvig.app.BASE_MARGIN_OCTUPLE
-import com.hedvig.app.BASE_MARGIN_TRIPLE
 import com.hedvig.app.BaseActivity
 import com.hedvig.app.HedvigApplication
 import com.hedvig.app.R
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -54,6 +67,7 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
 
     private val binding by viewBinding(ActivityLoggedInBinding::bind)
 
+    private val date = Date()
```
hedvigoscar, line 70: We should use `LocalDate` from the `java.time.*`-API, and I don't think it needs to be a class-level field 🙂

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -178,6 +197,28 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
                 }
             }
         }
+
+        if (shouldShowTooltip(getLastOpen())) {
+            if (LoggedInTabs.fromId(binding.bottomNavigation.selectedItemId) == LoggedInTabs.HOME) {
+                Handler(Looper.getMainLooper()).postDelayed({
```
hedvigoscar, line 203: What does using `Looper.getMainLooper()` do here?

app/src/main/res/values/theme.xml
```diff
@@ -80,6 +81,7 @@
     <attr name="textColorThird" format="reference" />
     <attr name="colorPlaceholder" format="reference" />
     <attr name="colorWarning" format="reference" />
+    <attr name="colorTooltip" format="reference" />
```
hedvigoscar, line 84: I think we can access `@color/colorTooltip` directly instead, we're not gaining any particular benefits from using this indirection

hedvigoscar (changes requested):
Mostly pretty good!

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
MathiasOhrstrom: Isnt it good to have it class-level since its used in two funktions (`onCreateOptionsMenu` and `shouldShowTooltip`) so we only have one instance of it?

hedvigoscar: It's alright to have it be locally used. If we're worried about instance re-use we can perhaps pass it between the functions. Class-level fields have a tendency to make things a lot more complicated over time, and hence we should avoid them unless entirely necessary 🙂

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
MathiasOhrstrom: Nothing that was needed 😅

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -38,22 +35,23 @@ import com.hedvig.app.util.apollo.ThemedIconUrls
...
 import com.hedvig.app.util.extensions.setLastOpen
 import com.hedvig.app.util.extensions.startClosableChat
+import com.hedvig.app.util.extensions.view.performOnTapHapticFeedback
 import com.hedvig.app.util.extensions.view.show
 import com.hedvig.app.util.extensions.view.updatePadding
 import com.hedvig.app.util.extensions.viewBinding
 import dev.chrisbanes.insetter.doOnApplyWindowInsets
 import dev.chrisbanes.insetter.setEdgeToEdgeSystemUiFlags
 import e
+import kotlinx.android.synthetic.main.activity_licenses.view.*
```
hedvigoscar, line 50: 🤔

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -198,22 +195,24 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
             }
         }
 
-        if (shouldShowTooltip(getLastOpen())) {
+        val currentEpochDay = LocalDate.now().toEpochDay()
```
hedvigoscar, line 198: `java.time` has a class called `Instant` which represents epoch-time. I can recommend that you use that instead: `Instant.now()`

hedvigoscar: Nevermind! I'm wrong

hedvigoscar (approved):

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -178,6 +191,30 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
...
+
+        val currentEpochDay = LocalDate.now().toEpochDay()
+        if (shouldShowTooltip(getLastOpen(), currentEpochDay)) {
+            if (LoggedInTabs.fromId(binding.bottomNavigation.selectedItemId) == LoggedInTabs.HOME) {
+                Handler().postDelayed({
+                    binding.toolbar.performOnTapHapticFeedback()
+                    ViewTooltip
+                        .on(binding.toolbar.menu.getItem(0).actionView)
+                        .autoHide(true, 5000)
+                        .corner(30)
```
hedvigoscar, line 203: Should we perhaps use `BASE_MARGIN` here? That would make it `8dp`

PR #325, opened 2020-09-25
fix/welcome crash

MathiasOhrstrom:
Item from design: []

### Screenshots

Screenshot 1
Screenshot 2
...

### Checklist


app/src/engineering/java/com/hedvig/app/DevelopmentActivity.kt
```diff
@@ -34,6 +43,14 @@ class DevelopmentActivity : AppCompatActivity(R.layout.activity_development) {
                             .putExtra(LoggedInActivity.EXTRA_IS_FROM_ONBOARDING, true)
                     )
                 },
+                DevelopmentScreenAdapter.DevelopmentScreenItem.Row("Logged in with whats new-screen") {
+                    unloadKoinModules(listOf(whatsNewModule))
```
hedvigoscar, line 47: Move this to its own mock activity, where we can perform proper cleanup. As it stands, this would cause other mock activities to be affected.

We can do this after the fact, however, as we want to merge this asap.

hedvigoscar (approved):
Fix issues post-merge, otherwise looks good.

PR #328, opened 2020-09-29
Only include resources in our supported languages

hedvigoscar:


MathiasOhrstrom (approved):
How does this work?

hedvigoscar:
It makes it so that only resources from the listed languages will be included in the final build: https://developer.android.com/studio/build/shrink-code#unused-alt-resources

PR #330, opened 2020-10-05
[HVG-120]: Fix bottom sheet not dismissing

hedvigoscar:
Also modernize the file upload bottom sheet slightly


hedvigoscar:
[HVG-120]

MathiasOhrstrom (approved):

PR #333, opened 2020-10-13
more visible status bar icons when bottomsheet is open

MathiasOhrstrom:


app/src/main/res/values/widget_themes.xml
```diff
@@ -13,6 +13,7 @@
         <item name="bottomSheetStyle">@style/Hedvig.Theme.BottomSheet</item>
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:navigationBarColor">?colorSurface</item>
+        <item name="android:windowLightStatusBar">false</item>
```
hedvigoscar, line 16: Should this be applied regardless of light/dark theme?

For reference, it looks correct on my device when using dark theme.

MathiasOhrstrom: I think since a dark statusbar is standard on darkmode you will have light icons when a bottomsheet is open, but we get sill have dark icons when using light theme. We allways want light icons regardless of theme when a bottomsheet is opened since we put a dark fade on the background

hedvigoscar: Makes sense. If you've verified it in both theme variants, I approve!

hedvigoscar (approved):

PR #357, opened 2020-10-27
HVG-255 Feature/insurance coverage

MathiasOhrstrom:


app/src/main/res/layout/contract_detail_row.xml
```diff
@@ -23,6 +23,7 @@
         android:layout_height="wrap_content"
         android:textAlignment="textEnd"
         android:textAppearance="?textAppearanceBody1"
+        android:textColor="?colorPrimary"
```
hedvigoscar, line 26: We should use a correctly predefined style. We should challenge this in the designs.

hedvigoscar (changes requested):

hedvigoscar (approved):

PR #358, opened 2020-10-27
HVG-256 documents

MathiasOhrstrom:


app/src/main/res/layout/contract_detail_documents_document.xml
```diff
@@ -1,10 +1,45 @@
...
     xmlns:tools="http://schemas.android.com/tools"
-    style="?materialButtonContainedLargeStyle"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:layout_marginStart="@dimen/base_margin_double"
-    android:layout_marginTop="@dimen/base_margin_double"
-    android:layout_marginEnd="@dimen/base_margin_double"
-    tools:text="@string/MY_DOCUMENTS_INSURANCE_CERTIFICATE" />
+    android:layout_marginBottom="@dimen/base_margin"
+    android:layout_height="80dp">
```
hedvigoscar, line 7: Can we allow this height to naturally come from the size of the content?

app/src/main/res/layout/contract_detail_documents_document.xml
```diff
@@ -1,10 +1,45 @@
...
+            app:srcCompat="@drawable/ic_document"
+            android:paddingEnd="@dimen/base_margin_double"
+            android:contentDescription="@null" />
+
+        <TextView
+            android:id="@+id/text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:textAppearance="?textAppearanceBody1"
+            android:textColor="?colorPrimary"
```
hedvigoscar, line 28: I think we have a text style that reflects these attributes (same size as B1 but colorPrimary) - check with design, I think this was simply mislabeled as B1

hedvigoscar (changes requested):

app/src/main/res/layout/contract_detail_documents_document.xml
MathiasOhrstrom: this is just temporary. Im waiting for another text to put here

hedvigoscar: Makes sense.

app/src/main/res/layout/contract_detail_documents_document.xml
```diff
@@ -1,10 +1,58 @@
...
     xmlns:tools="http://schemas.android.com/tools"
-    style="?materialButtonContainedLargeStyle"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:layout_marginStart="@dimen/base_margin_double"
-    android:layout_marginTop="@dimen/base_margin_double"
-    android:layout_marginEnd="@dimen/base_margin_double"
-    tools:text="@string/MY_DOCUMENTS_INSURANCE_CERTIFICATE" />
+    android:layout_marginBottom="@dimen/base_margin"
+    android:layout_height="@dimen/contract_document_card">
```
hedvigoscar, line 7: I still hold that we should allow this height to come naturally from its contents, not hard-coded 😉

app/src/main/res/layout/contract_detail_documents_document.xml
```diff
@@ -1,10 +1,58 @@
...
+            app:srcCompat="@drawable/ic_document"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toBottomOf="parent"
+            android:paddingEnd="@dimen/base_margin_double"
+            android:contentDescription="@null" />
+
+        <TextView
+            android:id="@+id/text"
+            android:layout_width="wrap_content"
```
hedvigoscar, line 28: These `TextView`s need to have `0dp` width, and an `end`-constraint. Potentially also `maxLines=1`+`ellipsize=end`

hedvigoscar (changes requested):

app/src/main/res/layout/contract_detail_documents_document.xml
MathiasOhrstrom: What is the reason behind that?

hedvigoscar: Because if circumstances cause the contents to become higher than the container, then the overflow will be cut off.

PR #359, opened 2020-10-27
Feature/contract details error

MathiasOhrstrom:
Item from design: []

### Checklist


app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/ContractDetailActivity.kt
```diff
@@ -74,11 +80,26 @@ class ContractDetailActivity : BaseActivity(R.layout.contract_detail_activity) {
...
+                    content.remove()
+                    error.root.apply {
+                        show()
+                        setBackgroundColor(context.colorAttr(R.attr.colorSurface))
+                    }
+                } else {
+                    content.show()
+                    error.root.remove()
+                    result.getOrNull()?.bindTo(binding.card)
+                    startPostponedEnterTransition()
```
hedvigoscar, line 100: We might need this row also when an error occurs - if we do not, there will be a large amount of delay upon attempting to navigate to this activity when it errors

hedvigoscar (approved):
Good solution!

Approved upon fixing comments

app/src/main/res/layout/generic_error.xml
```diff
@@ -2,7 +2,7 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
+    android:layout_height="match_parent"
```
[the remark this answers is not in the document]
MathiasOhrstrom: It is so the color we apply to it does not cut off at the end of its content. And yes it still works as normal on the other screens

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/ContractDetailActivity.kt
MathiasOhrstrom: do you mean like this? 
```model.data.observe(this) { result ->
            binding.apply {
                if (result.isFailure) {
                    content.remove()
                    error.root.apply {
                        show()
                        setBackgroundColor(context.colorAttr(R.attr.colorSurface))
                    }
                } else {
                    content.show()
                    error.root.remove()
                    result.getOrNull()?.bindTo(binding.card)
                }
                startPostponedEnterTransition()
            }
        }
```

hedvigoscar: Yes!

PR #363, opened 2020-10-29
correct widget theme

MathiasOhrstrom:


app/src/main/res/values/widget_themes.xml
```diff
@@ -45,7 +45,7 @@
     </style>
 
     <style name="ThemeOverlay.Hedvig.PSA" parent="">
-        <item name="android:textColorPrimary">@color/color_hedvig_black_67</item>
+        <item name="android:textColor">@color/color_hedvig_black_67</item>
```
hedvigoscar, line 48: We should prefer to change theme attributes in these theme overlays - could it be `textColorSecondary` that needs to be overriden?

MathiasOhrstrom: That is correct 🙂

hedvigoscar (approved):

PR #380, opened 2020-11-09
HVG-331, HVG-340: Text key changes

hedvigoscar:

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -211,7 +211,7 @@ dependencies {
     implementation "dev.chrisbanes:insetter-ktx:0.2.1"
 
     // markwon
-    implementation "io.noties.markwon:core:4.2.1"
+    implementation "io.noties.markwon:core:4.6.0"
 
     // adyen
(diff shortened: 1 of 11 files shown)


app/build.gradle
```diff
@@ -252,5 +252,7 @@ lokaliseProperties.load(new FileInputStream(rootProject.file("lokalise.propertie
 lokalise {
     id lokaliseProperties['id']
     token lokaliseProperties['token']
+
+    downloadConfig {}
 }
```
MathiasOhrstrom, line 257: Can you explain what this part do?

MathiasOhrstrom (approved):
✅

app/build.gradle
hedvigoscar: Yes! It creates a `DownloadConfig`-object for configuration purposes, but does not change any of the default values.

In Kotlin-terms it would be like doing the following:

```kotlin
data class DownloadConfig(
    val exampleConfigProp: Boolean = true
)

val downloadConfig = DownloadConfig()
```

PR #384, opened 2020-11-11
Chore: Update dependencies

hedvigoscar:

### DIFF
--- a/app/src/engineering/java/com/hedvig/app/mocks/MockKeyGearValuationViewModel.kt
+++ b/app/src/engineering/java/com/hedvig/app/mocks/MockKeyGearValuationViewModel.kt
@@ -1,6 +1,7 @@
 package com.hedvig.app.mocks
 
 import android.os.Handler
+import android.os.Looper.getMainLooper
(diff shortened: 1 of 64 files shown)


app/src/engineering/java/com/hedvig/app/mocks/MockKeyGearValuationViewModel.kt
```diff
@@ -40,7 +41,7 @@ class MockKeyGearValuationViewModel : KeyGearValuationViewModel() {
                 )
             )
         )
-        Handler().postDelayed({
+        Handler(getMainLooper()).postDelayed({
             // uploadResult.postValue()
         }, 500L)
```
MathiasOhrstrom, line 46: 👀

MathiasOhrstrom (approved):
Approved after checking out the weird comment. Also why do we have so many `.postDelayed` in out mock viewModels?

hedvigoscar:
It's to simulate some network latency. If you never test with network latency, you never actually see what users will see 🙂

PR #416, opened 2020-12-10
HVG-486 validation

MathiasOhrstrom:
Item from design: []

### Checklist


testdata/src/main/java/com/hedvig/app/testdata/feature/embark/builders/TextActionBuilder.kt
```diff
@@ -5,20 +5,31 @@ import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
...
-                ),
-                placeholder = placeholder
+                )
             )
         ),
         asEmbarkTextActionSet = null,
     )
+
+    companion object {
+        const val PERSONAL_NUMBER = "PersonalNumber"
```
hedvigoscar, line 28: Are we being delivered these masks as `String`s? 😞 

We might want to update the API in that case, these are really `enum`s in disguise 😬

hedvigoscar (changes requested):
Looks pretty good! A few things to address, however. 😉

testdata/src/main/java/com/hedvig/app/testdata/feature/embark/builders/TextActionBuilder.kt
MathiasOhrstrom: I asked Palmen if we should have these as enums instead but i think there was a reason they are strings but i dont remember why

hedvigoscar: Let's get some first-hand information! @palmenhq

palmenhq: We don't really have enum functionality in XML (which backs embark) so i don't think it makes sense to mask them as enums. If we do we would be to explode the server if a content editor adds something different than a valid enum value and i think it's just better (also for forward/backward compatibility reasons) to let them be strings and ignore any "invalid" value.

hedvigoscar: Well, the set of valid values on this string is in effect an enum, right? GraphQL-code gen generally takes into account that the valid enum values might change in the future and hence generates an '__UNKNOWN'-case for the enum.

To me, this seems like an example of a Stringly typed API, which I'm not in favour of 😕

palmenhq: Right. Although i don't fully agree i can't really say i have a strong opinion either :) feel free to open a pr in HedvigInsurance/embark or add it to the web backlog if you feel strongly for enums. I don't think we can put this as high prio from a web perspective with project bolt ongoing unfortunately :/

hedvigoscar (approved):

PR #427, opened 2021-01-08
HVG-742 removed system default

MathiasOhrstrom:


app/src/main/res/values/arrays.xml
```diff
@@ -11,32 +11,26 @@
         <item>system_default</item>
     </string-array>
     <string-array name="language_settings">
-        <item>@string/SETTINGS_LANGUAGE_SYSTEM_DEFAULT</item>
```
hedvigoscar, line 14: What happens if a user has `System default` as their setting and enter the settings screen? 🤔

app/src/main/java/com/hedvig/app/feature/settings/Language.kt
```diff
@@ -118,13 +114,19 @@ enum class Language {
...
-                    ?: SETTING_SYSTEM_DEFAULT
-            )
+            val sharedPref = PreferenceManager.getDefaultSharedPreferences(context)
+            val market = context.getMarket()
+            if (market != null) {
+                val selectedLanguage = sharedPref
+                    .getString("language", getAvailableLanguages(market).first().toString())
+                    ?: getAvailableLanguages(market).first().toString()
+                return if (selectedLanguage == SETTING_SYSTEM_DEFAULT) {
+                    from(getAvailableLanguages(market).first().toString())
```
hedvigoscar, line 124: Nice!

hedvigoscar (commented):
Reasonable implementation, I have one concern that I'd like us to investigate before merging, however.

app/src/main/res/values/arrays.xml
MathiasOhrstrom: Good catch! It says `Inget angett` and nothing is selected in the dialog

hedvigoscar: Perhaps we could, in `HedvigApplication::onCreate`, migrate these members to the 'default setting' that you added in `Language.kt`? 😉

app/src/main/java/com/hedvig/app/feature/settings/SettingsActivity.kt
```diff
@@ -37,9 +38,21 @@ import org.koin.android.viewmodel.ext.android.viewModel
 
 class SettingsActivity : BaseActivity(R.layout.activity_settings) {
     private val binding by viewBinding(ActivitySettingsBinding::bind)
+    private val marketProvider: MarketProvider by inject()
 
+    @SuppressLint("ApplySharedPref")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        val previousLanguage = PreferenceManager.getDefaultSharedPreferences(this).getString(SETTING_LANGUAGE, null)
```
hedvigoscar, line 47: Couldn't we do this in `HedvigApplication::onCreate`? 🤔 That way, everyone will get migrated from the legacy setting to a supported setting, right? 🤔

MathiasOhrstrom: That is much better, ill fix it

PR #434, opened 2021-01-14
HVG-395 Generic error screen

MathiasOhrstrom:
### Checklist


app/build.gradle.kts
```diff
@@ -276,7 +276,7 @@ dependencies {
     androidTestImplementation("androidx.test.espresso:espresso-core:3.3.0")
     androidTestImplementation("androidx.test:runner:1.3.0")
     androidTestImplementation("androidx.test:rules:1.3.0")
-    androidTestImplementation("androidx.test.ext:junit:1.1.2")
+    androidTestImplementation("androidx.test.ext:junit-ktx:1.1.2")
```
hedvigoscar, line 279: 🤔

hedvigoscar (commented):
Need to check the import thing out, otherwise this is fine I think

app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearViewModel.kt
```diff
@@ -14,16 +14,40 @@ import kotlinx.coroutines.flow.onEach
...
+
+    init {
+        load()
+    }
+
+    override fun load() {
+        viewModelScope.launch {
+            repository
+                .keyGearItems()
+                .onEach { response ->
```
hugokallstrom, line 37: It would be cool to throw an error in the repository if response.errors != null, instead of handling errors in two places here

PR #446, opened 2021-01-21
Attempt to repair tests failing in CI

hedvigoscar:


hedvigoscar:
Looks like we can get test stability back, but we're sacrificing two misbehaving tests in the process.

I leave it up to you if you want us to investigate the issues further, I'm pretty ok with letting these tests go, though.

hugokallstrom:
> Looks like we can get test stability back, but we're sacrificing two misbehaving tests in the process.
> 
> I leave it up to you if you want us to investigate the issues further, I'm pretty ok with letting these tests go, though.

Lets sacrifice them!

hugokallstrom (approved):

PR #448, opened 2021-01-22
Start using the right mutation for tokenizing payout details

hedvigoscar:
### Checklist


hugokallstrom:
Cool! Just read some of the DropInService.makePaymentsCall() docs and quite dont understand this bit: 

```
* This call is expected to be synchronous, as it already runs in the background, and the base class will handle messaging with the UI after it
* finishes based on the [CallResult]. If you want to make the call asynchronously, return [CallResult.ResultType.WAIT] on the type and call the
* [asyncCallback] method afterwards.
```

Can running coroutines inside that function give some unexpected behaviour?

hedvigoscar:
It could, had it not been for `runBlocking`, which converts the coroutine execution into a blocking operating 😉

PR #449, opened 2021-01-22
HVG-832 Refactor marketing navigation

hugokallstrom:
I have probably broken something here, not really sure how this is supposed to work. Thought it made more sense to update language/locale when pressing continue instead of selecting an item from the bottom sheet. But please shout if this is not the case! 

### Checklist


app/src/main/java/com/hedvig/app/feature/marketpicker/MarketPickerFragment.kt
```diff
@@ -8,29 +8,32 @@ import com.hedvig.app.databinding.FragmentMarketPickerBinding
...
 import com.hedvig.app.util.extensions.view.updateMargin
-import com.hedvig.app.util.extensions.viewBinding
 import dev.chrisbanes.insetter.doOnApplyWindowInsets
 import org.koin.android.ext.android.inject
 import org.koin.android.viewmodel.ext.android.sharedViewModel
 
 class MarketPickerFragment : Fragment(R.layout.fragment_market_picker) {
     private val viewModel: MarketPickerViewModel by sharedViewModel()
     private val marketingViewModel: MarketingViewModel by sharedViewModel()
-    private val binding by viewBinding(FragmentMarketPickerBinding::bind)
```
hedvigoscar, line 20: What's going on here? 🤔

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketSelectedFragment.kt
```diff
@@ -47,10 +48,18 @@ class MarketSelectedFragment : Fragment(R.layout.fragment_market_selected) {
                     setImageDrawable(context.compatDrawable(market.flag))
                 }
                 setHapticClickListener {
-                    viewModel.navigateTo(
-                        CurrentFragment.MARKET_PICKER,
-                        signUp to "marketButton"
-                    )
+                    val transaction = parentFragmentManager.beginTransaction()
```
hedvigoscar, line 51: Are we mixing navigation/back stack management strategies in this screen now? 🤔

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketPickerViewModel.kt
```diff
@@ -18,43 +18,31 @@ import com.hedvig.app.shouldOverrideFeatureFlags
...
             sharedPreferences.edit()
-                .putString(
-                    Market.MARKET_SHARED_PREF,
-                    data.market?.name
-                )
-                .commit()
-
-            sharedPreferences
-                .edit()
+                .putString(Market.MARKET_SHARED_PREF, data.market?.name)
```
hedvigoscar, line 39: I like that we're batching this edit together - perhaps we could improve it even further by using `core-ktx`: https://developer.android.com/reference/kotlin/androidx/core/content/package-summary#(android.content.SharedPreferences).edit(kotlin.Boolean,%20kotlin.Function1)

which would make this something like this:


```kotlin
sharedPreferences.edit(commit=true) { // putString twice }
```

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketPickerViewModel.kt
```diff
@@ -18,43 +18,31 @@ import com.hedvig.app.shouldOverrideFeatureFlags
...
-import com.hedvig.app.util.extensions.getStoredBoolean
 import kotlinx.coroutines.launch
 
 abstract class MarketPickerViewModel(private val context: Context) : ViewModel() {
     protected val _data = MutableLiveData<PickerState>()
     val data: LiveData<PickerState> = _data
     abstract fun uploadLanguage()
 
+    fun submitLanguageAndReload() {
+        uploadLanguage()
```
hedvigoscar, line 29: Let's investigate if this is the approach we want to take - I've got some additional context on why this thing was built this way 😬

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketPickerFragment.kt
hugokallstrom: Forgot to add this back after removing because of build errors, fixing

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketPickerViewModel.kt
hugokallstrom: Brilliant!

app/src/main/java/com/hedvig/app/feature/marketpicker/MarketSelectedFragment.kt
hugokallstrom: You are right, should navigate via view model here as well, fixing!

PR #452, opened 2021-01-28
Fix text action set default result not strictly speaking working

hedvigoscar:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -32,6 +32,7 @@ import com.hedvig.app.feature.embark.EmbarkViewModel
 import com.hedvig.app.feature.embark.EmbarkViewModelImpl
 import com.hedvig.app.feature.embark.passages.previousinsurer.PreviousInsurerViewModel
 import com.hedvig.app.feature.embark.passages.previousinsurer.PreviousInsurerViewModelImpl
+import com.hedvig.app.feature.embark.passages.textactionset.TextActionSetParameter
 import com.hedvig.app.feature.embark.passages.textactionset.TextActionSetViewModel
 import com.hedvig.app.feature.home.data.HomeRepository
 import com.hedvig.app.feature.home.service.HomeTracker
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/hedvig/app/feature/embark/passages/TextActionSetViewModel.kt
```diff
@@ -8,9 +8,28 @@ class TextActionSetViewModel : ViewModel() {
     private val _isValid = MutableLiveData<HashMap<Int, Boolean>>(hashMapOf())
     val isValid: LiveData<HashMap<Int, Boolean>> = _isValid
 
+    private val _inputs = MutableLiveData<Map<Int, String>>()
+    val inputs: LiveData<Map<Int, String>> = _inputs
+
```
hugokallstrom, line 13: Would it be possible to pass TextActionSetData as constructor parameter to the view model? I know it was difficult with dagger :(

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/embark/passages/TextActionSetViewModel.kt
hedvigoscar: Kind of easy actually: https://insert-koin.io/docs/reference/koin-android/viewmodel#viewmodel-and-injection-parameters

hedvigoscar: Also, I think that use-case is made easier through Hilt 😉

hugokallstrom: Cool! Should be able to pass `parameters: ParametersDefinition` to `Fragment.viewModel()` and then specify the type in the koin module?

hedvigoscar: Yes, there's good potential here, let me investigate 🙂

hedvigoscar: It's glorious!

PR #459, opened 2021-02-03
Market context dependency refactor

hugokallstrom:
- Add MarketManager to avoid dependency on context for tests

### Checklist


app/src/androidTest/java/com/hedvig/app/MarketTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.hedvig.app.feature.settings.MarketManager
+import com.hedvig.app.util.KoinMockModuleRule
+import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.Before
+import org.junit.Rule
+import org.koin.dsl.module
+
+abstract class MarketTest(val market: Market) : TestCase() {
```
hedvigoscar, line 13: It's probably more appropriate to make this a `Rule`, as they compose better 🙂

app/src/androidTest/java/com/hedvig/app/feature/payment/FreeMonthsCampaignTest.kt
```diff
@@ -60,7 +61,7 @@ class FreeMonthsCampaignTest : TestCase() {
                                 .fragments
                                 .monetaryAmountFragment
                                 .toMonetaryAmount()
-                                .format(context())
+                                .format(context(), Market.SE)
```
hedvigoscar, line 64: I fear that this might be incorrect in cases where the `MarketManager.market` returns something different than `Market.SE`. Perhaps we could have a function that loads the currently loaded market from the `MarketManager`? Given that we're working with test code, it doesn't have to be super-pretty, we can probably even do something like this:

```kotlin
inline fun market() = getKoin().get<MarketManager>().market
```

hedvigoscar (commented):
Looks mostly good, but we could do with a few tweaks to the test code wrt `Market`.

Given how hard this PR has been neglected by me, I can take it upon myself to implement the changes, if you'd like. Wdyt?

hugokallstrom:
> Looks mostly good, but we could do with a few tweaks to the test code wrt `Market`.
> 
> Given how hard this PR has been neglected by me, I can take it upon myself to implement the changes, if you'd like. Wdyt?

No problem, its valuable for me to learn more about testing anyways :) Fixing the comments!

hedvigoscar (approved):
Looks good, approved after formatting the code in the PR with `ktlint` (don't forget 😉 )

PR #460, opened 2021-02-03
HVG-867: Remove facebook sdk

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/offer/OfferTracker.kt
```diff
@@ -26,10 +22,6 @@ class OfferTracker(
     fun userDidSign(price: Double) {
```
hedvigoscar, line 22: Looks like this method might not even be necessary anymore 😬

hedvigoscar: Unless we feel like logging it to mixpanel 🤔 which we might

hedvigoscar (approved):

PR #465, opened 2021-02-04
HVG-891: Handle masks when loading values from store

hedvigoscar:
- Remask values loaded from store correctly
- Opportunistic refactor: Migrate this test to `TextActionScreen`

### Checklist


hugokallstrom (approved):
Excellent! ;)

PR #474, opened 2021-02-12
HVG-941: computed store values

hugokallstrom:
Adds support for computed store values in embark. 

Whats not included: 
Support for computed store values in responses. Please point me in the right direction on where I should implement this, `preProcessResponse` perhaps? Thanks

I also had problems verifying a `KTextInputLayout` - the test framework could not find the view for some reason. Please take a look in `ComputedValuesTest` 🙏 

### Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -273,6 +273,8 @@ dependencies {
     implementation("com.facebook.shimmer:shimmer:0.5.0")
 
     // test
+    testImplementation("junit:junit:4.13")
+    testImplementation("com.google.truth:truth:1.1.2")
     androidTestImplementation("androidx.test.espresso:espresso-core:3.3.0")
     androidTestImplementation("androidx.test:runner:1.3.0")
     androidTestImplementation("androidx.test:rules:1.3.0")
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -274,7 +274,6 @@ dependencies {
 
     // test
     testImplementation("junit:junit:4.13")
-    testImplementation("com.google.truth:truth:1.1.2")
     androidTestImplementation("androidx.test.espresso:espresso-core:3.3.0")
     androidTestImplementation("androidx.test:runner:1.3.0")
     androidTestImplementation("androidx.test:rules:1.3.0")
@@ -285,6 +284,7 @@ dependencies {
     androidTestImplementation(
         "com.apollographql.apollo:apollo-idling-resource:${Dependencies.Versions.apollo}"
     )
+    testImplementation("com.willowtreeapps.assertk:assertk-jvm:0.22")
     androidTestImplementation("com.willowtreeapps.assertk:assertk-jvm:0.22")
     androidTestImplementation("org.koin:koin-test:$koin_version")
     androidTestImplementation("org.awaitility:awaitility:4.0.2")
--- a/app/src/androidTest/java/com/hedvig/app/feature/embark/computedvalues/ComputedValuesTest.kt
+++ b/app/src/androidTest/java/com/hedvig/app/feature/embark/computedvalues/ComputedValuesTest.kt
@@ -1,6 +1,8 @@
 package com.hedvig.app.feature.embark.computedvalues
 
+import com.agoda.kakao.screen.Screen
 import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
+import com.hedvig.app.feature.embark.screens.EmbarkScreen
 import com.hedvig.app.feature.embark.screens.NumberActionScreen
 import com.hedvig.app.feature.embark.ui.EmbarkActivity
 import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_COMPUTED_VALUE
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -41,17 +42,25 @@ abstract class EmbarkViewModel(
     protected lateinit var storyData: EmbarkStoryQuery.Data
 
     private val store = HashMap<String, String>()
+    private lateinit var computedValues: Map<String, String>
```
hedvigoscar, line 45:
```suggestion
    private val computedValues: HashMap<String, String> = hashMap()
```

We can populate it when we set the initial state, later on 🙂

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -501,7 +523,7 @@ abstract class EmbarkViewModel(
 
 class EmbarkViewModelImpl(
     private val embarkRepository: EmbarkRepository,
-    tracker: EmbarkTracker,
+    tracker: EmbarkTracker
```
hedvigoscar, line 526:
```suggestion
    tracker: EmbarkTracker,
```

We should check your code style settings, it's weird that it's eating the trailing commas 😧

app/src/main/java/com/hedvig/app/feature/embark/computedvalues/TemplateExpressionCalculator.kt
```diff
@@ -0,0 +1,186 @@
+package com.hedvig.app.feature.embark.computedvalues
+
+object TemplateExpressionCalculator {
```
hedvigoscar, line 3: 🤯 kind of jealous that I didn't choose to implement this, ngl

app/build.gradle.kts
```diff
@@ -273,6 +273,8 @@ dependencies {
     implementation("com.facebook.shimmer:shimmer:0.5.0")
 
     // test
+    testImplementation("junit:junit:4.13")
+    testImplementation("com.google.truth:truth:1.1.2")
```
hedvigoscar, line 277: Should we check out using `assertk` over `truth`? The former seems to leverage Kotlin's features a bit more, wdyt?

app/src/test/java/com/hedvig/app/ComputedValuesTest.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.google.common.truth.Truth.assertThat
+import com.hedvig.app.feature.embark.computedvalues.TemplateExpressionCalculator
+import org.junit.Test
+
+class ComputedValuesTest {
+
+    private val emptyStore = hashMapOf<String, String>()
+
+    @Test
+    fun testSubstitutingANumber() {
```
hedvigoscar, line 12: Not necessary for merging, but when doing JDK tests with Kotlin, you can use the following fantastic syntax:

```suggestion
    fun `test substituting a number`() {
```

hedvigoscar: Also, given that this is a function in a test-suite more or less, I think having `test` in the name might be redundant 😄

app/src/androidTest/java/com/hedvig/app/feature/embark/computedvalues/ComputedValuesTest.kt
```diff
@@ -0,0 +1,51 @@
...
+            step("Enter value in first passage and submit") {
+                input {
+                    edit {
+                        typeText("1334")
+                    }
+                }
+                submit { click() }
+            }
+
+            step("Verify that computed value has been correctly parsed") {
```
hedvigoscar, line 42: Does this step make sense? When you submit the previous submit, will there not be an animation removing this passage? 😧

hedvigoscar (commented):
Really nice! A few things to address. I added a comment re: what I think might be wrong with the `KTextInputLayout`-thing.

re: interpolation for responses, I think if we simply replace the direct call to `store[...]` with `getFromStore(...)`, it should resolve our problems in a satisfactory manner 🙂

app/build.gradle.kts
```diff
@@ -273,6 +273,8 @@ dependencies {
     implementation("com.facebook.shimmer:shimmer:0.5.0")
 
     // test
+    testImplementation("junit:junit:4.13")
```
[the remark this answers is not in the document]
hugokallstrom: Added a gradle task in the yml config file

app/build.gradle.kts
hugokallstrom: Looked into `assertk`, its nice that they leverage Kotlin more but they are at version `0.23.1` at the moment while `truth` is at `1.1.2`. `truth` is also the framework [suggested](https://developer.android.com/training/testing/unit-testing/local-unit-tests) and built by Google which I think is a plus. And they also seem to support the framework well. 

I propose we run with `truth` for now, but if we feel that its too clunky with Kotlin we can change to `assertk`, sounds reasonable?

hedvigoscar: I think `assertk` is actually stable, despite the `0.*`-version 😉 Also, we already have `assertk` in the project, I think it could be valuable to not diverge on too many assertion-solutions 😄

hugokallstrom: Ah wasnt aware we already used `assertk`. Ill remove `truth`!

hedvigoscar (approved):

PR #475, opened 2021-02-12
HVG-932: Embark keyboard next

hugokallstrom:
Add listener for IME action done and navigate to next passage when pressed in `TextAction`, `TextActionSet` and `NumberAction`.

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/passages/numberaction/NumberActionFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/passages/numberaction/NumberActionFragment.kt
@@ -2,6 +2,8 @@ package com.hedvig.app.feature.embark.passages.numberaction
 
 import android.os.Bundle
 import android.view.View
+import android.view.inputmethod.EditorInfo
+import android.widget.TextView
 import androidx.core.os.bundleOf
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/hedvig/app/feature/embark/passages/numberaction/NumberActionFragment.kt
```diff
@@ -37,25 +40,37 @@ class NumberActionFragment : Fragment(R.layout.number_action_fragment) {
             input.doOnTextChanged { text, _, _, _ ->
                 numberActionViewModel.validate(text)
             }
+            input.setOnEditorActionListener(TextView.OnEditorActionListener { _, actionId, _ ->
```
hedvigoscar, line 43: Could we create an extension that generifies this? Something like

```kotlin
fun TextView.onImeAction(action: () -> Unit) = setOnEditorActionListener(TextView.OnEditorActionListener { _, actionId, _ ->
    if (actionId == EditorInfo.IME_ACTION_DONE) {
        action()
        return@OnEditorActionListener true
    }
    false
}
```

hedvigoscar (approved):
Approved after applying suggestion 🙂

PR #479, opened 2021-02-25
HVG-967: Add Apollo Studio-headers to GraphQL requests

hedvigoscar:


hugokallstrom (approved):

hugokallstrom:
Why do we need these headers?

hedvigoscar:
They will provide us with some metrics to segment on in Apollo Studio, which gives us some insight into how requests perform, amount of deprecated queries used etc.

PR #482, opened 2021-03-03
APP-29: Add change address button in home tab and only show in debug builds

hugokallstrom:
This PR adds a "Make changes" header together with a "Change your address" button in the home tab. Clicking the button should open the moving flow intro views (not in the scope of this PR). 

A new enum class `FeatureFlag` is introduced in order to hide/show features. This will make it possible to make PRs with develop as the base branch, instead of having long running feature branches (such as embark) where rebasing is required regularly. For now we are enabling the feature depending on build type. But functionality can be extended so that features can be toggled in the app, via backend etc. 

Not contained in this PR: String resources from Lokalise. 

### Checklist


app/src/androidTest/java/com/hedvig/app/feature/home/ChangeAddressButton.kt
```diff
@@ -0,0 +1,47 @@
+package com.hedvig.app.feature.home
```
hedvigoscar, line 1: Make sure to format your files before committing them 🙂

app/src/main/java/com/hedvig/app/util/FeatureFlag.kt
```diff
@@ -0,0 +1,17 @@
...
+    val key: String,
+    val title: String,
+    val explanation: String,
+    val enabled: Boolean = false,
+) {
+    MOVING_FLOW(
+        "moving_flow",
+        "Moving Flow",
+        "Lets a user change their address and get a new offer",
+        BuildConfig.BUILD_TYPE == "debug"
```
hedvigoscar, line 15: You can also use `isDebug()` from `ApplicationModule.kt`, which would enable this button in both dev- and staging-builds 🙂

app/src/main/res/layout/home_change_address.xml
```diff
@@ -0,0 +1,31 @@
...
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:app="http://schemas.android.com/apk/res-auto"
+        android:id="@+id/change_address_layout"
+        android:layout_height="wrap_content"
+        android:layout_width="match_parent"
+        android:layout_marginTop="@dimen/base_margin_sextuple"
+        android:layout_marginBottom="@dimen/base_margin_triple"
+        android:orientation="vertical">
+
+    <TextView
```
hedvigoscar, line 11: I've been thinking about making one generic H5-item layout to represent these 'Title 2'-items from the designs, and reusing that. Under that scheme, we could return one 'Title 2'-item with the appropriate text, and have a separate item for the 'Change your address'-thing. This would also remove the wrapping `LinearLayout`. What do you think about that? 🙂

app/src/androidTest/java/com/hedvig/app/feature/home/ChangeAddressButton.kt
```diff
@@ -0,0 +1,47 @@
...
+    val mockServerRule = ApolloMockServerRule(
+        LoggedInQuery.QUERY_DOCUMENT to apolloResponse { success(LOGGED_IN_DATA) },
+        HomeQuery.QUERY_DOCUMENT to apolloResponse { success(HOME_DATA_ACTIVE) },
+    )
+
+    @get:Rule
+    val apolloCacheClearRule = ApolloCacheClearRule()
+
+    @Test
+    fun shouldShowChangeAddressButton() {
```
hedvigoscar, line 34:
```suggestion
    fun shouldShowChangeAddressButton() = run {
```

app/src/androidTest/java/com/hedvig/app/feature/home/ChangeAddressButton.kt
hugokallstrom: Damn, sorry. Would it be valuable to automate this somehow?

hedvigoscar: I'm using Save Actions, but it's probably also possible to add some kind of pre-commit hook? 🤔

app/src/main/res/layout/home_change_address.xml
hugokallstrom: Sounds good!

hedvigoscar (approved):

PR #484, opened 2021-03-04
Feature/instant app

hugokallstrom:
### Checklist


onboarding/src/engineering/java/com/hedvig/onboarding/mocks/MockEmbarkActivity.kt
```diff
@@ -147,127 +149,127 @@ class EmbarkMockActivity : MockActivity() {
                 shouldLoad = true
                 mockedData = STORY_WITH_INCOMPATIBLE_ACTION
             }
-            startActivity(EmbarkActivity.newInstance(this@EmbarkMockActivity, this.javaClass.name))
+            startActivity(EmbarkActivity.newInstance(this@MockEmbarkActivity, this.javaClass.name))
```
hedvigoscar, line 152: These can generically be replaced with `context`, but whatever 🤷

onboarding/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,113 @@
...
+
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="http" />
+                <data
+                    android:host="instantapptest.dev.hedvigit.com"
```
hedvigoscar, line 37: Probably need to disable these 🙂

onboarding/src/main/java/com/hedvig/onboarding/chooseplan/OnboardingAdapter.kt
```diff
@@ -27,24 +19,26 @@ import com.hedvig.app.util.extensions.view.setHapticClickListener
...
-            throw Error("Unreachable")
-        }
+        com.hedvig.app.R.layout.generic_error -> ViewHolder.Error(parent)
+        else -> throw Error("Unreachable")
     }
 
     override fun getItemViewType(position: Int) = when (getItem(position)) {
         is OnboardingModel.Bundle -> R.layout.plan_card
-        OnboardingModel.Error -> R.layout.generic_error
+        OnboardingModel.Error -> com.hedvig.app.R.layout.generic_error
```
hedvigoscar, line 41: Is this needed? If not, lets do this to remove all the prefixes:

```suggestion
        OnboardingModel.Error -> R.layout.generic_error
```

hugokallstrom: Yeah that resource is in the app module so we need to specify the whole path

onboarding/src/main/AndroidManifest.xml
hugokallstrom: We are still using these for navigation, like so:

```
private const val DEFAULT_URL = "https://instantapptest.dev.hedvigit.com"

fun Activity.startActivityWithAppLink(appLink: AppLink) {
    val uri = getUri(appLink)
    val intent = Intent(Intent.ACTION_VIEW, uri)
    intent.addCategory(Intent.CATEGORY_BROWSABLE)
    startActivity(intent)
}

private fun getUri(appLink: AppLink): Uri {
    return appLink.query?.let {
        Uri.parse("$DEFAULT_URL/${appLink.path}?${it.first}=${it.second}")
    } ?: Uri.parse("$DEFAULT_URL/${appLink.path}")
}
```

I don't think we will be able to open these activities externally anyway since the appLinks.json isnt set up, or am I missing something?

hedvigoscar: We won't be able to open them externally when the app is not installed, but we will do so when it is installed.

Perhaps we can transition them to be on a proper hedvig.com-domain?

test/testutil/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    implementation(kotlin("stdlib", Dependencies.Versions.kotlin))
+
+    // Okhttp
+    val okhttp3_version = "4.9.1"
+    implementation("com.squareup.okhttp3:mockwebserver:$okhttp3_version")
+
+    // Koin
+    val koin_version = "2.2.2"
+    implementation("org.koin:koin-android:$koin_version")
+    implementation("org.koin:koin-test:$koin_version")
```
hedvigoscar, line 27: Perhaps we can `buildSrc` these version numbers 🙂

hedvigoscar:
RIP for now.

PR #485, opened 2021-03-05
APP-147: Animate embark keyboard

hugokallstrom:
Add inset listeners for ime keyboard. Unfortunately this only works for Android 11, lets hope they will backport it. Added util function for hiding keyboard, this also only works for Android 11. 

Finally used delays to wait between animations. 

### Checklist


hedvigoscar (approved):
Tests no longer seem to run, let's address that :slightly_smiling_face:

Other than that, it's DAMN GOOD! 🚀

PR #493, opened 2021-03-12
APP-10: Moving flow intro

hugokallstrom:
This PR contains the view shown when entering the moving flow (https://www.figma.com/file/Ds2hXzDbnUt9ZNhg4HSWH3/Moving-Flow-Android?node-id=94%3A0). 

We first check if the user has any upcoming agreement changes by querying `upcomingAgreementChange` (see `upcomingAgreement.graphql`). This is done in `GetUpcomingAgreementUseCase`, where we return a `UpcomingAgreementResult` with data suitable for the view to use. We populate the view with data from this query if an upcoming agreement is found.

If no upcoming agreement is found - we query `selfChangeEligibility` (`selfChangeEligibility.graphql`). If we receive blockers we show an error state. Otherwise we let the user continue with the moving flow by opening embark. 

Not present in this PR: 
* Starting embark when pressing "Continue to questions"
* Mock activity for testing in engineering mode (WIP)
* Some string resources missing for error states

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
@@ -4,6 +4,8 @@ import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.transition.TransitionManager
+import androidx.annotation.DrawableRes
+import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
 import com.hedvig.app.BaseActivity
 import com.hedvig.app.R
@@ -63,11 +65,15 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
     }
 
     private fun setViewState(viewState: ViewState): Any = when (viewState) {
-        Loading -> binding.spinner.loadingSpinner.show()
+        Loading -> {
+            binding.spinner.loadingSpinner.show()
+            binding.contentScrollView.remove()
+        }
         SelfChangeAddress -> setContent(
             titleText = getString(R.string.moving_intro_title),
             subtitleText = getString(R.string.moving_intro_description),
             buttonText = getString(R.string.moving_intro_open_flow_button_text),
+            buttonIcon = null,
             // TODO: Start Embark
             onContinue = { }
         )
@@ -75,12 +81,14 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
             titleText = getString(R.string.moving_intro_title),
             subtitleText = getString(R.string.moving_intro_manual_handling_description),
             buttonText = getString(R.string.moving_intro_manual_handling_button_text),
+            buttonIcon = R.drawable.ic_chat_white,
             onContinue = { openChat() }
         )
         is ChangeAddressInProgress -> setUpcomingChangeContent(
             titleText = getString(R.string.moving_intro_existing_move_title),
             subtitleText = getString(R.string.moving_intro_existing_move_description),
             buttonText = getString(R.string.moving_intro_manual_handling_button_text),
+            buttonIcon = R.drawable.ic_chat_white,
             onContinue = { openChat() },
             viewState.upcomingAgreementResult
         )
@@ -91,12 +99,14 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
                 is GeneralError -> viewState.error.message ?: "Could not continue, please try again later"
             },
             buttonText = "Try again",
+            buttonIcon = null,
             onContinue = { viewModel.reload() }
         )
         is SelfChangeError -> setContent(
             titleText = getString(R.string.error_dialog_title),
             subtitleText = viewState.error.message ?: "Could not continue, please try again later",
             buttonText = "Try again",
+            buttonIcon = null,
             onContinue = { viewModel.reload() }
         )
     }
@@ -107,9 +117,11 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
         titleText: String,
         subtitleText: String,
         buttonText: String?,
+        @DrawableRes buttonIcon: Int?,
         onContinue: () -> Unit
     ) = with(binding) {
         spinner.loadingSpinner.remove()
+        contentScrollView.show()
         image.show()
         title.text = titleText
         title.show()
@@ -117,6 +129,7 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
         subtitle.show()
         continueButton.text = buttonText
         continueButton.setOnClickListener { onContinue() }
+        continueButton.icon = buttonIcon?.let { ContextCompat.getDrawable(this@ChangeAddressActivity, it) }
         continueButton.isVisible = buttonText != null
     }
 
@@ -124,10 +137,12 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
         titleText: String,
         subtitleText: String,
         buttonText: String?,
+        @DrawableRes buttonIcon: Int?,
         onContinue: () -> Unit,
         upcomingAgreementResult: UpcomingAgreement
     ) = with(binding) {
         spinner.loadingSpinner.remove()
+        contentScrollView.show()
         image.remove()
 
         title.text = titleText
@@ -138,6 +153,7 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
         continueButton.text = buttonText
         continueButton.setOnClickListener { onContinue() }
         continueButton.isVisible = buttonText != null
+        continueButton.icon = buttonIcon?.let { ContextCompat.getDrawable(this@ChangeAddressActivity, it) }
 
         upcomingAddressLayout.upcomingAddressLayoutRoot.show()
         upcomingAddressLayout.addressLabel.text = upcomingAgreementResult.address.street
--- a/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
@@ -3,19 +3,35 @@ package com.hedvig.app.feature.home.ui.changeaddress
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.transition.TransitionManager
+import androidx.core.view.isVisible
 import com.hedvig.app.BaseActivity
 import com.hedvig.app.R
 import com.hedvig.app.databinding.ChangeAddressActivityBinding
-import com.hedvig.app.util.extensions.view.setHapticClickListener
+import com.hedvig.app.feature.chat.ui.ChatActivity
+import com.hedvig.app.feature.home.ui.changeaddress.ChangeAddressViewModel.ViewState
+import com.hedvig.app.feature.home.ui.changeaddress.ChangeAddressViewModel.ViewState.ChangeAddressInProgress
+import com.hedvig.app.feature.home.ui.changeaddress.ChangeAddressViewModel.ViewState.Loading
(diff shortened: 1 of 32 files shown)


app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
```diff
@@ -34,14 +52,118 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
...
+        spinner.loadingSpinner.remove()
+        contentScrollView.show()
+        image.show()
+        title.text = titleText
+        title.show()
+        subtitle.text = subtitleText
+        subtitle.show()
+        continueButton.text = buttonText
+        continueButton.setOnClickListener { onContinue() }
+        continueButton.icon = buttonIcon?.let { ContextCompat.getDrawable(this@ChangeAddressActivity, it) }
```
hedvigoscar, line 132:
```suggestion
        continueButton.icon = buttonIcon?.let { compatDrawable(it) }
```

There's a handy extension function for this purpose (needs to be imported, though 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
```diff
@@ -34,14 +52,118 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
...
+    private fun setContent(
+        titleText: String,
+        subtitleText: String,
+        buttonText: String?,
+        @DrawableRes buttonIcon: Int?,
+        onContinue: () -> Unit
+    ) = with(binding) {
+        spinner.loadingSpinner.remove()
+        contentScrollView.show()
+        image.show()
```
hedvigoscar, line 125:
```suggestion
        spinner.loadingSpinner.isVisible = false
        contentScrollView.isVisible = true
        image.isVisible = true
```

I'm partial to replacing the `show()`/`remove()`-extensions with the `isVisible`-extension, as the latter can lead to less code in certain cases, with the prime example being:

```kotlin
if (condition) {
    view.show()
} else {
    view.remove()
}

// vs

view.isVisible = condition
```

Admittedly, in this scenario the difference is small, but I think there are gains to be had by phasing out the former style with the latter entirely 😄

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
```diff
@@ -34,14 +52,118 @@ class ChangeAddressActivity : BaseActivity(R.layout.change_address_activity) {
...
+        contentScrollView.show()
+        image.remove()
+
+        title.text = titleText
+        title.show()
+        subtitle.text = subtitleText
+        subtitle.show()
+
+        continueButton.text = buttonText
+        continueButton.setOnClickListener { onContinue() }
```
hedvigoscar, line 154:
```suggestion
        continueButton.setHapticClickListener { onContinue() }
```

Perhaps we should discuss with our designers whether they still favor the haptics for everything-style. If not, we can maybe remove it as it's admittedly not really common practice for other apps

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.ChangeAddressInProgress
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.Loading
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.ManualChangeAddress
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.SelfChangeAddress
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.SelfChangeError
+import com.hedvig.app.feature.home.ui.changeaddress.ViewState.UpcomingAgreementError
+import kotlinx.coroutines.launch
+
+abstract class ChangeAddressViewModel: ViewModel() {
+    abstract val viewState: LiveData<ViewState>
```
hedvigoscar, line 18:
```suggestion
    protected val _viewState = MutableLiveData<ViewState>()
    val viewState: LiveData<ViewState>
```

using the `protected`-modifier we can provide a publishing-handle for all subclasses as we please. (I used to use this form but switched to what I'm suggesting as I usually always implemented the underlying `MutableLiveData` in both `Impl`s and `Mock`s 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+            _viewState.value = Loading
+            _viewState.value = createViewState()
+        }
+    }
+
+    private suspend fun createViewState(): ViewState {
+        return getUpComingAgreementState(
+            onNoUpcomingChange = ::getSelfChangeState
+        )
+    }
```
hedvigoscar, line 47:
```suggestion
    private suspend fun createViewState() =
        getUpComingAgreementState(
            onNoUpcomingChange = ::getSelfChangeState
        )
```

I'm partial to the functional form with an inferred return type, let me know if you feel differently 🙂

hedvigoscar: Also, is the underlying function ever called from anywhere but here? If not, maybe we can just inline it? 😬

app/src/main/res/layout/change_address_activity.xml
```diff
@@ -1,62 +1,104 @@
...
+
+        </androidx.constraintlayout.widget.ConstraintLayout>
+
+    </androidx.core.widget.NestedScrollView>
+
+    <include
+        android:id="@+id/spinner"
+        layout="@layout/loading_spinner" />
+
+    <com.google.android.material.button.MaterialButton
```
hedvigoscar, line 89:
```suggestion
    <Button
```

`<MaterialButton />` automatically gets used when using a `MaterialTheme` 😉

app/src/main/res/layout/home_change_address.xml
```diff
@@ -1,12 +1,13 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/change_address_button"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:drawablePadding="@dimen/base_margin_double"
+    android:foreground="?selectableItemBackground"
+    android:gravity="center_vertical"
+    android:padding="@dimen/base_margin_double"
+    android:text="@string/home_tab.editing_section_change_address_label"
+    android:textAppearance="@style/Hedvig.TextAppearance.Subtitle1"
```
hedvigoscar, line 12:
```suggestion
    android:textAppearance="?textAppearanceSubtitle1"
```

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetSelfChangeEligibilityUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.apollographql.apollo.ApolloClient
+import com.hedvig.android.owldroid.graphql.SelfChangeEligibilityQuery
+import com.hedvig.android.owldroid.type.SelfChangeBlocker
+import com.hedvig.app.feature.home.ui.changeaddress.GetSelfChangeEligibilityUseCase.SelfChangeEligibilityResult.Blocked
+import com.hedvig.app.feature.home.ui.changeaddress.GetSelfChangeEligibilityUseCase.SelfChangeEligibilityResult.Eligible
+import com.hedvig.app.feature.home.ui.changeaddress.GetSelfChangeEligibilityUseCase.SelfChangeEligibilityResult.Error
+import com.hedvig.app.util.apollo.QueryResult
+import com.hedvig.app.util.apollo.safeQuery
+
+class GetSelfChangeEligibilityUseCase(
```
hedvigoscar, line 12: How does a `UseCase` differ from a `Repository`? Is the difference merely in name? 🤔

hedvigoscar: Or is it that it crunches the data from the queries and returns what the `Activity`/`Fragment` previously used to do? 😄

hedvigoscar (commented):
A significant step towards our new suggested architecture! I like it. A few things to address, but nothing major 😄

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
hugokallstrom: I agree its better in the if/else example, but I still think 

`image.show()` reads better than `image.isVisible = true`. No harm in using both right?

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
hugokallstrom: Yes I'll talk to the designers about this :) Fixing for now though

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressViewModel.kt
hugokallstrom: Think this reads just a bit clearer tbh :)

app/src/main/res/layout/change_address_activity.xml
hugokallstrom: Changed because I wanted to be able to access `button.icon` in the activity

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressActivity.kt
hedvigoscar: I'm on the fence - but we can keep it as is 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressViewModel.kt
hedvigoscar: Fair enough!

app/src/main/res/layout/change_address_activity.xml
hedvigoscar: Oh! Hm 🤔 Perhaps that's an intentionally hidden implementation detail? Confusing to say the least.

hedvigoscar: No, it's very clearly a part of the exposed API. Fair enough. I'm satisfied with keeping the code as is!

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetSelfChangeEligibilityUseCase.kt
hugokallstrom: I think [this](https://stackoverflow.com/questions/43055247/what-is-the-difference-between-repositories-and-usecases) SO thread explains it quite nice: 

"Simply, Use-Cases handles your business logic, Repositories are the data layer which you store and access data."

For example, `ChangeAddressViewModel` calls use cases in order to construct a `ViewState`, we could skip the use cases and do this in the view model as well but I think use cases encapsulate business logic quite well. If we had local storage (eg. room), we could have a `MemberRepository` which would handle network and database requests. The use case would then call `memberRepository.selfChangeEligibility()` instead of `apolloClient.query(SelfChangeEligibilityQuery()).safeQuery()`. 

Hope it makes sense!

hedvigoscar: It does indeed! 🙇

hedvigoscar (approved):

PR #494, opened 2021-03-22
Remove api version dependency for hide key board extension

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
+++ b/app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
@@ -78,21 +78,18 @@ fun Context.hideKeyboard(view: View) {
     inputMethodManager.hideSoftInputFromWindow(view.windowToken, 0)
 }
 
-@RequiresApi(Build.VERSION_CODES.R)
-suspend fun Context.hideKeyboardIfVisible(view: View, inputView: View, delayMillis: Long = 0) {
-    if (view.rootWindowInsets?.isVisible(WindowInsets.Type.ime()) == true) {
-        hideKeyboard(inputView)
-        delay(delayMillis)
-    }
+suspend fun Context.hideKeyboardWithDelay(inputView: View, delayMillis: Long = 0) {
+    hideKeyboard(inputView)
+    delay(delayMillis)
 }
 
 fun Context.triggerRestartActivity(activity: Class<*> = SplashActivity::class.java) {
     val startActivity = Intent(this, activity)
     startActivity.flags =
         Intent.FLAG_ACTIVITY_NEW_TASK or
-        Intent.FLAG_ACTIVITY_CLEAR_TASK or
-        Intent.FLAG_ACTIVITY_CLEAR_TOP or
-        Intent.FLAG_ACTIVITY_SINGLE_TOP
+            Intent.FLAG_ACTIVITY_CLEAR_TASK or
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
```diff
@@ -78,9 +78,13 @@ fun Context.hideKeyboard(view: View) {
     inputMethodManager.hideSoftInputFromWindow(view.windowToken, 0)
 }
 
-@RequiresApi(Build.VERSION_CODES.R)
-suspend fun Context.hideKeyboardIfVisible(view: View, inputView: View, delayMillis: Long = 0) {
-    if (view.rootWindowInsets?.isVisible(WindowInsets.Type.ime()) == true) {
+suspend fun Context.hideKeyboardWithDelay(inputView: View, delayMillis: Long = 0) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+        if (inputView.rootView.rootWindowInsets?.isVisible(WindowInsets.Type.ime()) == true) {
```
hedvigoscar, line 83: I think there's actually a compat-variant of this API! We should check that out, in that case we could move to only having one code path

hedvigoscar (approved):
Looks good enough to merge, really, but I would like it if we investigated the Compat-approach 🙂

app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
hugokallstrom: Investigating!

hugokallstrom: We need `androidx.core:core-ktx:1.5.0-beta03` In order to access `WindowInsetsCompat.Type`. Should we upgrade or wait for stable?

hedvigoscar: I'm ok with upgrading 🙂

hugokallstrom: Done!

app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
```diff
@@ -78,9 +79,8 @@ fun Context.hideKeyboard(view: View) {
     inputMethodManager.hideSoftInputFromWindow(view.windowToken, 0)
 }
 
-@RequiresApi(Build.VERSION_CODES.R)
-suspend fun Context.hideKeyboardIfVisible(view: View, inputView: View, delayMillis: Long = 0) {
-    if (view.rootWindowInsets?.isVisible(WindowInsets.Type.ime()) == true) {
+suspend fun Context.hideKeyboardWithDelay(inputView: View, delayMillis: Long = 0) {
+    if (inputView.rootView.rootWindowInsets?.isVisible(WindowInsetsCompat.Type.ime()) == true) {
```
hedvigoscar, line 83: This needs to use the compat-equivalents, or it will break 😞

hedvigoscar: i.e the `rootWindowInsets`-compat ones

hugokallstrom: Ah good catch, fixed

PR #495, opened 2021-03-22
APP-243: Market selected preferences

hugokallstrom:
Check market shared string preference to determine if market has been selected instead of using separate boolean preference

### Checklist


app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -136,14 +136,8 @@ class SplashActivity : BaseActivity(R.layout.activity_splash) {
         val market = sharedPreferences.getString(Market.MARKET_SHARED_PREF, null)
         when (loginStatus) {
             LoginStatus.ONBOARDING -> {
-                if (market == null) {
```
hedvigoscar, line 139: 😬

hedvigoscar (approved):

PR #503, opened 2021-03-31
APP-266: Embark text action set ime action

hugokallstrom:
Refactor text action set to use linear layout instead of recycler view and handle ime actions.

I realised that this could have probably been done in the adapter as well, by checking position in the view holder. But I think replacing the recyclerview in this case is fine since we don't expect the number of input views to be big enough to really benefit from a recyclerview. In this case we could perhaps make the input view a custom view and use it in both text action set and text action fragments. 

### Checklist


app/src/main/java/com/hedvig/app/feature/embark/passages/textactionset/TextActionFragment.kt
```diff
@@ -90,19 +95,61 @@ class TextActionSetFragment : Fragment(R.layout.fragment_text_action_set) {
...
-                key,
-                data.placeholders[index],
-                data.mask[index],
-                key?.let { model.getFromStore(it) },
-            )
+    private fun createInputViews(): List<View> {
+        return data.keys.mapIndexed { index, key ->
+            val inputView = LayoutInflater.from(requireContext()).inflate(R.layout.embark_input_item, binding.inputContainer, false)
+            val inputLayout = inputView.findViewById<TextInputLayout>(R.id.textField)
+            val inputEditText = inputView.findViewById<TextInputEditText>(R.id.input)
```
hedvigoscar, line 102: Let's use `ViewBinding`, and let it do the finding of views 😉

app/src/main/java/com/hedvig/app/feature/embark/passages/textactionset/TextActionFragment.kt
```diff
@@ -90,19 +95,61 @@ class TextActionSetFragment : Fragment(R.layout.fragment_text_action_set) {
...
-    private fun textFieldData(data: TextActionSetParameter) =
-        data.keys.mapIndexed { index, key ->
-            TextFieldData(
-                key,
-                data.placeholders[index],
-                data.mask[index],
-                key?.let { model.getFromStore(it) },
-            )
+    private fun createInputViews(): List<View> {
+        return data.keys.mapIndexed { index, key ->
```
hedvigoscar, line 99: I'm partial to the functional syntax for method declaration, in particular when the first statement in a method is a return-statement:

```suggestion
    private fun createInputViews() = data.keys.mapIndexed { index, key ->
```

(note that merging this suggestion requires one curly bracket to be removed at the end of the method 😉 )

app/src/main/java/com/hedvig/app/feature/embark/passages/textactionset/TextActionFragment.kt
```diff
@@ -90,19 +95,61 @@ class TextActionSetFragment : Fragment(R.layout.fragment_text_action_set) {
...
+
+            if (index < data.keys.size - 1) {
+                inputEditText.imeOptions = EditorInfo.IME_ACTION_NEXT
+            } else {
+                inputEditText.imeOptions = EditorInfo.IME_ACTION_DONE
+            }
+
+            inputEditText.onImeAction {
+                if (textActionSetViewModel.isValid.value == true) {
+                    viewLifecycleScope.launch {
```
hedvigoscar, line 137: This probably needs to be implemented exactly the same as how the submit-button is implemented, otherwise it will handle multiple clicks on the IME submit improperly.

hedvigoscar (commented):
Approved upon addressing suggestions 🙂

app/src/main/java/com/hedvig/app/feature/embark/passages/textactionset/TextActionFragment.kt
hugokallstrom: The IME seems to handle multiple clicks really well actually, dont think we need to change this 👍

app/src/main/java/com/hedvig/app/feature/embark/passages/textactionset/TextActionFragment.kt
```diff
@@ -95,57 +97,52 @@ class TextActionFragment : Fragment(R.layout.fragment_text_action_set) {
         delay(PASSAGE_ANIMATION_DELAY_MILLIS)
     }
 
-    private fun createInputViews(): List<View> {
-        return data.keys.mapIndexed { index, key ->
-            val inputView = LayoutInflater.from(requireContext()).inflate(R.layout.embark_input_item, binding.inputContainer, false)
-            val inputLayout = inputView.findViewById<TextInputLayout>(R.id.textField)
-            val inputEditText = inputView.findViewById<TextInputEditText>(R.id.input)
+    private fun createInputViews(): List<View> = data.keys.mapIndexed { index, key ->
+        val inputView = EmbarkInputItemBinding.inflate(layoutInflater)
```
hedvigoscar, line 101: We can still pass `container` and `attachToParent` here for more efficient layouting 🙂

PR #506, opened 2021-04-06
APP-24: Number action set

hugokallstrom:
Parse number action set and use same fragment and viewmodel for `NumberAction` and `NumberActionSet`.

### Checklist


hedvigoscar (approved):
Approved upon addressing comments 🙂

PR #507, opened 2021-04-06
APP-5: Contract details change address button

hugokallstrom:
Add change address button for Swedish apartment in insurance detail view.

Should this be added in any other contract than Swedish apartment?

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoAdapter.kt
+++ b/app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoAdapter.kt
@@ -5,6 +5,7 @@ import android.view.ViewGroup
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import com.hedvig.app.R
+import com.hedvig.app.databinding.ChangeAddressButtonBinding
 import com.hedvig.app.databinding.ContractDetailRowBinding
 import com.hedvig.app.databinding.ContractDetailYourInfoHeaderBinding
 import com.hedvig.app.databinding.YourInfoChatButtonBinding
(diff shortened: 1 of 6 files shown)


app/src/main/res/layout/change_address_button.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Button xmlns:android="http://schemas.android.com/apk/res/android"
```
hedvigoscar, line 2: Should we create a generic outlined button-layout file, and use that whenever we need one of these?

I know this breaks the norm for the repository, but I think it could be wise to move towards that kind of structure, as we currently have a lot of layouts that basically are just `TextView`s or `Button`s with the same style that only really differ in what `android:text` they have. Something to think about, for sure 🙂

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoAdapter.kt
```diff
@@ -93,16 +96,29 @@ class YourInfoAdapter :
...
+
+        class ChangeAddressButton(parent: ViewGroup) :
+            ViewHolder(parent.inflate(R.layout.change_address_button)) {
+            private val binding by viewBinding(YourInfoChatButtonBinding::bind)
+            override fun bind(data: YourInfoModel) {
+                binding.root.apply {
+                    setHapticClickListener {
+                        context.startActivity(ChangeAddressActivity.newInstance(context))
+                    }
+                }
```
hedvigoscar, line 119: We don't need the additional scope from `apply` here, I think. 🙂 

```suggestion
                binding.root.setHapticClickListener {
                    context.startActivity(ChangeAddressActivity.newInstance(context))
                }
```

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoAdapter.kt
```diff
@@ -93,16 +96,29 @@ class YourInfoAdapter :
...
                     setHapticClickListener {
                         context.startActivity(ChatActivity.newInstance(context, true))
                     }
                 }
             }
         }
+
+        class ChangeAddressButton(parent: ViewGroup) :
+            ViewHolder(parent.inflate(R.layout.change_address_button)) {
+            private val binding by viewBinding(YourInfoChatButtonBinding::bind)
```
hedvigoscar, line 113: This `ViewBinding` does not match the layout file, which could lead to some serious issues down the line 😨

hedvigoscar (commented):
A few things to address, but generally looks good 🙂

app/src/main/res/layout/change_address_button.xml
hugokallstrom: Yes agreed, feels a bit out of scope for this PR though. Compose will be so nice in this regard 💯

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoAdapter.kt
hugokallstrom: Wops good find!

hedvigoscar (approved):

PR #510, opened 2021-04-12
APP-263: Embark offer redirect

hugokallstrom:
We are now getting keys from `offerRedirect` which we use instead of the hard coded `"quoteId"`-key. These are used because we need multiple offer-ids for eg. NO combo.
 

### Checklist

### DIFF
--- a/app/src/androidTest/java/com/hedvig/app/feature/embark/redirect/OfferRedirectTest.kt
+++ b/app/src/androidTest/java/com/hedvig/app/feature/embark/redirect/OfferRedirectTest.kt
@@ -4,7 +4,6 @@ import com.agoda.kakao.screen.Screen.Companion.onScreen
 import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
 import com.hedvig.app.feature.embark.screens.EmbarkScreen
 import com.hedvig.app.feature.embark.ui.EmbarkActivity
-import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_EXTERNAL_REDIRECT
 import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_OFFER_REDIRECT
 import com.hedvig.app.util.ApolloCacheClearRule
 import com.hedvig.app.util.ApolloMockServerRule
--- a/app/src/androidTest/java/com/hedvig/app/feature/embark/redirect/OfferRedirectTest.kt
+++ b/app/src/androidTest/java/com/hedvig/app/feature/embark/redirect/OfferRedirectTest.kt
@@ -1,0 +1,41 @@
+package com.hedvig.app.feature.embark.redirect
+
+import com.agoda.kakao.screen.Screen.Companion.onScreen
+import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
+import com.hedvig.app.feature.embark.screens.EmbarkScreen
+import com.hedvig.app.feature.embark.ui.EmbarkActivity
+import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_EXTERNAL_REDIRECT
+import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_OFFER_REDIRECT
+import com.hedvig.app.util.ApolloCacheClearRule
+import com.hedvig.app.util.ApolloMockServerRule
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/hedvig/app/feature/embark/ui/EmbarkActivity.kt
```diff
@@ -235,13 +245,13 @@ class EmbarkActivity : BaseActivity(R.layout.activity_embark) {
         return UpgradeAppFragment.newInstance()
     }
 
-    private fun showWebOffer() {
+    private fun showWebOffer(keys: List<String?>) {
```
hedvigoscar, line 248: Can we take `List<String>` here and filter it at the call site?

app/src/main/java/com/hedvig/app/feature/webonboarding/WebOnboardingActivity.kt
```diff
@@ -145,11 +157,16 @@ class WebOnboardingActivity : BaseActivity(R.layout.activity_web_onboarding) {
...
+            context: Context,
+            webPath: String?,
+            offer: Boolean = false,
+            quoteId: List<String?>
+        ): Intent {
             val intent = Intent(context, WebOnboardingActivity::class.java)
             intent.putExtra(WEB_PATH, webPath)
             intent.putExtra(OFFER, offer)
-            intent.putExtra(QUOTE_ID, quoteId)
+            intent.putExtra(QUOTE_ID, quoteId.toArrayList())
```
hedvigoscar, line 169: `null` is probably not wise to send away here, we should avoid it if possible 🙂

hedvigoscar (changes requested):
I think it would be wise to add some tests for this component 🙂

hugokallstrom:
> I think it would be wise to add some tests for this component 🙂

For EmbarkActivity?

hedvigoscar:
For `EmbarkOfferRedirect` 😉

app/src/androidTest/java/com/hedvig/app/feature/embark/redirect/OfferRedirectTest.kt
```diff
@@ -0,0 +1,41 @@
+package com.hedvig.app.feature.embark.redirect
+
+import com.agoda.kakao.screen.Screen.Companion.onScreen
+import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
+import com.hedvig.app.feature.embark.screens.EmbarkScreen
+import com.hedvig.app.feature.embark.ui.EmbarkActivity
+import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_EXTERNAL_REDIRECT
```
hedvigoscar, line 7:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/embark/ui/EmbarkActivity.kt
```diff
@@ -93,10 +93,10 @@ class EmbarkActivity : BaseActivity(R.layout.activity_embark) {
     private fun showNextView(embarkData: EmbarkModel, passage: EmbarkStoryQuery.Passage?) {
         val offerKeys = embarkData.passage?.offerRedirect?.data?.keys
         if (offerKeys != null && offerKeys.isNotEmpty()) {
-            val offerIds = offerKeys.map { key -> key?.let { model.getFromStore(it) } }
+            val offerIds = offerKeys.mapNotNull { key -> key?.let { model.getFromStore(it) } }
             showWebOffer(offerIds)
         } else if (embarkData.passage?.externalRedirect?.data?.location == EmbarkExternalRedirectLocation.OFFER) {
-            val key = model.getFromStore("quoteId")
+            val key = model.getFromStore("quoteId") ?: ""
```
hedvigoscar, line 99: I think it might be warranted to error out here. Do you agree?

hedvigoscar (approved):
Approved upon merging suggestions!

app/src/main/java/com/hedvig/app/feature/embark/ui/EmbarkActivity.kt
hugokallstrom: Totally agree, fixing

PR #519, opened 2021-04-21
APP-372: Start embark from moving flow intro 

hugokallstrom:
Use updated schema to query an embark story id and use in order to start embark for moving flow. Note that the current id is not the final story to be used. 

### Checklist


hedvigoscar (approved):
🚀

PR #522, opened 2021-04-22
APP-374: Support Expression-based responses

hedvigoscar:
As it turns out, these are functionally equivalent to `EmbarkMessage`.
However, we need to support both.

I also did some opportunistic refactoring. Nice!

- Kill SubExpression in favor of BasicExpression
- Make the MockEmbarkViewModel work again
- Support `EmbarkResponseExpression`

### Checklist


hugokallstrom (approved):
Nice :)

PR #527, opened 2021-04-27
APP-386: Refactor all dependencies to use `buildSrc`

hedvigoscar:
This PR will be the basis on which I do some dependency update work.


hugokallstrom (approved):
Very nice!!

PR #535, opened 2021-04-30
APP-414: Prevent missing Play Store from crashing app when attempting to open play store

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/ratings/RatingsDialog.kt
```diff
@@ -64,7 +65,11 @@ class RatingsDialog : DialogFragment() {
                 when (choice) {
                     RatingsChoice.YES -> {
                         tracker.rate()
```
hugokallstrom, line 67: Can probably extract this as a function to Util as well, like `tryOpenPlayStore()` or something

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/ratings/RatingsDialog.kt
hedvigoscar: Agreed. I'll do that instead I think

PR #536, opened 2021-05-03
APP-415: Upload test reports

hedvigoscar:
### Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -78,6 +78,12 @@ jobs:
           target: google_apis
           profile: pixel_3a
           script: ./gradlew connectedDebugAndroidTest
+      - name: Upload test reports
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-reports
+          path: app/build/reports
   lint:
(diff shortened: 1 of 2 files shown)


hugokallstrom (approved):
What does this do? :)

hedvigoscar:
Good question 😬 This PR makes it so that after running tests in Github Actions, regardless of whether the tests pass or fail, the test results get uploaded as an artifact to Github Actions.

That means that when a test fails, we can download the test report and get some proper context to what actually went wrong. I think this will aid us in maintaining test suite stability over time 🙂

hedvigoscar:
For example, the tests fail in this PR. Now I can download the test report and dig through it to see what's up 🙂

hugokallstrom:
> Good question 😬 This PR makes it so that after running tests in Github Actions, regardless of whether the tests pass or fail, the test results get uploaded as an artifact to Github Actions.
> 
> That means that when a test fails, we can download the test report and get some proper context to what actually went wrong. I think this will aid us in maintaining test suite stability over time 🙂

Thats great! 👍

PR #542, opened 2021-05-05
Offer view state

hugokallstrom:
Observe view state to decouple network model from view.

Upcoming PRs will remove `OfferQuery.Data` from `OfferModel` to further decrease coupling between view and network layer.

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferExt.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferExt.kt
@@ -1,0 +1,17 @@
+package com.hedvig.app.feature.offer
+
+import com.apollographql.apollo.api.Response
+import com.hedvig.android.owldroid.graphql.OfferQuery
+
+fun Response<OfferQuery.Data>.priceOrNull() = data
+    ?.lastQuoteOfMember
+    ?.asCompleteQuote
+    ?.insuranceCost
+    ?.fragments
+    ?.costFragment
+    ?.monthlyNet
+    ?.fragments
+    ?.monetaryAmountFragment
+    ?.amount
+    ?.toBigDecimal()
+    ?.toDouble()
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
@@ -1,0 +1,27 @@
+package com.hedvig.app.feature.offer
+
+import com.hedvig.android.owldroid.graphql.OfferQuery
+import com.hedvig.app.feature.offer.ui.OfferModel
+
+class OfferItemsBuilder {
+
+    companion object {
+        fun createItems(data: OfferQuery.Data): List<OfferModel> {
+            return listOfNotNull(
+                OfferModel.Header(data),
+                OfferModel.Info,
+                OfferModel.Facts(data),
+                OfferModel.Perils(data),
+                OfferModel.Terms(data),
+                data.lastQuoteOfMember.asCompleteQuote?.currentInsurer?.let { currentInsurer ->
+                    if (currentInsurer.switchable == true) {
+                        OfferModel.Switcher(currentInsurer.displayName)
+                    } else {
+                        null
+                    }
+                },
+                OfferModel.Footer
+            )
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -1,22 +1,26 @@
...
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
 import java.time.LocalDate
 
 abstract class OfferViewModel : ViewModel() {
-    abstract val data: MutableLiveData<OfferQuery.Data>
+    abstract val viewState: LiveData<ViewState>
```
hedvigoscar, line 23: Let's make this non-abstract - the base class can contain both the producer and consumer-handles of the `LiveData`, as such:

```kotlin
protected val _viewState = MutableLiveData<ViewState>()
val viewState: LiveData<ViewState> = _viewState
```

app/src/main/java/com/hedvig/app/feature/offer/OfferExt.kt
```diff
@@ -0,0 +1,17 @@
+package com.hedvig.app.feature.offer
+
+import com.apollographql.apollo.api.Response
+import com.hedvig.android.owldroid.graphql.OfferQuery
+
+fun Response<OfferQuery.Data>.priceOrNull() = data
```
hedvigoscar, line 6: I think we're going to have to switch away from `lastQuoteOfMember` pretty soon - perhaps we can make this extension on `Quote` and sum it up on the affected `Quote`s later on?

app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
```diff
@@ -0,0 +1,27 @@
+package com.hedvig.app.feature.offer
+
+import com.hedvig.android.owldroid.graphql.OfferQuery
+import com.hedvig.app.feature.offer.ui.OfferModel
+
+class OfferItemsBuilder {
+
+    companion object {
+        fun createItems(data: OfferQuery.Data): List<OfferModel> {
```
hedvigoscar, line 9: I'd be ok with making this a top-level function, or making the `OfferItemsBuilder` an `object`, to get rid of the `companion object` 🙂

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -81,52 +81,39 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
+        storeBoolean(IS_VIEWING_OFFER, false)
+        startActivity(
+            Intent(
+                this@OfferActivity,
+                LoggedInActivity::class.java
+            ).apply {
+                putExtra(LoggedInActivity.EXTRA_IS_FROM_ONBOARDING, true)
+                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK)
+            }
```
hedvigoscar, line 113: Could we move this `Intent`-creation to use `LoggedInActivity::newInstance`? I think that would limit the amount of `Intent`-parameters to an amount that is easier to reason about 🙂

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -150,4 +178,11 @@ class OfferViewModelImpl(
             response.getOrNull()?.data?.let { offerRepository.removeStartDateFromCache(it) }
         }
     }
+
+    override fun trackUserSign() {
+        viewModelScope.launch {
+            val price = offerRepository.offer().firstOrNull()?.priceOrNull() ?: 0.0
```
hedvigoscar, line 184: I don't think that `0.0` is an appropriate default here - perhaps we could do something like this, instead?

```suggestion
            val price = offerRepository.offer().firstOrNull()?.priceOrNull() ?: return@launch
```

hedvigoscar (changes requested):
Good stuff!

app/src/main/java/com/hedvig/app/feature/offer/OfferExt.kt
hugokallstrom: Yes! I think thats suited another PR though, I'll create an issue for it.

hedvigoscar (approved):

PR #548, opened 2021-05-18
APP-438: Improve build performance

hedvigoscar:
- Disable unused build features
- Delete a large amount of unused resources
- Add cache fix-plugin
- Limit use of `BuildConfig`
- Delete some unused GraphQL files
- Formatting
- Add Gradle-doctor plugin

### Checklist


hugokallstrom (approved):
Just curious, why change from using build config fields to string resources?

hedvigoscar:
Ultimately, I'd like to move away from using `BuildConfig` altogether as it takes some time to generate and can unnecessarily invalidate the build cache. Strings are more static in that way, I figured.

Plus we get less noise in the `build.gradle.kts`-file 😄

PR #549, opened 2021-05-18
Query activeContractBundles to get embark story id for moving flow. Rename usecase.

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/embark/ui/EmbarkActivity.kt
```diff
@@ -109,6 +109,13 @@ class EmbarkActivity : BaseActivity(R.layout.activity_embark) {
         if (offerKeys != null && offerKeys.isNotEmpty()) {
             val offerIds = offerKeys.mapNotNull { model.getFromStore(it) }
             showWebOffer(offerIds)
```
hugokallstrom, line 111: I just check for the "Offer" passage here. Can we do this in a neater way?

hedvigoscar: There should be an `ExternalRedirect` here, I believe, that instructs us where to go, and probably also which keys to provide as context to that external redirect location. 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetAddressChangeStoryIdUseCase.kt
```diff
@@ -1,21 +1,25 @@
...
 ) {
 
     suspend operator fun invoke(): SelfChangeEligibilityResult {
-        return when (val result = apolloClient.query(SelfChangeEligibilityQuery()).safeQuery()) {
+        return when (val result = apolloClient.query(ActiveContractBundlesQuery()).safeQuery()) {
             is QueryResult.Success ->
-                result.data?.selfChangeEligibility?.embarkStoryId
+                result.data
+                    .activeContractBundles
+                    .firstOrNull()
```
hedvigoscar, line 20: I suppose first is going to be fine for now, but long term we likely want this to be able to address any of the contract bundles.

That's a project for when we have UI for handling multiple bundles though. But let's keep it in mind 🙂

hedvigoscar (approved):
lgtm

PR #550, opened 2021-05-20
Query bundled quotes to get pending address change

hugokallstrom:
### Checklist


apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/home.graphql
```diff
@@ -73,4 +73,23 @@ query HomeQuery($locale: Locale!, $languageCode: String!) {
     }
     body
   }
+
+  quoteBundle(input: $quoteBundleInput) {
```
hugokallstrom, line 77: This whole query now fails if we dont supply a list of ids here, should we ask back-end to handle that error better? Perhaps return some sort of `QuoteBundleResult` instead of `QuoteBundleInput`?

PR #554, opened 2021-05-24
APP-442: Fix update language never being called for default language

hedvigoscar:
- Fix language never being sent to backend for default language selection
- Verify that it happens for unknown market as well

### Checklist


hugokallstrom (approved):
LGTM!

PR #555, opened 2021-05-25
APP-7: Change Address Result-Screen, in Compose!

hedvigoscar:
### Checklist


app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
```diff
@@ -0,0 +1,127 @@
...
+            val date: LocalDate,
+        ) : Result()
+
+        @Parcelize
+        object UnderwritingLimitsHit : Result()
+    }
+
+    companion object {
+        private const val RESULT = "RESULT"
+
```
hugokallstrom, line 55: To make this view even more general, we could pass a data class containing a drawable res (illustration), title, paragraph and button text + icon. We could then reuse it for some of the intro screens as well. We could also use finish with result and let the caller handle what should happen if the button is pressed. What do you think?

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
```diff
@@ -0,0 +1,127 @@
...
+                            Spacer(Modifier.size(ButtonDefaults.IconSpacing))
+                            Text(stringResource(R.string.moving_uw_failure_button_text))
+                        }
+                    )
+                }
+            }
+        }
+    }
+}
+
```
hugokallstrom, line 109: Are there any drawbacks for having previews in the activity?

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
```diff
@@ -0,0 +1,105 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.google.android.material.composethemeadapter.MdcTheme
+
```
hugokallstrom, line 26: This is brilliant 😍

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
```diff
@@ -0,0 +1,105 @@
...
+@Composable
+fun ITPB(
```
hugokallstrom, line 28: I think we should be careful with naming composables since they are so general in nature. The name should be easily searchable and explain well what they are used for. Perhaps something like `ImageWithTitleAndParagraph`? Maybe we should discuss naming conventions for composables more in depth 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
```diff
@@ -0,0 +1,105 @@
+package com.hedvig.app.feature.home.ui.changeaddress
+
```
hugokallstrom, line 2: Since this composable can be used outside of the changeaddress-package, would it make sense to have some kind of `ui-composable`-package were we put all general composables?

hugokallstrom (changes requested):
Psyched about this!! Good job 🙌

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
hedvigoscar: My thinking was that the ITPB was generic enough for use in Activities, and hence we did not need to make the Activity reusable 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
hedvigoscar: I'm not sure, to be honest. Probably not?

I guess something that could be worth investigating is what the compose compiler does with Previews. Preferably, they should be stripped in release targets, right? 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
hedvigoscar: I agree that the name is scuffed, which is one of the reasons why this is still a draft PR 😅

ITPB = Illustration, Title, Paragraph, Button

I'm discussing better terminology with Zak on Figma. I'll update it once we get there 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
hedvigoscar: Agreed. This component is made to be generic enough to join some sort of design system-package.

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
hugokallstrom: Yeah that makes sense :)

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ChangeAddressResultActivity.kt
hugokallstrom: Yeah exactly. But it probably makes sense to have the previews in the same file since they are so tightly coupled with the actual view.

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/ITPB.kt
hedvigoscar: That's alright, I made the draft PR so that I could gather this type of feedback 😉

hedvigoscar: Zak suggested Briefing Screen - what do you think about that name?

PR #562, opened 2021-05-31
APP-456: Query multi action variable, fetch and cast stored multi action values

hugokallstrom:
This got quite messy, suggestion on improvements very welcome.

### Checklist

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -149,6 +149,7 @@ dependencies {
     androidTestImplementation(Libs.AndroidX.Test.runner)
     androidTestImplementation(Libs.AndroidX.Test.rules)
     androidTestImplementation(Libs.AndroidX.Test.junit)
+    testImplementation("org.json:json:20210307")
 
     implementation(Libs.materialComponents)
     implementation(Libs.flexbox)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -146,10 +146,10 @@ dependencies {
     androidTestImplementation(Libs.AndroidX.Espresso.intents)
     androidTestImplementation(Libs.AndroidX.Espresso.contrib)
     testImplementation(Libs.AndroidX.Test.junit)
+    testImplementation(Libs.jsonTest)
     androidTestImplementation(Libs.AndroidX.Test.runner)
     androidTestImplementation(Libs.AndroidX.Test.rules)
     androidTestImplementation(Libs.AndroidX.Test.junit)
-    testImplementation("org.json:json:20210307")
 
     implementation(Libs.materialComponents)
     implementation(Libs.flexbox)
--- a/app/src/main/java/com/hedvig/app/feature/embark/util/VariableExtractor.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/util/VariableExtractor.kt
@@ -10,14 +10,20 @@ object VariableExtractor {
 
     fun extractVariables(variables: List<GraphQLVariablesFragment>, valueStore: ValueStore): JSONObject {
         var regularVariables = variables.mapNotNull {
-            it.asEmbarkAPIGraphQLSingleVariable?.let {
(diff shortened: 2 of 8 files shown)


app/build.gradle.kts
```diff
@@ -149,6 +149,7 @@ dependencies {
     androidTestImplementation(Libs.AndroidX.Test.runner)
     androidTestImplementation(Libs.AndroidX.Test.rules)
     androidTestImplementation(Libs.AndroidX.Test.junit)
+    testImplementation("org.json:json:20210307")
```
hugokallstrom, line 153: Had to include this in order to use JSONObject in unit tests

app/src/main/java/com/hedvig/app/feature/embark/util/VariableExtractor.kt
```diff
@@ -0,0 +1,123 @@
...
+            extractMultiActionVariables(it, valueStore)
+        }
+
+        multiActionVariables.forEach {
+            regularVariables = regularVariables.plus(it)
+        }
+        return regularVariables
+    }
+
+    private fun extractRegularVariables(
```
hedvigoscar, line 29: Could we split this into two methods, `extractSingleVariable`/`extractGeneratedVariable` which as input take `GraphQLVariablesFragment.AsGraphQLSingleVariable`/`GraphQLVariablesFragment.AsGeneratedVariable`? That would remove the need for having a possible null return for this/these functions.

app/src/main/java/com/hedvig/app/feature/embark/util/VariableExtractor.kt
```diff
@@ -0,0 +1,123 @@
...
+        return null
+    }
+
+    private fun extractMultiActionVariables(
+        variable: GraphQLVariablesFragment,
+        valueStore: ValueStore
+    ): JSONObject? {
+        variable.asEmbarkAPIGraphQLMultiActionVariable?.let { multiActionVariables ->
+            val multiActionItems = valueStore.getMultiActionItems(multiActionVariables.key)
+            val casted = multiActionItems.map { multiActionMap ->
```
hedvigoscar, line 57: If we squint a little at this function, it looks like it does the same job as `extractVariables` does at top-level, but with a `ValueStore` scoped to the Multi Action-item. Perhaps we could make that the reality in code, as well? Then we would not have to express the procedure twice, so to speak.

We could accomplish this by translating the `GraphQLVariablesFragment.AsGraphQLMultiActionVariable` into a `GraphQLVariablesFragment`, creating a static `ValueStore` from the `multiActionItems`, and then finally calling `extractVariables` with those two parameters as the argument.

What are your thoughts on that approach?

app/src/test/java/com/hedvig/app/feature/embark/VariableExtractorTest.kt
```diff
@@ -0,0 +1,259 @@
...
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import com.hedvig.android.owldroid.fragment.GraphQLVariablesFragment
+import com.hedvig.android.owldroid.type.EmbarkAPIGraphQLSingleVariableCasting
+import com.hedvig.app.feature.embark.util.VariableExtractor
+import org.json.JSONObject
+import org.junit.Test
+
+class VariableExtractorTest {
```
hedvigoscar, line 11: 🚀

hedvigoscar (commented):
Looks mostly good I think!

app/src/main/java/com/hedvig/app/feature/embark/util/VariableExtractor.kt
hugokallstrom: Yes I was thinking the same, but had difficulty creating the fragment in `embarkStory.graphql` - how would that look? Thanks

hedvigoscar (approved):

hedvigoscar:
Please address the failing test before merging, though 🤔

PR #566, opened 2021-06-03
Remove quotes from parsed computed store value

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/computedvalues/TemplateExpressionCalculator.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/computedvalues/TemplateExpressionCalculator.kt
@@ -126,7 +126,7 @@ object TemplateExpressionCalculator {
     private fun evaluateAbstractExpression(expression: Expression, store: HashMap<String, String>): String = when (expression) {
         is Expression.StoreKeyExpression -> store[expression.key] ?: ""
         is Expression.NumberConstantExpression -> expression.constant.toBigDecimal().stripTrailingZeros().toString()
-        is Expression.StringConstantExpression -> expression.constant
+        is Expression.StringConstantExpression -> expression.constant.replace("'", "")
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/hedvig/app/ComputedValuesTest.kt
```diff
@@ -62,6 +62,13 @@ class ComputedValuesTest {
         assertThat(result).isEqualTo("1337")
     }
 
+    @Test
+    fun `test remove quote from string`() {
```
hedvigoscar, line 66: I think this test could be named slightly more descriptively 😉 

```suggestion
    fun `should evaluate string constants correctly`() {
```

app/src/main/java/com/hedvig/app/feature/embark/computedvalues/TemplateExpressionCalculator.kt
```diff
@@ -126,7 +126,7 @@ object TemplateExpressionCalculator {
     private fun evaluateAbstractExpression(expression: Expression, store: HashMap<String, String>): String = when (expression) {
         is Expression.StoreKeyExpression -> store[expression.key] ?: ""
         is Expression.NumberConstantExpression -> expression.constant.toBigDecimal().stripTrailingZeros().toString()
-        is Expression.StringConstantExpression -> expression.constant
+        is Expression.StringConstantExpression -> expression.constant.replace("'", "")
```
hedvigoscar, line 129: Given that we're matching on a single character, we can use the `char`-based `String::replace` instead of the `String`-based one 🙂 

```suggestion
        is Expression.StringConstantExpression -> expression.constant.replace('\'', '')
```

(with reservations for that this might not compile exactly as suggested 😬)

hedvigoscar (approved):
Two very minor improvements, apply them if you agree. 🙂

PR #569, opened 2021-06-07
Feature/persist quote ids from change address result

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
+++ b/app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
@@ -5,7 +5,6 @@ import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
 import com.hedvig.app.feature.embark.screens.EmbarkScreen
 import com.hedvig.app.feature.embark.ui.EmbarkActivity
 import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_GRAPHQL_MUTATION_AND_OFFER_REDIRECT
-import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_GRAPHQL_MUTATION_AND_SINGLE_VARIABLE
 import com.hedvig.app.testdata.feature.embark.data.VARIABLE_MUTATION
 import com.hedvig.app.util.ApolloCacheClearRule
 import com.hedvig.app.util.ApolloMockServerRule
--- a/app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
+++ b/app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
@@ -11,10 +11,10 @@ import com.hedvig.app.util.ApolloMockServerRule
 import com.hedvig.app.util.LazyIntentsActivityScenarioRule
 import com.hedvig.app.util.apolloResponse
 import com.hedvig.app.util.context
+import com.hedvig.app.util.jsonArrayOf
 import com.hedvig.app.util.jsonObjectOf
 import com.hedvig.app.util.stub
 import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
-import org.json.JSONArray
 import org.junit.Rule
 import org.junit.Test
 
@@ -29,10 +29,6 @@ class ListVariableTest : TestCase() {
             success(jsonObjectOf("hello" to jsonArrayOf("1", "2", "3")))
         }
     )
-
-    private fun jsonArrayOf(vararg items: Any) = JSONArray().apply {
-        items.forEach { put(it) }
-    }
 
     @get:Rule
     val apolloCacheClearRule = ApolloCacheClearRule()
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -68,10 +70,18 @@ abstract class EmbarkViewModel(
...
 
+    fun putInStore(key: String, value: List<String>) {
+        valueStore.put(key, value)
+    }
+
     fun getPrefillFromStore(key: String) = valueStore.prefill.get(key)
 
     fun getFromStore(key: String) = valueStore.get(key)
 
+    fun getQuoteIds(): List<String> = valueStore.getList("quoteIds")
```
hugokallstrom, line 81: Can we get around hard coding these keys?

hedvigoscar: Yes! The `Redirect`-component should specify from where in the store to load parameters, hard-coding just won't do 😞

app/src/main/java/com/hedvig/app/util/JSONObject.kt
```diff
@@ -48,3 +53,11 @@ operator fun JSONObject.plus(other: JSONObject): JSONObject {
...
     return clone
 }
+
+fun JSONArray.toStringArray(): List<String> {
+    val list: MutableList<String> = ArrayList()
+    for (i in 0 until length()) {
+        list.add(get(i).toString())
+    }
+    return list
+}
```
hedvigoscar, line 63:
```suggestion
fun JSONArray.toStringArray() = map { it.toString() }
```

hedvigoscar (changes requested):
I think we need to rethink this on an Embark-level, hardcoding this stuff would be disasterous I think 😕

app/src/main/java/com/hedvig/app/util/JSONObject.kt
hugokallstrom: Cant seem to iterate over a JSONArray 🤔

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
hugokallstrom: Rewrote the `getListFromStore` function in the viewmodel like this: 
```
    fun getListFromStore(keys: List<String>): List<String> {
        return keys.map {
            valueStore.getList(it) ?: listOfNotNull(valueStore.get(it))
        }.flatten()
    }
```

app/src/main/java/com/hedvig/app/util/JSONObject.kt
hedvigoscar: Oh damn, my bad. Maybe it could be worth implementing an extension that provides an iterator for `JSONArray`? With that, we could make a lot more types of iterations on them 🙂

hugokallstrom: Yeah we could do something like
```
fun <T> JSONArray.toArray(action: () -> T): List<T> {
    val list: MutableList<T> = ArrayList()
    for (i in 0 until length()) {
        list.add(action())
    }
    return list
}
```

But I think we can tackle that problem in case we need it?

hedvigoscar: Personally, I'd implement a `JSONArrayIterator`, we have a few examples of custom iterators in the project. (lint now tells me that they are incorrectly implemented, so we have that to deal with, but whatever)

Having iterators available for this is generally the correct level of abstraction to create, in my opinion, but if you feel that it's entirely unnecessary, we can skip it. 🙂

app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
```diff
@@ -0,0 +1,62 @@
+package com.hedvig.app.feature.embark.api.graphqlmutation
+
+import com.agoda.kakao.screen.Screen
+import com.hedvig.android.owldroid.graphql.EmbarkStoryQuery
+import com.hedvig.app.feature.embark.screens.EmbarkScreen
+import com.hedvig.app.feature.embark.ui.EmbarkActivity
+import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_GRAPHQL_MUTATION_AND_OFFER_REDIRECT
+import com.hedvig.app.testdata.feature.embark.data.STORY_WITH_GRAPHQL_MUTATION_AND_SINGLE_VARIABLE
```
hedvigoscar, line 8:
```suggestion
```

app/src/androidTest/java/com/hedvig/app/feature/embark/api/graphqlmutation/ListVariableTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @get:Rule
+    val apolloMockServerRule = ApolloMockServerRule(
+        EmbarkStoryQuery.QUERY_DOCUMENT to apolloResponse { success(STORY_WITH_GRAPHQL_MUTATION_AND_OFFER_REDIRECT) },
+        VARIABLE_MUTATION to apolloResponse {
+            success(jsonObjectOf("hello" to jsonArrayOf("1", "2", "3")))
+        }
+    )
+
+    private fun jsonArrayOf(vararg items: Any) = JSONArray().apply {
```
hedvigoscar, line 34: We can extract this function to live next to `jsonObjectOf`.

hedvigoscar (approved):
Approved upon addressing comments!

hedvigoscar:
Tests are no longer green though, that needs to be addressed before merging 😬

PR #573, opened 2021-06-09
Delay starting transition to avoid glitchy input layouts

hugokallstrom:
I looked around in the docs and there is a callback [onFinishInflate](https://developer.android.com/reference/android/view/View#onFinishInflate()), but could not get it to work properly. This works even though it is a bit hacky. Damn transitions... 

### Checklist


hedvigoscar (approved):
Weird, but aight! ✅

PR #582, opened 2021-06-16
APP-452: Offer important documents

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
```diff
@@ -0,0 +1,93 @@
...
+    override fun onBindViewHolder(holder: DocumentsViewHolder, position: Int) {
+        when (val item = getItem(position)) {
+            is DocumentItems.Document -> (holder as DocumentViewHolder).bind(item)
+            is DocumentItems.Header -> (holder as TitleViewHolder).bind(item)
+        }
+    }
+
+    abstract class DocumentsViewHolder(view: View) : RecyclerView.ViewHolder(view)
+
+    inner class TitleViewHolder(view: View) : DocumentsViewHolder(view) {
```
hedvigoscar, line 47: Making these `inner class`es isn't necessarily a bad thing, but it does give them a bit more privilege than they necessarily should have (they now have access to the entire `DocumentAdapter` instead of only the `MarketManager` and the `trackClick`-function).

I really liked when you used the approach of passing these dependencies to the `ViewHolder` constructor - what do you think about continuing to use that pattern? 🙂

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
```diff
@@ -0,0 +1,93 @@
...
+        fun bind(header: DocumentItems.Header) {
+            binding.text.text = itemView.context.getString(header.stringRes)
+        }
+    }
+
+    inner class DocumentViewHolder(view: View) : DocumentsViewHolder(view) {
+        private val binding by viewBinding(DocumentBinding::bind)
+
+        fun bind(document: DocumentItems.Document) {
+            val button = binding.button
```
hedvigoscar, line 59: We're only really using the `button`-variable in one place, should we inline this declaration? 🙂

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
```diff
@@ -0,0 +1,93 @@
...
+            val title = document.getTitle(itemView.context)
+            val subTitle = document.getSubTitle(itemView.context)
+
+            binding.text.text = title
+            binding.subtitle.text = subTitle
+            binding.subtitle.isVisible = subTitle != null
+
+            button.setHapticClickListener {
```
hedvigoscar, line 67:
```suggestion
            binding.button.setHapticClickListener {
```

app/src/main/java/com/hedvig/app/feature/documents/DocumentItems.kt
```diff
@@ -0,0 +1,28 @@
+package com.hedvig.app.feature.documents
+
+import android.content.Context
+import androidx.annotation.StringRes
+
+sealed class DocumentItems {
+
+    data class Header(@StringRes val stringRes: Int) : DocumentItems()
```
hedvigoscar, line 8: Here's an interesting dilemma: making this `data class` take a `@StringRes` gives it a tight coupling with the android view-layer 🤔 

I'm not sure that I care enough to think that we should change that, but it's something to think about for sure.

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
```diff
@@ -0,0 +1,93 @@
...
+                // TODO Quick fix for getting new terms and conditions
+                val uri = Uri.parse(
+                    if (marketManager.market == Market.SE &&
+                        document.type == DocumentItems.Document.Type.TERMS_AND_CONDITIONS
+                    ) {
+                        "https://www.hedvig.com/se/villkor"
+                    } else {
+                        document.url
+                    }
+                )
```
hedvigoscar, line 80: Should we lift up this logic to another layer? I think perhaps having the place where we instantiate the `Document` could pre-calculate the `url` for us 🙂

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
```diff
@@ -0,0 +1,93 @@
...
+                try {
+                    it.context.openUri(uri)
+                } catch (throwable: Throwable) {
+                    itemView.context.showAlert(
+                        title = R.string.error_dialog_title,
+                        message = R.string.component_error,
+                        positiveAction = {}
+                    )
+                }
```
hedvigoscar, line 89: Should we lift this routine out to an extension function? 🙂

hedvigoscar (commented):
I really like how this turned out with `ConcatAdapter`, well done!

A few improvements to look at, on the whole very good 😄

app/src/main/java/com/hedvig/app/feature/offer/TermsAdapter.kt
```diff
@@ -29,7 +29,9 @@ class TermsAdapter(
         private val binding by viewBinding(OfferTermsAreaButtonBinding::bind)
         fun bind(terms: OfferQuery.InsuranceTerm, tracker: OfferTracker, marketManager: MarketManager) {
             val button = binding.button
-            button.text = terms.displayName + " ↗"
+            binding.text.text = terms.displayName
+            // TODO: Parse subtitle from bundledquote api
```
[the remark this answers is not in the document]
hugokallstrom: I didn't see this in the design, so thought I'd remove it. I'll ask designers.

app/src/main/res/layout/activity_offer.xml
```diff
@@ -13,11 +13,11 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:clipToPadding="false"
+        android:paddingBottom="120dp"
```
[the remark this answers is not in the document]
hugokallstrom: Its for not hiding content below the sign-button, not the best solution. Do you have any ideas? Can we use insets perhaps?

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
hugokallstrom: Good feedback, fixing

app/src/main/java/com/hedvig/app/feature/documents/DocumentItems.kt
hugokallstrom: Yeah I think it makes more sense though to pass the stringRes than to inject a context in the viewModel creating the items.

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
hugokallstrom: Yes!!!

app/src/main/java/com/hedvig/app/feature/documents/DocumentAdapter.kt
hugokallstrom: Absolutely! :)

PR #585, opened 2021-06-17
APP-461: Query `quoteBundle` instead of `lastQuoteOfMember`

hedvigoscar:
This refactor was highly non-trivial. Damn.

### Checklist


hedvigoscar:
Given the complex scope of this PR, I decided to try to round the PR out here. There are a lot of outstanding issues with the state of the code for the feature right now though, here is a list of what I was able to think of:

- Determine the title in the header (maybe from backend?)
- Add new switching-sections
- Remake perils to work with multiple quotes
- Make choosing start date work with multiple quotes (complex!)
- Add Network error-UI
- Add Loading-UI

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -46,19 +47,40 @@ abstract class OfferViewModel : ViewModel() {
...
     override val autoStartToken = MutableLiveData<SignOfferMutation.Data>()
     override val signStatus = MutableLiveData<SignStatusFragment>()
     override val signError = MutableLiveData<Boolean>()
 
     init {
-        load()
+        if (_quoteIds.isEmpty()) {
+            viewModelScope.launch {
+                val idResult = offerRepository.quoteIdOfLastQuoteOfMember().safeQuery()
+                if (idResult !is QueryResult.Success) {
```
hugokallstrom, line 64: Can we move this OfferRepository (something like `GetAddressChangeStoryIdUseCase`) and return our own Result class, and then map the result to a viewstate? Perhaps thats the scope of another PR :)

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hedvigoscar: I gave it a shot in a new commit, check it out!

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetQuoteIdsUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+) {
+    sealed class Result {
+        data class Success(
+            val ids: List<String>,
+        ) : Result()
+
+        object Error : Result()
+    }
+
+    suspend operator fun invoke(ids: List<String>): Result {
```
hugokallstrom, line 18: Nice! Could we move this request from offerRepository to this useCase as well? 
```
fun offer(ids: List<String>) = apolloClient
        .query(offerQuery(ids))
        .watcher()
        .toFlow()   
```

And then use the result with this function in the viewmodel (passing our Result instead of a Response)
```
    private fun toViewState(response: Response<OfferQuery.Data>): ViewState {
        return response.errors?.let {
            ViewState.Error.GeneralError(it.firstOrNull()?.message)
        } ?: response.data?.let { data ->
            if (data.contracts.isNotEmpty()) {
                ViewState.HasContracts
            } else {
                val items = OfferItemsBuilder.createItems(data)
                ViewState.OfferItems(items)
            }
        } ?: ViewState.Error.EmptyResponse
    }
```

Then we would only need to make changes in this class when removing support for `lastQuoteOfMember`.

hugokallstrom (approved):

PR #589, opened 2021-06-22
APP-520: Revert changes and show web offer activity

hugokallstrom:
### Checklist


hedvigoscar (approved):
We may want to hide the current behaviour behind a feature flag, but let's do that later on.

PR #592, opened 2021-06-28
APP-511: Per-Quote Detail Screen

hedvigoscar:
### Checklist


app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -199,4 +215,21 @@ class OfferViewModelImpl(
             response.getOrNull()?.data?.let { offerRepository.removeStartDateFromCache(quoteIds, it) }
         }
     }
+
```
hugokallstrom, line 218: I think you can notify the view here with some kind of event instead of returning from the function. Can make it tricky to follow if the error result notifies the view but the success result returns as usual.

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -199,4 +215,21 @@ class OfferViewModelImpl(
...
+    override suspend fun getQuoteDetailItems(
+        id: String,
+    ): QuoteDetailItems? = when (val result = getQuoteUseCase(quoteIds, id)) {
+        GetQuoteUseCase.Result.Error -> {
+            _viewState.value = ViewState.Error.GeneralError("")
```
hugokallstrom, line 223: Should we present some kind of general error here?

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -90,11 +93,32 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
             offerToolbar.setNavigationOnClickListener { onBackPressed() }
             offerToolbar.setOnMenuItemClickListener(::handleMenuItem)
 
+            val openQuoteDetails = { quoteID: String ->
+                lifecycleScope.launch {
+                    val quoteDetailItems = model.getQuoteDetailItems(quoteID) ?: return@launch
```
hugokallstrom, line 98: So instead of handling the result here, we update the view with a "open details activity"-event.

hugokallstrom (approved):
Very nice!!

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hedvigoscar: I agree - I'll remake this code to use a more event-based architecture.

I noticed that doing it that way will fix an issue related to the use of `lifecycleScope` too (which dies when the activity dies, ie when rotating the screen/resizing the window).

Additionally, I noticed that we have an issue with how the events are architected today. I'll make a proposal on how to change it and present it to you when it's done 🙂

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -90,33 +103,49 @@ class OfferViewModelImpl(
...
-            } else {
-                val topOfferItems = OfferItemsBuilder.createTopOfferItems(data)
-                val perilItems = OfferItemsBuilder.createPerilItems(data.quoteBundle.quotes)
-                val insurableLimitsItems = OfferItemsBuilder.createInsurableLimits(data.quoteBundle.quotes)
-                val documentItems = OfferItemsBuilder.createDocumentItems(data.quoteBundle.quotes)
-                val bottomOfferItems = OfferItemsBuilder.createBottomOfferItems()
-                ViewState.OfferItems(topOfferItems, perilItems, documentItems, insurableLimitsItems, bottomOfferItems)
-            }
-        } ?: ViewState.Error.EmptyResponse
+    private fun toDataOrError(response: Response<OfferQuery.Data>): Either<OfferQuery.Data, Event> {
```
hugokallstrom, line 134: I think we should try to move this down to `OfferRepository`, and return som kind of `OfferResult` instead of ``Response<OfferQuery.Data>` from `fun offer(ids: List<String>)`. This will be in line with not exposing the network layer to the view layer. Something like this (in `OfferRepository`):

```
fun offer(ids: List<String>) = apolloClient
        .query(offerQuery(ids))
        .watcher()
        .toFlow()
        .map(::toDataOrError) 
```     

What do you think?

hugokallstrom (changes requested):

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hedvigoscar: I see your point.

As the code is written now, the `toDataOrError`-function returns `Either<OfferQuery.Data, OfferViewModel.Event>`. I guess it would not make sense to expose the viewmodel-layer to the network layer by continuing to have the flow return `OfferViewModel.Event` in its either-return. Would it be overkill to introduce one more error-type and map from it to an `Event` at the `ViewModel`-layer? 🤔

hugokallstrom (approved):

PR #593, opened 2021-06-30
APP-498: Offer Redesign, FAQ section

hedvigoscar:
### Checklist


app/src/main/java/com/hedvig/app/feature/offer/ui/OfferModel.kt
```diff
@@ -28,13 +28,18 @@ sealed class OfferModel {
...
     data class QuoteDetails(
         val name: String,
         val id: String,
     ) : OfferModel()
 
-    sealed class Paragraph : OfferModel() {
-        object Coverage: Paragraph()
-    }
+    data class FAQ(
+        val items: List<Pair<String, String>>
```
hugokallstrom, line 41: Could perhaps refactor the Pair to a data class like 
```
data class FAQItem(
  val headLine: String,
  val body: String
)
```
Could make adding/removing fields easier later.

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferModel.kt
hedvigoscar: Agreed.

I think I may actually have made a mental note to do that while implementing this feature, but of course, I forgot it. 😬

PR #595, opened 2021-07-05
APP-542: Set sign method view state

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
@@ -7,6 +7,7 @@ import android.os.Bundle
 import android.transition.TransitionManager
 import android.view.MenuItem
 import android.widget.Toast
+import androidx.core.content.ContextCompat
 import androidx.core.view.doOnNextLayout
 import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
@@ -34,6 +35,7 @@ import com.hedvig.app.feature.perils.PerilsAdapter
 import com.hedvig.app.feature.settings.MarketManager
 import com.hedvig.app.feature.settings.SettingsActivity
 import com.hedvig.app.service.LoginStatusService.Companion.IS_VIEWING_OFFER
+import com.hedvig.app.util.extensions.showErrorDialog
 import com.hedvig.app.util.extensions.startClosableChat
 import com.hedvig.app.util.extensions.storeBoolean
 import com.hedvig.app.util.extensions.view.hide
@@ -100,7 +102,8 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
                 fragmentManager = supportFragmentManager,
                 tracker = tracker,
                 marketManager = marketManager,
-                removeDiscount = model::removeDiscount
+                onRemoveDiscount = model::removeDiscount,
+                onSign = ::onSign
             )
             val perilsAdapter = PerilsAdapter(
                 fragmentManager = supportFragmentManager,
@@ -116,7 +119,8 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
                 fragmentManager = supportFragmentManager,
                 tracker = tracker,
                 marketManager = marketManager,
-                removeDiscount = model::removeDiscount
+                onRemoveDiscount = model::removeDiscount,
+                onSign = ::onSign
             )
 
             val concatAdapter = ConcatAdapter(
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
```diff
@@ -2,10 +2,12 @@ package com.hedvig.app.feature.offer.ui
 
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.content.ContextCompat
```
hedvigoscar, line 5:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -7,6 +7,7 @@ import android.os.Bundle
 import android.transition.TransitionManager
 import android.view.MenuItem
 import android.widget.Toast
+import androidx.core.content.ContextCompat
```
hedvigoscar, line 10:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
```diff
@@ -57,7 +60,7 @@ class OfferAdapter(
     }
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
-        holder.bind(getItem(position), fragmentManager, tracker, removeDiscount, marketManager)
+        holder.bind(getItem(position), fragmentManager, tracker, onRemoveDiscount, onSign, marketManager)
```
hedvigoscar, line 63: Perhaps we could provide `onSign` in the `ViewHolder`-`constructor` instead? You've introduced that pattern before and I think it makes the `ViewHolder`-binding process a whole lot cleaner 🙂

app/src/main/java/com/hedvig/app/feature/offer/ui/SignMethodExt.kt
```diff
@@ -0,0 +1,34 @@
...
+    SignMethod.APPROVE_ONLY,
+    SignMethod.NORWEGIAN_BANK_ID, // Deprecated
+    SignMethod.DANISH_BANK_ID, // Deprecated
+    SignMethod.UNKNOWN__ -> null
+}
+
+fun MaterialButton.bindWithSignMethod(signMethod: SignMethod) {
+    text = context.getString(signMethod.checkoutTextRes())
+    signMethod.checkoutIconRes()?.let {
+        val icon = ContextCompat.getDrawable(context, it)
```
hedvigoscar, line 31:
```suggestion
        val icon = context.compatDrawable(it)
```

app/src/main/java/com/hedvig/app/feature/offer/ui/SignMethodExt.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.hedvig.app.R
+import java.lang.IllegalArgumentException
+
+fun SignMethod.checkoutTextRes() = when (this) {
+    SignMethod.SWEDISH_BANK_ID -> R.string.OFFER_SIGN_BUTTON
+    SignMethod.SIMPLE_SIGN -> R.string.OFFER_CHECKOUT_BUTTON
+    SignMethod.APPROVE_ONLY -> R.string.OFFER_APPROVE_CHANGES
+    SignMethod.NORWEGIAN_BANK_ID, // Deprecated
+    SignMethod.DANISH_BANK_ID, // Deprecated
+    SignMethod.UNKNOWN__ -> throw IllegalArgumentException("Could not parse checkout string for $this")
```
hedvigoscar, line 16: This `throw` makes it so that we need to guard against calling this function on a higher layer, which makes it a bit dangerous 😕 We also do not have this guard in place in the `OfferActivity`, meaning that introducing a new sign method would make any app user opening the Offer Screen crash instead of having a more graceful degradation of their experience.

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -152,28 +156,45 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
+    private fun onSign(signMethod: SignMethod) {
+        when (signMethod) {
+            SignMethod.SWEDISH_BANK_ID -> {
+                binding.signButton.setHapticClickListener {
+                    tracker.floatingSign()
+                    OfferSignDialog.newInstance().show(
+                        supportFragmentManager,
+                        OfferSignDialog.TAG
+                    )
+                }
```
hedvigoscar, line 187: This does not seem right I think - shouldn't we be triggering the `OfferSignDialog` in this case instead of setting up the click listener? 😬

hedvigoscar (changes requested):

app/src/main/java/com/hedvig/app/feature/offer/ui/SignMethodExt.kt
hugokallstrom: Very true, good catch. Fixing that.

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
hugokallstrom: Yes 😅 This is fixed in APP-543!

hedvigoscar (approved):

PR #600, opened 2021-07-07
APP-505: Switching Sections

hedvigoscar:
Previous insurers from Insurely are intentionally scoped out, as we do not
support insurely in the app right now.

### Checklist


hugokallstrom (approved):
👌

PR #601, opened 2021-07-07
Offer discount fix

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
@@ -43,7 +43,8 @@ object OfferItemsBuilder {
                     .fragments
                     .monetaryAmountFragment
                     .toMonetaryAmount(),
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
```diff
@@ -43,7 +43,8 @@ object OfferItemsBuilder {
                     .fragments
                     .monetaryAmountFragment
                     .toMonetaryAmount(),
-                incentiveDisplayValue = null,
+                incentiveDisplayValue = data.redeemedCampaigns.mapNotNull { it.fragments.incentiveFragment.displayValue },
+                hasCampaigns = data.redeemedCampaigns.isNotEmpty(),
```
hugokallstrom, line 47: I think this will contain campaign for eg. Norways bundle? If that is the case we should use some other way of showing "Remove discount". I'll investigate.

hedvigoscar (approved):

PR #602, opened 2021-07-08
APP-507: Offer Redesign, Error State

hedvigoscar:
### Checklist


app/src/main/java/com/hedvig/app/feature/offer/usecase/RefreshQuotesUseCase.kt
```diff
@@ -0,0 +1,19 @@
...
+    private val offerRepository: OfferRepository
+) {
+    sealed class Result {
+        object Success : Result()
+        object Error : Result()
+    }
+
+    suspend operator fun invoke(ids: List<String>) = when (offerRepository.refreshOfferQuery(ids).safeQuery()) {
+        is QueryResult.Success -> Result.Success
+        else -> Result.Error
```
hugokallstrom, line 17: We could use the error message from safeQuery if we want more specific messages

hugokallstrom (approved):

PR #605, opened 2021-07-08
APP-563: Show snackbar when approved quote is successful

hugokallstrom:
### Checklist


hedvigoscar:
This thing doesn't really make sense to me - we already have the result screen, and now a snackbar as well? 🤔

hugokallstrom:
> This thing doesn't really make sense to me - we already have the result screen, and now a snackbar as well? 🤔

You are right. Should not be necessary when we show the success screen 😬 I'll ask Hiro!

hedvigoscar:
Should we close this PR? @hugokallstrom

PR #609, opened 2021-07-12
APP-8: Show address result screens

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/ApproveQuotesUseCase.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class ApproveQuotesUseCase(
+    private val apolloClient: ApolloClient,
+    private val offerRepository: OfferRepository
+) {
+
+    sealed class ApproveQuotesResult {
+        data class Success(val date: LocalDate?) : ApproveQuotesResult()
+        sealed class Error : ApproveQuotesResult() {
+            data class GeneralError(val message: String?) : Error()
```
hedvigoscar, line 18: How about we add a default argument here? Then we don't need to have the form `GeneralError(null)` later 😄 

```suggestion
            data class GeneralError(val message: String? = null) : Error()
```

hedvigoscar (approved):

PR #613, opened 2021-07-13
Checkout progress

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
@@ -4,10 +4,12 @@ import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.transition.TransitionManager
+import androidx.appcompat.app.AlertDialog
 import androidx.core.view.isVisible
 import androidx.core.widget.addTextChangedListener
 import androidx.lifecycle.flowWithLifecycle
 import androidx.lifecycle.lifecycleScope
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import com.google.android.material.textfield.TextInputLayout
 import com.hedvig.app.BaseActivity
 import com.hedvig.app.R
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
```diff
@@ -36,6 +38,7 @@ class CheckoutActivity : BaseActivity(R.layout.activity_checkout) {
     }
     private val viewModel: CheckoutViewModel by viewModel { parametersOf(parameter.quoteIds) }
     private val binding by viewBinding(ActivityCheckoutBinding::bind)
+    private lateinit var progressDialog: AlertDialog
```
hedvigoscar, line 41: How come we're making this `lateinit`? 🤔 Couldn't we create this where it's used?

app/src/main/res/layout/progress_dialog.xml
```diff
@@ -0,0 +1,22 @@
...
+        android:layout_height="@dimen/loading_spinner_height"
+        android:layout_margin="@dimen/base_margin" />
+
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_centerVertical="true"
+        android:layout_margin="@dimen/base_margin"
+        android:layout_toEndOf="@+id/progressBar"
+        android:text="Signing..." />
```
hedvigoscar, line 21: 👮

hedvigoscar (changes requested):

app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
hugokallstrom: Unfortunately not, we need the activity to be created before we can create the dialog

hedvigoscar: I mean, can't we create it where we invoke it? That must be after `onCreate`, right? 🤔

hugokallstrom: We want to be able to dismiss the dialog if we receive an error, thats why I created it here and then just do show/dismiss when handling an event

app/src/main/res/layout/progress_dialog.xml
hugokallstrom: Good catch, created a string res 🙏

hedvigoscar: Alright, fair enough.

app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
hedvigoscar: Alright, fair enough.

PR #614, opened 2021-07-13
Various moving flow fixes

hugokallstrom:
Please check commit messages to see what fixes this PR contains

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
@@ -16,6 +16,7 @@ import com.hedvig.app.feature.claims.ui.commonclaim.EmergencyData
 import com.hedvig.app.feature.home.service.HomeTracker
 import com.hedvig.app.feature.loggedin.ui.LoggedInViewModel
 import com.hedvig.app.feature.loggedin.ui.ScrollPositionListener
+import com.hedvig.app.feature.settings.Market
 import com.hedvig.app.feature.settings.MarketManager
 import com.hedvig.app.util.FeatureFlag
 import com.hedvig.app.util.extensions.view.updatePadding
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -157,7 +158,7 @@ class HomeFragment : Fragment(R.layout.home_fragment) {
                     if (pendingAddress != null && pendingAddress.isNotBlank()) {
                         add(HomeModel.PendingAddressChange(pendingAddress))
                     }
-                    if (FeatureFlag.MOVING_FLOW.enabled) {
+                    if (FeatureFlag.MOVING_FLOW.enabled && marketManager.market == Market.SE) {
```
hedvigoscar, line 161: Should we move this logic up to the feature flag itself? Now we're spreading it out in various places in the code, right? 🤔

app/src/main/res/layout/activity_checkout.xml
```diff
@@ -100,7 +100,7 @@
                 android:hint="@string/simple_sign_login.text_field.label"
                 app:layout_constraintStart_toStartOf="parent"
                 app:layout_constraintTop_toBottomOf="@id/emailInputContainer"
-                app:placeholderText="DDMMYYXXXX">
+                app:placeholderText="@string/simple_sign_login.text_field.helper_text">
```
hedvigoscar, line 103: 👏

hedvigoscar (approved):
Approved upon addressing question 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
hugokallstrom: I was thinking about that, but the feature flag will be removed when we release moving flow and we want to keep the market logic - so thats why they are separated here.

hedvigoscar: Perhaps we should keep the flag until we reach general availability of the feature? 🙂

PR #616, opened 2021-07-13
Feature flags

hugokallstrom:
### Checklist


hedvigoscar:
I'm kind of confused about the architecture of this feature flag-implementation 🤔 perhaps we could discuss it later today?

hugokallstrom:
> I'm kind of confused about the architecture of this feature flag-implementation 🤔 perhaps we could discuss it later today?

Yes!

hedvigoscar:
As mentioned in our call earlier today:

- Should we re-architect this solution using dependency injection? 🙂
- Should we try to isolate the debug feature flags into the debug-source set, in order to not accidentally load them in the production environment? 🤔

PR #626, opened 2021-07-16
APP-615: Table headers

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/table/TableAdapter.kt
```diff
@@ -19,11 +19,11 @@ class TableAdapter : ListAdapter<TableAdapter.UpcomingAgreementItem, TableAdapte
 
     fun setTable(table: Table) {
         submitList(
-            listOf(
```
hedvigoscar, line 22: A pattern I've adopted when the declarative approach becomes a bit messy is to simply create an `ArrayList` and use `if`s to conditionally `add` things to the final list. Could be something to consider here if you agree 😄

hedvigoscar (approved):
Approved! Acting on comment is optional

PR #641, opened 2021-07-26
APP-718: Termination text

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceContractExtensions.kt
+++ b/app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceContractExtensions.kt
@@ -122,7 +122,7 @@ fun InsuranceQuery.Contract.bindTo(
     }
 
 private fun InsuranceContractCardBinding.removeBackgroundColor() {
-    container.setBackgroundColor(container.context.colorAttr(android.R.attr.colorBackground))
+    container.setBackgroundColor(container.context.getColor(R.color.hedvig_light_gray))
     blur.remove()
 }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceContractExtensions.kt
```diff
@@ -12,19 +14,24 @@ import com.hedvig.app.util.extensions.view.show
...
                 firstStatusPill.setText(R.string.DASHBOARD_INSURANCE_STATUS_INACTIVE_NO_STARTDATE)
+                typeOfContract.bindBackgroundColor(this)
             }
             contractStatus.asActiveInFutureStatus?.let { activeInFuture ->
                 firstStatusPill.show()
                 firstStatusPill.text = firstStatusPill.resources.getString(
                     R.string.DASHBOARD_INSURANCE_STATUS_INACTIVE_STARTDATE,
                     dateTimeFormatter.format(activeInFuture.futureInception)
                 )
+                removeBackgroundColor()
```
hedvigoscar, line 34: While we're touching this code - I think we should use the gray background color that's been requested in other tickets instead of removing the background color 😄 It's linked in APP-667.

app/src/main/res/layout/contract_detail_activity.xml
```diff
@@ -50,6 +51,16 @@
                         android:id="@+id/cardContainer"
                         layout="@layout/insurance_contract_card" />
 
+                    <TextView
+                        android:id="@+id/termination_info"
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:padding="@dimen/base_margin_double"
+                        android:text="@string/INSURANCE_DETAILS_TERMINATION_INFO"
+                        android:textSize="13sp"
```
hedvigoscar, line 60: We should use a `textAppearance` here I think, instead of inlining a `textSize` - looks like the specified text style does not exist in our design system, we should probably check that with Hiro 😄

hedvigoscar (approved):
Approved upon addressing comments 🙂

PR #644, opened 2021-07-27
APP-648: Use new chat button for offer and logged in activity

hugokallstrom:
device-2021-07-27-111900

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
@@ -6,6 +6,7 @@ import android.graphics.drawable.PictureDrawable
 import android.os.Bundle
 import android.transition.TransitionManager
 import android.view.MenuItem
+import androidx.core.view.get
 import androidx.core.view.isVisible
 import androidx.lifecycle.flowWithLifecycle
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -262,11 +263,17 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
-        binding.offerToolbar.menu.clear()
+        val menu = binding.offerToolbar.menu
+        menu.clear()
         when (loginStatus) {
             LoginStatus.ONBOARDING,
             LoginStatus.IN_OFFER -> binding.offerToolbar.inflateMenu(R.menu.offer_menu)
-            LoginStatus.LOGGED_IN -> binding.offerToolbar.inflateMenu(R.menu.offer_menu_logged_in)
+            LoginStatus.LOGGED_IN -> {
+                binding.offerToolbar.inflateMenu(R.menu.offer_menu_logged_in)
+                menu.getItem(0).actionView.setOnClickListener {
```
hedvigoscar, line 273: Does this not happen automatically? 😮

app/src/main/res/drawable/ic_chat_background.xml
```diff
@@ -0,0 +1,15 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
hedvigoscar, line 1: Has this `VectorDrawable` been `avocado`ified?

Also, should it use color attributes, instead of hard-coded colors?

hedvigoscar (approved):

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
hugokallstrom: I think its some weird behaviour because we are using an action view for the item. We are having the same behaviour in LoggedInActivity.

hedvigoscar: Ah, makes sense!

PR #648, opened 2021-07-28
APP-606: Support action-level API calls

hedvigoscar:
This implementation uses the same code path as passage-level API.
It's not ideal, but it will get the job done for now!

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
@@ -103,7 +103,23 @@ abstract class EmbarkViewModel(
         }.flatten()
     }
 
-    fun navigateToPassage(passageName: String) {
+    fun submitAction(nextPassageName: String, submitIndex: Int = 0) {
+        data.value?.passage?.let { currentPassage ->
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -103,7 +103,23 @@ abstract class EmbarkViewModel(
         }.flatten()
     }
 
-    fun navigateToPassage(passageName: String) {
+    fun submitAction(nextPassageName: String, submitIndex: Int = 0) {
```
hugokallstrom, line 106: Okay so this is where the magic happens ;) I kind of understand, but still having a hard time wrapping my head around embark. So nice to have tests though, good job!

hugokallstrom (approved):
LGTM!

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
hedvigoscar: That is indeed where the magic happens.

Embark could indeed do with a nice refactor, it's grown a bit hairy. 😬

PR #649, opened 2021-07-28
APP-665: Refactor LoginStatusService

hugokallstrom:
Refactor LoginStatusService, hide sharedPreferences implementation. 

### Checklist


app/src/main/java/com/hedvig/app/feature/chat/viewmodel/UserViewModel.kt
```diff
@@ -42,15 +46,25 @@ class UserViewModel(
         }
     }
 
-    fun logout(callback: () -> Unit) {
+    fun logout(context: Context, callback: () -> Unit) {
```
hedvigoscar, line 49: While we're at it - should we make this a `suspend fun` instead of being callback-based?

Also, should we remove the dependency on `context` for the caller, and instead allow `LoginStatusService` to handle the deletion of the `authenticationToken`?

app/src/main/java/com/hedvig/app/service/LoginStatusService.kt
```diff
@@ -1,44 +1,52 @@
...
+import android.content.SharedPreferences
 import com.apollographql.apollo.ApolloClient
 import com.apollographql.apollo.coroutines.await
 import com.hedvig.android.owldroid.graphql.ContractStatusQuery
 import com.hedvig.app.util.extensions.getAuthenticationToken
-import com.hedvig.app.util.extensions.getStoredBoolean
-import com.hedvig.app.util.extensions.isLoggedIn
-import com.hedvig.app.util.extensions.setIsLoggedIn
+
+private const val SHARED_PREFERENCE_IS_LOGGED_IN = "shared_preference_is_logged_in"
```
hedvigoscar, line 10: Should we put these in the `companion object`?

hedvigoscar (approved):

app/src/main/java/com/hedvig/app/feature/chat/viewmodel/UserViewModel.kt
hugokallstrom: Created a LogoutUseCase and removed the callback

app/src/main/java/com/hedvig/app/service/LoginStatusService.kt
hugokallstrom: I think it makes sense to have top level here since we only have one class in the file, feels unnecessary to create companion object overhead

app/src/main/java/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
```diff
@@ -293,4 +313,17 @@ class ChatViewModel(
             gifs.postValue(response.getOrNull()?.data)
         }
     }
+
+    fun restartChat() {
```
hugokallstrom, line 317: This is a bit gnarly, tried to keep the same logic as in /develop but still dont get a proper restart. I think we should look into this more since it feels somewhat broken both in this branch and in develop - first the app restarts and after a delay the chat starts again, but with no option to go back.

app/src/main/java/com/hedvig/app/service/LoginStatusService.kt
hedvigoscar: My preference isn't super-big towards `companion object`s, the only thing that I think talks particularly in favor of `companion object`s are that Android Studio lists the file as `LoginStatusService` instead of `LoginStatusService.kt` when there's only one class in the file 😅 

What kind of overhead are you talking about for the `companion object`? The added lines, or something the compiler generates? 🤔

app/src/main/java/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
hedvigoscar: I agree that this is probably broken, unfortunately. 😞

app/src/main/java/com/hedvig/app/service/LoginStatusService.kt
hugokallstrom: Thats a good point :) I was thinking both the added lines and the added java code generated by the compiler, but lets go with companion objects for now since that's what we're used to

app/src/androidTest/java/com/hedvig/app/feature/loggedin/TerminatedTest.kt
```diff
@@ -45,7 +45,7 @@ class TerminatedTest : TestCase() {
 
     @Test
     fun shouldOpenWithHomeTabWhenUserIsNotTerminated() = run {
-        every { loginStatusService.isLoggedIn }.returns(true)
+        every { loginStatusService.isLoggedIn }.returns(false)
```
hedvigoscar, line 48: This doesn't seem to be in line with what the test says 😮

app/src/main/java/com/hedvig/app/service/LoginStatusService.kt
hedvigoscar: Out of curiousity, I checked what the compiler generated for either case (`companion object` vs file-level `const`s) - they generate more or less the same code, actually - file-level `const`s will generate something like this:

```kotlin
static class ${FileName}Kt {
    private static final CONST_VARIABLE = "CONST_VARIABLE";
}
```

while `companion object` generates something like this:

```kotlin

class FileNameKt { // Parent class
    static class Companion {
        private static final CONST_VARIABLE = "CONST_VARIABLE";
    }
}
```

hugokallstrom: Cool!

app/src/androidTest/java/com/hedvig/app/feature/loggedin/TerminatedTest.kt
hugokallstrom: Not really sure what this is testing but it was actually returning isLoggedIn == false before refactoring 🤔

hedvigoscar: It's testing that terminated members get to see the logged in-screen with the terminated-state. Is this no longer the case after the refactor?

PR #650, opened 2021-07-29
APP-730: Add feature flag provider and inject when necessary

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/util/featureflags/FeatureRuntimeBehavior.kt
```diff
@@ -0,0 +1,32 @@
+package com.hedvig.app.util.featureflags
+
+import androidx.annotation.VisibleForTesting
+import com.hedvig.app.feature.settings.MarketManager
+import java.util.concurrent.CopyOnWriteArrayList
+
+class FeatureRuntimeBehavior(
```
hedvigoscar, line 7: This name still doesn't entirely make sense to me 😅 It appears to have the role of providing feature flags, I kind of wish it was named `FeatureFlagProvider` 😬

app/src/main/java/com/hedvig/app/util/featureflags/TestFeatureFlagProvider.kt
```diff
@@ -0,0 +1,16 @@
+package com.hedvig.app.util.featureflags
+
+object TestFeatureFlagProvider : FeatureFlagProvider {
```
hedvigoscar, line 3: Is this still needed when we're using DI? I guess we can probably just mock on the DI-level:

```kotlin
val mock = mocck<FeatureRuntimeBehavior>()
every { mock.isFeatureEnabled(Feature.MOVING_FLOW) } returns true
```

app/src/main/java/com/hedvig/app/util/featureflags/FeatureRuntimeBehavior.kt
```diff
@@ -0,0 +1,32 @@
...
+    val marketManager: MarketManager,
+    isDebugBuild: Boolean
+) {
+
+    @VisibleForTesting
+    internal val providers = CopyOnWriteArrayList<FeatureFlagProvider>()
+
+    init {
+        if (isDebugBuild) {
```
hedvigoscar, line 16: It would be super-cool to provide these via multi-binding and have them reside only in the appropriate source sets, but Koin doesn't really support multi-binding 😞 I think we can skip it for now, but it would be really cool!

hedvigoscar (approved):
Few questions, but mostly ready to merge I think!

app/src/main/java/com/hedvig/app/util/featureflags/FeatureRuntimeBehavior.kt
hugokallstrom: Agreed, changing :)

app/src/main/java/com/hedvig/app/util/featureflags/TestFeatureFlagProvider.kt
hugokallstrom: Right, that makes more sense.

PR #652, opened 2021-08-02
Remove tabs transition

hugokallstrom:
I noticed the transition between tabs were quite glitchy. This is because we need to postpone enter transitions, and wait until the recylerview in the enter fragment has been laid out. After trying different approaches for this and getting it to work quite well, I removed the transition just to compare the differences. Conclusion: the tabs feel more responsive and less glitchy when disabling the transitions altogether.

Please try it out and see what you think. I could spend some more time getting the original transitions to work if we really want to, but I think this is works well - especially when we have loading states for each tab. 

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
@@ -7,8 +7,6 @@ import android.os.Bundle
 import android.os.Handler
 import android.view.Menu
 import android.view.MenuItem
-import android.view.Window
-import androidx.core.view.WindowCompat
 import androidx.dynamicanimation.animation.FloatValueHolder
 import androidx.dynamicanimation.animation.SpringAnimation
 import androidx.dynamicanimation.animation.SpringForce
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -127,7 +124,7 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
                 supportFragmentManager
                     .beginTransaction()
                     .replace(R.id.tabContent, id.fragment)
-                    .commitAllowingStateLoss()
+                    .commit()
```
hedvigoscar, line 127: Is this correct? 🤔

hedvigoscar (commented):
Agreed, this hasn't been looking totally correct in a long time.

hedvigoscar (approved):

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
hugokallstrom: Good catch, fixing

PR #655, opened 2021-08-03
APP-727: In app review dialog

hugokallstrom:
Used datastore instead of shared preferences here, mostly because of built-in flow support. Works really good! Will need to test the review dialog in a test track though, not possible from a debug build. 

### Checklist


app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
```diff
@@ -35,9 +40,14 @@ abstract class LoggedInViewModel : ViewModel() {
 }
 
 class LoggedInViewModelImpl(
-    private val loggedInRepository: LoggedInRepository
+    private val loggedInRepository: LoggedInRepository,
+    chatClosedTracker: ChatEventStore
 ) : LoggedInViewModel() {
 
+    override val shouldOpenReviewDialog = chatClosedTracker.observeChatClosedCounter()
```
hugokallstrom, line 47: Thinking if we should reset this counter somewhere? Perhaps after a new version?

app/build.gradle.kts
```diff
@@ -156,6 +158,9 @@ dependencies {
     implementation(Libs.materialComponents)
     implementation(Libs.flexbox)
 
+    implementation(Libs.playCore)
+    implementation(Libs.playKtx)
```
hedvigoscar, line 162: It's often possible to only include the `ktx`-artifact, and get both artifacts for the price of one. Try it out! 🙂

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
hedvigoscar: It's probably fine as is, for now. Could be wise to reset it on logging out though, I think!

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
```diff
@@ -21,6 +24,8 @@ abstract class LoggedInViewModel : ViewModel() {
     private val _toolbarInset = MutableLiveData<Int>()
     val toolbarInset: LiveData<Int> = _toolbarInset
 
+    abstract val shouldOpenReviewDialog: LiveData<Boolean>
```
hedvigoscar, line 27: This thing looks like it may be representing an event, rather than data, right? 🤔 There's a tricky problem with `LiveData` and events, in that any data represented by `LiveData` needs to be repeatable on configuration change. If not, some unpleasant bugs can occur. Imagine the following scenario:

1. We publish `true` on this `LiveData`
2. The user gets the rating dialog
3. The user dismisses the dialog
4. The user rotates the screen/resizes the window/changes the device theme
5. The dialog re-appears, because the `LiveData` re-triggers post-configuration change

The solution I've employed most times to remedy this issue is to use `SharedFlow`, which is more appropriate for events (that should not repeat after configuration change).

hedvigoscar: Additionally, I've been favoring `StateFlow` and `SharedFlow` as a replacement entirely for `LiveData` - iirc, Google now recommends these over `LiveData` for Kotlin-codebases 😉

hedvigoscar: Finally, should we make this non-abstract? I've been favoring the producer-consumer pairs of mutable/non-mutable in the abstract class, as we generally never put anything but an underlying `Mutable{StateFlow|SharedFlow|LiveData}` in there at the end of the day 🤷

app/src/main/java/com/hedvig/app/feature/chat/data/ChatEventStore.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.intPreferencesKey
+import e
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.map
+
+interface ChatEventStore {
```
hedvigoscar, line 12: Are we getting any mileage out of segregating the `interface` here? 🤔 Usually, I find that `mockk` can replicate the interface of any given `class` without any real hassle, and as such I've mostly skipped extracting an interface, unless I have a few underlying implementations that I want to support 🙂

buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -53,6 +58,9 @@ object Libs {
     const val combineTuple = "com.github.Zhuinden:livedata-combinetuple-kt:1.2.1"
     const val fragmentViewBindingDelegate = "com.github.Zhuinden:fragmentviewbindingdelegate-kt:1.0.0"
 
+    const val playCore = "com.google.android.play:core:1.10.0"
+    const val playKtx = "com.google.android.play:core-ktx:1.8.1"
```
hedvigoscar, line 62: Is this a version mismatch? 🤔

app/src/main/java/com/hedvig/app/util/extensions/ActivityExtensions.kt
```diff
@@ -156,3 +159,21 @@ fun Activity.makeACall(uri: Uri) {
...
     startActivity(intent)
 }
+
+suspend fun Activity.showReviewDialog(
+    ratingsTracker: RatingsTracker,
+    delayMillis: Long = 2000
+) {
+    delay(delayMillis)
+    val manager = ReviewManagerFactory.create(this)
+    val request = manager.requestReviewFlow()
```
hedvigoscar, line 169: Was there a coroutine-based extension for this one in Play Core-ktx?

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
hugokallstrom: Fixed, thanks for the comments!

app/src/main/java/com/hedvig/app/feature/chat/data/ChatEventStore.kt
hugokallstrom: I suppose this is more future proofing, if we want to change the implementation later its nice to have an interface in place. Also nice if we want to inject debug implementations.

app/src/main/java/com/hedvig/app/util/extensions/ActivityExtensions.kt
hugokallstrom: Not that I'm aware of

PR #658, opened 2021-08-05
Catch numberformatexception and try to parse float

hugokallstrom:
Fixes this crash: https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/4ad3af41ecc1cce93bb294d93a998cd3?time=last-seven-days&sessionEventKey=610AD92A0080000145F345392B62FF70_1571086014019419984

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/util/VariableExt.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/util/VariableExt.kt
@@ -5,8 +5,8 @@ import com.hedvig.android.owldroid.type.EmbarkAPIGraphQLSingleVariableCasting
 import com.hedvig.android.owldroid.type.EmbarkAPIGraphQLVariableGeneratedType
 import com.hedvig.app.util.toJsonArray
 import com.hedvig.app.util.toJsonObject
-import org.json.JSONObject
 import java.util.UUID
+import org.json.JSONObject
 
 fun GraphQLVariablesFragment.Variable.toFragment() = GraphQLVariablesFragment(
     asEmbarkAPIGraphQLSingleVariable = asEmbarkAPIGraphQLSingleVariable1?.let { single ->
@@ -42,7 +42,13 @@ fun GraphQLVariablesFragment.AsEmbarkAPIGraphQLSingleVariable.createSingleVariab
 ): Pair<String, Any>? {
     val value = when (as_) {
         EmbarkAPIGraphQLSingleVariableCasting.STRING -> storeValue
-        EmbarkAPIGraphQLSingleVariableCasting.INT -> storeValue.toInt()
+        EmbarkAPIGraphQLSingleVariableCasting.INT -> {
+            try {
+                storeValue.toInt()
+            } catch (exception: NumberFormatException) {
+                storeValue.toFloat()
+            }
+        }
         EmbarkAPIGraphQLSingleVariableCasting.BOOLEAN -> storeValue.toBoolean()
         // Unsupported generated types are ignored for now.
         EmbarkAPIGraphQLSingleVariableCasting.UNKNOWN__ -> null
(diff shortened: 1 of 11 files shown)


hedvigoscar (commented):
What's the root cause of this issue? How come we are getting floats in the first place? Is this the correct layer to handle the issue, or should it be handled more closely to where we gather the input?

hedvigoscar:
Also, there are some red actions, make sure to address them 😉

hugokallstrom:
> What's the root cause of this issue? How come we are getting floats in the first place? Is this the correct layer to handle the issue, or should it be handled more closely to where we gather the input?

Yeah good point. Where do you think we get float values from? User input?

hedvigoscar:
Good question. Perhaps the input value will depend on vendor/keyboard on what kind of type its entered as? Currently, we validate using `toIntOrNull`, but we save with a non-coerced value (https://github.com/HedvigInsurance/android/blob/9f93f9eca3eb1ffb3cbd4dfa70aa2a64b5363f0f/app/src/main/java/com/hedvig/app/feature/embark/passages/numberactionset/NumberActionViewModel.kt#L18 and https://github.com/HedvigInsurance/android/blob/9f93f9eca3eb1ffb3cbd4dfa70aa2a64b5363f0f/app/src/main/java/com/hedvig/app/feature/embark/passages/numberactionset/NumberActionViewModel.kt#L37 respectively). We could perhaps coerce it before saving, that would make sure that we have valid data 'within the bounds' of the system, so to speak 🙂

hugokallstrom:
Good ideas. I also found this after some digging: 
```
private fun evaluateAbstractExpression(expression: Expression, store: HashMap<String, String>): String =
        when (expression) {
            is Expression.StoreKeyExpression -> store[expression.key] ?: ""
            is Expression.NumberConstantExpression -> expression.constant.toBigDecimal().stripTrailingZeros().toString()
...
```

If we have a computed store value we do `toBigDecimal`, can this cause problems if we have the same key when extracting variables for an api call?

hedvigoscar:
Interesting. Yes, that can probably cause issues as well, for sure.

hugokallstrom:
> Interesting. Yes, that can probably cause issues as well, for sure.

Do you think we need the bigdecimal conversion there? Are we doing floating point operations for expressions? I tried switching to int and all tests pass 🤔

app/src/main/java/com/hedvig/app/feature/embark/util/VariableExt.kt
```diff
@@ -42,7 +42,14 @@ fun GraphQLVariablesFragment.AsEmbarkAPIGraphQLSingleVariable.createSingleVariab
 ): Pair<String, Any>? {
     val value = when (as_) {
         EmbarkAPIGraphQLSingleVariableCasting.STRING -> storeValue
-        EmbarkAPIGraphQLSingleVariableCasting.INT -> storeValue.toInt()
+        EmbarkAPIGraphQLSingleVariableCasting.INT -> {
+            try {
+                storeValue.toInt()
+            } catch (exception: NumberFormatException) {
+                // The stored value can in some cases be floats, eg. for computed store values
+                storeValue.toFloat()
```
hedvigoscar, line 50: We're requested to cast the variable to `Int`, is this really correct? I think we should probably try to ensure the correct data types on the other border of the data - as the data enters the system - not as the data leaves the system 🙂

hedvigoscar (changes requested):

app/src/main/java/com/hedvig/app/feature/embark/util/VariableExt.kt
hugokallstrom: Yeah agree, but I think that means that we can't return floats from the expression evaluator. Please see my comment above!

hedvigoscar: I see. Still, this would have to be 
```suggestion
                storeValue.toFloat().toInt()
```

right? If not, we would get an upstream API error, providing the API with a `Float` when an `Int` is required.

hedvigoscar (approved):

PR #659, opened 2021-08-05
APP-740: Do not use delegate when binding view

hugokallstrom:
Trying to fix this: https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/dcb21560ba705b789ee6f32fffa67fdf?time=last-thirty-days&sessionEventKey=61093E3B024A00012F17C310AF9B5648_1570634230566486193

I think it may have something to do with starting bankId and then returning to this fragment and trying to access bindings which no longer exist (because view no longer exists?). Tried using "dont keep activities" with no success. Removing the binding delegate was the only thing I could think of worth trying.

Do you have any ideas what might cause this?

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/trustly/TrustlyConnectFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/trustly/TrustlyConnectFragment.kt
@@ -22,12 +22,10 @@ import com.hedvig.app.feature.connectpayin.ConnectPaymentViewModel
 import com.hedvig.app.feature.connectpayin.TransitionType
 import com.hedvig.app.feature.connectpayin.showConfirmCloseDialog
 import com.hedvig.app.util.onBackPressedCallback
-import com.zhuinden.fragmentviewbindingdelegatekt.viewBinding
 import org.koin.androidx.viewmodel.ext.android.sharedViewModel
 import org.koin.androidx.viewmodel.ext.android.viewModel
 
 class TrustlyConnectFragment : Fragment(R.layout.trustly_connect_fragment) {
-    private val binding by viewBinding(TrustlyConnectFragmentBinding::bind)
     private val trustlyViewModel: TrustlyViewModel by viewModel()
     private val connectPaymentViewModel: ConnectPaymentViewModel by sharedViewModel()
 
@@ -66,6 +64,8 @@ class TrustlyConnectFragment : Fragment(R.layout.trustly_connect_fragment) {
                 })
             )
         }
+
+        val binding = TrustlyConnectFragmentBinding.bind(view)
 
         binding.apply {
             toolbar.apply {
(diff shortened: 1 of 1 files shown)


hedvigoscar:
I've looked into this issue before (but not managed to completely solve it) - and I suspect that the issue is that `WebViewClient` can be called out-of-lifecycle. With this fix in place, I still think we might crash, as the underlying view should still be destroyed (or perhaps we're leaking the view at this point?). Perhaps some guard rails against out-of-lifecycle calls could remedy the issue?

app/src/main/java/com/hedvig/app/feature/trustly/TrustlyConnectFragment.kt
```diff
@@ -107,14 +111,19 @@ class TrustlyConnectFragment : Fragment(R.layout.trustly_connect_fragment) {
...
-                                binding.root,
-                                MaterialFadeThrough()
-                            )
-                            loadingContainer.isVisible = false
-                            trustly.isVisible = true
-                            hasLoadedWebView = true
+                        lifecycleScope.launch {
+                            // Need to check lifecycle here since we are starting other activities in onPageStarted
+                            viewLifecycleOwner.withStateAtLeast(Lifecycle.State.CREATED) {
+                                if (!hasLoadedWebView) {
```
hugokallstrom, line 117: I added some lifecycle checks here. Do you think this could work?

hedvigoscar: Probably. Maybe we can simplify it to `if (viewLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.Created) { ... }`? I worry that creating a coroutine scope here might complicate things even more.

hugokallstrom: Sounds good

app/src/main/java/com/hedvig/app/feature/trustly/TrustlyConnectFragment.kt
```diff
@@ -107,14 +108,16 @@ class TrustlyConnectFragment : Fragment(R.layout.trustly_connect_fragment) {
...
                         super.onPageFinished(view, url)
-                        if (!hasLoadedWebView) {
-                            TransitionManager.beginDelayedTransition(
-                                binding.root,
-                                MaterialFadeThrough()
-                            )
-                            loadingContainer.isVisible = false
-                            trustly.isVisible = true
-                            hasLoadedWebView = true
+                        if (viewLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.CREATED)) {
```
hugokallstrom, line 111: Changed to doing this synchronously

hedvigoscar: Looks good! We can simplify the if-statement to one-level probably, but this is good enough to merge 😄

hedvigoscar (approved):

PR #664, opened 2021-08-09
APP-743: Show embark error messages

hugokallstrom:
Not sure if checking action == null && messages.notEmpty is enough here. I want to not show the "update app" screen and instead show the error messages, for example if we input co-insured > 6. What do you think? 

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
@@ -21,6 +21,8 @@ import com.hedvig.app.util.getWithDotNotation
 import com.hedvig.app.util.plus
 import com.hedvig.app.util.safeLet
 import com.hedvig.app.util.toStringArray
+import java.util.Stack
+import kotlin.math.max
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -21,6 +21,8 @@ import com.hedvig.app.util.getWithDotNotation
 import com.hedvig.app.util.plus
 import com.hedvig.app.util.safeLet
 import com.hedvig.app.util.toStringArray
+import java.util.Stack
```
hedvigoscar, line 24: This change seems to happen frequently on your commits 🤔 maybe there's something strange in your IDE settings? 🤔

hedvigoscar (commented):
I'm not sure that I'm understanding what we're doing here - is there no action on these passages? If so, I think that's an issue in the underlying embark passage we're attempting to render 🤔

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
hugokallstrom: Oh, will check it out

hugokallstrom:
> I'm not sure that I'm understanding what we're doing here - is there no action on these passages? If so, I think that's an issue in the underlying embark passage we're attempting to render 🤔

Exactly, we dont get any action. How is the flow supposed to be? Can we fix it in angel?

hedvigoscar:
We should indeed be able to fix it in angel - every passage is required to provide an action. A lack of action is an indication of incompatibility - that there has been a new action implemented upstream, which we do not yet support - hence requesting that the user updates the app is correct.

PR #668, opened 2021-08-10
Add dependencies for jetpack compose

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/ui/compose/Theme.kt
```diff
@@ -0,0 +1,50 @@
...
+    onError = Color.White,
+    isLight = false
+)
+
+@Composable
+fun HedvigTheme(
+    darkTheme: Boolean = isSystemInDarkTheme(),
+    content: @Composable () -> Unit
+) {
+    MaterialTheme(
```
hedvigoscar, line 46: Should we use `MdcTheme` from https://material-components.github.io/material-components-android-compose-theme-adapter/ to bridge our existing styles to `compose`? 🙂

PR #678, opened 2021-08-12
Do not show "when current plan expires" if user has selected start date

hugokallstrom:
### Checklist


hedvigoscar (approved):
Nice!

Bonus points if you run `ktlint "\!**/generated/**" "**/src/**/*.kt*"` before merging 😄

hugokallstrom:
> Also, looks like the unit tests didn't like this change 🙈

Forgot we had unit tests for this 😅 fixed!

PR #681, opened 2021-08-17
APP-750: Insurely intro screen

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/PreviousInsurerFragment.kt
```diff
@@ -45,7 +61,11 @@ class PreviousInsurerFragment : Fragment(R.layout.previous_insurer_fragment) {
                 onShowInsurers()
             }
             continueButton.setHapticClickListener {
-                onContinue()
+                if (featureManager.isFeatureEnabled(Feature.INSURELY_EMBARK)) {
```
hedvigoscar, line 64: I think the insurely-functionality needs to be split out of this fragment - this fragment powers `EmbarkPreviousInsuranceProviderAction`, while insurely should be shown upon `EmbarkExternalInsuranceProviderAction`. 😬

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.material.Scaffold
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import com.hedvig.app.R
+import com.hedvig.app.ui.compose.composables.TopAppBarWithBack
+import com.hedvig.app.ui.compose.theme.HedvigTheme
+
+class AskForPriceInfoActivity : ComponentActivity() {
```
hedvigoscar, line 18: Should this use `BaseActivity`? If not, I believe we might be loading the incorrect locale here.

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+                        parameter.selectedInsuranceProvider,
+                        onSkipRetrievePriceInfo = ::finishWithResult,
+                        onNavigateToRetrievePrice = ::startRetrievePriceActivity
+                    )
+                }
+            }
+        }
+    }
+
+    private fun startRetrievePriceActivity() {
```
hedvigoscar, line 48: Should we use activity-based screens here? Perhaps we could experiment with Composables as screens for this flow? 🙂

app/src/main/java/com/hedvig/app/ui/compose/theme/Colors.kt
```diff
@@ -13,4 +13,7 @@ val lavender400 = Color(0xffBE9BF3)
 val errorLight = Color(0xffDD2727)
 val errorDark = Color(0xffE24646)
 
+val textColorPrimary = Color(0xAB121212)
```
hedvigoscar, line 16: We should probably use https://material-components.github.io/material-components-android-compose-theme-adapter/, in order to avoid having multiple sources of truth for what our theme is 😅

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoContent.kt
```diff
@@ -0,0 +1,60 @@
...
+            modifier = Modifier.padding(top = baseMarginQuadruple),
+            text = stringResource(R.string.insurely_confirmation_continue_button_text),
+            onClick = onNavigateToRetrievePriceInfo
+        )
+        SecondaryTextButton(
+            text = stringResource(R.string.insurely_intro_skip_button_text),
+            onClick = onSkipRetrievePriceInfo
+        )
+        Text(
+            text = "Genom att klicka på \"Ja, hämta info\" godkänner du att Hedvig hämtar data från Trygg Hansa " +
```
hedvigoscar, line 54: We should use a `stringResource` here

hedvigoscar (commented):
Exciting to finally get to use Compose! 🚀

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/PreviousInsurerFragment.kt
hugokallstrom: Aha! So previous insurance and external insurance are different use cases? What are the difference?

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
hugokallstrom: Thats a really good idea, I'll look into it

app/src/main/java/com/hedvig/app/ui/compose/theme/Colors.kt
hugokallstrom: Yes we are using the adapter now! These colors are unused, I'll remove them.

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
hugokallstrom: Good catch, fixing

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/PreviousInsurerFragment.kt
hedvigoscar: They are indeed different! The former merely allows the selection of your previous insurer (no additional interactions), while the latter combines selection of your previous insurer with the possibility of fetching some details about that former insurer with Insurely.

hugokallstrom: I see. Can we look at `InsuranceProvider.hasExternalCapabilities` and only start the insurely flow if its set? And then use the same embark fragment (`EmbarkPreviousInsuranceProviderAction`)

hedvigoscar: I think we need to look at whether the action in the passage is `EmbarkPreviousInsuranceProviderAction` or `EmbarkExternalInsuranceProviderAction` and use that as the basis for whether we should show the insurely UI, but we can probably re-use the same fragment for the implementation 🙂

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
hugokallstrom: Looked into this and we need to target sdk version 31, which causes some problems at the moment. Some of our dependencies like adyen and shake bugs needs to have `exported = true/false` for each activity in their respective manifest. This field is missing, hence we get

`android:exported needs to be explicitly specified for <activity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined.`

There seems to be some work arounds for this but I'm not sure how stable those are. Should we dig into this more or just wait with upgrading to 31?

hedvigoscar: Wait, how come we need to upgrade to SDK 31? 😮

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/PreviousInsurerFragment.kt
hugokallstrom: Ok cool, I'll add this in another PR if that makes sense?

hedvigoscar: I guess we could technically do that, as long as we remember to fix it before we toggle the feature flag on 😄

app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/askforprice/AskForPriceInfoActivity.kt
hugokallstrom: Because of navigation-compose (https://developer.android.com/jetpack/compose/navigation). From my understanding that is the way to go if we want to navigate between composables

hedvigoscar: Is SDK 31 a requirement for using Compose Navigation? 😮

StylianosGakis: From some point on all androidx libraries started building against api 31. [Accompanist](https://github.com/google/accompanist/releases/tag/v0.17.0) release notes is one example. But we can compile against api 31 and still have targetApi as 30. That shouldn't bring problems.

hugokallstrom:
I'll merge this and create another PR with navigation 👍

PR #688, opened 2021-08-26
APP-749: Stop using `signOffer`

hedvigoscar:
- Stop using `signOffer`
- Bonus: Make it c o m p o s e
- Animate the size of the dialog
- Remove the checkout step
- Add tracking/logged in-state management
- Remove some unneeded code

### Checklist


app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
```diff
@@ -0,0 +1,177 @@
...
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ) = ComposeView(requireContext()).apply {
+        setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnLifecycleDestroyed(viewLifecycleOwner))
+        setContent {
+            val viewState by model.viewState.collectAsState()
+            HedvigTheme {
+                SwedishBankIdSignDialog(
+                    text = stringResource(
```
hugokallstrom, line 71: Think it makes sense to extract this to a separate function with viewState as param

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
```diff
@@ -0,0 +1,177 @@
...
+                            startActivity(
+                                Intent(
+                                    Intent.ACTION_VIEW, bankIdUri
+                                )
+                            )
+                        }
+                    }
+                    SwedishBankIdSignViewModel.Event.StartDirectDebit -> {
+                        marketManager
+                            .market
```
hugokallstrom, line 103: Should we check that the current market is SE here since this is for SwedishBankId? Perhaps that is too defensive :)

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignViewModel.kt
```diff
@@ -0,0 +1,127 @@
...
+            }
+            BankIdStatus.COMPLETE -> {
+                when (status.signState) {
+                    SignState.INITIATED, SignState.IN_PROGRESS -> null
+                    SignState.COMPLETED -> {
+                        ViewState.Success
+                    }
+                    else -> ViewState.Error
+                }
+            }
```
hugokallstrom, line 123: Should this result in an error as well?

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/usecase/ManuallyRecheckSwedishBankIdSignStatusUseCase.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val apolloClient: ApolloClient,
+) {
+    suspend operator fun invoke(): SignStatusFragment? {
+        val result = apolloClient
+            .query(SignStatusQuery())
+            .safeQuery()
+
+        return when (result) {
+            is QueryResult.Error -> {
+                e { "Error manually rechecking sign status: ${result.message}" }
```
hugokallstrom, line 20: Perhaps this could return a result as well, containing `Success` with `status.collectStatus?.status`, `status.collectStatus?.code` and `status.signState` (which we use in `SwedishBankIdSignViewModel.toViewStateOrNull`. And in addition to Success - Error and Empty states?

hugokallstrom (changes requested):

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/usecase/ManuallyRecheckSwedishBankIdSignStatusUseCase.kt
hedvigoscar: Given that this only re-fetches the data in case of missed subscription messages, it is slightly more lenient on purpose.

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignViewModel.kt
hedvigoscar: The `else`-case covers `UNKNOWN__` and `null` (which is probably just a legacy schema mistake) - if we error here, then any new `BankIdStatus`es would cause error UI to show up when we perhaps do not want it to.

app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutActivity.kt
```diff
@@ -15,6 +15,7 @@ import com.hedvig.app.BaseActivity
 import com.hedvig.app.R
 import com.hedvig.app.databinding.ActivityCheckoutBinding
 import com.hedvig.app.feature.loggedin.ui.LoggedInActivity
+import com.hedvig.app.feature.swedishbankid.sign.SwedishBankIdSignDialog
```
hedvigoscar, line 18:
```suggestion
```

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
hedvigoscar: Something like `getTextByViewState(viewState: ViewState)`?

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
hedvigoscar: Good question! What would the consequences of being in the wrong market be? 🤔

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
hugokallstrom: Something like that yeah!

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
hugokallstrom: I suppose we only want to start `TrustlyConnectPayinActivity` here, and if we (for some reason) dont have a market this will just do nothing

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignViewModel.kt
hugokallstrom: Aha I see

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
hedvigoscar: I kind of like the 'looser' coupling we get from using `Market::connectPayin` 😉 if we change what `Activity` we launch for connecting payin for the Swedish market, it's kind of good to have it in one place I think 🤔 

Personally, I feel like the consequences of having the wrong market will be, before this happens, that the member simply will not be able to sign (as a member outside this market will not have a Swedish BankID). What do you think? 🙂

hugokallstrom: Yeah I agree, fair points!

PR #689, opened 2021-08-31
APP-759: Add cross-sell card in insurance tab

hedvigoscar:
- Preparative refactor
- Implement UI in Compose
- Query CrossSells
- Add a test

### Checklist

- [X] Functionality is covered by a unit test


app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
```diff
@@ -219,24 +298,133 @@ class InsuranceAdapter(
 
     companion object {
         private const val TRANSITION_NAME = "contract_card"
+
+        private const val CROSS_SELL = 1
+        private const val SUBHEADING = 2
     }
 }
```
hugokallstrom, line 306: Would it make sense to extract these composables to their own class/classes to increase visibility?

app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
```diff
@@ -219,24 +298,133 @@ class InsuranceAdapter(
...
+/*
+ * Note: This Composable uses hardcoded colors due to difficulties with
+ * declaring a particular component to be in dark theme instead of the
+ * default. When we update `HedvigTheme` to be Compose-first instead of
+ * XML-Theme first, we can reconfigure the theme for this composable to
+ * be `dark` no matter what the system value is.
+ */
+@Composable
+fun CrossSell(
+    data: InsuranceModel.CrossSell,
```
hugokallstrom, line 340: I think we can make this reusable by passing the primitive types instead of `InsuranceModel.CrossSell` - like `backgroundBlurHash`, `backgroundUrl` etc.

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
```diff
@@ -0,0 +1,80 @@
...
+        }
+        insuranceViewModel
+            .data
+            .flowWithLifecycle(viewLifecycle)
+            .onEach { bind(it) }
+            .launchIn(viewLifecycleScope)
+    }
+
+    private fun bind(viewState: InsuranceViewModel.ViewState) {
+        val adapter = binding.insuranceRecycler.adapter as? InsuranceAdapter ?: return
```
hugokallstrom, line 61: Should this throw instead of returning?

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
```diff
@@ -0,0 +1,80 @@
...
+        when (viewState) {
+            InsuranceViewModel.ViewState.Error -> {
+                adapter.submitList(
+                    listOf(
+                        InsuranceModel.Header,
+                        InsuranceModel.Error
+                    )
+                )
+            }
+            InsuranceViewModel.ViewState.Loading -> {
```
hugokallstrom, line 73: Missing loading state

app/src/test/java/com/hedvig/app/feature/insurance/ui/tab/ItemsTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import assertk.assertThat
+import assertk.assertions.contains
+import assertk.assertions.containsNone
+import com.hedvig.app.feature.insurance.ui.InsuranceModel
+import com.hedvig.app.testdata.dashboard.INSURANCE_DATA
+import com.hedvig.app.testdata.dashboard.INSURANCE_DATA_WITH_CROSS_SELL
+import com.hedvig.app.util.containsNoneOfType
+import com.hedvig.app.util.containsOfType
+import org.junit.Test
+
```
hugokallstrom, line 12: Brilliant.

hugokallstrom (changes requested):

app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
hedvigoscar: Classes are less commonly used in Compose, but we can definitely move them to their own files 🙂

app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
hedvigoscar: I'd agree if we had more use-cases for this particular type of card - for now, I'd point to the rule of three 😉 

https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
hedvigoscar: This code only looks new, it's actually exactly the code that was in `InsuranceFragment` before the file was moved 😬

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
hedvigoscar: Same here - this is the same code as was in place before 🙃

app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
hugokallstrom: Fair enough :)

app/src/main/java/com/hedvig/app/feature/insurance/ui/InsuranceAdapter.kt
hugokallstrom: Yeah if they are tightly coupled with this adapter then it makes sense to have them here. But I think we should try to make composables as reusable as possible from the start, so that you dont end up with many different variants.

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
hugokallstrom: Aha!

hugokallstrom (approved):

PR #692, opened 2021-09-07
APP-767: Generify Contract Detail-Screen

hedvigoscar:
This removes all dependencies we have on the type of a contract,
allowing us to add new contracts without any client-side implementation.

### Checklist

- [X] Functionality is covered by a unit test


app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoItems.kt
```diff
@@ -0,0 +1,32 @@
+package com.hedvig.app.feature.insurance.ui.detail.yourinfo
+
+import com.hedvig.android.owldroid.graphql.InsuranceQuery
+import com.hedvig.app.feature.insurance.ui.detail.ContractDetailViewModel
+import com.hedvig.app.feature.table.intoTable
+import com.hedvig.app.util.apollo.toUpcomingAgreementResult
+
+fun yourInfoItems(
```
hugokallstrom, line 8: Perhaps rename this to something like `createYourInfoItems` or `constructYourInfoItems`

hugokallstrom (approved):
Really nice

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/yourinfo/YourInfoItems.kt
hedvigoscar: I think this comes down to thinking about the name in a procedural context versus in a functional context. When reading the calling code, I read `yourInfoItems()` as being directly substitutable for the items in the list, not as being a creation procedure. I'm not strictly speaking against `create*`/`construct*`, but they do make less sense to me 😅

PR #693, opened 2021-09-08
Delete some code related to coverage and KeyGear

hedvigoscar:
We unfortunately had some breaking schema changes due to a plan downgrade on GraphCMS.
Fortunately, though, these breaking changes are contained to KeyGear, which as a feature has
few enough users to not matter.

This commit removes the offending code, and makes it so that KeyGear will mostly work
without these fields in place.

### Checklist


hugokallstrom:
Thanks for the explanation 👍

hugokallstrom (approved):

PR #695, opened 2021-09-13
Feature/app 771 accident insurance result screens

StylianosGakis:
### Checklist

Available to preview in engineering mode through showkase > Components > Cross Sell result


app/src/main/AndroidManifest.xml
```diff
@@ -186,6 +186,13 @@
         <activity
             android:name=".feature.offer.ui.checkout.CheckoutActivity"
             android:windowSoftInputMode="adjustResize" />
+        <activity
+            android:name=".feature.crossselling.ui.CrossSellingResultActivity"
+            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+            android:launchMode="singleTask"
+            android:screenOrientation="portrait"
+            android:theme="@style/Hedvig.Theme"
+            android:windowSoftInputMode="adjustResize" />
```
StylianosGakis, line 195: I simply copied the standard configuration from the rest of the activities. I assume this is what we use everywhere unless there are some specific requirements for some screens.

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.hedvig.app.BaseActivity
+import kotlinx.parcelize.Parcelize
+import java.time.LocalDate
+
+class CrossSellingResultActivity : BaseActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val crossSellingResult: CrossSellingResult =
+            intent.getParcelableExtra(CROSS_SELLING_RESULT) ?: CrossSellingResult.Error
```
StylianosGakis, line 17: Don't think the the case that the thing wouldn't exist inside the parcelable is even possible, please do comment if we want to handle this somewhat differently

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import androidx.compose.ui.unit.dp
+import com.hedvig.app.R
+import com.hedvig.app.ui.compose.HedvigTheme
+import com.hedvig.app.ui.compose.designsystem.LargeContainedButton
+import com.hedvig.app.ui.compose.designsystem.LargeOutlinedButton
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+// TODO text resources still missing
```
StylianosGakis, line 32: Do we wait for string resources to exist before merge?

app/src/main/AndroidManifest.xml
hedvigoscar: That's a sensible thing to do 😉

However, this is how we 'most recently' do this 🙂

```suggestion
        <activity
            android:name=".feature.crossselling.ui.CrossSellingResultActivity"
            android:windowSoftInputMode="adjustResize" />
```

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
hedvigoscar: I think it's fine to throw an error here, as there's a clear 'misuse' of the `Activity` if there is no `CROSS_SELLING_RESULT` on the `Intent`.

Something like this, perhaps:

```kotlin
val crossSellingResult = intent.getParcelableExtra<CrossSellingResult>(CROSS_SELLING_RESULT) ?: throw IllegalArgumentException("Programmer error: CROSS_SELLING_RESULT not provided to ${this.javaClass.name}")
```

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+            CrossSellingResultScreen(crossSellingResult)
+        }
+    }
+
+    companion object {
+        fun newInstance(context: Context, crossSellingResult: CrossSellingResult): Intent {
+            return Intent(context, CrossSellingResultActivity::class.java).apply {
+                putExtra(CROSS_SELLING_RESULT, crossSellingResult)
+            }
+        }
```
hedvigoscar, line 28: I'm a big fan of the functional syntax for functions in kotlin that simply have one return statement:

```suggestion
        fun newInstance(context: Context, crossSellingResult: CrossSellingResult) = Intent(context, CrossSellingResultActivity::class.java).apply {
                putExtra(CROSS_SELLING_RESULT, crossSellingResult)
            }
```

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        private const val CROSS_SELLING_RESULT = "CROSS_SELLING_RESULT"
+    }
+}
+
+sealed class CrossSellingResult : Parcelable {
```
hedvigoscar, line 34: Perhaps we can move this to its own file 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
hedvigoscar: Yes 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+                ) {
+                    Image(
+                        painter = painterResource(R.drawable.ic_chat_white),
+                        contentDescription = null,
+                    )
+                    Spacer(Modifier.width(8.dp))
+                    Text("Open chat")
+                }
+                LargeOutlinedButton(
+                    onClick = {},
```
hedvigoscar, line 142: I think this needs to be wired up to something 😉

app/src/main/java/com/hedvig/app/ui/compose/designsystem/LargeOutlinedButton.kt
```diff
@@ -0,0 +1,46 @@
...
+
+@Composable
+fun LargeOutlinedButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    content: @Composable RowScope.() -> Unit,
+) {
+    OutlinedButton(
+        onClick = onClick,
+        modifier = modifier,
```
hedvigoscar, line 25: What do you think about putting `.fillMaxWidth()` on this modifier? I think every use-case for the `Large Outlined Button` is actually `fillMaxWidth()`, this might save us some effort on the call-sites 🤔

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+                R.string.purchase_confirmation_new_insurance_today_app_state_title,
+                crossSellingResult.insuranceType
+            )
+        }
+    }
+    val descriptionText = when (crossSellingResult) {
+        CrossSellingResult.Error -> "Your purchase couldn't be completed.\nContact us in the chat."
+        is CrossSellingResult.Success -> {
+            when {
+                crossSellingResult.startingDate.dayOfMonth <= LocalDate.now().dayOfMonth -> {
```
hedvigoscar, line 84: Is this logic correct? Shouldn't it be `crossSellingResult.startingDay == LocalDate.now()`?

(also, we should inject a `Clock` here in order to not have a hard dependency on the system clock 😄)

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+) {
+    HedvigTheme {
+        Box(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(horizontal = 16.dp)
+        ) {
+            InformationSection(
+                crossSellingResult = crossSellingResult,
+                modifier = Modifier
```
hedvigoscar, line 45: Given that the only use-case for `InformationSection` is inside `CrossSellingResultScreen`, perhaps we could inline the modifier? Or is it hoisted because we lack the `BoxScope` inside `InformationSection`?

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+                ) {
+                    Text("Done")
+                }
+            }
+        }
+    }
+}
+
+@Preview(
+    showSystemUi = true,
```
hedvigoscar, line 161: I'm generally a fan of not showing the system UI in previews as it takes up valuable screen real estate. What's your opinion on showing the system UI? 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+fun CrossSellingResultScreen(
+    crossSellingResult: CrossSellingResult
+) {
+    HedvigTheme {
+        Box(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(horizontal = 16.dp)
+        ) {
+            InformationSection(
```
hedvigoscar, line 43: I wonder if this `Composable` is abstracting along the right axis.

I think it could be made a lot simpler by having a high-level check on `CrossSellResult`s variant, and then barely making any logic checks at later stages 🤔 

If you'd like, we can pair programming and I can show what I mean in practice 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
```diff
@@ -0,0 +1,180 @@
...
+                        R.string.purchase_confirmation_new_insurance_active_in_future_app_state_description,
+                        activationDate
+                    )
+                }
+            }
+        }
+    }
+    Column(
+        modifier = modifier,
+        verticalArrangement = Arrangement.spacedBy(18.dp),
```
hedvigoscar, line 101: Perhaps moreso a comment to the designers, but I'm not sure that `18 dp` is a size we want to use in practice. There are some issues with translating the spacing in Figma to implementation that I'd like to show you as well, in particular with regards to the concept of `line height`, which does not work as Figma thinks it does in android 🙄

hedvigoscar (changes requested):
Good-looking first PR! 🎉

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
StylianosGakis: In this particular case, not showing it looks like this
image
and showing it looks like this
image
That is because this screen is .fillMaxSize() anyway since it has a box filling the entire screen and the two different items are aligned to top and bottom respectively.

If anything might need changing here might be the CrossSellingResultScreen implementation if you have some ideas to do this better

hedvigoscar: Oh, strange! Looks like the screen is less tall when system UI is shown? 😄 then the saving are moot I guess, this works totally fine 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
StylianosGakis: I think I see what you mean. Calculating what the two texts and the icon should be higher up and simply passing them down to InformationSection instead of passing the CrossSellResult and doing it all down there. We can look into it yes.

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
StylianosGakis: Yeah, I can show you how I got that number from Figma and you can tell me if I made some wrong assumptions on that process.

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
StylianosGakis: InformationSection could absolutely be an extension function to BoxScope which would allow this. This is in fact how I wrote it initially. The reason the modifier is hoisted is because then by looking at the Box() composable you can immediately know how they are arranged without looking at the two smaller composables themselves. In fact I'd say they themselves shouldn't really care how they are aligned in the box scope, and should just know how to render themselves, and the modifier coming from above can tell them their position in the box scope.
Does the thought process make sense? We can look at it together too and see the alternatives

app/src/main/AndroidManifest.xml
StylianosGakis: How does it work without providing the theme to the activity? Don't we get problems with that?

hedvigoscar: If we do not provide a theme on the `<activity />`-level, the `Activity` will use the `android:theme` defined for the `<application />` 🙂

app/src/main/java/com/hedvig/app/ui/compose/designsystem/LargeContainedButton.kt
```diff
@@ -20,7 +21,7 @@ fun LargeContainedButton(
 ) {
     Button(
         onClick = onClick,
-        modifier = modifier,
+        modifier = Modifier.fillMaxWidth().then(modifier),
```
hedvigoscar, line 24: Nice trick!

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
StylianosGakis: Absolutely, but this would be part of "wiring it up to the rest of the app" which I'd need some help on. Since right now there's no "normal" way to go into this screen anyway

PR #696, opened 2021-09-15
APP-787: Select the correct post-sign screen on Offer-screen

hedvigoscar:
This approach has its limitations, and should probably be considered to be
temporary.

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
@@ -1,6 +1,5 @@
 package com.hedvig.app.feature.offer
 
-import android.content.Context
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.hedvig.android.owldroid.graphql.OfferQuery
@@ -11,7 +10,6 @@ import com.hedvig.android.owldroid.type.SignMethod
 import com.hedvig.app.authenticate.LoginStatus
 import com.hedvig.app.authenticate.LoginStatusService
 import com.hedvig.app.feature.documents.DocumentItems
-import com.hedvig.app.feature.home.ui.changeaddress.result.ChangeAddressResultActivity
 import com.hedvig.app.feature.insurablelimits.InsurableLimitItem
 import com.hedvig.app.feature.offer.quotedetail.buildDocuments
 import com.hedvig.app.feature.offer.quotedetail.buildInsurableLimits
@@ -25,7 +23,6 @@ import com.hedvig.app.feature.offer.usecase.GetQuoteUseCase
 import com.hedvig.app.feature.offer.usecase.GetQuotesUseCase
 import com.hedvig.app.feature.offer.usecase.RefreshQuotesUseCase
 import com.hedvig.app.feature.perils.PerilItem
-import com.hedvig.app.feature.settings.Market
 import e
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.MutableSharedFlow
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
@@ -1,14 +1,17 @@
 package com.hedvig.app.feature.offer
 
+import android.content.Context
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.hedvig.android.owldroid.graphql.OfferQuery
 import com.hedvig.android.owldroid.graphql.RedeemReferralCodeMutation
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -198,20 +198,40 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
+                                        ?.let { startActivity(it) }
+                                }
+                                OfferViewModel.PostSignScreen.MOVE -> {
+                                    ChangeAddressResultActivity.newInstance(
+                                        this@OfferActivity,
+                                        ChangeAddressResultActivity.Result.Success(event.startDate),
+                                    )
+                                }
+                                OfferViewModel.PostSignScreen.CROSS_SELL -> {
+                                    // TODO: Wire up Cross-Sell Result screen here
```
hedvigoscar, line 215: This is where we would wire up the cross-sell result screen @StylianosGakis

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -198,20 +198,40 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
...
+                                OfferViewModel.PostSignScreen.MOVE -> {
+                                    startActivity(
+                                        ChangeAddressResultActivity.newInstance(
+                                            this@OfferActivity,
+                                            ChangeAddressResultActivity.Result.Error
+                                        )
+                                    )
+                                }
+                                OfferViewModel.PostSignScreen.CROSS_SELL -> {
+                                    // TODO: Wire up Cross-Sell Result Screen here
```
hedvigoscar, line 232: This is where we would wire up the cross-sell result screen @StylianosGakis

hedvigoscar:
Note that I've not been able to test this yet, as the API for Cross-sells is temporarily down. 😬

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -48,16 +53,36 @@ abstract class OfferViewModel : ViewModel() {
...
+            val postSignScreen: PostSignScreen,
+            val bundleDisplayName: String,
         ) : Event()
 
         data class StartSwedishBankIdSign(val quoteIds: List<String>, val autoStartToken: String) : Event()
 
         object DiscardOffer : Event()
     }
 
+    enum class PostSignScreen {
```
hugokallstrom, line 71: Would it make more sense to have this enum in `GetPostSignDependenciesUseCase`? That would decouple network types from the viewmodel.

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -163,10 +186,22 @@ class OfferViewModelImpl(
     override fun approveOffer() {
         viewModelScope.launch {
             _viewState.value = _viewState.value.copy(isLoading = true)
+            val postSignDependencies = getPostSignDependenciesUseCase(quoteIds)
```
StylianosGakis, line 189: When using `invoke` functions without explicitly calling `.invoke()` it makes it impossible for AS to find the usage when doing Cmd + B on the `invoke()` function itself.
It does work on latest IntelliJ version so it seems like an AS bug, but we all know these can take many months before they are integrated to AS too.
I would be happy if we considered explicitly calling `invoke()` on the UseCase functions as it doesn't make it any less readable or whatever, and it really helps with navigating the code (finding usages of the useCase in one go instead of Searching for the name of the useCase as a string)

But wouldn't protest if you don't like it, it's just a mild inconvenience

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hedvigoscar: Sure!

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hedvigoscar: Either style is fine by me. Maybe you can weigh in, @hugokallstrom?

hugokallstrom: Explicitly calling invoke() makes sense tbh. Its not as pretty but worth it being able to navigate via Cmb + B.

hedvigoscar:
```suggestion
            val postSignDependencies = getPostSignDependenciesUseCase.invoke(quoteIds)
```

hedvigoscar: `invoke` it is.

PR #697, opened 2021-09-16
APP-791: only show gross price for cross selling offer

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
@@ -35,7 +35,7 @@ object OfferItemsBuilder {
                     data.netMonthlyCost()
                 },
                 originalPremium = data.grossMonthlyCost(),
-                hasDiscountedPrice = !(data.grossMonthlyCost() - data.netMonthlyCost()).isZero &&
+                hasDiscountedPrice = !data.grossMonthlyCost().isEqualTo(data.netMonthlyCost()) &&
                     !data.quoteBundle.appConfiguration.ignoreCampaigns,
                 incentiveDisplayValue = data
                     .redeemedCampaigns
--- a/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
+++ b/app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
@@ -9,6 +9,7 @@ import com.hedvig.app.feature.offer.ui.OfferModel
 import com.hedvig.app.feature.offer.ui.changestartdate.getStartDate
 import com.hedvig.app.feature.offer.ui.changestartdate.getStartDateLabel
 import com.hedvig.app.feature.offer.ui.changestartdate.toChangeDateBottomSheetData
+import com.hedvig.app.feature.offer.ui.checkoutTextRes
 import com.hedvig.app.feature.offer.ui.faq.FAQItem
 import com.hedvig.app.feature.offer.ui.grossMonthlyCost
 import com.hedvig.app.feature.offer.ui.netMonthlyCost
@@ -41,7 +42,9 @@ object OfferItemsBuilder {
                     .mapNotNull { it.fragments.incentiveFragment.displayValue },
                 hasCampaigns = data.redeemedCampaigns.isNotEmpty(),
                 changeDateBottomSheetData = data.quoteBundle.inception.toChangeDateBottomSheetData(),
+                checkoutTextRes = data.checkoutTextRes(),
                 signMethod = data.signMethodForQuotes,
+                approveButtonTerminology = data.quoteBundle.appConfiguration.approveButtonTerminology,
                 showCampaignManagement = data.quoteBundle.appConfiguration.showCampaignManagement,
                 ignoreCampaigns = data.quoteBundle.appConfiguration.ignoreCampaigns,
                 gradientRes = when (data.quoteBundle.appConfiguration.gradientOption) {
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -1,11 +1,13 @@
 package com.hedvig.app.feature.offer
 
+import androidx.annotation.StringRes
```
StylianosGakis, line 3: Are these automatic checks by ktlint fixed if you run 
ktlint -F "\!**/generated/**" "**/src/**/*.kt*”
on your terminal?

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -286,8 +287,9 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
         }
     }
 
-    private fun setSignState(signMethod: SignMethod) {
-        binding.signButton.bindWithSignMethod(signMethod)
+    private fun setSignButtonState(signMethod: SignMethod, checkoutTextRes: Int?) {
+        binding.signButton.text = checkoutTextRes?.let(::getString) ?: ""
+        binding.signButton.icon = signMethod.checkoutIconRes()?.let(::compatDrawable)
         binding.signButton.setHapticClickListener {
```
StylianosGakis, line 293: A bit unrelated to this PR. But I've seen we use haptic click listener quite a lot. In compose we don't quite have a simple `. setHapticClickListener`, we should look into this especially now that we have our first buttons in Compose too.

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
StylianosGakis: I realise now that on the PR view it's not visible, but while reviewing ktlint was complaining about the ordering of some imports, and I wonder if that's a case of IntelliJ auto-formatter formatting them differently from what ktlint likes

app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
```diff
@@ -27,15 +29,24 @@ object OfferItemsBuilder {
...
                     .getStartDateLabel(data.quoteBundle.appConfiguration.startDateTerminology),
-                netMonthlyCost = data.netMonthlyCost(),
-                grossMonthlyCost = data.grossMonthlyCost(),
+                premium = if (data.quoteBundle.appConfiguration.ignoreCampaigns) {
+                    data.grossMonthlyCost()
+                } else {
+                    data.netMonthlyCost()
+                },
+                originalPremium = data.grossMonthlyCost(),
+                hasDiscountedPrice = !(data.grossMonthlyCost() - data.netMonthlyCost()).isZero &&
```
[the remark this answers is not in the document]
hugokallstrom: Yes! Fixed 👍

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
hugokallstrom: My IDE does not complain on the ordering here 🤔 Do we have different settings?

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
hugokallstrom: Very good point. Let's talk more about this with designers.

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
StylianosGakis: Right, my IDE wasn't either, but while being on the review mode inside GitHub I saw a bunch of warnings like in this picture here
image
but now they are gone. I think the IDE doesn't complain about the ordering the same way that ktlint does. But I would hope the Ktlint IDE plugin should indicate these as warnings by default too 🤔

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferModel.kt
```diff
@@ -20,15 +21,20 @@ sealed class OfferModel {
         val startDate: OfferStartDate,
         @StringRes
         val startDateLabel: Int,
-        val netMonthlyCost: MonetaryAmount,
-        val grossMonthlyCost: MonetaryAmount,
+        @StringRes
+        val checkoutTextRes: Int?,
```
hedvigoscar, line 25: Something I've been thinking about: similar to how we're attempting to avoid data-layer objects in the android-layer, I think we should avoid android-objects in the viewmodel-layer. My reasoning is that if we ever want to attempt to re-use the viewmodel - which should be doable, given that it's more or less agnostic to the rendering-method - we would struggle to do so if we have android-specifics in the viewmodel.

Given that line of reasoning, I've been thinking that perhaps we should return enums describing what type of text we're looking to display, and then have the view-layer interpret what that means in terms of android-specific resources.

What do you think about this? @hugokallstrom @StylianosGakis

app/src/main/java/com/hedvig/app/util/BroadcastReceiver.kt
```diff
@@ -6,7 +6,7 @@ import android.content.Intent
 
 inline fun newBroadcastReceiver(crossinline closure: (context: Context?, intent: Intent?) -> Unit):
     BroadcastReceiver = object : BroadcastReceiver() {
-    override fun onReceive(context: Context?, intent: Intent?) {
-        closure(context, intent)
+        override fun onReceive(context: Context?, intent: Intent?) {
```
hedvigoscar, line 9: Perhaps you need to run `ktlint applyToIDEAProject` again? 🤔

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferModel.kt
StylianosGakis: Sounds very reasonable as an idea. How about we try it out and see how it feels.

hugokallstrom: Yes sounds great. Ill fix!

PR #698, opened 2021-09-17
APP-792: Refactor dynamic link logic and handle new path "insurance"

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -47,88 +52,7 @@ class SplashActivity : BaseActivity(R.layout.activity_splash) {
 
     override fun onNewIntent(intent: Intent) {
         super.onNewIntent(intent)
-        handleFirebaseDynamicLink(intent, null)
```
StylianosGakis, line 50: Before, we were using the intent coming from `onNewIntent` to pass to `FirebaseDynamicLinks.getInstance().getDynamicLink(intent)` . Now we are using `Activity.getIntent()` instead since we don't pass it from here, but the documentation for it says "Return the intent that started this activity." but in the case where we get the deep link from onNewIntent() the one that started the activity wouldn't be the same as the one passed inside the onNewIntent() right?

Would this be a bug in case the activity is already open get get a new intent with a deep link, or does the Intent returned from `Activity.getIntent()` get updated after the `onNewIntent` happens? The docs on `onNewIntent` seem to mention that "Note that getIntent still returns the original Intent." so I just looking at the code makes me think this would make these links not work if the activity is already open 🤔 What do you think? @hugokallstrom

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
```diff
@@ -0,0 +1,88 @@
...
+import com.hedvig.app.feature.settings.Market
+import com.hedvig.app.feature.settings.MarketManager
+
+class DynamicLinkHandler(
+    val context: Context,
+    val marketManager: MarketManager,
+    val startDefaultActivity: (LoginStatus?) -> Unit
+) {
+
+    enum class DynamicLink(val path: String) {
```
StylianosGakis, line 22: I like having the possibilities as an type what we can do an exhaustive when on, good idea 👍

StylianosGakis (commented):
Please do take a good look at least on the onNewIntent() comment, on first sight it appears that there may be a problem on the change

app/src/main/java/com/hedvig/app/SplashActivity.kt
```diff
@@ -30,10 +27,18 @@ class SplashActivity : BaseActivity(R.layout.activity_splash) {
     private val loggedInService: LoginStatusService by inject()
     private val marketManager: MarketManager by inject()
     private val binding by viewBinding(ActivitySplashBinding::bind)
+    private lateinit var dynamicLinkHandler: DynamicLinkHandler
```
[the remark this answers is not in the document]
hugokallstrom: I wanted to pass this activities context as a constructor param, but maybe it makes more sense to pass it as a function param and inject the class through koin.

app/src/main/java/com/hedvig/app/SplashActivity.kt
hugokallstrom: I see! I think this was an issue in the original code as well though. 

`handleFirebaseDynamicLink(intent, null)` this will call the function with loginStatus as null, so the checks in `handleReferralsDeepLink()`, `handleDirectDebitDeepLink()` and `handleForeverDeepLink()` will call `startDefaultActivity()` with null loginStatus as parameter. This will finally call `handleFirebaseDynamicLink(intent, loginStatus)` in `navigateToActivity()`. A bit hard to follow but I think thats whats going on 😅

app/src/main/java/com/hedvig/app/SplashActivity.kt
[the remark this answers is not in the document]
StylianosGakis: Does it need to be this activity context or just any? Koin can provide the application's context by itself. But if we need the activity it's as you said maybe not super straightforward

app/src/main/java/com/hedvig/app/SplashActivity.kt
StylianosGakis: Yeah, we don't want it to pass null all the time either.
How about then having `getLoginStatusAndNavigate` and `navigateToActivity` receive a nullable intent parameter? That way if the deep link comes from the onNewIntent it can use the correct one, and if it doesn't we can do like
```kotlin
val intent = intentComingFromParamters ?: this@Activity.intent
dynamicLinkHandler.handleIntent(intent, loginStatus)
```
so that we actually use the correct intent in this case.

app/src/main/java/com/hedvig/app/SplashActivity.kt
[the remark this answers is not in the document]
hugokallstrom: When starting activities using the application context we can get unexpected back stack behaviour - thats why we need the local context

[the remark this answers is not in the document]
StylianosGakis: Right I didn't know about this, interesting 🤔

app/src/main/java/com/hedvig/app/SplashActivity.kt
hugokallstrom: Good idea, fixed 👍

app/src/main/java/com/hedvig/app/SplashActivity.kt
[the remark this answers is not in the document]
hedvigoscar: I agree with Stelios here - we should probably use Koin to initialize this dependency.

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    fun handleIntent(
+        intent: Intent,
+        loginStatus: LoginStatus?
+    ) {
+        FirebaseDynamicLinks.getInstance()
+            .getDynamicLink(intent)
+            .addOnSuccessListener { pendingDynamicLinkData ->
+                // This will actually be null in some cases
```
hedvigoscar, line 36: I think this has been tagged with `@RecentlyNonNull`, which has made this comment a lie 😄

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+                if (pendingDynamicLinkData != null && pendingDynamicLinkData.link != null) {
+                    val link = pendingDynamicLinkData.link
+                    when (link?.pathSegments?.getOrNull(0)) {
```
hedvigoscar, line 39: Perhaps the `DynamicLink`-enum could own this code, perhaps with a `DynamicLink::from`-function on the `companion object`?

hedvigoscar (approved):
Approved with nits

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
hugokallstrom: Removing the check :)

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
hugokallstrom: Yes, fixed

app/src/main/java/com/hedvig/app/SplashActivity.kt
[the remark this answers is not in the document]
hugokallstrom: I went back to the drawing board. Refactored to use extension functions instead and passing the required parameters. Much cleaner IMO.

app/src/main/java/com/hedvig/app/service/DynamicLinkHandler.kt
hedvigoscar: Very rude of google to add that annotation in that case 😞

app/src/main/java/com/hedvig/app/service/DynamicLink.kt
```diff
@@ -0,0 +1,55 @@
+package com.hedvig.app.service
+
+import android.content.Intent
```
hedvigoscar, line 3: `ktlint -F`

hedvigoscar (approved):
Approved with nit

PR #705, opened 2021-09-21
Fix/app 780 wrong theme on cross sell subheader

StylianosGakis:
### Checklist

Enabling the MdcTheme text colors just doesn't play well with what we're doing. It stops recognising what surface the composables are on top of and simply puts the values that are I guess specified in our theme. So all these may be true
* I understand something wrong about the MdcTheme implementation.
* Our theme doesn't contain the correct values for the text colors
* Some other changes are needed in our theme file?
* Some other weird bug?

For now it looks like this by simply not setting `setTextColors: Boolean` to true.

Light mode | Dark Mode
-|-
 Insurance_Cross_Sell  |  Insurance_Cross_Sell_Dark

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
+++ b/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultScreen.kt
@@ -1,5 +1,6 @@
 package com.hedvig.app.feature.crossselling.ui
 
+import android.content.res.Configuration.UI_MODE_NIGHT_YES
 import androidx.compose.foundation.Image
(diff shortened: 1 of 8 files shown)


hugokallstrom (commented):
Is this fixed by just removing `setTextColors` and `setDefaultFontFamily` when creating `MdcTheme`?

StylianosGakis:
> Is this fixed by just removing `setTextColors` and `setDefaultFontFamily` when creating `MdcTheme`?

Yes that's the thing. That one always seems to work and I don't even know how, even with it set as true. But a lot of other things are messed up when it's true like the the button here
image

hedvigoscar (approved):

PR #706, opened 2021-09-22
APP-797: Refresh insurance after purchase

hugokallstrom:
### Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-@HedvigInsurance/android
+*       @HedvigInsurance/android


StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -440,6 +456,7 @@ class LoggedInActivity : BaseActivity(R.layout.activity_logged_in) {
             putExtra(INITIAL_TAB, initialTab)
             putExtra(EXTRA_IS_FROM_ONBOARDING, isFromOnboarding)
             putExtra(SHOW_RATING_DIALOG, showRatingDialog)
+            putExtra(FORCE_RELOAD, forceReload)
```
hedvigoscar, line 459: It feels a little bit architecturally 'strange' to me that the `LoggedInActivity` owns the cache invalidation here 🤔 I guess it's technically less repetition to not have it be called from the places that actually cause the cache invalidation, but it's still a bit astonishing 😬 

What do you think @hugokallstrom @StylianosGakis?

hedvigoscar (commented):
Some architectural questions arose when reviewing. We should hammer them out before merging

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
StylianosGakis: I am a bit out of the loop about what this cache invalidation entails, but with that said, sure maybe when the invalidation needs to be done for whatever reason, it could be a standalone thing that will can be asked to run from whenever.
Why is it bound to the Activity now?
Does it have to do with the guarantee that it will complete in the scope of the Activity's lifetime?
Was it just a convenient place to place it at the moment?

hugokallstrom: My reasoning for using a parameter here is to keep the logic in one place, and the effects of clearing the cache as close to the logic as possible. Before, the cache got cleared in different places - `ChangeAddressResultActivity`, `CheckoutViewModel` etc. And I had forgot to add additional logic for other purchase flows such as accident. I think it makes more sense to start a "fresh" activity eg. reloading all data. 

Maybe I'm wrong though! Where would you like the cache invalidation to take place? In the UseCases?

hedvigoscar: I think the use-case is probably a good place to do it, yes 🙂

PR #707, opened 2021-09-23
APP-799: Onboarding click listener

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/dismissiblepager/DismissiblePager.kt
+++ b/app/src/main/java/com/hedvig/app/feature/dismissiblepager/DismissiblePager.kt
@@ -48,7 +48,10 @@ abstract class DismissiblePager : DialogFragment() {
         dismiss()
     }
 
-    open fun onLastPageButton() {}
+    @CallSuper
+    open fun onLastPageButton() {
+        dismiss()
+    }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/hedvig/app/feature/dismissiblepager/DismissiblePager.kt
```diff
@@ -48,7 +48,10 @@ abstract class DismissiblePager : DialogFragment() {
         dismiss()
     }
 
-    open fun onLastPageButton() {}
+    @CallSuper
+    open fun onLastPageButton() {
+        dismiss()
+    }
```
StylianosGakis, line 54: Hey, are we sure we want all of the children to inherit this behaviour? Instead of making it abstract and make them decide for themselves if they want to necessarily dismiss.

On the HowClaimsWorkDialog.kt it seems like the last button opens a bottom sheet, and maybe we don't want to dismiss the dialog at the same time. Since if the user clicks that, then clicks somewhere else accidentally dismissing the bottom sheet they will wonder where the dialog has gone. Not sure what the expected behaviour is for sure though.

hugokallstrom: Good catch. Fixed!

hedvigoscar (approved):

StylianosGakis:
Hey nice, but now the WelcomeDialog also will not dismiss right? And we do want it to dismiss in that case, just like in the WhatsNewDialog.

That's why I thought making onLastPageButton abstract and *forcing* each child to decide for themselves might make the most sense. It will help avoid small little bugs like these.

hugokallstrom:
> Hey nice, but now the WelcomeDialog also will not dismiss right? And we do want it to dismiss in that case, just like in the WhatsNewDialog.
> 
> That's why I thought making onLastPageButton abstract and _forcing_ each child to decide for themselves might make the most sense. It will help avoid small little bugs like these.

Your are absolutely right. Fixed again!

PR #713, opened 2021-09-23
App 806 accident wrong date text

StylianosGakis:
### Checklist


app/src/main/java/com/hedvig/app/BaseActivity.kt
```diff
@@ -22,7 +23,9 @@ abstract class BaseActivity : AppCompatActivity {
     private val marketManager: MarketManager by inject()
 
     override fun onResume() {
-        tracker.screenView(javaClass.simpleName)
+        val currentClassSimpleName = javaClass.simpleName
+        tracker.screenView(currentClassSimpleName)
+        d { "Navigated to $currentClassSimpleName" }
```
hedvigoscar, line 28:
```suggestion
        tracker.screenView(javaClass.simpleName)
```

If we want to log analytics events, we should do so by adding a `TrackerSink` in debug-builds 🙂

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
```diff
@@ -17,10 +21,14 @@ class ChangeDateBottomSheetViewModel(
 ) : ViewModel() {
 
     private val _viewState = MutableStateFlow<ViewState>(ViewState.Inceptions(data.inceptions))
-    val viewState: StateFlow<ViewState> = _viewState
+    val viewState: StateFlow<ViewState> = _viewState.asStateFlow()
```
hedvigoscar, line 24:
```suggestion
    val viewState = _viewState.asStateFlow()
```

Only one is really necessary 🙂

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
```diff
@@ -17,10 +21,14 @@ class ChangeDateBottomSheetViewModel(
...
 
     private val selectedDates = mutableMapOf<String, LocalDate>()
 
+    fun shouldOpenDatePicker(): Boolean {
+        return _viewState.value !is ViewState.Loading
+    }
```
hedvigoscar, line 30:
```suggestion
    fun shouldOpenDatePicker() = viewState.value !is ViewState.Loading
```

Two-part suggestion:

1. When reading, it's probably a good idea to use the read-handle instead of the write-handle. Not super-important when internally in the `ViewModel`, but one could for example imagine that we `map` on the `viewState` at a later stage, and then we would risk reading the 'incorrect' value here.
2. I'm partial to the functional syntax for functions when there is only one statement in a function, but this is personal preference. I like how concise it is, it makes it more readable to me 🙂

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
```diff
@@ -57,12 +65,16 @@ class ChangeDateBottomSheetViewModel(
         viewModelScope.launch {
             tracker.changeDateContinue()
             _viewState.value = ViewState.Loading(true)
-            val results = selectedDates.map {
-                editStartDateUseCase.setStartDate(
-                    id = it.key,
-                    idsInBundle = data.idsInBundle,
-                    date = it.value
-                )
+            val results = coroutineScope {
```
hedvigoscar, line 68: Why are we opening a new scope here? 🤔

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheet.kt
```diff
@@ -77,7 +82,8 @@ class ChangeDateBottomSheet : BottomSheetDialogFragment() {
             currentInsurerDisplayName = inception.currentInsurer?.displayName,
             startDate = inception.startDate,
             switchable = inception.currentInsurer?.switchable ?: false,
-            datePickerListener = {
+            datePickerListener = datePickerListener@{
```
hedvigoscar, line 85: What made this check required? 🤔

hedvigoscar (approved):
Approved with nits

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheet.kt
StylianosGakis: Couldn't do a local return inside the lambda without explicitly naming it like this.
The name it had by itself was `bind` so I could do `return@bind` but that one felt confusing, you have to go and look for what is named bind which is some lines further up.
Naming it like this makes it super clear to understand what you are doing a local return on.
But anyway, `return@bind` is an option too if you think this solution is too weird. Labling an anonymous function like this is definitely not something you do every day. Maybe even something you shouldn't? I don't have any strong personal opinions on this or examples of when it's not a good idea?

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
StylianosGakis: This comes from this article (https://elizarov.medium.com/structured-concurrency-722d765aa952) I read from Elizarov the other day.
Jump to the bottom of it or Cmd + F for "You cannot simply write async { … } anymore, you have to provide a scope."
Without it, if one of the async fails, it fails silently, the rest continue, but we do want them all to be done successfully normally right? I can do some testing around this to see how it behaves when something goes bad

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
StylianosGakis: I agree with both of these. An oversight by me, fixed now.

app/src/main/java/com/hedvig/app/BaseActivity.kt
StylianosGakis: I don't think I follow you here.
The one thing I added was `d { "Navigated to $currentClassSimpleName" }`, I am logging the activity change because it makes it super easy for me to open the log and see which screen I am inside to navigate to the appropriate class. Slimber makes it so that this should be free for production builds right? What is the issue with this one?

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheetViewModel.kt
hedvigoscar: Fair enough

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheet.kt
hedvigoscar: I meant why we needed to check `shouldOpenDatePicker` 😬

StylianosGakis: Oh 🤦‍♂️
So what happens if we do not do this check: 
After you click on Save button, it gets disabled thus you can not click it again while the result is loading. And this does take a solid half a second or more most of the times. But in the meantime, you can actually click on the date picker again. And then you get the date dialog, and you're confused since even if you do make a change, when the save result comes back, the dialog will just close on your face and the bottom sheet will be redacted. 
So I thought just like we disable the button, we should also disable the date picker.

hedvigoscar: Sensible! But perhaps we could do it in the same way as we do for the dismissibility of the dialog? 😬

app/src/main/java/com/hedvig/app/BaseActivity.kt
hedvigoscar: It will indeed be free, but I think a more sustainable way to do this would be to add a `TrackerSink` that logs all analytics events, that way we do not have to manually add these kinds of log-statements at locations where we track 🙂

hedvigoscar: the `TrackerSink` could look something like this:

```kotlin
class DebugLogTrackerSink: TrackerSink {
	override fun track(eventName: String, properties: JSONObject?) {
        d { "Track $eventName, properties: ${properties}" }
    }
    override fun identify(id: String) {
        d { "Identify: $id" }
    }
}
```

app/src/main/java/com/hedvig/app/feature/offer/ui/changestartdate/ChangeDateBottomSheet.kt
StylianosGakis: I started out by trying to do it this way. But we create this view programmatically inside the .Inceptions case of the view state, there is no reference to it outside of there. And setting its parent view as non clickable, or disabled etc has no effect on the click listener. The only way I found to fix it is this way, maybe I am missing something however.

hedvigoscar: Fair enough, then this is probably the best approach indeed.

PR #718, opened 2021-09-27
APP-842: Hide change address button

hugokallstrom:
### Checklist


hedvigoscar (changes requested):
If we introduce this change, we move away from being generic on contract type, and regress to requiring implementation in the app for each new contract type. I think we should solve this issue in a different layer, perhaps giraffe?

hugokallstrom:
> If we introduce this change, we move away from being generic on contract type, and regress to requiring implementation in the app for each new contract type. I think we should solve this issue in a different layer, perhaps giraffe?

You are right! I'll close this and start on a giraffe PR.

PR #722, opened 2021-09-28
Refactor composable buttons

hugokallstrom:
### Checklist


StylianosGakis (approved):

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
```diff
@@ -11,15 +11,33 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.dimensionResource
 import androidx.compose.ui.tooling.preview.Preview
-import com.airbnb.android.showkase.annotation.ShowkaseComposable
 import com.hedvig.app.R
 import com.hedvig.app.ui.compose.theme.HedvigTheme
 
 @Composable
-fun LargeContainedButton(
+fun PrimaryTextButton(
```
hedvigoscar, line 18: I think we should use the terminology from the design system - that is:

- (Large) Contained Button
- (Large) Outlined Button
- Text Button

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
```diff
@@ -11,15 +11,33 @@ import androidx.compose.runtime.Composable
...
+    modifier: Modifier = Modifier,
+    text: String,
+    onClick: () -> Unit
+) {
+    PrimaryButton(
+        modifier = modifier,
+        content = {
```
hedvigoscar, line 25: With this structure, we're unable to add icons in buttons. I quite like the ability to just be able to add arbitrary Composables in buttons, that ensures we have all the flexibility we need when using them 😄

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
```diff
@@ -11,15 +11,33 @@ import androidx.compose.runtime.Composable
...
+            Text(
+                text = text,
+                style = MaterialTheme.typography.button,
+                color = MaterialTheme.colors.onSecondary
```
hedvigoscar, line 29: This is actually implicit when adding a Text-Composable as a child to Button, through the magic of LocalContentColor and LocalTextStyle 🙂

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
```diff
@@ -33,29 +51,24 @@ fun LargeContainedButton(
...
+fun PrimaryButtonPreview() {
     HedvigTheme {
-        LargeContainedButton({}) {
-            Text("Contained Button (Large)")
-        }
+        PrimaryTextButton(text = "Contained Button (Large)") {}
     }
 }
 
-@ShowkaseComposable(skip = true)
```
hedvigoscar, line 48: We probably do want this to be skipped, as it won't actually render as dark in Showkase.

hedvigoscar (changes requested):

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
hugokallstrom: My thinking is `PrimaryTextButton` is specifically a text button, and `PrimaryButton` is a more general implementation where you can add custom content

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
hugokallstrom: Very reasonable.

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/SecondaryButtons.kt
```diff
@@ -16,26 +16,22 @@ import com.hedvig.app.R
 import com.hedvig.app.ui.compose.theme.HedvigTheme
 
 @Composable
-fun SecondaryTextButton(
+fun LageOutlinedTextButton(
```
hedvigoscar, line 19:
```suggestion
fun LargeOutlinedTextButton(
```

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
hedvigoscar: I'm still a bit skeptical about this implementation - as it is stated, the usage of it becomes quite 'un-composy' if you will. The flexibility of manipulating the contents as desired is missing, and since the `onClick` is the last argument of the function, the following code does not do what would be expected of canonical compose code:

```kotlin
LargeContainedButton(
    text = "example",
) {
    Text("example") // This actually won't compile, because the tail argument is now the click listener instead of a `content`-Composable
}
```

hedvigoscar (approved):
Approved with nits

app/src/main/java/com/hedvig/app/ui/compose/composables/buttons/PrimaryButtons.kt
StylianosGakis: I think [this article](https://medium.com/androiddevelopers/pushing-the-right-buttons-in-jetpack-compose-124cb4b17197) is a perfect explanation of how the button started with the API surface of passing in a text etc. The shortcomings and confusions that this brought to them. And how in the end, using the "slot API" came to the rescue to finally end up in the API we currently have for a button.
With that said, it's not a crazy idea to provide a "simple" button in our codebase just *just* takes a string. But we are as you said going away from the freedom the slot apis give us and might pollute out autocompletion results with non-obvious functions that don't follow the rest of the composable APIs standards.

hedvigoscar: I think that article captures it perfectly 😄

hugokallstrom: Great article. I would argue though that their example of a `LoginButton` is exactly what this `TextButton` is - a wrapper function around the compose api. Instead of passing a content lambda you pass a string. Compare this with another composable `TopAppBarWithBack`: 
```
@Composable
fun TopAppBarWithBack(
    onClick: () -> Unit,
    title: String
) {
    TopAppBar(
        title = {
            Text(
                text = title,
                style = MaterialTheme.typography.h6
            )
        },
        navigationIcon = {
            IconButton(
                onClick = { onClick() },
                content = {
                    Icon(
                        imageVector = Icons.Filled.ArrowBack,
                        contentDescription = null
                    )
                }
            )
        },
        backgroundColor = MaterialTheme.colors.background,
        elevation = 0.dp,
    )
}
```

Here we also don't pass the content since we want every `TopAppBarWithBack` to have the same content, ie. a back icon.

PR #723, opened 2021-09-29
Cross sell card click listener

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
+++ b/app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
@@ -4,6 +4,7 @@ import android.content.res.Configuration.UI_MODE_NIGHT_YES
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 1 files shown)


StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
```diff
@@ -70,7 +71,11 @@ fun CrossSell(
             ),
             contentDescription = null,
             contentScale = ContentScale.Crop,
-            modifier = Modifier.fillMaxSize(),
+            modifier = Modifier
+                .fillMaxSize()
+                .clickable(
```
hedvigoscar, line 76: Does this make the card have a ripple? 🙂

hedvigoscar (approved):
Approved with nits

app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
StylianosGakis: There is a different overload of the `Card` composable that does have an `onClick`, but it is currently annotated with `@ExperimentalMaterialApi` (not sure why, maybe they're likely to change its API).
So if this one doesn't behave as expected maybe that one is the solution.

hugokallstrom: It does make a ripple! :)

PR #726, opened 2021-09-29
App 876 change address should not be shown for accident

StylianosGakis:


testdata/src/main/java/com/hedvig/app/testdata/dashboard/builders/InsuranceDataBuilder.kt
```diff
@@ -125,6 +125,7 @@ class InsuranceDataBuilder(
                 statusPills = emptyList(),
                 detailPills = emptyList(),
                 gradientOption = TypeOfContractGradientOption.GRADIENT_ONE,
+                supportsAddressChange = true,
```
StylianosGakis, line 128: Not quite sure about the mocked data. It feels like it being defaulted to true is fine for now, but if not, we'd need to go into each instantiation of this object and put true/false manually since we don't have giraffe to do it for us.
Is this good enough for now? What are your thoughts?

hedvigoscar: You could allow it to be passed into the `InsuranceDataBuilder`, and default it there. I think that would be good enough 🙂

testdata/src/main/java/com/hedvig/app/testdata/feature/insurance/builders/InsuranceContractBuilder.kt
```diff
@@ -21,7 +21,7 @@ class InsuranceContractBuilder(
     private val detailsTable: TableFragment = TableFragmentBuilder().build(),
 ) {
 
-    fun build() = InsuranceQuery.Contract(
+    fun build(): InsuranceQuery.Contract = InsuranceQuery.Contract(
```
hedvigoscar, line 24:
```suggestion
    fun build() = InsuranceQuery.Contract(
```

This type is implicit from the functional syntax, and can therefore be omitted for the sake of readability 🙂

hedvigoscar (approved):
Approved with nits

testdata/src/main/java/com/hedvig/app/testdata/feature/insurance/builders/InsuranceContractBuilder.kt
StylianosGakis: It got changed by default by the IDE when I extracted a local variable which made this function to not have this functional syntax. Changed it back but forgot to remove the explicit return type. Fixing it now.

testdata/src/main/java/com/hedvig/app/testdata/dashboard/builders/InsuranceDataBuilder.kt
StylianosGakis: Agreed!

PR #728, opened 2021-10-04
Remove showkase preview annotations in order for release build to compile

hugokallstrom:
### Checklist


hedvigoscar (changes requested):
This PR removes too much, I think - we only need to remove the annotation `@ShowkaseComposable`, nothing more 🙂

hugokallstrom:
> This PR removes too much, I think - we only need to remove the annotation `@ShowkaseComposable`, nothing more 🙂

Ok, then we will get duplicate entries in Showkase. Is that preferably to removing dark mode preview in AS editor?

hedvigoscar:
> > This PR removes too much, I think - we only need to remove the annotation `@ShowkaseComposable`, nothing more 🙂
> 
> Ok, then we will get duplicate entries in Showkase. Is that preferably to removing dark mode preview in AS editor?

Disregard my reply, I missed that it only removed the dark previews.

hedvigoscar (approved):

PR #730, opened 2021-10-04
fix/remove hard coded terms and conditions link

hugokallstrom:
### Checklist


hugokallstrom:
> Approved with caveat: Does the backend return the correct data already? If yes, we can merge now, if not, let's hold off a little bit.

Not yet, holding off a bit until it does.

PR #736, opened 2021-10-05
Remove border on theme to avoid buttons in card to get a border. Use style for card instead

hugokallstrom:
Please see the button "Add payment method" in the info card below. This PR removes the border.
device-2021-10-05-151538
device-2021-10-05-151453

### DIFF
--- a/app/src/main/res/layout/home_info_card.xml
+++ b/app/src/main/res/layout/home_info_card.xml
@@ -66,7 +66,7 @@
 
         <Button
             android:id="@+id/action"
-            style="@style/Hedvig.Button.Text"
+            style="?materialButtonTextStyle"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center_horizontal"
--- a/app/src/main/res/layout/home_info_card.xml
+++ b/app/src/main/res/layout/home_info_card.xml
@@ -9,7 +9,8 @@
     android:layout_marginTop="@dimen/base_margin_double"
     android:layout_marginEnd="@dimen/base_margin_double"
     android:padding="@dimen/base_margin_double"
-    android:theme="@style/ThemeOverlay.Hedvig.WarningCard">
+    android:theme="@style/ThemeOverlay.Hedvig.WarningCard"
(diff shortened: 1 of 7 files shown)


hugokallstrom:
> I could use a little bit more context on this issue 😅

Sorry! Attached screenshots.

app/src/main/res/layout/home_info_card.xml
```diff
@@ -65,7 +66,7 @@
 
         <Button
             android:id="@+id/action"
-            style="?materialButtonTextStyle"
```
hedvigoscar, line 68: Is this correct? 🤔

app/src/main/res/values/styles.xml
```diff
@@ -118,4 +118,10 @@
         <item name="android:paddingTop">12dp</item>
     </style>
 
+    <style name="Hedvig.Theme.InfoCard">
```
hedvigoscar, line 121: This style does not contain the background color for warning cards, and it appears to be used for warning cards, but is named info card 🤔

hedvigoscar (commented):
A few questions came to mind!

app/src/main/res/layout/home_info_card.xml
hugokallstrom: Oh we have this `<item name="materialButtonTextStyle">@style/Hedvig.Button.Text</item>` - changing back.

app/src/main/res/values/styles.xml
hugokallstrom: So this style is only for setting the card properties. For setting the colors we are still using the themes, like this:

```
android:theme="@style/ThemeOverlay.Hedvig.RenewalCard"
style="@style/Hedvig.Theme.InfoCard"
```

hugokallstrom: Having a theme at the top level with `<item name="strokeWidth">1dp</item>` will apply a stroke on every child, in this case the buttons.

hedvigoscar: Sensible.

hedvigoscar (approved):
Approved with previously mentioned nits

PR #743, opened 2021-10-06
APP-837: Fix ripple color on white buttons

hedvigoscar:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
+++ b/app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
@@ -30,7 +30,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import coil.compose.rememberImagePainter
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
```diff
@@ -143,10 +164,11 @@ private val previewData = InsuranceModel.CrossSell(
 )
 @Composable
 fun CrossSellPreview() {
+    val context = LocalContext.current
     HedvigTheme {
         CrossSell(
             data = previewData,
-            onCtaClick = {}
+            onCtaClick = { context.makeToast("Doing stuff") }
```
StylianosGakis, line 171: Right, this was left from testing, should prob remove, which will remove the makeToast extension fun import too

app/src/main/java/com/hedvig/app/feature/insurance/ui/CrossSell.kt
```diff
@@ -111,22 +117,37 @@ fun CrossSell(
...
+                    }
                 }
             }
         }
     }
 }
 
+private object DarkRippleTheme : RippleTheme {
+    @Composable
+    override fun defaultColor() = Color(0x1f000000)
```
StylianosGakis, line 142: Maybe this also deserves a comment with a link to the exact package and line of where this is originally defined in the deep theme hierarchy? Because otherwise it also feels like a magic number.

Was this the link? https://cs.android.com/android/platform/superproject/+/master:prebuilts/sdk/current/support/v7/appcompat/res/values/values.xml;l=59?q=%20lang:xml%201f000000

PR #744, opened 2021-10-06
APP-863: missing red dot on the add more coverage title

StylianosGakis:
The idea behind this PR:

When required to add a second badge that was backed by the same data, it made it hard to show one of them, while hiding the other, because the logic was that they were shown under the same conditions, but they were hidden in different conditions.

Enter NotificationBadgeService and NotificationBadge.
The NotificationBadgeService serves as a layer on top of the datastore to provide a simpler API that consumers can more confidently use without needing to get keys using companion object constants that don't ensure compile-time type safety.
All this can instead be hidden inside NotificationBadge which knows about its own key in the datastore, and the service that makes use of this information.

With this came the detail of also needing to do a migration on the datastore too, inside app/src/main/java/com/hedvig/app/datastore/SeenCrossSellsPreferencesDataMigration.kt , this was interesting I couldn't find any info on how to do this properly online, feel free to chime in with ideas/links if you know more about it.

This feels like too many changes for just adding another badge. But first I couldn't make it work with 1 single datastore entry. Then I split them up and made another entry in the datastore and then one thing brought the next and here we are 😅


app/src/main/java/com/hedvig/app/feature/insurance/ui/Subheading.kt
```diff
@@ -1,20 +1,28 @@
...
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.hedvig.app.ui.compose.theme.HedvigTheme
 
 @Composable
-fun Subheading(text: String) {
-    Text(
-        text = text,
-        style = MaterialTheme.typography.h6,
+fun Subheading(
```
hugokallstrom, line 21: I think it makes sense to create some kind of `NotificationSubheading`, instead of adding more functionality to this composable. Imagine when we want to add `showIcon: Boolean`, `showStrikethrough: Boolean` etc. - this composable will contain too much logic.

Perhaps you could even wrap `Subheading`  in `NotificationSubheading`.

app/src/main/java/com/hedvig/app/service/badge/Seen.kt
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.app.service.badge
+
+@JvmInline
+value class Seen private constructor(val seen: Boolean) {
```
hugokallstrom, line 4: Beautiful

app/src/main/java/com/hedvig/app/datastore/SeenCrossSellsPreferencesDataMigration.kt
```diff
@@ -0,0 +1,45 @@
...
+
+class SeenCrossSellsPreferencesDataMigration : DataMigration<Preferences> {
+    private val oldCrossSellSetKey = stringSetPreferencesKey("SEEN_CROSS_SELLS")
+
+    override suspend fun shouldMigrate(currentData: Preferences): Boolean {
+        return try {
+            val oldCrossSellStrings = currentData[oldCrossSellSetKey] ?: return false
+            return oldCrossSellStrings.isNotEmpty()
+        } catch (e: ClassCastException) {
+            false
```
hugokallstrom, line 17: Log error here

hugokallstrom (changes requested):
This is great :)

app/src/main/java/com/hedvig/app/datastore/SeenCrossSellsPreferencesDataMigration.kt
StylianosGakis: The idea is that if the DataStore does contain data of the type Set<String> it means that it is storing information as we did before.

Optimally I wanted to check if it simply has data on the key `stringSetPreferencesKey("SEEN_CROSS_SELLS")` and that does return true (a function called .contains() on Preferences or smth like that), but unfortunately, this also returns true *after* the migration when the key is actually then `booleanPreferencesKey("SEEN_CROSS_SELLS")` so it just looks at the string but not the type contained. So with the .get approach, unfortunately it throws an exception and I am using it for the normal flow, not as an exceptional thing that would happen very rarely.

So this implementation seemed the only way available from the available API to see if the data stored in the datastore is of the old type, meaning with the key "SEEN_CROSS_SELLS" and at the same time containing the old set of strings.
So I'm using it just to run the migration when this is the case.

app/src/main/java/com/hedvig/app/feature/insurance/ui/Subheading.kt
```diff
@@ -1,20 +1,23 @@
...
+    text: String,
+    contentBeforeText: (@Composable () -> Unit)? = null,
```
hedvigoscar, line 18:
```suggestion
    contentBeforeText: (@Composable () -> Unit)? = null,
```

I think we're probably better off just separating these two, the API for `Subheading` became a bit needlessly complex 😬

app/src/main/java/com/hedvig/app/feature/loggedin/service/TabNotificationService.kt
```diff
@@ -1,52 +1,47 @@
...
+import com.hedvig.app.service.badge.NotificationBadgeService
+import com.hedvig.app.service.badge.Seen
+import com.hedvig.app.service.badge.isSeen
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
 class TabNotificationService(
     private val getCrossSellsUseCase: GetCrossSellsUseCase,
-    private val dataStore: DataStore<Preferences>,
+    private val notificationBadgeService: NotificationBadgeService,
```
hedvigoscar, line 13: I'm a little confused - don't `NotificationBadgeService` and `TabNotificationService` kind of deal with the same subject matter? 🤔 Should they be separated, or should they be the same? 🤔

app/src/main/java/com/hedvig/app/service/badge/NotificationBadgeService.kt
```diff
@@ -0,0 +1,50 @@
+package com.hedvig.app.service.badge
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+interface NotificationBadgeService {
```
hedvigoscar, line 9: I'm personally not a big fan of single-implementation interfaces - I feel like they generally just add a layer of abstraction that never actually delivers any concrete value.

The way I see it, in Java/C#, there was an 'acceptable reason' to create them in that mocking classes was far easier with single implementation-interfaces when testing, but `mockk` can mock any given non-open class just fine, making that reason kind of moot.

Some more reading on the subject: https://www.symphonious.net/2011/06/18/the-single-implementation-fallacy/ and https://martinfowler.com/bliki/InterfaceImplementationPair.html

app/src/main/java/com/hedvig/app/datastore/SeenCrossSellsPreferencesDataMigration.kt
```diff
@@ -0,0 +1,45 @@
+package com.hedvig.app.datastore
+
+import androidx.datastore.core.DataMigration
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.stringSetPreferencesKey
+import com.hedvig.android.owldroid.type.TypeOfContract
+import com.hedvig.app.service.badge.NotificationBadge
+
+class SeenCrossSellsPreferencesDataMigration : DataMigration<Preferences> {
```
hedvigoscar, line 9: Given that we're not live with the tab notification yet, we can probably omit the migration 🙂

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/service/badge/NotificationBadge.kt
```diff
@@ -0,0 +1,28 @@
...
+    sealed class BottomNav(key: Preferences.Key<Boolean>) : NotificationBadge(key) {
+        object CrossSellOnInsuranceFragment : BottomNav(booleanPreferencesKey("SEEN_CROSS_SELLS"))
+    }
+
+    companion object {
+        fun fromPotentialCrossSells(
+            potentialCrossSells: Set<TypeOfContract>
+        ): List<BottomNav> =
+            potentialCrossSells.mapNotNull { typeOfContract ->
+                if (typeOfContract == TypeOfContract.SE_ACCIDENT) {
```
hedvigoscar, line 21: This breaks the contract that we set out for the cross-sell tab notification: this only checks for `SE_ACCIDENT`, but the business rule is that we should show the tab notification any time there is a cross-sell available that the member has not seen 🙂

hedvigoscar (changes requested):
Sorry that my review arrives a bit late, it's been quite a busy day at the office 😧

app/src/main/java/com/hedvig/app/service/badge/Seen.kt
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.app.service.badge
+
+@JvmInline
```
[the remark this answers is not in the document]
StylianosGakis: I think it should recognise value classes, as it probably could also use inline classes when they were named that since Kotlin 1.3. It'd be a long time to still not have any support for it?
As far as the @JvmInline goes, the annotation itself is forced when using the `value` keyword for reasons explained in the [KEEP](https://github.com/Kotlin/KEEP/blob/master/notes/value-classes.md#project-valhalla)

app/src/main/java/com/hedvig/app/service/badge/NotificationBadge.kt
StylianosGakis: Ugh okay, I misunderstood this a bit, let me go back and think about it a bit again. Might have dug a bit of a hole for myself.

app/src/test/java/com/hedvig/app/feature/loggedin/service/TabNotificationServiceTest.kt
```diff
@@ -1,57 +1,50 @@
...
-    private fun mockedDataStore(data: Set<String>): DataStore<Preferences> {
-        val mockPreferences = mockk<Preferences>()
-        every { mockPreferences[TabNotificationService.SEEN_CROSS_SELLS_KEY] } returns data
-        val mockDataStore = mockk<DataStore<Preferences>>()
-        every { mockDataStore.data } returns flowOf(mockPreferences)
-        return mockDataStore
+    private fun mockedCrossSellNotificationBadgeService(
+        returnValue: Set<TypeOfContract>
+    ): CrossSellNotificationBadgeService {
+        val mock = mockk<CrossSellNotificationBadgeService>()
```
hedvigoscar, line 46: Generally, try to avoid mocking types we own - this often leads to overmocking, resulting in unreliable and hard-to-maintain tests.

A good rule is that things that directly does I/O or interacts with platform code should be mocked 🙂

PR #746, opened 2021-10-07
APP-944: Cross Sell-tracks

hedvigoscar:
Draft because we're missing one track: `offer_viewed`.
Let's discuss how to best do this together.

Also, let's discuss how to practically track only once when the screen is
opened, and not for example when configuration changes (and the activity gets
recreated).

- Track checkout-button clicks
- Track tab views
- Rename event for screen view-tracks
- Fix the tab tracking
- Track clicks on cross-sell cards
- Track cross sell-result screen

### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
@@ -13,8 +13,8 @@ import java.time.Clock
 import java.time.format.DateTimeFormatter
 
 class CrossSellingResultActivity : BaseActivity() {
-
     private val clock: Clock by inject()
+    private val tracker: CrossSellTracker by inject()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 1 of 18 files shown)


hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellingResultActivity.kt
```diff
@@ -9,20 +9,25 @@ import com.hedvig.app.feature.chat.ui.ChatActivity
...
+        get() = intent.getParcelableExtra(CROSS_SELLING_RESULT)
             ?: throw IllegalArgumentException(
                 "Programmer error: CROSS_SELLING_RESULT not provided to ${this.javaClass.name}"
             )
+    private val model: CrossSellResultViewModel by viewModel { parametersOf(crossSellingResult) }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
 
+        tracker.view(crossSellingResult)
```
StylianosGakis, line 30: Now we're doing this here and [here](https://github.com/HedvigInsurance/android/pull/746/files#diff-5d00ecd4c87d651b28a245689421c2c663fdce75ed428074b8516c994f9b2f21R10) at the same time right? It's the same line of code twice. Are we sure we want to have the tracker inside the Activity as well? I thought just in the init() of the VM would be enough? What am I missing? @hedvigoscar

PR #752, opened 2021-10-11
APP-916: Update dependencies

hugokallstrom:
### Checklist

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -6,6 +6,6 @@ repositories {
     google()
 }
 dependencies {
-    implementation("com.android.tools.build:gradle:7.0.0")
+    implementation("com.android.tools.build:gradle:7.0.2")
     implementation(kotlin("gradle-plugin", version = "1.5.30"))
 }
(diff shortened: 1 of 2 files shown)


buildSrc/build.gradle.kts
```diff
@@ -7,5 +7,5 @@ repositories {
 }
 dependencies {
     implementation("com.android.tools.build:gradle:7.0.0")
```
hedvigoscar, line 9:
```suggestion
    implementation("com.android.tools.build:gradle:7.0.2")
```

buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -89,17 +88,17 @@ object Libs {
         const val tracking = "com.google.firebase:firebase-analytics-ktx"
     }
 
-    const val mixpanel = "com.mixpanel.android:mixpanel-android:5.9.1"
+    const val mixpanel = "com.mixpanel.android:mixpanel-android:6.0.0-beta1"
```
hedvigoscar, line 91: Is this safe? 😬

hedvigoscar (approved):
Approved with nits

buildSrc/src/main/kotlin/Libs.kt
hugokallstrom: Moved to stable release

PR #754, opened 2021-10-12
APP-957: Add file upload variable functionality

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/embark/util/VariableExt.kt
```diff
@@ -53,6 +53,9 @@ fun GraphQLVariablesFragment.AsEmbarkAPIGraphQLSingleVariable.createSingleVariab
         EmbarkAPIGraphQLSingleVariableCasting.BOOLEAN -> storeValue.toBoolean()
         // Unsupported generated types are ignored for now.
         EmbarkAPIGraphQLSingleVariableCasting.UNKNOWN__ -> null
+        // FILE is not handled here since we need to create a separate multipart request body for
```
hedvigoscar, line 56: Should we add an `else`-case here? On master, we currently do not have an `else`-case, and that makes the code break if we introduce new enum-members 😬

PR #759, opened 2021-10-14
APP-966: Feature/show claim status

StylianosGakis:
### Checklist

- [X] Functionality is accessible in engineering mode - Home -> Claims (Heading) -> Various options for different claim status

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/ClaimTypeData.kt
+++ b/app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/ClaimTypeData.kt
@@ -9,11 +9,9 @@ data class ClaimTypeData(val displayableText: DisplayableText) {
     companion object {
         fun fromHomeQueryClaim(homeQueryClaim: HomeQuery.Claim): ClaimTypeData {
             return ClaimTypeData(
-                when {
-                    // TODO get the type from backend? It exists in Hope already, but not on GraphQL
-                    //  context: https://hedviginsurance.slack.com/archives/C01SCHY7W1W/p1634557500017900
-                    // claimStatus.perilType != null -> claimStatus.perilType.toProperReadableString
-                    homeQueryClaim.status == ClaimStatus.SUBMITTED -> {
+                // TODO add the claim type title when the copy is addressed https://hedvig.atlassian.net/browse/APP-995
+                when (homeQueryClaim.status) {
+                    ClaimStatus.SUBMITTED -> {
                         DisplayableText(R.string.claim_type_new_insurance_case)
                     }
                     else -> {
--- a/app/src/main/java/com/hedvig/app/util/compose/DisplayableText.kt
+++ b/app/src/main/java/com/hedvig/app/util/compose/DisplayableText.kt
@@ -8,6 +8,7 @@ import com.hedvig.app.ui.compose.theme.LocalMarket
 import com.hedvig.app.util.apollo.format
 import javax.money.MonetaryAmount
 
+// TODO fix this at https://hedvig.atlassian.net/browse/APP-994
 sealed class DisplayableText private constructor() {
     @Composable
     abstract fun text(): String


app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -151,6 +151,10 @@ class HomeFragment : Fragment(R.layout.home_fragment) {
                 val items = mutableListOf<HomeModel>().apply {
                     addAll(listOfNotNull(*psaItems(successData.importantMessages).toTypedArray()))
                     add(HomeModel.BigText.Active(firstName))
+                    // TODO think about showing this in other states as well? Not just on isActive(...)
+                    if (successData.claimStatus.isNotEmpty()) {
+                        add(HomeModel.ClaimStatus(successData.claimStatus))
+                    }
```
StylianosGakis, line 157: Need to check design-wise if these cards should also be shown when there isn't an active contract.
It feels like if someone has terminated their contract a day ago but they still have an ongoing claim from a week ago it should still be visible? Need to discuss this.

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/Pills.kt
```diff
@@ -0,0 +1,103 @@
...
+            horizontalArrangement = Arrangement.Center,
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            // TODO uppercase it in a different way? Check user's preference on language maybe?
+            val currentLocale = Locale.current
+            Text(
+                text.uppercase(java.util.Locale(currentLocale.language, currentLocale.region)),
+                style = MaterialTheme.typography.caption,
+                maxLines = 1
+            )
```
StylianosGakis, line 82: Is there a better way to force uppercase in this scenario? We do have limited language options anyway, so would looking at the user selected language be the best approach? Is there some other way that I am not aware of?

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/ClaimTypeData.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@JvmInline
+value class ClaimTypeData(val text: String) {
+    companion object {
+        @Composable
+        fun fromClaimStatus(claimStatus: HomeQuery.ClaimStatus): ClaimTypeData {
+            return ClaimTypeData(
+                when {
+                    // TODO get the type from backend? It exists in Hope already, but not on GraphQL 
+                    // claimStatus.perilType != null -> claimStatus.perilType.toProperReadableString
```
StylianosGakis, line 17: Hope does contain information on "Claim Type" when looking at a specific claim, something that doesn't seem to be available in the GraphQL query.
Need to edit giraffe? Not sure, but need to add this so that not all the claims show "(New) Insurance case"

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/PillData.kt
```diff
@@ -0,0 +1,76 @@
...
+                    PillData(
+                        stringResource(R.string.home_claim_card_pill_reopened),
+                        PillType.Contained(ClaimStatusColors.Pill.reopened)
+                    ),
+                    PillData(
+                        stringResource(R.string.home_claim_card_pill_claim),
+                        PillType.Outlined
+                    ),
+                )
+                ClaimStatus.UNKNOWN__ -> throw IllegalArgumentException()
```
StylianosGakis, line 72: Are the claimStatus options (or any enum) supposed to be a non-breaking change in the GraphQL schema? Should we just not render anything instead of crashing when a new one is added? Probably yes right?

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/RelatedContractTypeData.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import androidx.compose.runtime.Composable
+import com.hedvig.android.owldroid.graphql.HomeQuery
+
+@JvmInline
+value class RelatedContractTypeData(val text: String) {
+    companion object {
+        @Composable
+        fun fromClaimStatus(claimStatus: HomeQuery.ClaimStatus): RelatedContractTypeData {
+            // TODO Is displayName appropriate?
```
StylianosGakis, line 11: For example for Swedish house we get "Home Insurance Renter" but is this what we're supposed to show to the user? Do we have a nicer human-readable version of these, or are these the appropriate ones?

app/src/main/java/com/hedvig/app/ui/compose/theme/Colors.kt
```diff
@@ -10,13 +15,60 @@ val background = Color(0xffF6F6F6)
...
+
+@Composable
+fun hedvigContentColorFor(backgroundColor: Color): Color {
+    // TODO check how to properly provide contentColor colors for non-material colors. This is a workaround for now.
+    return when (backgroundColor) {
+        lavender400 -> Color.Black
+        foreverOrange500 -> Color.Black
+        else -> contentColorFor(backgroundColor)
+    }
+}
```
StylianosGakis, line 74: Context: Text on custom colors is off in dark mode without this.
So this is something that I am not familiar with.
Using contentColorFor() works great but *just* when using the normal material colors, since under the hood all it does is check if the color is primary, it uses onPrimary for the content, if it's secondary it uses onSecondary etc. But here we have unique colors that aren't part of the material spec, just some unique color for the pills. This workaround works for now, but is there a better way to do this? Is there a function that can infer from a color's luminescence or whatever if the content color should be white or black?

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCards.kt
```diff
@@ -0,0 +1,65 @@
...
+    require(homeQueryClaimStatusList.isNotEmpty())
+
+    val claimStatusDataList: List<ClaimStatusData> = homeQueryClaimStatusList.map {
+        ClaimStatusData.fromHomeQueryClaimStatus(it)
+    }
+
+    ClaimStatusCards(claimStatusDataList)
+}
+
+@JvmName("ClaimStatusCardsWithConvertedInputData")
```
StylianosGakis, line 30: Apparently having the same function with the same name and the only parameter being a list counts as the same JVM signature since I assume the generic type gets erased and it can't tell the difference.
This @JvmName setting manually feels a bit weird, is it ok to keep it like this?

app/src/main/java/com/hedvig/app/feature/home/ui/HomeModel.kt
```diff
@@ -36,6 +36,10 @@ sealed class HomeModel {
         object Terminated : BodyText()
     }
 
+    data class ClaimStatus(
+        val claims: List<HomeQuery.ClaimStatus>
```
hedvigoscar, line 40: A recent, but arguably good, practice we've started following in the app is that we remap from network-layer types to our own types. I know that this is not what we're currently doing on this screen, but I think we should do it here as well (i.e create our own `Claim`-type that we use here, or something of the sort) 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCards.kt
hedvigoscar: I think we can do the data remapping at a higher level, that will circumvent this issue entirely 😄 The type does indeed get erased, from the JVM's perspective it's only `List<Any>`

app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
hedvigoscar: I agree, but double-check with Hiro just to be safe.

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/Pills.kt
hedvigoscar: This is probably the right way to do it, but shouldn't we have been provided with this text already pre-uppercased? 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/ClaimProgressData.kt
```diff
@@ -0,0 +1,104 @@
...
+                            ClaimProgressType.CurrentlyActive,
+                        )
+                    }
+                }
+                ClaimStatus.REOPENED -> Triple(
+                    ClaimProgressType.PastInactive,
+                    ClaimProgressType.Reopened,
+                    ClaimProgressType.FutureInactive,
+                )
+                ClaimStatus.UNKNOWN__ -> throw IllegalArgumentException()
```
hedvigoscar, line 100: We need to handle `UNKNOWN__` a bit more gracefully, since this will be matched by all old clients that have not updated their apps in the event of a new `ClaimStatus` 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/PillData.kt
hedvigoscar: What to do when encountering an unknown `enum`-case needs to be evaluated for each use-case - here, I'm not quite sure what's best. Check with Hiro, perhaps 🙂

app/src/main/java/com/hedvig/app/ui/compose/theme/Colors.kt
```diff
@@ -10,13 +15,60 @@ val background = Color(0xffF6F6F6)
...
+    @Composable get() = colorResource(R.color.lavender_400)
+val lavender900: Color
+    @Composable get() = colorResource(R.color.lavender_900)
+
+val foreverOrange300: Color
+    @Composable get() = colorResource(R.color.forever_orange_300)
+val foreverOrange500: Color
+    @Composable get() = colorResource(R.color.forever_orange_500)
+
+object ClaimStatusColors {
```
hedvigoscar, line 40: Perhaps these should live in a different package 🤔

app/src/main/java/com/hedvig/app/ui/compose/theme/Colors.kt
hedvigoscar: I'm not quite sure about the particular use-case for this function, but this is probably good enough, I guess? 😄

app/src/main/res/layout/home_start_claim_contained.xml
```diff
@@ -4,7 +4,7 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:layout_marginStart="@dimen/base_margin_double"
-    android:layout_marginTop="@dimen/base_margin_quintuple"
+    android:layout_marginTop="18dp"
```
hedvigoscar, line 7: Where did this change come from? 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/RelatedContractTypeData.kt
hedvigoscar: I think this is indeed the appropriate one 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/RelatedContractTypeData.kt
```diff
@@ -0,0 +1,16 @@
...
+            val relatedContract = claimStatus.contract?.displayName ?: String()
```
hedvigoscar, line 12:
```suggestion
            val relatedContract = claimStatus.contract?.displayName ?: ""
```

Or does `String()` have other benefits? 😄

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/Pills.kt
```diff
@@ -0,0 +1,103 @@
...
+    ) {
+        Row(horizontalArrangement = Arrangement.spacedBy(8.dp)) {
+            pillData.forEach { pillData: PillData ->
+                Pill(
+                    text = pillData.text,
+                    pillType = pillData.type,
+                )
+            }
+        }
+        // Uncomment when the card is clickable to go to detail screen
```
hedvigoscar, line 39:
```suggestion
        // TODO: Uncomment when the card is clickable to go to detail screen
```

Just so that we get some highlighting on it 😉

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/ClaimTypeData.kt
hedvigoscar: If it's not in the API, we should discuss it with the implementing backend team (MX) 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/PillData.kt
```diff
@@ -0,0 +1,75 @@
...
+                                add(
+                                    PillData(
+                                        stringResource(R.string.claim_decision_paid),
+                                        PillType.Contained(MaterialTheme.colors.primary)
+                                    )
+                                )
+                                claimStatus.payout?.let { payout: HomeQuery.Payout ->
+                                    add(
+                                        PillData(
+                                            "${payout.amount} ${payout.currency}",
```
hedvigoscar, line 39: Use `MonetaryAmount.format(Context)` here 🙂 You can find it in `com.hedvig.app.util.apollo`

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/PillData.kt
```diff
@@ -0,0 +1,75 @@
...
+
+data class PillData(val text: String, val type: PillType) {
+
+    sealed class PillType {
+        data class Contained(val color: Color) : PillType()
+        object Outlined : PillType()
+    }
+
+    companion object {
+        @Composable
```
hedvigoscar, line 21: Doesn't it become a little bit architecturally weird to have a `data class`-factory function that is `@Composable`? 😬

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/data/RelatedContractTypeData.kt
```diff
@@ -0,0 +1,16 @@
+package com.hedvig.app.feature.home.ui.claimstatus.data
+
+import androidx.compose.runtime.Composable
+import com.hedvig.android.owldroid.graphql.HomeQuery
+
+@JvmInline
+value class RelatedContractTypeData(val text: String) {
```
hedvigoscar, line 7: How much mileage are we getting from having this be a `value class` wrapping a `String`? 🙂

app/src/main/java/com/hedvig/app/feature/home/ui/HomeModel.kt
StylianosGakis: Yeap, absolutely reasonable

PR #761, opened 2021-10-18
APP-863 missing red dot on the add more coverage title

StylianosGakis:
### Checklist

- [X] Functionality is accessible in engineering mode : Insurance -> [Cross-Sells] Cross-Sell with card badge notification

This PR is a follow-up on #744, as it one was merged but then reverted.


StylianosGakis:
This PR is dropped in favor of #764 which actually adds the new commits on top of a commit that reverts that revert.

PR #763, opened 2021-10-18
APP-942: Change CrossSell card implementation according to the new design

StylianosGakis:
New look: 
image


StylianosGakis:
Two concerns.
1. The new line break doesn't look all that nice with the new layout.
2. The gradient we have going from bottom to the top as a gradient helped with the readablity of the text, it doesn't do so anymore since the white text is at the top. In figma the background picture looks a bit more blackened in general, but I can't understand what kind of gradient we want instead. It just shows no color 50%, what does this even mean? @hedvigoscar 
image

PR #778, opened 2021-10-29
APP-994: Remove DisplayableText workaround

StylianosGakis:
This also removes the need to pass market and language through a
CompositionLocal
Allow the composables to simply receive the string that they should
display without any need for them to manipulate the data


hugokallstrom (approved):
LGTM

PR #779, opened 2021-11-01
Add string resources

hugokallstrom:
### Checklist


StylianosGakis (approved):
Maybe we don't need to add the named argument in this simple one argument function? I know autocompletion adds it by itself I don't love that 😅
But literally a non-issue, keep it that way if you prefer it

hedvigoscar:
Agreed, I prefer `stringResource(R.string.name_of_string)` over `stringResource(id = R.string.name_of_resource)` as well.

hedvigoscar (approved):
Approved with same nit as @StylianosGakis

PR #780, opened 2021-11-01
APP-998: Ask for permissions when pressing start record button

hugokallstrom:
### Checklist

Tried to implement compose-style permission handling (more info [here](https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/permissions/RequestPermissionSample.kt)), but didnt quite get it to work 100%. Used our already existing implementation since we are using a fragment here anyway. We can look at the compose implementation together later if you want.

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderFragment.kt
@@ -1,5 +1,7 @@
 package com.hedvig.app.feature.embark.passages.audiorecorder
 
+import android.Manifest
+import android.content.pm.PackageManager
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.ViewGroup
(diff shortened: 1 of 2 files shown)


hedvigoscar:
What happened when you attempted to use the accompanist-variant? I'm assuming that we probably want to use that long-term 🤔

hedvigoscar:
Barring that, perhaps the `registerForActivityResult`-API is preferrable to this one. 🤔

app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderFragment.kt
```diff
@@ -54,8 +58,36 @@ class AudioRecorderFragment : Fragment() {
...
+            model.startRecording()
+        } else {
+            requireActivity().askForPermissions(
+                arrayOf(Manifest.permission.RECORD_AUDIO),
+                REQUEST_AUDIO_PERMISSION
+            )
+        }
+    }
+
+    override fun onRequestPermissionsResult(
```
hedvigoscar, line 72: Do we have any fallback in case the user has denied the permission? 🤔

hedvigoscar (approved):
LGTM, but check out the `ActivityResult`-API, if applicable here.

app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderFragment.kt
```diff
@@ -25,6 +25,16 @@ class AudioRecorderFragment : Fragment() {
     private val model: AudioRecorderViewModel by viewModel()
     private val clock: Clock by inject()
 
+    private val permission = Manifest.permission.RECORD_AUDIO
+
+    private val permissionResultLauncher = registerForActivityResult(ActivityResultContracts.RequestPermission()) {
+        if (it) {
```
hedvigoscar, line 31:
```suggestion
    private val permissionResultLauncher = registerForActivityResult(ActivityResultContracts.RequestPermission()) { permissionGranted ->
        if (permissionGranted) {
```

hedvigoscar (approved):

PR #782, opened 2021-11-02
APP-1006: Check for empty string when evaluating null expression

hugokallstrom:
### Checklist


hugokallstrom:
Talking to MX if they can return null for member phone number instead

StylianosGakis (approved):
Just gotta make sure that we want to fix it here, meaning that iOS will probably need to make the same fix too.

PR #783, opened 2021-11-02
APP-972: Cross-sell notification support

hugokallstrom:
### Checklist


apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/crossSells.graphql
```diff
@@ -1,7 +1,41 @@
-query CrossSells {
+query CrossSells($locale: Locale!) {
   activeContractBundles {
     potentialCrossSells {
+      title
```
hedvigoscar, line 4: Given the large amount of duplication we have now, should we use a GraphQL Fragment? Perhaps a separate query would be appropriate for the use-case? 🤔

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt
```diff
@@ -69,3 +104,10 @@ data class CrossSellData(
         )
     }
 }
+
+fun CrossSellType.toCrossSellType() = when (this) {
+    CrossSellType.ACCIDENT -> CrossSellData.CrossSellType.ACCIDENT
```
hedvigoscar, line 109: I think having us be explicit about the `CrossSellType` will erode some of the genericity of this solution - currently, any type of cross-sell is supported. If we enable `sealed class for enums` for the `CrossSellType`, we can simply match on the `rawValue`, and we never have to build in support client-side. 😄

app/src/main/java/com/hedvig/app/service/push/managers/CrossSellNotificationManager.kt
```diff
@@ -0,0 +1,108 @@
...
+            val crossSell = getCrossSell(type)
+            if (crossSell != null) {
+                createIntentAndNotify(
+                    context = context,
+                    title = title,
+                    body = body,
+                    crossSell = crossSell
+                )
+            } else {
+                e { "Could not find cross-sell with type: $type" }
```
hedvigoscar, line 40: Should we fall back to showing a notification, but simply opening to the insurance-tab? 🤔

app/src/main/java/com/hedvig/app/service/push/managers/CrossSellNotificationManager.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private fun parseCrossSellType(type: String?) = when (type) {
+        "ACCIDENT" -> CrossSellData.CrossSellType.ACCIDENT
+        "TRAVEL" -> CrossSellData.CrossSellType.TRAVEL
+        "HOME_CONTENT" -> CrossSellData.CrossSellType.HOME_CONTENT
+        else -> CrossSellData.CrossSellType.UNKNOWN
+    }
+
+    private val id = AtomicInteger(100)
```
hedvigoscar, line 102: Do we need this `AtomicInteger`-thing for this type of notification? I was under the impression that it was only necessary for the chat-notifications 🤔

hedvigoscar (changes requested):
Some architectural concerns to be considered before we go ahead, primarily the genericity-question 🙂

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt
hugokallstrom: Good idea

app/src/main/java/com/hedvig/app/service/push/managers/CrossSellNotificationManager.kt
hugokallstrom: Yes! 👍

apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/crossSells.graphql
hugokallstrom: Yeah I suppose a fragment is the way to go. They are so clunky to work with though..

StylianosGakis: What is it that you feel that Fragments are clunky?

hedvigoscar: I agree that `object.fragments.someThingFragment.field` is less nice than `object.field` 😞 The tradeoff of not having to have duplicate methods acting on the data (which we do have a few times in this PR, mind you 😄) is worth it in my experience 🙂

hugokallstrom: Its so unfortunate that you cant reference fragments without their "parent" query, like `CrossSells.Fragments.{...}` and `Insurance.Fragments.{...}` :( I agree its still worth it though. Fixing now!

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/service/push/Util.kt
```diff
@@ -25,5 +26,18 @@ fun setupNotificationChannel(
             }
         )
     }
+
+fun getImmutablePendingIntentFlags(): Int = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+    PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
```
hedvigoscar, line 31: `FLAG_IMMUTABLE` can be specified on lower SDKs, right? 🤔

app/src/main/java/com/hedvig/app/service/push/Util.kt
```diff
@@ -25,5 +26,18 @@ fun setupNotificationChannel(
...
+} else {
+    PendingIntent.FLAG_UPDATE_CURRENT
+}
+
+fun getMutablePendingIntentFlags(): Int = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
hedvigoscar, line 36: 👍

hedvigoscar (approved):

PR #790, opened 2021-11-05
Set read timeout to 0 for uploads to avoid timeouts on client

hugokallstrom:
### Checklist


StylianosGakis (commented):
How would this behave in the case where we do actually lose internet connection right before we make this request?

hedvigoscar:
@StylianosGakis A complete loss of connection would likely trigger an error, but a connection that stalls forever would likely just never respond in the UI for the user.

Then again, if we set this value to a very high one, our users would still likely just give up after a certain amount of time, probably.

I think this change is fine, but there is one thing that we probably do want to add long-term: Upload progress, if at all possible.

hedvigoscar (approved):

PR #792, opened 2021-11-05
Update build config to only have showkase on non production builds

StylianosGakis:
Updates after input taken from https://github.com/airbnb/Showkase/issues/189 issue


vinaygaba (commented):
You can also move your showkaseRoot implementation into the debug sourceSet instead of adding it in main.

hedvigoscar:
@vinaygaba We actually have it under the `engineering`-sourceSet, which is not included in release-builds 🙂

hedvigoscar:
Also, thanks for checking our little codebase out 😉

hedvigoscar (approved):
lgtm!

PR #794, opened 2021-11-08
APP-999: Track Cross-Sell Notifications

hedvigoscar:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellTracker.kt
+++ b/app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellTracker.kt
@@ -27,4 +27,22 @@ class CrossSellTracker(
             }
         }
     }
+
+    fun notificationReceived(crossSell: CrossSellData?) {
+        trackingFacade.track(
+            "cross_sell_notification_receive",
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellTracker.kt
```diff
@@ -27,4 +27,22 @@ class CrossSellTracker(
             }
         }
     }
+
+    fun notificationReceived(crossSell: CrossSellData?) {
```
hugokallstrom, line 31: Can we pass only `crossSellType` here instead of `CrossSellData`?

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/crossselling/ui/CrossSellTracker.kt
hedvigoscar: We can 🙂 I intentionally let these methods accept the entire `CrossSellData`-object here, as the way I saw it, the calling code should not concern themselves with which part of the `CrossSellData` should be tracked 😅

PR #796, opened 2021-11-08
APP-1018: Remove termination info which was showing faulty information

StylianosGakis:
Context:
https://hedviginsurance.slack.com/archives/CPCUHMMJQ/p1635932572116000
Optimally this should be shown in the case where specifically a previous
contract was over after the member went through the moving flow, but
there is no way to know that when we get a new clean contract, therefore
it was decided to show nothing instead.
The information about the old contract being active 30 days after the
moving is presented at the end of the moving flow.

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/ContractDetailActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/ContractDetailActivity.kt
@@ -101,9 +101,6 @@ class ContractDetailActivity : BaseActivity(R.layout.contract_detail_activity) {
                             error.root.isVisible = false
                             val contract = viewState.data
                             contract.bindTo(cardContainer, marketManager)
-                            terminationInfo.isVisible =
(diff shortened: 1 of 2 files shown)


hedvigoscar (approved):
Was this the last use of `contractStatusFragment` in `insurance.graphql`? If yes, bonus credits if you remove it 🔪

StylianosGakis:
> Was this the last use of `contractStatusFragment` in `insurance.graphql`? If yes, bonus credits if you remove it 🔪

Still used:
https://github.com/HedvigInsurance/android/blob/8e0b27311cbe0e3c607faeb6a878ebc6e21d53e2/app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/Items.kt#L15
https://github.com/HedvigInsurance/android/blob/8e0b27311cbe0e3c607faeb6a878ebc6e21d53e2/app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/Items.kt#L42

as a way to find and filter terminated contracts.

Unless we have another way to extract that information, we will have to use it.

hedvigoscar:
Fair enough! Proceed as is, in that case 🙂

PR #805, opened 2021-11-10
APP-1044: Add France market

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/util/featureflags/FeatureManager.kt
```diff
@@ -1,29 +1,27 @@
 package com.hedvig.app.util.featureflags
 
 import androidx.annotation.VisibleForTesting
-import com.hedvig.app.feature.settings.MarketManager
+import com.hedvig.app.feature.settings.Market
+import com.hedvig.app.isDebug
 import java.util.concurrent.CopyOnWriteArrayList
 
-class FeatureManager(
```
hugokallstrom, line 7: I had to remove marketManager injection here because of circular dependencies. Passing it to the function instead.

StylianosGakis (commented):
All looks good to me, but still probably a good idea for @hedvigoscar to see the logic around the different markets

app/src/main/java/com/hedvig/app/feature/settings/MarketManager.kt
```diff
@@ -21,6 +23,7 @@ class MarketManagerImpl(
         Market.SE,
         Market.NO,
         Market.DK,
+        if (isFeatureEnabled(Feature.FRANCE_MARKET, market)) Market.FR else null
```
hedvigoscar, line 26: Would it perhaps be simpler to simply make a `isDebug`-check here, and keep the rest of the architecture as-was? 🤔

hedvigoscar (commented):
Some concerns about the architecture, rest looks good 🙂

app/src/main/java/com/hedvig/app/feature/settings/MarketManager.kt
hugokallstrom: Yes that makes more sense. Will revert to that.

PR #806, opened 2021-11-11
APP-1038: Generic Auth, Email Input

hedvigoscar:
### Checklist

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
@@ -10,7 +10,7 @@ class CreateOtpAttemptUseCase(
 ) {
     sealed class Result {
         data class Success(
-            val token: String,
+            val id: String,
         ) : Result()
 
         object Error : Result()
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
@@ -1,0 +1,29 @@
+package com.hedvig.app.feature.genericauth
+
+import com.apollographql.apollo.ApolloClient
+import com.hedvig.android.owldroid.graphql.CreateOtpAttemptMutation
+import com.hedvig.app.util.apollo.QueryResult
+import com.hedvig.app.util.apollo.safeQuery
+
+class CreateOtpAttemptUseCase(
+    private val apolloClient: ApolloClient,
+) {
+    sealed class Result {
+        data class Success(
+            val token: String,
+        ) : Result()
+
+        object Error : Result()
+    }
+
+    suspend operator fun invoke(email: String) = when (
+        val result = apolloClient
+            .mutate(CreateOtpAttemptMutation(email = email))
+            .safeQuery()
+    ) {
+        is QueryResult.Success -> {
+            Result.Success(result.data.login_createOtpAttempt)
+        }
+        else -> Result.Error
+    }
+}
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
@@ -4,6 +4,7 @@ import com.apollographql.apollo.ApolloClient
 import com.hedvig.android.owldroid.graphql.CreateOtpAttemptMutation
 import com.hedvig.app.util.apollo.QueryResult
 import com.hedvig.app.util.apollo.safeQuery
+import e
 
 class CreateOtpAttemptUseCase(
     private val apolloClient: ApolloClient,
@@ -24,6 +25,9 @@ class CreateOtpAttemptUseCase(
         is QueryResult.Success -> {
             Result.Success(result.data.login_createOtpAttempt)
         }
-        else -> Result.Error
+        is QueryResult.Error -> {
+            result.message?.let { e { it } }
+            Result.Error
+        }
     }
 }
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
@@ -46,7 +46,7 @@ fun EmailInputScreen(
     onClear: () -> Unit,
     onBlur: () -> Unit,
     inputValue: String,
-    error: GenericAuthViewModel.ViewState.TextFieldError?,
+    error: String?,
 ) {
     Column(
         modifier = Modifier.fillMaxSize(),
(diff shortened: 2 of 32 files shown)


app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun submitEmail() {
+        viewModelScope.launch {
+            // TODO: Set Loading-state, once one exists
+            val email = viewState.value.input
+            when (val result = createOtpAttemptUseCase.invoke(email)) {
+                is CreateOtpAttemptUseCase.Result.Success -> {
+                    // TODO: Show the OTP input screen, maintain the token in the ViewModel
+                }
+                CreateOtpAttemptUseCase.Result.Error -> {
+                    _viewState.update { it.copy(error = ViewState.TextFieldError.INVALID_EMAIL) }
```
hugokallstrom, line 67: This should not be invalid email I suppose?

app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
```diff
@@ -0,0 +1,179 @@
...
+
+@Composable
+fun EmailInputScreen(
+    onUpClick: () -> Unit,
+    onInputChanged: (String) -> Unit,
+    onSubmitEmail: () -> Unit,
+    onClear: () -> Unit,
+    onBlur: () -> Unit,
+    inputValue: String,
+    error: GenericAuthViewModel.ViewState.TextFieldError?,
```
hugokallstrom, line 49: Pass string here instead of ViewState? To avoid coupling this screen to GenericAuthViewModel

app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    suspend operator fun invoke(email: String) = when (
+        val result = apolloClient
+            .mutate(CreateOtpAttemptMutation(email = email))
+            .safeQuery()
+    ) {
+        is QueryResult.Success -> {
+            Result.Success(result.data.login_createOtpAttempt)
+        }
+        else -> Result.Error
```
hugokallstrom, line 27: Get error message here

hugokallstrom (changes requested):

app/src/main/java/com/hedvig/app/util/compose/ModifierExt.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import android.view.KeyEvent
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.input.key.onKeyEvent
+import androidx.compose.ui.platform.LocalSoftwareKeyboardController
+
+@OptIn(ExperimentalComposeUiApi::class)
+fun Modifier.submitOnEnter(action: () -> Unit) = composed {
```
StylianosGakis, line 11: Closing the keyboard is more of a "hidden side effect" behind a pretty innocent looking name on the function. Might catch us unprepared sometime in the future when we forget it also does that. Do you think it's important to denote that it does that in the name?

app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
StylianosGakis: Definitely a good idea about not passing down the GraphQL type all the way down here. And as far as the dialog goes, probably Hiro will have an opinion right?

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
hedvigoscar: True - perhaps we could request some kind of `network error`-copy from Hiro.

app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
hedvigoscar: I guess we could do the error -> string-conversion at the calling layer? Having the parent `Activity` be the one dependent on the `ViewModel`. I'll make that change!

app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
hedvigoscar: The error-message isn't user-presentable, right? That was my motivation for not extracting it. Perhaps logging it could be valid, though?

app/src/main/java/com/hedvig/app/util/compose/ModifierExt.kt
hedvigoscar: I'd consider it par for the course when submitting using `Enter`, but perhaps it's not? 🤔

StylianosGakis: Hmm yeah fair point, Enter kinda implies that, as it's the default keyboard behaviour.

app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCase.kt
hugokallstrom: In `<T> ApolloCall<T>.safeQuery()` we currently get the exception message or error message from the response. We should map these error states to user readable strings though, but for now I think its better to go with what we have. It makes debugging so much easier.

hugokallstrom: Or we could just map this error to a general network error string :) And log the exception message. That probably makes more sense.

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
hedvigoscar: Come to think of it, don't we have a generic network error-translation we could utilize here? @hugokallstrom

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+        _viewState.update { previousState ->
+            val newState = previousState.copy(touched = previousState.dirty)
+            val error = computeError(newState, newState.input)
+            newState.copy(error = error)
+        }
+    }
+
+    fun submitEmail() {
+        viewModelScope.launch {
+            // TODO: Set Loading-state, once one exists
```
hugokallstrom, line 69: Mimic what iOS did here and create a button with a loading state? ;)

StylianosGakis: Yeah if this is something Hiro can also add to Figma if we think we want to go that way it'd be great

hedvigoscar: It would be fun, but until we have an asset that I can animate I'm going to skip this 😬

PR #807, opened 2021-11-12
APP-1039: Generic Auth, OTP Input

hugokallstrom:
### Checklist

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -314,7 +314,8 @@
 
         <activity
             android:name=".feature.genericauth.otpinput.OtpInputActivity"
-            android:windowSoftInputMode="adjustResize"/>
+            android:exported="true"
+            android:windowSoftInputMode="adjustResize" />
 
         <meta-data
             android:name="com.google.firebase.messaging.default_notification_icon"
@@ -368,6 +369,10 @@
         <intent>
             <action android:name="android.intent.action.VIEW" />
             <data android:scheme="https" />
+        </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="mailto" />
         </intent>
     </queries>
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -314,6 +314,7 @@
 
         <activity
             android:name=".feature.genericauth.otpinput.OtpInputActivity"
+            android:exported="true"
             android:windowSoftInputMode="adjustResize"/>
 
         <meta-data
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -312,6 +312,10 @@
             android:name=".feature.genericauth.GenericAuthActivity"
             android:windowSoftInputMode="adjustResize" />
 
+        <activity
+            android:name=".feature.genericauth.otpinput.OtpInputActivity"
+            android:windowSoftInputMode="adjustResize"/>
+
         <meta-data
             android:name="com.google.firebase.messaging.default_notification_icon"
             android:resource="@drawable/ic_hedvig_h" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -314,7 +314,6 @@
 
         <activity
             android:name=".feature.genericauth.otpinput.OtpInputActivity"
-            android:exported="true"
             android:windowSoftInputMode="adjustResize"/>
 
         <meta-data
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -75,7 +75,9 @@ import com.hedvig.app.feature.embark.passages.textaction.TextActionViewModel
 import com.hedvig.app.feature.genericauth.CreateOtpAttemptUseCase
 import com.hedvig.app.feature.genericauth.GenericAuthViewModel
 import com.hedvig.app.feature.genericauth.otpinput.OtpInputViewModel
+import com.hedvig.app.feature.genericauth.otpinput.ReSendOtpCodeUseCase
 import com.hedvig.app.feature.genericauth.otpinput.ReSendOtpCodeUseCaseImpl
+import com.hedvig.app.feature.genericauth.otpinput.SendOtpCodeUseCase
 import com.hedvig.app.feature.genericauth.otpinput.SendOtpCodeUseCaseImpl
 import com.hedvig.app.feature.home.data.GetHomeUseCase
 import com.hedvig.app.feature.home.service.HomeTracker
@@ -610,8 +612,8 @@ val useCaseModule = module {
     single { GraphQLQueryUseCase(get()) }
     single { GetCrossSellsUseCase(get(), get()) }
     single { CreateOtpAttemptUseCase(get()) }
-    single { SendOtpCodeUseCaseImpl(get()) }
-    single { ReSendOtpCodeUseCaseImpl(get()) }
+    single<SendOtpCodeUseCase> { SendOtpCodeUseCaseImpl(get()) }
+    single<ReSendOtpCodeUseCase> { ReSendOtpCodeUseCaseImpl(get()) }
 }
 
 val cacheManagerModule = module {
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -74,6 +74,7 @@ import com.hedvig.app.feature.embark.passages.textaction.TextActionParameter
 import com.hedvig.app.feature.embark.passages.textaction.TextActionViewModel
 import com.hedvig.app.feature.genericauth.CreateOtpAttemptUseCase
 import com.hedvig.app.feature.genericauth.GenericAuthViewModel
+import com.hedvig.app.feature.genericauth.otpinput.OtpInputViewModel
 import com.hedvig.app.feature.home.data.GetHomeUseCase
 import com.hedvig.app.feature.home.service.HomeTracker
 import com.hedvig.app.feature.home.ui.HomeViewModel
@@ -374,6 +375,7 @@ val viewModelModule = module {
         CrossSellDetailViewModel(notificationMetadata, crossSell, get())
     }
     viewModel { GenericAuthViewModel(get()) }
+    viewModel { OtpInputViewModel() }
 }
 
 val choosePlanModule = module {
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -75,6 +75,8 @@ import com.hedvig.app.feature.embark.passages.textaction.TextActionViewModel
 import com.hedvig.app.feature.genericauth.CreateOtpAttemptUseCase
 import com.hedvig.app.feature.genericauth.GenericAuthViewModel
 import com.hedvig.app.feature.genericauth.otpinput.OtpInputViewModel
+import com.hedvig.app.feature.genericauth.otpinput.ReSendOtpCodeUseCase
+import com.hedvig.app.feature.genericauth.otpinput.SendOtpCodeUseCase
 import com.hedvig.app.feature.home.data.GetHomeUseCase
 import com.hedvig.app.feature.home.service.HomeTracker
 import com.hedvig.app.feature.home.ui.HomeViewModel
@@ -375,7 +377,7 @@ val viewModelModule = module {
         CrossSellDetailViewModel(notificationMetadata, crossSell, get())
     }
     viewModel { GenericAuthViewModel(get()) }
-    viewModel { OtpInputViewModel() }
+    viewModel { (otpId: String, credential: String) -> OtpInputViewModel(otpId, credential, get(), get(), get()) }
 }
 
 val choosePlanModule = module {
@@ -608,6 +610,8 @@ val useCaseModule = module {
     single { GraphQLQueryUseCase(get()) }
     single { GetCrossSellsUseCase(get(), get()) }
     single { CreateOtpAttemptUseCase(get()) }
+    single { SendOtpCodeUseCase(get()) }
+    single { ReSendOtpCodeUseCase(get()) }
 }
 
 val cacheManagerModule = module {
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -75,8 +75,8 @@ import com.hedvig.app.feature.embark.passages.textaction.TextActionViewModel
 import com.hedvig.app.feature.genericauth.CreateOtpAttemptUseCase
 import com.hedvig.app.feature.genericauth.GenericAuthViewModel
 import com.hedvig.app.feature.genericauth.otpinput.OtpInputViewModel
-import com.hedvig.app.feature.genericauth.otpinput.ReSendOtpCodeUseCase
-import com.hedvig.app.feature.genericauth.otpinput.SendOtpCodeUseCase
+import com.hedvig.app.feature.genericauth.otpinput.ReSendOtpCodeUseCaseImpl
+import com.hedvig.app.feature.genericauth.otpinput.SendOtpCodeUseCaseImpl
 import com.hedvig.app.feature.home.data.GetHomeUseCase
 import com.hedvig.app.feature.home.service.HomeTracker
 import com.hedvig.app.feature.home.ui.HomeViewModel
@@ -610,8 +610,8 @@ val useCaseModule = module {
     single { GraphQLQueryUseCase(get()) }
     single { GetCrossSellsUseCase(get(), get()) }
     single { CreateOtpAttemptUseCase(get()) }
-    single { SendOtpCodeUseCase(get()) }
-    single { ReSendOtpCodeUseCase(get()) }
+    single { SendOtpCodeUseCaseImpl(get()) }
+    single { ReSendOtpCodeUseCaseImpl(get()) }
 }
 
 val cacheManagerModule = module {
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthActivity.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthActivity.kt
@@ -10,6 +10,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.ui.res.stringResource
 import com.hedvig.app.BaseActivity
 import com.hedvig.app.R
+import com.hedvig.app.feature.genericauth.otpinput.OtpInputActivity
 import com.hedvig.app.ui.compose.theme.HedvigTheme
 import com.hedvig.app.util.extensions.compatSetDecorFitsSystemWindows
 import org.koin.androidx.viewmodel.ext.android.viewModel
@@ -23,7 +24,13 @@ class GenericAuthActivity : BaseActivity() {
 
         setContent {
             val viewState by model.viewState.collectAsState()
+            val events = model.eventsFlow.collectAsState(initial = null)
+
             HedvigTheme {
+                when (val value = events.value) {
+                    is GenericAuthViewModel.Event.SubmitEmailSuccess -> startOtpInputActivity(value)
+                }
+
                 EmailInputScreen(
                     onUpClick = ::finish,
                     onInputChanged = model::setInput,
@@ -46,6 +53,11 @@ class GenericAuthActivity : BaseActivity() {
                 R.string.login_text_input_email_error_not_valid
         }
     )
+
+    private fun startOtpInputActivity(value: GenericAuthViewModel.Event.SubmitEmailSuccess) {
+        val intent = OtpInputActivity.newInstance(this, value.id, value.credential)
+        startActivity(intent)
+    }
 
     companion object {
         fun newInstance(context: Context) = Intent(context, GenericAuthActivity::class.java)
--- a/app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
+++ b/app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
@@ -3,14 +3,23 @@ package com.hedvig.app.feature.genericauth
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.hedvig.app.feature.embark.EMAIL_REGEX
+import kotlinx.coroutines.channels.Channel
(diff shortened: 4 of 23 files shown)


app/src/main/java/com/hedvig/app/ui/compose/composables/ErrorDialog.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.compose.material.AlertDialog
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.ui.res.stringResource
+import com.hedvig.app.R
+
+@Composable
+fun ErrorDialog(
```
hedvigoscar, line 12: Where is this used? 🤔

app/src/main/java/com/hedvig/app/ui/compose/composables/FullScreenProgressOverlay.kt
```diff
@@ -0,0 +1,81 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.ExperimentalUnitApi
+import androidx.compose.ui.unit.dp
+import com.hedvig.app.R
+import com.hedvig.app.ui.compose.theme.HedvigTheme
+
+@OptIn(ExperimentalUnitApi::class, ExperimentalAnimationApi::class)
+@Composable
+fun FullScreenProgressOverlay(show: Boolean) {
```
hedvigoscar, line 35: 📸

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/SendOtpCodeUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+        else -> OtpErrorEvent.UNKNOWN
+    }
+
+    sealed class OtpResult {
+        data class Success(val authToken: String) : OtpResult()
+        data class NetworkError(val message: String?) : OtpResult()
+        data class OtpError(val error: ErrorEvent) : OtpResult()
+    }
+
+    enum class OtpErrorEvent(@StringRes private val resourceId: Int) : ErrorEvent {
```
hedvigoscar, line 45: I feel like the view-layer might be leaking out a bit when string-resources find themselves down in use-cases 🤔

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthActivity.kt
```diff
@@ -20,7 +21,13 @@ class GenericAuthActivity : BaseActivity() {
 
         setContent {
             val viewState by model.viewState.collectAsState()
+            val events = model.eventsFlow.collectAsState(initial = null)
+
             HedvigTheme {
+                when (val value = events.value) {
+                    is GenericAuthViewModel.Event.SubmitEmailSuccess -> startOtpInputActivity(value)
+                }
```
StylianosGakis, line 29: I think receiving one time events as state and doing the `startOtpInputActivity` in composition is dangerous as "composables should be side-effect free" https://developer.android.com/jetpack/compose/side-effects .
Since the event will be saved as state, if the composable recomposes for whatever reason (there's a lot of reasons why it potentially could even if it maybe doesn't right now) it will trigger this again.
We should be subscribing to the state outside of composition in some way. Maybe something like (I wrote this outside of the IDE not sure if it compiles but you get the idea):
```kotlin
LaunchedEffect(Unit) {
    model.eventsFlow // This could be named just events too, "...Flow" suffix is maybe not necessary
        .collect { event -> // Or use `collectLatest` if we want to drop events and not chain them
            when (val value = model.eventsFlow) {
                is GenericAuthViewModel.Event.SubmitEmailSuccess -> startOtpInputActivity(value)
            }
        }
}
```

app/src/main/java/com/hedvig/app/util/extensions/ActivityExtensions.kt
```diff
@@ -133,6 +133,11 @@ private fun Activity.openAppSettings() {
     startActivity(intent)
 }
 
+fun Activity.openEmail(title: String) {
+    val intent = Intent(Intent.ACTION_SEND)
```
hedvigoscar, line 137: How does this work? 🤔

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
```diff
@@ -3,14 +3,23 @@ package com.hedvig.app.feature.genericauth
...
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 
 class GenericAuthViewModel(
     private val createOtpAttemptUseCase: CreateOtpAttemptUseCase,
 ) : ViewModel() {
+    private val eventChannel = Channel<Event>(Channel.UNLIMITED)
+    val eventsFlow = eventChannel.receiveAsFlow()
```
StylianosGakis, line 17: `_events` and `events` is usually what we've been naming those, I find the type suffix to not be super necessary since we get IDE type hints and all that jazz anyway.
I think Oscar thinks the same too, but not too strong opinion on this.

hedvigoscar (commented):
Some architectural questions. Mostly looking real good 😎

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputActivity.kt
```diff
@@ -0,0 +1,117 @@
...
+                    modifier = Modifier.systemBarsPadding(top = true),
+                ) {
+                    val viewState by model.viewState.collectAsState()
+                    val events = model.eventsFlow.collectAsState(initial = null)
+                    val openDialog = remember { mutableStateOf(false) }
+
+                    LaunchedEffect(events.value) {
+                        when (events.value) {
+                            is OtpInputViewModel.Event.Success -> startLoggedIn()
+                            OtpInputViewModel.Event.CodeResent -> launch {
```
StylianosGakis, line 69: Is the `launch` necessary here since we're already in the coroutine scope of LaunchedEffect?

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/SendOtpCodeUseCase.kt
hugokallstrom: Yeah I agree.. Do you know of any good solutions for error handling?

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputActivity.kt
```diff
@@ -0,0 +1,117 @@
...
+                                delay(1000)
+                                val message = getString(R.string.login_snackbar_code_resent)
+                                scaffoldState.snackbarHostState.showSnackbar(message)
```
StylianosGakis, line 72: How does this snackbar leave? When a new event state comes in cancelling the previous coroutine job? Do we want it to stay there forever? Otherwise can we do something like
```kotlin
withTimeout(5000) {
    scaffoldState.snackbarHostState.showSnackbar(message)
}
```
To show it for 5000 (or whatever) milliseconds?

app/src/main/java/com/hedvig/app/ui/compose/composables/ErrorDialog.kt
hugokallstrom: In OtpInputActivity

app/src/main/AndroidManifest.xml
```diff
@@ -312,6 +312,10 @@
             android:name=".feature.genericauth.GenericAuthActivity"
             android:windowSoftInputMode="adjustResize" />
 
+        <activity
```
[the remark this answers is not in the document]
hugokallstrom: Yeah maybe we should look at migrating to compose navigation

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputActivity.kt
```diff
@@ -0,0 +1,117 @@
...
+                        TopAppBarWithBack(
+                            onClick = ::onBackPressed,
+                            title = stringResource(R.string.login_navigation_bar_center_element_title)
+                        )
+                    },
+                    scaffoldState = scaffoldState,
+                    modifier = Modifier.systemBarsPadding(top = true),
+                ) {
+                    val viewState by model.viewState.collectAsState()
+                    val events = model.eventsFlow.collectAsState(initial = null)
```
StylianosGakis, line 63: Probably the same comment as above about saving events as Compose State<T>. Even though here you're launching a new LaunchedEffect when the value changes maybe we want to have one approach that we always take for event handling.

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputViewModel.kt
```diff
@@ -0,0 +1,110 @@
...
+        val errorEvent: ErrorEvent? = null,
+        val loadingResend: Boolean = false,
+        val loadingCode: Boolean = false
+    )
+
+    sealed class Event {
+        data class Success(val authToken: String) : Event()
+        object ShowDialog : Event()
+        object CodeResent : Event()
+        object None : Event()
```
StylianosGakis, line 40: Why do we need None as an event? Is it to clear past events? If that's so, it probably stems from the fact that we're saving the event as a state?

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/SendOtpCodeUseCase.kt
```diff
@@ -0,0 +1,54 @@
...
+            OtpResult.Success(it.accessToken)
+        } ?: OtpResult.OtpError(OtpErrorEvent.UNKNOWN)
+    }
+
+    private fun parseErrorCode(errorCode: String) = when (errorCode) {
+        "TOO_MANY_ATTEMPTS" -> OtpErrorEvent.TOO_MANY_ATTEMPTS
+        "ALREADY_COMPLETED" -> OtpErrorEvent.ALREADY_COMPLETED
+        "EXPIRED" -> OtpErrorEvent.EXPIRED
+        "WRONG_OTP" -> OtpErrorEvent.WRONG_OTP
+        else -> OtpErrorEvent.UNKNOWN
```
StylianosGakis, line 36: Can we have these strigns as an Enum, or as a GraphQL type that the backend gives us or as anything typed really so that we don't forget to fix it whenever we change it?

app/src/main/java/com/hedvig/app/ui/compose/composables/ErrorDialog.kt
```diff
@@ -0,0 +1,33 @@
...
+    show: MutableState<Boolean>,
```
StylianosGakis, line 13: Since we only need the "setting" part of the mutableState and we never use `show.value` here, wouldn't this be better off as a callback to change the state that is hoisted up?

StylianosGakis (changes requested):
Super nice! It looks great in general, just have some concerns mainly all revolving around events and compose. I really think they need to be addressed and have us all be in the same page because if I am right in what I wrote we may have a problem with triggering events as it is now.

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
StylianosGakis: Compose navigation is heavily tied with Jetpack Navigation which I am not sure if it can work as a thing in the "leaf" nodes. I've only seen it encompass the entire application navigation graph. But might be worth looking into.

app/src/main/java/com/hedvig/app/util/extensions/ActivityExtensions.kt
hugokallstrom: Good catch, this needed more testing. Fixed now

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthActivity.kt
hugokallstrom: Thanks for clarifying!! I'll read up and fix 👍

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
hugokallstrom: Good point :) Will change to the standard we use.

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputActivity.kt
hugokallstrom: It disappears automatically

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputViewModel.kt
hugokallstrom: Yeah this is ugly. I couldnt find a better solution, see here: https://stackoverflow.com/a/69358848

app/src/main/java/com/hedvig/app/ui/compose/composables/ErrorDialog.kt
hugokallstrom: Good idea

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/SendOtpCodeUseCase.kt
hugokallstrom: Yeah that would be ideal. I'll try and find a better solution for this.

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputViewModel.kt
StylianosGakis: Aha I see. Can you test this with the solution given above? Of receiving the event inside the LaunchedEffect and never saving it in a state therefore not relying on clearing it and setting it again as all unique events should be consumed even if they are the same since they come from a Channel and consumed right away.

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
hedvigoscar: It's not impossible to use it for this use-case.

Note that implementing navigation on your own with Compose is completely trival, so the investment need not be that big 😄

Here's an example:

```kotlin
sealed class NavigationDestination {
    object Example1: NavigationDestination()
    object Example2: NavigationDestination()
}

@Composable
fun Navigation() {
    var currentDestination by remember { mutableStateOf(NavigationDestination.Example1) }
    val navigateTo = { navigationDestination: NavigationDestination -> currentDestination = navigationDestination }
    when (currentDestination) {
        NavigationDestination.Example1 -> { Example1Composable(navigateTo) }
        NavigationDestination.Example2 -> { Example2Composable(navigateTo) }
    }
}
```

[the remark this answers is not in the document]
hedvigoscar: caveat: I didn't compile this code, I simply wrote it inline in this comment.

Navigation in Compose is actually that simple 😅

app/src/main/java/com/hedvig/app/util/extensions/ActivityExtensions.kt
```diff
@@ -133,6 +136,35 @@ private fun Activity.openAppSettings() {
...
+        val openInChooser = Intent.createChooser(intentChooser, title)
+
+        // Then create a list of LabeledIntent for the rest of the registered email apps
+        val emailApps = resInfo.toLabeledIntentArray(packageManager)
+
+        // Add the rest of the email apps to the picker selection
+        openInChooser.putExtra(Intent.EXTRA_INITIAL_INTENTS, emailApps)
+        startActivity(openInChooser)
+    } else {
+        Timber.e("No email app found")
```
hedvigoscar, line 158:
```suggestion
        e { "No email app found" }
```

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/SendOtpCodeUseCase.kt
hedvigoscar: It depends a bit on the use-case.

For attaching diagnostic information for developer purposes, I think having something like `interface Error { val message: String }` could be ok, but for more result-style types I prefer having the types be specified as members of a sealed class, and then matched upon later on.

It is a style of error handling that can be slightly verbose at times, but I find the trade-off worth it (no 'invisible' error-handling, like with exceptions, which often can turn into simply not handling errors).

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
hedvigoscar: Agreed - the style is commonly called hungarian notation, and I believe it to be superfluous when using a modern IDE and a well-behaved statically typed language 😄

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
StylianosGakis: Oh if we mean like this then absolutely it's trivial. Could even slap a CrossFade composable around this and get animations too.
But I don't know how well this handles backstacks, restoration after process death, deep links to the screens we'll be making and much more I'm missing probably. Depending on how much we do this the more things we'd need to consider. And the more we might do it the more we'd need to consider a proper investment in androidx-navigation.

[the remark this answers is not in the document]
hugokallstrom: That's true Stelios. I suggest we take a look at navigation together as a separate PR.

app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -606,6 +610,8 @@ val useCaseModule = module {
     single { GraphQLQueryUseCase(get()) }
     single { GetCrossSellsUseCase(get(), get()) }
     single { CreateOtpAttemptUseCase(get()) }
+    single { SendOtpCodeUseCaseImpl(get()) }
+    single { ReSendOtpCodeUseCaseImpl(get()) }
```
hedvigoscar, line 614:
```suggestion
    single<SendOtpCodeUseCase> { SendOtpCodeUseCaseImpl(get()) }
    single<ReSendOtpCodeUseCase> { ReSendOtpCodeUseCaseImpl(get()) }
```

Isn't this necessary in order to bind against the `interface`? 🤔

app/src/test/java/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
```diff
@@ -0,0 +1,211 @@
...
+    @ExperimentalCoroutinesApi
+    @get:Rule
+    var mainCoroutineRule = MainCoroutineRule()
+
+    private var otpResult: OtpResult = OtpResult.Success("authtest")
+    private var resendOtpResult: ResendOtpResult = ResendOtpResult.Success("authtest")
+
+    @Before
+    fun setup() {
+        viewModel = OtpInputViewModel(
```
hedvigoscar, line 36: Can't this simply be declared top-level? 🤔

app/src/test/java/com/hedvig/app/util/coroutines/MainCoroutineRule.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.TestCoroutineScope
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@ExperimentalCoroutinesApi
+class MainCoroutineRule(
+    private val dispatcher: TestCoroutineDispatcher = TestCoroutineDispatcher()
```
hedvigoscar, line 14: Should this `dispatcher` be exposed, in order to allow access to `runBlockingTest`? 🤔

app/src/test/java/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
```diff
@@ -0,0 +1,211 @@
...
+                override suspend fun invoke(credential: String): ResendOtpResult {
+                    return resendOtpResult
+                }
+            }
+        )
+    }
+
+    @ExperimentalCoroutinesApi
+    @Test
+    fun testNetworkError() = runBlockingTest {
```
hedvigoscar, line 61: Should these tests use the `runBlockingTest`-method from the `TestCoroutineDispatcher` in the `MainCoroutineRule`? 🤔

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+    private val credential: String,
+    private val authenticationTokenService: AuthenticationTokenService,
+    private val sendOtpCodeUseCase: SendOtpCodeUseCase,
+    private val reSendOtpCodeUseCase: ReSendOtpCodeUseCase,
+) : ViewModel() {
+    private val _viewState = MutableStateFlow(ViewState(credential = credential))
+    val viewState = _viewState.asStateFlow()
+
+    private val _events = Channel<Event>(Channel.UNLIMITED)
+    val events = _events.receiveAsFlow()
```
hedvigoscar, line 24: I'm slightly scared that this is incorrect and will drop events, which is embarrasing as I was the one who put this pattern in place in the first place.

It appears to be really hard to figure out what the truly correct method is, but I think the correct code is probably the following:

```suggestion
    val events: ReceiveChannel = _events
```

this could then be simplified to the following:

```kotlin
fun <T> Channel<T>.asReceiver(): ReceiveChannel<T> = this
```

with the consumer being converted to:

```kotlin
model
	.events
	.consumeAsFlow()
	.flowWithLifecycle(lifecycleOwner)
	.onEach { /* ... */ }
	.launchIn(lifecycleScope)
```

If my fear is correct, then we risk the same data loss that we got when using `SharedFlow` for events happening here as well, which would not be ideal.

We don't need to remedy this issue in the context of this PR, but I could take it upon myself to refactor all uses of the current pattern to what I believe the be the correct method.

Check my code before we commit to that, though! @hugokallstrom @StylianosGakis

hedvigoscar (approved):
LGTM!

app/src/main/java/com/hedvig/app/feature/genericauth/otpinput/OtpInputViewModel.kt
hugokallstrom: That sounds reasonable imo!

PR #810, opened 2021-11-22
Fix nulls being stored as `"null"` in Embark Store

hedvigoscar:
### Checklist


StylianosGakis (approved):
So how does it look like in embark after parsing the null as a proper null instead of "null"?
Is this a first step to enable the solution?

PR #812, opened 2021-11-23
APP-1089: Fetch insurance providers 

hugokallstrom:
### Checklist


StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/GetInsuranceProvidersUseCase.kt
```diff
@@ -17,8 +22,22 @@ data class InsuranceProvider(
     val name: String
 )
 
-class GetInsuranceProvidersUseCaseImpl : GetInsuranceProvidersUseCase {
+class GetInsuranceProvidersUseCaseImpl(
+    private val apolloClient: ApolloClient,
+    private val localeManager: LocaleManager
+) : GetInsuranceProvidersUseCase {
```
hedvigoscar, line 28: Do we need this Interface-Implementation Pair, or can we inline the interface into the parent class?

app/src/test/java/com/hedvig/app/feature/embark/passages/ExternalInsurerViewModelTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    private var insuranceProvidersResult: InsuranceProvidersResult =
+        InsuranceProvidersResult.Success(insuranceProviders)
+
+    @ExperimentalCoroutinesApi
+    @get:Rule
+    var mainCoroutineRule = MainCoroutineRule()
+
+    private val getInsuranceProvidersUseCase = object : GetInsuranceProvidersUseCase {
```
hugokallstrom, line 30: Implementing the interface here, in order to test different results (and not depending on apollo and network). I think this works really well! The test does not need to know how the result is created, since we are testing the view model. We also have the possibility to change the implementation of the use case to use a database, another networking solution etc. without affecting the view model test.

hedvigoscar: I see! How does this differ from ```kotlin
val getInsuranceProvidersUseCase = mockk<GetInsuranceProvidersUseCase()

every { getInsuranceProvidersUseCase.getInsuranceProviders() } returns insuranceProvidersResult
```

though? 😄

hugokallstrom: Good point, mocking is much cleaner. Fixed!

hedvigoscar (approved):

PR #816, opened 2021-11-26
POC: Robolectric Tests

hedvigoscar:
### Checklist


hugokallstrom (approved):
Cool! What are the benefits of using Roboelectric compared to what we have now for UI tests?

hedvigoscar:
Tests using Robolectric run on the JVM instead of on an emulator, making them significantly faster. The tradeoff is that you're not actually running your tests on a real device, which means that you may miss some code being tested.

hugokallstrom:
Sounds like a reasonable tradeoff :)

PR #817, opened 2021-11-26
APP-1107: Store selected insurance provider

hugokallstrom:
### Checklist

### DIFF
--- a/apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/embarkStory.graphql
+++ b/apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/embarkStory.graphql
@@ -137,7 +137,8 @@ query EmbarkStory($name: String!, $locale: String!) {
             }
             skip {
               ...EmbarkLinkFragment
-            }
+            },
+            storeKey
           }
         }
         ... on EmbarkNumberAction {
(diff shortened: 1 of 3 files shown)


StylianosGakis (commented):
Not quite familiar with all this. Maybe @hedvigoscar should take a look?

apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/embarkStory.graphql
```diff
@@ -137,7 +137,8 @@ query EmbarkStory($name: String!, $locale: String!) {
             }
             skip {
               ...EmbarkLinkFragment
-            }
+            },
```
hedvigoscar, line 140: From where are we actually getting the list of insurance providers to show in this thing? 🤔

hedvigoscar (approved):
One question, LGTM though

apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/embarkStory.graphql
hugokallstrom: Good questions. That would be from 
```
insuranceProviders(locale: $locale) {
    id
    name
    externalCollectionId
  }
```
I suppose it would be better if this was done in embark and populated on `EmbarkExternalInsuranceProviderAction`.

hedvigoscar: Agreed! I believe we wrote an extension in giraffe for this last time we had to deal with `insuranceProviders`, perhaps we should do the same here as well?

PR #819, opened 2021-11-30
APP-752: Insurely authentication and collection status.

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/authenticate/insurely/InsurelyAuthViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.coroutines.launch
+
+class InsurelyAuthViewModel(
+    private val reference: String,
+    private val getDataCollectionUseCase: GetDataCollectionUseCase
+) : ViewModel() {
+
+    private val _viewState = MutableStateFlow(ViewState())
+    val viewState: StateFlow<ViewState>
+        get() = _viewState
```
StylianosGakis, line 18: We usually write this as
```kotlin
val viewState = _viewState.asStateFlow()
```
Infers the type well by itself too, pretty neat.
And might be good to keep it consistent in case we want to search for all such occasions in the project for whatever reason.

app/src/main/java/com/hedvig/app/authenticate/insurely/InsurelyAuthViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    data class ViewState(
+        val autoStartToken: String? = null,
+        val authStatus: DataCollectionResult.Success.CollectionStatus? = null,
+        val error: DataCollectionResult.Error? = null,
+    )
```
StylianosGakis, line 24: Does the Error case maybe need to be mutually exclusive from the other states?
Is there a particular reason an error can exist at the same time that either of the other two variables?
If not a sealed class maybe fits this case better, what do you think?

app/src/main/java/com/hedvig/app/authenticate/insurely/InsurelyAuthViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    init {
+        viewModelScope.launch {
+            getDataCollectionUseCase.getCollectionStatus(reference).collect { result ->
```
StylianosGakis, line 28: I was thinking something about these cases.
Since the we launch this on the viewModelScope, on a hot StateFlow, this would still be actively listening to the subscription even if there are no subscribers to it right?

Inspired from [something I read a while back](https://medium.com/androiddevelopers/migrating-from-livedata-to-kotlins-flow-379292f419fb) under the section "The WhileSubscribed strategy" the entire class could also look like this:
```kotlin
class InsurelyAuthViewModel(
    reference: String,
    getDataCollectionUseCase: SubscribeToDataCollectionUseCase,
) : ViewModel() {

    val viewState: StateFlow<ViewState> = getDataCollectionUseCase
        .getCollectionStatus(reference)
        .map { result ->
            when (result) {
                is DataCollectionResult.Error -> ViewState.Error
                is DataCollectionResult.Success.NorwegianBankId -> ViewState.Success(
                    autoStartToken = result.norwegianBankIdWords,
                    authStatus = result.status,
                )
                is DataCollectionResult.Success.SwedishBankId -> ViewState.Success(
                    autoStartToken = result.autoStartToken,
                    authStatus = result.status,
                )
            }
        }
        .stateIn(
            scope = viewModelScope,
            started = SharingStarted.WhileSubscribed(5000),
            initialValue = ViewState.Initial,
        )


    sealed class ViewState {
        object Initial : ViewState() // Since I saw the the initial was just the model with all parameters being null maybe a plain "Initial" state is adequate?
        object Error : ViewState()
        data class Success(
            val autoStartToken: String?,
            val authStatus: DataCollectionResult.Success.CollectionStatus,
        ) : ViewState()
    }
}
```
With the extra benefit that the class is now even smaller.

Any opinions on this? @hedvigoscar you too.

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -9,23 +9,50 @@ import androidx.compose.runtime.Composable
...
+        intent.getParcelableExtra(PARAMETER)
+            ?: InsuranceProviderParameter("Test")
+            ?: throw Error("Programmer error: DATA is null in ${this.javaClass.name}")
+    }
+
+    private val viewModel: RetrievePriceViewModel by viewModel { parametersOf(parameter) }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        supportFragmentManager.setFragmentResultListener(
```
StylianosGakis, line 41: Nice that this worked!

StylianosGakis (approved):
Great! Some more stylistic comments from me, just merge if you want to.

app/src/main/java/com/hedvig/app/authenticate/insurely/InsurelyAuthViewModel.kt
hedvigoscar: Good points, and it looks good, I think.

hugokallstrom: Agree, looks 👌

PR #821, opened 2021-11-30
APP-1119: Insurely failure screen

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -63,9 +61,23 @@ class RetrievePriceInfoActivity : BaseActivity() {
...
+                            ErrorDialog(
+                                onDismiss = {},
+                                message = stringResource(id = event.errorResult.getStringResource())
+                            )
+                        }
+                        is RetrievePriceViewModel.Event.AuthInformation -> {
+                            InsurelyDialog.newInstance(event.reference)
+                                .show(supportFragmentManager, AuthenticateDialog.TAG)
+                        }
+                    }
```
StylianosGakis, line 77: Collecting events as State is kinda scary. And more specifically using this state inside the when() statement which is inside the compose scope (not inside a lambda) means that if this recomposes for whatever reason (always assume it will) it will trigger InsurelyDialog.show more than once. You could test this by animating something in the composable for example and see .show get called a gazillion times 😅
So the suggestion here is two-fold:
* Never do side-effects inside the composable scope if we're not in a lambda which is not composable (Inside the onClick: () -> Unit on a Button for example is a perfectly fine place to have side-effects)
* Do not collect events as state and do something like this instead:
```kotlin
LaunchedEffect(Unit) {
    viewModel.events.collect { event -> // or collectLatest if you don't care about potentially dropping older events
        when(event) { 
            // etc 
        }
    }
}
```
I am not 100% sure this approach is flawless or if it can be improved further, but it's probably a good first step until and if we find any reason not to do it this way.

StylianosGakis (approved):
Looking good, with a comment on some compose-event-state shenanigans that need to be changed. Added a proposal on the change.

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
StylianosGakis: And since then you'd be on non-compose land, and you'd still want to show the ErrorDialog, I'd probably flip a boolean like: `val showError by rememberSaveable { mutableStateOf(false) }` (Maybe saveable so that the dialog doesn't close itself on configuration change etc.? I think a simple `remember` would lose state on config change)
And then also flip it again inside the `onDismiss = {}` to make it dismissable by changing the local state.

hedvigoscar: Collecting events in a SideEffect looks correct. The rememberSaveable parts are a bit trickier - when you show a dialog, it will generally survive configuration changes, but not process death.

StylianosGakis: Is that necessarily true though? If we're presenting some important dialog to them, and they get a phone call, talk for X minutes and come back, do we want the dialog to disappear?

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -125,11 +124,13 @@ fun RetrievePriceScreen(
                 RetrievePriceContent(
                     onRetrievePriceInfo = viewModel::onRetrievePriceInfo,
                     onIdentityInput = viewModel::onIdentityInput,
+                    onDismissError = viewModel::onDismissError,
                     input = viewState.input,
                     title = viewState.market?.titleRes()?.let { stringResource(it) } ?: "",
                     placeholder = viewState.market?.placeHolderRes()?.let { stringResource(it) } ?: "",
                     label = viewState.market?.labelRes()?.let { stringResource(it) } ?: "",
                     inputErrorMessage = viewState.inputError?.errorTextKey?.let { stringResource(it) },
+                    errorMessage = viewState.inputError?.errorTextKey?.let { stringResource(it) }
```
hugokallstrom, line 133: I think the error state makes sense to embed in the content composable here, since the error state originates from the viewmodel.

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -53,6 +54,18 @@ class RetrievePriceInfoActivity : BaseActivity() {
...
+            .flowWithLifecycle(lifecycle)
+            .onEach { event ->
+                when (event) {
+                    is RetrievePriceViewModel.Event.AuthInformation -> {
+                        InsurelyDialog.newInstance(event.reference)
+                            .show(supportFragmentManager, AuthenticateDialog.TAG)
+                    }
+                }
+            }
+            .launchIn(lifecycleScope)
```
hugokallstrom, line 67: Moved this outside of the composable

StylianosGakis (approved):

PR #825, opened 2021-12-07
APP-746: Insurely result request in offer screen

StylianosGakis:
### Checklist

- [X] Functionality is accessible in engineering mode: Offer -> Bottom of the page


app/src/engineering/java/com/hedvig/app/feature/offer/MockOfferViewModel.kt
```diff
@@ -61,41 +74,66 @@ class MockOfferViewModel : OfferViewModel() {
...
         var shouldError = false
-        var mockData = OFFER_DATA_SWEDISH_APARTMENT
+        var mockData: OfferMockData = OfferMockData(OFFER_DATA_SWEDISH_APARTMENT)
+        var mockRefreshEvery2Seconds = false
+
+        data class OfferMockData(
+            val offer: OfferQuery.Data,
+            val dataCollectionValue: SubscribeToDataCollectionUseCase.Status? = null,
+            val externalInsuranceData: DataCollectionResultQuery.Data? = null,
+        ) {
```
StylianosGakis, line 123: Since the data now come from 3 different queries, just bunched them up in a class

app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -419,9 +422,31 @@ val marketingModule = module {
 }
 
 val offerModule = module {
-    viewModel<OfferViewModel> { (ids: List<String>, shouldShowOnNextAppStart: Boolean) ->
-        OfferViewModelImpl(ids, get(), get(), get(), get(), get(), get(), get(), shouldShowOnNextAppStart, get(), get())
+    viewModel<OfferViewModel> { parametersHolder: ParametersHolder ->
+        val (
+            ids: List<String>, shouldShowOnNextAppStart: Boolean, insurelyDataCollectionReferenceUuid: String?,
```
StylianosGakis, line 427: Only `insurelyDataCollectionReferenceUuid` changed here, but added named args because it was hard to reason about the order

app/src/main/java/com/hedvig/app/feature/offer/OfferItemsBuilder.kt
```diff
@@ -92,69 +72,217 @@ object OfferItemsBuilder {
...
-                    OfferModel.CurrentInsurer(
-                        displayName = currentInsurer.displayName,
-                        associatedQuote = if (bundle.quotes.size > 1) {
-                            associatedQuote
-                        } else {
-                            null
+        }
+        if (showInsuranceSwitchableStates) {
+            addAll(currentInsuranceSwitchableStates(bundle.quotes))
+        }
```
StylianosGakis, line 139: This part was simply moved from the bottom items to the top items

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -144,6 +149,7 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
             )
 
             binding.offerScroll.adapter = concatAdapter
+            binding.offerScroll.itemAnimator = ViewHolderReusingDefaultItemAnimator()
```
StylianosGakis, line 152: To enable the compose animations not being hid by the RV diffing animation

app/src/main/java/com/hedvig/app/feature/offer/ui/composable/insurely/InsurelyCard.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.hedvig.app.feature.offer.ui.OfferModel.InsurelyCard.FailedToRetrieve
+import com.hedvig.app.feature.offer.ui.OfferModel.InsurelyCard.Loading
+import com.hedvig.app.feature.offer.ui.OfferModel.InsurelyCard.Retrieved
+import com.hedvig.app.ui.compose.theme.HedvigTheme
+import com.hedvig.app.ui.compose.theme.hedvigBlack12percent
+import com.hedvig.app.ui.compose.theme.hedvigContentColorFor
+import com.hedvig.app.util.compose.preview.previewData
+import java.util.UUID
+
+// TODO string resources
```
StylianosGakis, line 28: Waiting with this since the design doesn't seem to be set in stone anyway

app/src/main/java/com/hedvig/app/feature/offer/ui/composable/insurely/RetrievedInfo.kt
```diff
@@ -0,0 +1,132 @@
...
+                    )
+                } else {
+                    "YOUR INSURANCE${data.insuranceProvider?.let { " WITH $it" }}"
+                },
+                style = MaterialTheme.typography.caption
+            )
+        }
+        Spacer(Modifier.height(24.dp))
+        Text(
+            text = "${data.totalNetPremium}", // todo locale formatting
```
StylianosGakis, line 61: Will depend how the design ends up with. And if we want to force the currency before/after the number or if we can let it be as the current locale likes it

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -154,22 +160,35 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
                 .viewState
                 .flowWithLifecycle(lifecycle)
                 .onEach { viewState ->
+                    binding.progressBar.isVisible = viewState is OfferViewModel.ViewState.Loading
+                    binding.offerScroll.isVisible = viewState !is OfferViewModel.ViewState.Loading
+                    when (viewState) {
+                        is OfferViewModel.ViewState.Loading -> {}
```
hugokallstrom, line 166: show progress?

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
```diff
@@ -418,4 +498,27 @@ class OfferAdapter(
...
+                oldItem is OfferModel.CurrentInsurancesHeader && newItem is OfferModel.CurrentInsurancesHeader -> {
+                    // Should only display 1 CurrentInsurancesHeader ever
+                    true
+                }
+                else -> {
+                    oldItem == newItem
+                }
+            }
+
+            override fun areContentsTheSame(oldItem: OfferModel, newItem: OfferModel) = oldItem == newItem
```
hugokallstrom, line 521: Can we check if contents are same here?

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
```diff
@@ -418,4 +498,27 @@ class OfferAdapter(
...
+            override fun areItemsTheSame(oldItem: OfferModel, newItem: OfferModel): Boolean = when {
+                oldItem is OfferModel.InsurelyCard && newItem is OfferModel.InsurelyCard -> {
+                    oldItem.id == newItem.id
+                }
+                oldItem is OfferModel.CurrentInsurancesHeader && newItem is OfferModel.CurrentInsurancesHeader -> {
+                    // Should only display 1 CurrentInsurancesHeader ever
+                    true
+                }
+                else -> {
+                    oldItem == newItem
```
hugokallstrom, line 517: Can we check id here?

app/src/main/java/com/hedvig/app/feature/offer/usecase/ExternalInsuranceDataCollectionUseCase.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.hedvig.app.util.apollo.QueryResult
+import com.hedvig.app.util.apollo.safeQuery
+
+class ExternalInsuranceDataCollectionUseCase(
+    private val apolloClient: ApolloClient,
+) {
+    sealed class Result {
+        abstract val referenceUuid: String
+
+        data class Success(override val referenceUuid: String, val data: DataCollectionResultQuery.Data) : Result()
```
hugokallstrom, line 14: Map `DataCollectionResultQuery.Data` to another data class instead so we dont depend on GraphQL in the viewmodel

app/src/main/java/com/hedvig/app/feature/offer/usecase/insurelydatacollection/SubscribeToDataCollectionUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+    operator fun invoke(referenceUuid: String): Flow<Status> {
+        return apolloClient
+            .subscribe(DataCollectionStatusSubscription(referenceUuid))
+            .safeSubscription()
+            .map { queryResult ->
+                when (queryResult) {
+                    is QueryResult.Error -> Status.Error(referenceUuid)
+                    is QueryResult.Success -> Status.Content(
+                        referenceUuid,
+                        DataCollectionResult.fromDto(queryResult.data.dataCollectionStatusV2),
```
hugokallstrom, line 33: Very good :)

app/src/main/java/com/hedvig/app/util/LCE.kt
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.app.util
+
+sealed class LCE<out T> {
```
hugokallstrom, line 3: Cool! Do you think we can use this in other parts of the app? 

Is "LCE" a name from the functional world, like Left/Right?

hugokallstrom (changes requested):
GJ!

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
StylianosGakis: The functionality is the same as it was before the changes. The lines 163-164 are handling the loading state so there's nothing to put inside this case.

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
StylianosGakis: Since every single OfferModel subclass is a data class they should be handled properly with the built-in == implementation, so this should work as intended. Just like the GenericDiffUtilItemCallback does

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferAdapter.kt
StylianosGakis: Not all of them (if any) have a proper ID, so just carried over the exact same functionality as we had before (GenericDiffUtilItemCallback) which simply checks for object equality. Putting an ID to all the items is a bit out of scope for this and also maybe not needed since we're not really using the RV in this screen to update any of those items on the fly

app/src/main/java/com/hedvig/app/util/LCE.kt
StylianosGakis: Hmm yeah I haven't decided 100% on this either.

LCE is just short for Loading/Content/Error and all it does in this case is save us from creating yet another model in the case where whatever we're doing is adequately modeled with these three states.
Technically if anything else in the app is similarly modeled with the exact same 3 possible outcomes, they could use this LCE in order to just save time and the mental and IDE overhead of having to import yet another type which just contains those 3 cases inside of it, with its own package name etc which makes auto-completion harder.
Example of something we have in multiple places in the app:
image

Technically the same can be done for a result type where we can migrate all of our custom Result objects into [Insert whatever library here].
But all in all, this mostly saves on an extra type that has to live inside the class we're working on. 
image

In the end, whenever any of the cases broke out of the "Can be represented by these 2-3 standard cases" then we could go custom as we're doing already. That's why I'm not sure on it, I am afraid if we do it wrong it will just arbitrary constraints that don't need to be there in the first place.
But yeah I don't know, I don't have a strong opinion on it as of now, just trying out things and see what makes the most sense and we can decide on something all together if we feel like doing so.

hugokallstrom: I like it - and have thought of something similar. For this PR though I think we can go with what we currently have (eg have it in the viewmodel). Lets continue the discussion in slack!

StylianosGakis: Reverted for now as suggested. We can always come back to this in the future if we have more ideas! 😊

hugokallstrom (approved):
LGTM!

PR #831, opened 2021-12-13
Change font roughly done to get a feeling for it

StylianosGakis:
Do not merge!
Made just to play around with it and get a feeling of how it'd look like.


StylianosGakis:
Closed until we actually do this with the final font

PR #832, opened 2021-12-14
APP-1143, APP-1144, APP-1145, APP-1146: Add events for onboarding, offer and chat

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
```diff
@@ -285,3 +288,23 @@ fun Context.tryOpenUri(uri: Uri) {
...
+    val options =
+        ActivityOptionsCompat.makeCustomAnimation(
+            this,
+            R.anim.chat_slide_up_in,
+            R.anim.stay_in_place
+        )
+
+    ActivityCompat.startActivity(this, intent, options.toBundle())
+    trackingFacade.track("click_start_chat")
+}
```
StylianosGakis, line 310: Hmm this now makes it so that wherever we need to open the chat, we need to have a hold of the trackingFacade which might be a bit awkward in some contexts (let's say inside a composable inside a ViewHolder).
Would it make sense that we instead log this event inside ChatActivity instead, where it can fetch its own instance of the tracker? If not, is it a problem with duplicate events or something similar?

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/util/extensions/ContextExtensions.kt
hugokallstrom: You're right, I'll change! Good input :)

app/src/main/java/com/hedvig/app/feature/chat/ui/ChatActivity.kt
```diff
@@ -77,7 +79,8 @@ class ChatActivity : BaseActivity(R.layout.activity_chat) {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        tracker.closeChat()
+        trackingFacade.track("click_start_chat")
```
StylianosGakis, line 82: Do we need to add `if (savedInstanceState == null) { trackingFacade.track("click_start_chat") }` here to only show this on the first time the activity starts? Wouldn't we get duplicate events on the event of a process death?
I know I am the one that asked for this change but it'd be a shame if we get wrong duplicate events because of me 😅

PR #838, opened 2021-12-15
APP-1159: Persist ssn in value store for insurely

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -106,17 +110,17 @@ class RetrievePriceInfoActivity : BaseActivity() {
 @Composable
 fun RetrievePriceScreen(
     viewModel: RetrievePriceViewModel = viewModel(),
-    onContinue: (String?) -> Unit,
+    onContinue: (String?, String?) -> Unit,
```
StylianosGakis, line 113: Hmm maybe a good idea to put names on these two Strings? Like `(referenceResult: String?, ssn: String?)` To help not mess up the order of the parameters by accident in the future?

StylianosGakis (approved):

PR #841, opened 2021-12-16
APP-1147: Show details for every quote in bundle

hugokallstrom:
### Checklist

I removed the expandable content view here. The implementation was very shoehorned into the recycler view and didnt work well with showing multiple detail tables. It also caused some animation issues (with the toolbar for example). I don't think the user experience takes a big hit from this, and we remove quite a lot of complexity. 


StylianosGakis (approved):
Aha so is *this* what caused this weird buzzing behavior Hiro observed?
I agree it's not such a big difference, and even if it is, we'd probably much easier be able recreate this in Compose without the problems :D

PR #843, opened 2021-12-16
APP-1166: Weird dot when waiting for old insurance switch

StylianosGakis:
Claims Status cards were basically acting up when an empty list was sent into them because I didn't account for it. And the HorizontalPagerIndicator dots were showing one dot even though a size of 0 was sent into it 🤷‍♂️
This was the second bug that was introduced from the last QA by a list being empty when it should never be so, so I thought maybe actually bringing in the NonEmptyList is a net positive in the long run.


app/build.gradle.kts
```diff
@@ -188,39 +188,13 @@ dependencies {
...
-    implementation(libs.androidx.other.startup)
-    androidTestImplementation(libs.androidx.espresso.core)
-    androidTestImplementation(libs.androidx.espresso.intents)
-    androidTestImplementation(libs.androidx.espresso.contrib)
-    testImplementation(libs.androidx.test.junit)
-    testImplementation(libs.jsonTest)
-    androidTestImplementation(libs.androidx.test.runner)
-    androidTestImplementation(libs.androidx.test.rules)
-    androidTestImplementation(libs.androidx.test.junit)
-
```
StylianosGakis, line 218: I happened to be in the build.gradle.kts and I saw that we have all these written twice, must have been when I merged some stuff and messed this up 😅

app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -113,12 +115,10 @@ class HomeFragment : Fragment(R.layout.home_fragment) {
                 return@observe
             }
             if (isPending(successData.contracts)) {
-                val items = mutableListOf<HomeModel>().apply {
+                val items = buildList {
```
StylianosGakis, line 118: A bit of a nicer way to build those lists. I know we have to add the optIn for the experimental stdlib api but they will be stable in Kotlin 1.6 https://kotlinlang.org/docs/whatsnew16.html#stable-collection-builders so there's no risk of needing to migrate off of it anymore so we can use it.
Benefits mainly come from it returning the list as a non-mutable one outside of the lambda, so no way to accidentally mutate it when you shouldn't.

app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -213,9 +207,14 @@ class HomeFragment : Fragment(R.layout.home_fragment) {
...
     }
 
-    private fun claimStatusCards(successData: HomeQuery.Data): HomeModel.ClaimStatus = HomeModel.ClaimStatus(
-        successData.claimStatusCards.map(ClaimStatusCardData::fromClaimStatusCardsQuery)
-    )
+    private fun MutableList<HomeModel>.addClaimStatusCardsIfApplicable(successData: HomeQuery.Data) {
+        if (featureManager.isFeatureEnabled(Feature.CLAIMS_STATUS).not()) return
+        NonEmptyList.fromList(successData.claimStatusCards)
+            .map { claimStatusCardsQuery -> claimStatusCardsQuery.map(ClaimStatusCardData::fromClaimStatusCardsQuery) }
+            .tap { claimStatusCardDataList ->
```
StylianosGakis, line 214: `tap` function signature looks like this:
```
/**
* The given function is applied as a fire and forget effect
* if this is a `some`.
* When applied the result is ignored and the original
* Some value is returned
*
* Example:
* ```
* Some(12).tap { println("flower") } // Result: prints "flower" and returns: Some(12)
* none<Int>().tap { println("flower") }  // Result: does not print anything and returns: None
* ```
*/
public inline fun tap(f: (A) -> Unit): Option<A> =
when (this) {
    is None -> this
    is Some -> {
        f(this.value)
        this
    }
}
```

Basically short for having to do a when on the result and do nothing on `None` and what you want to do on `Some`

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCards.kt
```diff
@@ -15,6 +15,8 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import arrow.core.NonEmptyList
```
hugokallstrom, line 18: Do you think it would be difficult to roll our own implementation of NonEmptyList?

app/src/main/java/com/hedvig/app/feature/home/ui/HomeFragment.kt
hugokallstrom: Nice

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCards.kt
StylianosGakis: Technically it would not be difficult, but we'd just set ourselves up for more work in the long run by not taking advantage of all the nice things that come out of the box from arrow.
It's not like it's some obscure lib that nobody uses either, it's the most popular Koltin library that does this.
Plus we're not taking in the more specialized arrow components like Arrow Optics that require a kapt implementation too which would slow down builds, we just take core which can be simply seen as an extension to the kotlin stdlib.

Is there some particular reason why you would want to avoid it that you're thinking about? If you want Oscars input too, we did have a discussion some weeks ago where I mentioned looking into arrow core and he said he was positive about looking into it at least.

hugokallstrom: That sounds good. I haven't looked into arrow that much so just wanted to know what its about :) Thanks for the explanation! Go for it! 👍

hugokallstrom (approved):

PR #844, opened 2021-12-17
APP-1169: Connect payment at sign (NO + DK)

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/adyen/payin/AdyenConnectPayinActivity.kt
```diff
@@ -66,7 +60,7 @@ class AdyenConnectPayinActivity : BaseActivity(R.layout.fragment_container_activ
                             PostSignExplainerFragment.newInstance(ConnectPayinType.ADYEN)
                         )
                         .commitAllowingStateLoss()
-                is ConnectPaymentScreenState.Connect -> startAdyenPayment()
+                is ConnectPaymentScreenState.Connect -> startAdyenPayment(marketManager.market, paymentMethods)
```
StylianosGakis, line 63: Since this is accessing a lateinit var, is there any guarantee that the lines further down
```kotlin
adyenConnectPayinViewModel.paymentMethods.observe(this) {
            paymentMethods = it
```
are called before?
I guess we were doing this before as well though right?

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/adyen/payin/AdyenConnectPayinActivity.kt
hugokallstrom: Yeah you are right this is quite messy. The PR does not change any of the logic related to this though so let's assume it works as intended. We wont need this activity much longer anyway, since we are moving to connecting payment in offer screen.

PR #845, opened 2021-12-17
APP-1160: Remove prefix before insurance name

StylianosGakis:
### Checklist

### DIFF
--- a/apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/providerStatus.graphql
+++ b/apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/providerStatus.graphql
@@ -1,0 +1,8 @@
+query providerStatus {
+  externalInsuranceProvider {
+    providerStatusV2 {
+      insuranceProvider
+      insuranceProviderDisplayName
+    }
+  }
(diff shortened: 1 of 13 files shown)


apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/providerStatus.graphql
```diff
@@ -0,0 +1,8 @@
+query providerStatus {
```
hugokallstrom, line 1: Do we need a separate query for this? Is it because we cant combine it with a subscription?

StylianosGakis: The subscription does not actually give us the human-readable insurance name :( Has to do with the fact that insurely itself has a separate endpoint to provide this information. I will give you more context in Slack.

hugokallstrom (approved):

StylianosGakis:
Might wanna look into the ignored tests after this is done. https://github.com/apollographql/apollo-kotlin/issues/3737 It would make this so much simpler

PR #852, opened 2022-01-17
APP-1241: Name parameters and add collection id to "Other" insurer item in order to show info dialog and not crash.

hugokallstrom:
### Checklist


StylianosGakis (approved):
If the crash was due to the null collectionId, here:
https://github.com/HedvigInsurance/android/blob/5cc1cf6c41c2f7fd1b6a0f7b4d47a449f17dc7ab/app/src/main/java/com/hedvig/app/feature/embark/passages/previousinsurer/PreviousInsurerFragment.kt#L81
It seems like if it's null at any point a lot of things don't like that, when should it be null? Can we consider making the collectionId not be null? Or at least not have a default null value to avoid doing this mistake again?

hugokallstrom:
> 

We are using `PreviousInsurer` for `PreviousInsuranceProviderAction`, which uses the same adapter but without Insurely. In that case we are not getting collectionId.

PR #862, opened 2022-01-19
Bump action-android-lint to latest version

StylianosGakis:
Might as well update this too while I'm changing the pr.yml file


StylianosGakis:
v2 of the lint seems to be quite a lot more strict with some stuff that we've been ignoring. Gonna turn it into a draft for now and revisit it later

oscarnylander:
Fixed in #1031

PR #863, opened 2022-01-20
APP-1222: Start free text chat from offer 

hugokallstrom:
### Checklist


app/src/main/java/com/hedvig/app/feature/chat/usecase/TriggerFreeTextChatUseCase.kt
```diff
@@ -1,21 +1,31 @@
...
-
-        if (result is QueryResult.Error) {
-            result.message?.let { e { it } }
-        }
+            .toEither { FreeTextError.NetworkError }
+            .flatMap { data ->
+                data.triggerFreeTextChat
+                    .rightIfNotNull { FreeTextError.CouldNotTrigger }
+                    .map { FreeTextSuccess }
+            }
```
StylianosGakis, line 22: Right I was the one who OKed this, but I was thinking about it and I realized I made a mistake sorry.

This right now would return `FreeTextSuccess` if the value of `data.triggerFreeTextChat` is not null. And `FreeTextError.CouldNotTrigger` only if it is null.

However, the thing is it could be non null, but a `false` in which case we'd also want to return `FreeTextError.CouldNotTrigger`. So we're returning success now in case the backend returns a `false` value.

(As a side note, it's very weird that the return type of the mutation is a *nullable* boolean, I'd guess it should just return false if it failed, not `null`, but anyway)

This should be expressed in a different way, one idea could be:
```kotlin
.flatMap { data ->
    val didTriggerFreeTextChat = data.triggerFreeTextChat ?: false
    if (didTriggerFreeTextChat) {
        FreeTextSuccess.right()
    } else {
        FreeTextError.CouldNotTrigger.left()
    }
}
```
But turns out there is a convenience function on the companion object of Either called `conditionally` and all this could be expressed like this
```suggestion kotlin
            .flatMap { data ->
                val didTriggerFreeTextChat = data.triggerFreeTextChat ?: false
                Either.conditionally(
                    test = didTriggerFreeTextChat,
                    ifFalse = { FreeTextError.CouldNotTrigger },
                    ifTrue = { FreeTextSuccess }
                )
            }
```
Btw this could also be written as
```kotlin
Either.conditionally(didTriggerFreeTextChat, { FreeTextError.CouldNotTrigger }, { FreeTextSuccess })
```
or
```kotlin
Either.conditionally(
    didTriggerFreeTextChat,
    { FreeTextError.CouldNotTrigger },
    { FreeTextSuccess }
)
```
or
```
Either.conditionally(
    didTriggerFreeTextChat,
    ifFalse = { FreeTextError.CouldNotTrigger },
    ifTrue = { FreeTextSuccess }
)
```
I don't know what our preference might be. At first it's a mental overhead to always remember that the failure goes on the left (first argument) and the success on the right side, but as all of us become accustomed to the pattern we might find the named arguments unnecessary.

StylianosGakis (approved):
Nice cleanup too!
Noticed a bug that I suggested the other day 😅 Posted a suggestion for a fix, please address it before merging.

app/src/main/java/com/hedvig/app/feature/chat/usecase/TriggerFreeTextChatUseCase.kt
hugokallstrom: Super, will fix :) Thank you

PR #864, opened 2022-01-20
APP-1233: Embark top app bar options showing sign in

StylianosGakis:
Must've been a case of the loggedIn state not being updated at the right time in some occasions.
Making the value observable should remove any potential problems of not asking for the value at the correct times. Now it should ask for the value on initialization, and then on every sharedPref change, since we're attaching a listener to the SharedPreferences.


hugokallstrom (approved):
LGTM

PR #865, opened 2022-01-21
APP-1228: Home tab error after sign

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutViewModel.kt
```diff
@@ -144,7 +147,15 @@ class CheckoutViewModel(
...
             val event = when (result) {
                 is SignQuotesUseCase.SignQuoteResult.Error -> Event.Error(result.message)
-                SignQuotesUseCase.SignQuoteResult.Success -> Event.CheckoutSuccess
+                SignQuotesUseCase.SignQuoteResult.Success -> {
+                    loginStatusService.isLoggedIn = true
+                    loginStatusService.isViewingOffer = false
+                    // Delay sending success in order for the signed quotes to be added on the member
+                    // Sending success instantly will start HomeFragment, but the member will not have
+                    // updated contracts.
+                    delay(5000)
```
StylianosGakis, line 156: This problem exists because LoggedInActivity queries the data at the start right?
And the member may never pull to refresh to see the new data
Plus that data doesn't come from a subscription to update by itself.

So the tradeoff is that we prefer making this screen delay for an arbitrary amount of time and then almost always have the data available to make a good first impression of LoggedInActivity compared to going there directly and most of the time not showing the new contract.

How big of a deal is it to consider adding a simple polling mechanism in the loggedInScreen, or a simple system of fetching the data at progressively slower intervals (example 0, 1, 2, 4, 10, 20, stop) seconds. And if the same data comes in every time the value can be conflated to not show any change on the screen.

Because 5 seconds do certainly feel sluggish imo, and in the end it may not even be enough if the data isn't ready at that 5 second mark.

StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/offer/ui/checkout/CheckoutViewModel.kt
hugokallstrom: I agree that this solution is not the best - it works though and does not require much additional complexity. I suggest we go with this for now, as a temporary solution to the problem. Next step is involving back-end in order to build a more scalable solution.

StylianosGakis: 👍

PR #866, opened 2022-01-21
APP-1237: Alter safeQuery to handle error responses

StylianosGakis:
When graphQL returns an ok 200 result with partial data but an error in
the errors body, before it was considered a successful response,
breaking some embark flows. This change makes it so that such responses
are interpreted as a failed response, making the logic continue on the
unhappy path instead.

Should probably test as many other embark flows as possible before releasing with this change, I tried going through some of them to see that we were not relying on this behavior and it's looking okay so far, but I'm never sure with making embark changes. Let's see what the tests have to say too.


hugokallstrom (approved):
Superb!

PR #868, opened 2022-01-25
APP-1296: Center the title text if subtitle is non existent

StylianosGakis:
Sometimes the backend may return an empty subtitle. Best we can do
client side is to make the UI look a bit smoother.

https://hedvig.atlassian.net/jira/software/c/projects/APP/boards/6?modal=detail&selectedIssue=APP-1296&quickFilter=51


hugokallstrom (approved):
LGTM!

PR #869, opened 2022-01-25
APP-1311: Add the PAID pill when the claim is paid out

StylianosGakis:
Before, since the monetaryAmount can't be shown yet due to backend
limitations, the pill was hidden altogether. This simply changes the
handling so that the pill does show only with no money amount shown.


hugokallstrom (approved):
LGTM

PR #870, opened 2022-01-25
APP-1308: Fix documents tab for pending members

hugokallstrom:
Removed dependencies on GraphQL classes in view layer. Refactored the state handling and created a new use case which returns the state for contract detail view model.


app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/ContractDetailViewModel.kt
```diff
@@ -1,118 +1,40 @@
...
+                is Either.Right -> ViewState.Success(insurance.value)
             }
-        }
-    }
-
-    private fun createDocumentItems(contract: InsuranceQuery.Contract): List<DocumentItems.Document> {
-        return if (contract.currentAgreement?.asAgreementCore?.status == AgreementStatus.PENDING) {
-            // Do not show anything if status is pending
-            // TODO: Show error state
-            emptyList()
```
hugokallstrom, line 99: Previously, we didnt show anything if the agreement status was pending

app/src/main/java/com/hedvig/app/feature/insurance/ui/detail/documents/DocumentsFragment.kt
```diff
@@ -23,7 +27,17 @@ class DocumentsFragment : Fragment(R.layout.contract_detail_documents_fragment)
...
+                .flowWithLifecycle(lifecycle)
+                .onEach { viewState ->
+                    when (viewState) {
+                        ContractDetailViewModel.ViewState.Error -> {}
+                        ContractDetailViewModel.ViewState.Loading -> {}
+                        is ContractDetailViewModel.ViewState.Success -> {
+                            documentsAdapter.submitList(viewState.state.documentsViewState.documents)
+                        }
+                    }
+                }.launchIn(viewLifecycleScope)
```
StylianosGakis, line 40: I think the previous implementation meant that in case the documentsList was empty, the empty list was submitted to the documentsAdapter.
With the new implementation, if somehow we first get a success with a nonEmpty list of documents it will be submitted, and later a Loading/Error state will not change those already existing documents.
I am missing some context here, but is this possible? If it's not, this is okay as it is. If this is possible, are we okay with this interaction? If yes, then again it's no problem. If however we don't want to do that, we may want to add `documentsAdapter.submitList(emptyList())` in the Error/Loading states

StylianosGakis (approved):
Suuuper nice cleanup 🥳

PR #871, opened 2022-01-25
APP-1309: String res for pending title

hugokallstrom:


hugokallstrom:
> I have no idea how lint could've failed with only two string res changes 😅

Think it crashed: 
* What went wrong:
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

PR #872, opened 2022-01-26
APP-1303: Home screen spacing issues

StylianosGakis:
Some combination of items shown in the home screen was not cooperating and the spacing dependencies between all the different combinations make it hard to add it to a specific item.
This simply adds a Space item to serve the purposes of solving this combination which didn't look nice

Accessible through engineering mode: Home > Terminated with ongoing past claims

| before | after |
| :---: | :---: |
| - |  |


StylianosGakis:
Touches some of the same parts as you did to remove the feature flags for the claim status cards. It's probably best if this is merged into develop after the release branch gets merged back to develop so that I can take care of the merge conflicts instead of you having to deal with them while also finishing the release

hedvigoscar (approved):

hedvigoscar:
CI is bright red though, might want to see to that 😬

PR #876, opened 2022-01-28
APP-1334: Show notification on referral badge if campaign is active 

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/loggedin/service/TabNotificationService.kt
```diff
@@ -2,30 +2,43 @@ package com.hedvig.app.feature.loggedin.service
...
+            referralsNotificationBadgeService.shouldShowNotification()
+        ) { shouldShowCrossSellNotification: Boolean,
+            shouldShowReferralNotification: Boolean ->
+
+            buildSet {
+                if (shouldShowCrossSellNotification) add(LoggedInTabs.INSURANCE)
+                if (shouldShowReferralNotification) add(LoggedInTabs.REFERRALS)
             }
+        }
     }
```
StylianosGakis, line 27: Looking super nice in general. There's only only thing that I think is good to change which happens due to how I wrote the CrossSellNotificationBadgeService::getUnseenCrossSells.
It is a suspending function, that returns a flow, which is kind of an unnecessary thing, as the flow is supposed to be cold and it can call the suspending functions it needs to after it is actually being collected somewhere.
Not only that, but that change from me now makes all these functions that really don't need to be suspending, like the one I am highlighting here also be suspending.
I am putting up a PR to propose a solution to this which I made real quick to make this conversation easier. I am up to discussion if there's something else I am missing which would make this a bad idea.
PR mentioned: https://github.com/HedvigInsurance/android/pull/877

hugokallstrom: I see, that PR looks good :) Thanks

StylianosGakis (approved):
Looking goood

PR #878, opened 2022-01-28
APP-1340: Add string resources for referral campaign

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/referrals/ui/tab/ReferralsViewModel.kt
```diff
@@ -79,22 +74,10 @@ class ReferralsViewModelImpl(
         }
     }
 
-    private suspend fun createTopBar(): ViewState.Success.TopBarState? {
+    private suspend fun shouldShowCampaignBar(): Boolean {
+        val remoteConfig = RemoteConfig()
```
StylianosGakis, line 78: Can we pass the RemoteConfig as a parameter to the ViewModel so that we don't have to do this every time?

app/src/main/res/layout/fragment_referrals.xml
```diff
@@ -16,7 +16,7 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:clipToPadding="false"
-            android:paddingTop="?actionBarSize"
+            android:paddingTop="100dp"
```
StylianosGakis, line 19: What was actionBarSize doing before? Is it just a design change or did the previous value not work for some reason?

StylianosGakis (approved):
lgtm, just had two questions

app/src/main/res/layout/fragment_referrals.xml
hugokallstrom: Had to adjust the padding here, the campaign bar was somehow laid out below the toolbar. Very strange, looked into different solutions and this worked best.

PR #881, opened 2022-02-01
APP-1179: Remove claims chat

hugokallstrom:
Did some cleaning here. Removed unused code in a couple of places.


hedvigoscar (approved):
LGTM

app/src/main/java/com/hedvig/app/feature/chat/data/ChatRepository.kt
```diff
@@ -176,8 +179,26 @@ class ChatRepository(
         apolloClient.mutate(EditLastResponseMutation()).await()
 
     suspend fun triggerFreeTextChat() =
```
StylianosGakis, line 181: Hey with functions with return types, which are also public (not private), since the types are more interesting to see than the implementation if all you need to know is the function signature, I think it's super important to explicitly write the return type.
I like it for the scenario of: I am using/want to use this function and I expect it does what it says and it has no bugs so I am not interested in going through the implementation and understanding it but I want to know what it returns. Without a return type, this means I need to dig into it and locate the possible return types (could be many points if there are early returns in the function. With a return type, I can read the function signature which is all in 1-3 lines and go on with my day.
Sorry for being *that* guy but I really really like explicit return types for public functions which are used from other places 😅

StylianosGakis (approved):
Looking great! Got one thing that I would like to discuss to make our lives a bit easier 😊

app/src/main/java/com/hedvig/app/feature/chat/data/ChatRepository.kt
hedvigoscar: I'm not strictly speaking against doing this, but as for solving your particular issue, couldn't you just use `F1` on the function to see the return type? 🤔

hedvigoscar: Example: 
image

hugokallstrom: Thats a nice shortcut, but I agree with you Stelios that it increases readability a lot if you write the type explicitly. It can be ok if the function name itself give away the type, like "getTitleText". Ill fix!

StylianosGakis: Nice shortcut indeed! I'll just note a couple more points why I feel so strongly about this though 😅
- You'd have to press the shortcut every time
- You'd need to hover my mouse over it and click an extra key instead of just existing in that file.
- If I am changing the implementation of the function which in turn automatically changes the inferred return type, I am not warned about it instantly, but only when I compile for example and the call-site complains. (Often this also means that I've done something wrong if I am changing the return type by changing the implementation. I've have had many examples like this)

Sorry for being so pushy about it, but as time has passed I've realized that the amount of times I've regretted putting an explicit return type on public functions is 0. Can't say the same about doing the opposite unfortunately.

PR #883, opened 2022-02-02
APP-1342: Use remote feature flag provider for referral campaign

hugokallstrom:


app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    private var seCampaignVisible: Boolean = false
+    private var noCampaignVisible: Boolean = false
+    private var dkCampaignVisible: Boolean = false
+
+    init {
+        CoroutineScope(Dispatchers.IO).launch {
+            seCampaignVisible = remoteConfig.fetch().seCampaignVisible
+            noCampaignVisible = remoteConfig.fetch().noCampaignVisible
+            dkCampaignVisible = remoteConfig.fetch().dkCampaignVisible
```
StylianosGakis, line 25: Doesn't fetch() return an object that contains all the variables by itself? Here we're doing 3 fetches but all three will make a network request as I understand right? If it can be 1 request it'd be better to be like that. If for some reason the have to be separate requests it'd be a good idea to wrap them all in `async {}` and put them in a list and do `.awaitAll()` on them so that they run in parallel at least instead of sequentially.

app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    override fun isFeatureEnabled(feature: Feature) = when (feature) {
```
StylianosGakis, line 29: Since xxCampaignVisible are asynchronously populated, I guess we just have to assume that it's populated by the time we as for isFeatureEnabled right?

otherwise we can make this suspending too, and have some sort of guarantee that it's waiting for the init{} function to finish first.

app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    override val priority = REMOTE_PRIORITY
+
+    private var seCampaignVisible: Boolean = false
+    private var noCampaignVisible: Boolean = false
+    private var dkCampaignVisible: Boolean = false
+
+    init {
+        CoroutineScope(Dispatchers.IO).launch {
+            seCampaignVisible = remoteConfig.fetch().seCampaignVisible
```
hedvigoscar, line 23: Does `RemoteConfig::fetch` carry a penalty on a per-call basis? If yes, this could be a viable option:
```suggestion
            val remoteConfigData = remoteConfig.fetch()
            seCampaignVisible = remoteConfigData.seCampaignVisible
```

If not, disregard this comment, unless you feel that it improves readability.

hedvigoscar (approved):
lgtm

StylianosGakis (approved):
Looking nice, just a couple of questions for RemoteFeatureFlagProvider which we may want to look into?

app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
hugokallstrom: Good catch, will fix

app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
hugokallstrom: Yeah will definitely want to only do one call here 😅 My mistake!

app/src/main/java/com/hedvig/app/util/featureflags/RemoteFeatureFlagProvider.kt
hugokallstrom: Yeah we assume the network call har returned. We should probably do this behind a splash screen since its likely that we will depend on some features being fetched when the app starts. But doing this for now should be fine.

PR #884, opened 2022-02-03
APP-1216: Device id header

hugokallstrom:
Do we want to replace the member id tracking with this id as well? Maybe @tarikstafford or @sampettersson?

### DIFF
--- a/app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
+++ b/app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
@@ -1,0 +1,44 @@
+package com.hedvig.app.authenticate
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+import java.util.UUID
+
+interface DeviceIdStore {
+    fun observeDeviceId(): Flow<String>
+}
+
+class DeviceIdStoreImpl(
+    private val dataStore: DataStore<Preferences>
+) : DeviceIdStore {
+
+    private val key = stringPreferencesKey("hedvig-device-id")
+
+    init {
+        CoroutineScope(Dispatchers.IO).launch {
+            val deviceId = observeDeviceId().firstOrNull()
+            if (deviceId == null || deviceId == "") {
+                generateDeviceId()
+            }
+        }
+    }
+
+    override fun observeDeviceId(): Flow<String> {
+        return dataStore.data.map { it[key] ?: "" }
+    }
+
+    private suspend fun generateDeviceId() {
+        val deviceId = UUID.randomUUID().toString()
+        dataStore.edit {
+            it[key] = deviceId
+        }
+    }
+}
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -24,6 +24,8 @@ import com.google.firebase.analytics.ktx.analytics
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.messaging.FirebaseMessaging
 import com.hedvig.app.authenticate.AuthenticationTokenService
+import com.hedvig.app.authenticate.DeviceIdStore
+import com.hedvig.app.authenticate.DeviceIdStoreImpl
 import com.hedvig.app.authenticate.LoginStatusService
 import com.hedvig.app.authenticate.LogoutUseCase
 import com.hedvig.app.authenticate.SharedPreferencesAuthenticationTokenService
@@ -213,6 +215,7 @@ import com.hedvig.app.terminated.TerminatedTracker
 import com.hedvig.app.util.LocaleManager
 import com.hedvig.app.util.apollo.ApolloTimberLogger
 import com.hedvig.app.util.apollo.CacheManager
+import com.hedvig.app.util.apollo.DeviceIdInterceptor
 import com.hedvig.app.util.apollo.SunsettingInterceptor
 import com.hedvig.app.util.featureflags.FeatureManager
 import com.mixpanel.android.mpmetrics.MixpanelAPI
@@ -289,6 +292,7 @@ val applicationModule = module {
                         .build()
                 )
             }
+            .addInterceptor(DeviceIdInterceptor(get()))
         if (isDebug()) {
             val logger = HttpLoggingInterceptor { message ->
                 if (message.contains("Content-Disposition")) {
@@ -726,4 +730,8 @@ val dataStoreModule = module {
             }
         )
     }
+}
+
+val deviceIdStoreModule = module {
+    single<DeviceIdStore> { DeviceIdStoreImpl(get()) }
 }
--- a/app/src/main/java/com/hedvig/app/feature/di/KoinInitializer.kt
+++ b/app/src/main/java/com/hedvig/app/feature/di/KoinInitializer.kt
@@ -15,6 +15,7 @@ import com.hedvig.app.clockModule
 import com.hedvig.app.coilModule
 import com.hedvig.app.connectPaymentModule
 import com.hedvig.app.dataStoreModule
+import com.hedvig.app.deviceIdStoreModule
 import com.hedvig.app.embarkModule
 import com.hedvig.app.embarkTrackerModule
 import com.hedvig.app.externalInsuranceModule
@@ -107,6 +108,7 @@ class KoinInitializer : Initializer<KoinApplication> {
                 externalInsuranceModule,
                 insurelyAuthModule,
                 campaignModule,
+                deviceIdStoreModule,
             )
         )
     }
--- a/app/src/main/java/com/hedvig/app/util/apollo/DeviceIdInterceptor.kt
+++ b/app/src/main/java/com/hedvig/app/util/apollo/DeviceIdInterceptor.kt
@@ -1,0 +1,41 @@
+package com.hedvig.app.util.apollo
+
+import com.hedvig.app.authenticate.DeviceIdStore
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+import okhttp3.Interceptor
+import okhttp3.Request
+import okhttp3.Response
+
+private const val HEADER_NAME = "hedvig-device-id"
+
+class DeviceIdInterceptor(
+    private val deviceIdStore: DeviceIdStore
+) : Interceptor {
+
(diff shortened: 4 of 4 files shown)


app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
```diff
@@ -0,0 +1,44 @@
...
+class DeviceIdStoreImpl(
+    private val dataStore: DataStore<Preferences>
+) : DeviceIdStore {
+
+    private val key = stringPreferencesKey("hedvig-device-id")
+
+    init {
+        CoroutineScope(Dispatchers.IO).launch {
+            val deviceId = observeDeviceId().firstOrNull()
+            if (deviceId == null || deviceId == "") {
```
StylianosGakis, line 28: stdlib has a nice isNullOrBlank()/isNullOrEmpty() extension on `String?`

StylianosGakis (approved):
LGTM

Apollo 3.0 intercept function is suspending so this will make it even easier and remove the need for a mutable nullable deviceId inside the interceptor but call the datastore directly

sampettersson:
Yeah, tracking id should be sent to `hAnalytics`.

app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.datastore.preferences.core.stringPreferencesKey
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+import java.util.UUID
+
+interface DeviceIdStore {
```
hedvigoscar, line 15: Is this interface/implementation pair necessary? Are there more than one implementation, or do we expect one to exist?

hedvigoscar (commented):
One question, looks good in all other regards 🙂

app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
StylianosGakis: Hmm about this, I've thought about it for some time and I personally found myself to lean towards enjoying having such interfaces sometimes.
Aside from allowing a bit of an easier path to testing without the use of mocks (which I sometimes still find kinda confused by when reading the tests), I mostly like the fact that I can see that if a class is injected with such an interface, I can look into its contract and see the public functions that I can use and I don't necessarily have to know more than that (maybe I shouldn't, to keep to its contract that other classes may be interested in as well).

Now I can also see the negatives of this, like having to do a double navigation if I am in fact interested in the implementation details for some reason. But I also feel like the IDE has our backs on this by allowing us to do Cmd+Option+B which finds the interface usages instead of the interface itself so this is kind of negated.

Now with all that said, I haven't been doing this in the codebase myself so far (I've actively tried not to do it at least, maybe sometimes it slipped), to avoid doing something that we haven't discussed before and all agreed together, but I am only saying all this to say that I am not against it 😅

Do you have any other points for why we've been avoiding it in general Oscar? I'd love to hear them.

hedvigoscar: There's been a good amount written on this subject - the reason I'm against it is because it serves no actual purpose. In Java, this pattern was quite common, as creating mock objects in tests that was accepted by the compiler was a bit more challenging. Over time, this became part of 'just how code is written', in a sort of cargo culty-manner.

Kotlin does not suffer from the Java-flaw (mockk can create a mock-object of more or less any class, which will be readily accepted, simply by invoking `val mock = mocck<ClassWeWantToMock>()`.

Hence, we lose the purpose of having this level of abstraction. We are now only stuck with the costs (more code, indirection that must be kept in working memory when analyzing the code).

Others have written more on the subject than I have, probably more coherently:

https://martinfowler.com/bliki/InterfaceImplementationPair.html

https://mortoray.com/2014/08/01/the-false-abstraction-antipattern/

StylianosGakis: Completely fair points, just one thought. This is all based on the fact that we can mock the classes in an easy manner therefore we gain nothing from this pair. But this is implying that we are using mocks for tests. In case we did for some tests not use mocks, but we used some test doubles, that would qualify for a valid case for such an interface right?

I was reading this the other day https://martinfowler.com/articles/mocksArentStubs.html and along with how confusing mocks felt like the first time I got to use them compared to a test double, I was always leaning towards the path of making interfaces for things and then making test implementations for them, therefore making those interfaces useful.

But as I mentioned before, this is all on an idea level for me, I can't say that I've tried and tested both ways extensively and I am a proponent of one of them. In this particular code snippet, I wouldn't object to removing the interface until such a usage existed (where as the first article says as well, your IDE can just perform the 'extract interface' function for free for you)

hugokallstrom: I've been thinking about this a bit, and I don't think there is a right or wrong here tbh. The purpose for `DeviceIdStore` here is to hide implementation details, so looking at the function declaration is all you need to do in order to understand what the implementation of this interface is doing. The naming here is essential for that - its a store for a device id, with one function which lets you observe that device id. You don't know where or how that device id is stored, it can be from network or locally - you don't care! 

I also think interfaces are good at showing what the implementations should be used for - if using a class its very easy to just make a function public, for example `generateDeviceId()`. And "pollute" the public interface with functions which should only be private. 

Having an interface here also tells me that there are implementation details that probably will change, right now we are using `DataStore` but this could very well change to be fetched remotely. If/when we are doing that change, its easy to just add another implementation of the interface. This goes against [yigni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), but I still think its good to signal to other developers that this is an implementation that could (and probably will) change. Another example for this is the `TrackerSink` interface. It has multiple implementations - but even if it didn't I would still think an interface would be appropriate.

Lastly - using test doubles for smaller interfaces instead of mocks can be easier to both write and read. See `OtpInputViewModelTest` for an example.

TL;DR 
- Interfaces can tell you something about the implementations (even if there is only one implementation) - it's probably likely to change, be careful with adding more functions etc. 
- Test doubles are easy to implement with smaller interfaces.

app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun observeDeviceId(): Flow<String>
+}
+
+class DeviceIdDataStore(
```
hugokallstrom, line 19: Change the name here to better reflect implementation

app/src/main/java/com/hedvig/app/authenticate/DeviceIdStore.kt
hedvigoscar: >Completely fair points, just one thought. This is all based on the fact that we can mock the classes in an easy manner therefore we gain nothing from this pair. But this is implying that we are using mocks for tests. **In case we did for some tests not use mocks, but we used some test doubles**, that would qualify for a valid case for such an interface right?

I think you could probably use `mockk` in order to produce any of the levels of test doubles mentioned in the article - `mocck<T>` produces a class that implements the same implicit interface as `T`.

>I've been thinking about this a bit, and I don't think there is a right or wrong here tbh. The purpose for DeviceIdStore here is to hide implementation details, so looking at the function declaration is all you need to do in order to understand what the implementation of this interface is doing. The naming here is essential for that - its a store for a device id, with one function which lets you observe that device id. You don't know where or how that device id is stored, it can be from network or locally - you don't care!

>I also think interfaces are good at showing what the implementations should be used for - if using a class its very easy to just make a function public, for example generateDeviceId(). And "pollute" the public interface with functions which should only be private.

Well, the API surface stays the same, right? Any given caller would still see the same thing, making the implementation details de-facto hidden.

>Having an interface here also tells me that there are implementation details that probably will change, right now we are using DataStore but this could very well change to be fetched remotely. If/when we are doing that change, its easy to just add another implementation of the interface. This goes against [yigni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), but I still think its good to signal to other developers that this is an implementation that could (and probably will) change.

I'd argue that `observeDeviceId(): Flow<String>` already communicates this well enough.

>Another example for this is the TrackerSink interface. It has multiple implementations - but even if it didn't I would still think an interface would be appropriate.

`TrackerSink` does have multiple different implementations, however. At that point, an interface becomes the right tool for the job, I think.

>Lastly - using test doubles for smaller interfaces instead of mocks can be easier to both write and read. See OtpInputViewModelTest for an example.

I think we could re-express that test more briefly and achieve the same functionality using `mockk`, actually:

```kotlin
    private var authToken: String? = "testToken"

    private var otpResult: OtpResult = OtpResult.Success("authtest")
    private var resendOtpResult: ResendOtpResult = ResendOtpResult.Success("authtest")
    private val viewModel = OtpInputViewModel(
        otpId = "testId",
        credential = "test@email.com",
        authenticationTokenService = object : AuthenticationTokenService {
            override var authenticationToken: String?
                get() = authToken
                set(value) {
                    authToken = value
                }
        },
        sendOtpCodeUseCase = object : SendOtpCodeUseCase {
            override suspend fun invoke(otpId: String, otpCode: String): OtpResult {
                delay(100)
                return otpResult
            }
        },
        reSendOtpCodeUseCase = object : ReSendOtpCodeUseCase {
            override suspend fun invoke(credential: String): ResendOtpResult {
                delay(100)
                return resendOtpResult
            }
        }
    )
```

vs

```kotlin
private var authToken: String? = "testToken"

private var otpResult: OtpResult = OtpResult.Success("authtest")
private var resendOtpResult: ResendOtpResult = ResendOtpResult.Success("authtest")
val authenticationTokenService = mockk<AuthenticationTokenService>
init { every { authenticationTokenService.authToken } answers { authToken } }
val sendOtpCodeUseCase = mockk<SendOtpCodeUseCase>
init {
	coEvery { sendOtpCodeUseCase.invoke(any()) } coAnswers {
		delay(100)
		otpResult
	}
}
val reSendOtpCodeUseCase = mockk<ReSendOtpCodeUseCase>
init {
	coEvery { reSendOtpCodeUseCase.invoke(any()) coAnswers {
		delay(100)
		resendOtpResult
	}
}

private val viewModel = OtpInputViewModel(
	otpId = "testId",
	credential = "test@email.com",
	authenticationTokenService = authenticationTokenService,
	sendOtpCodeUseCase = sendOtpCodeUseCase,
	reSendOtpCodeUseCase = reSendOtpCodeUseCase,
)
```

note that I haven't compiled this code, but it likely runs with some minor tweaks on top.

>TL;DR

Don't worry about providing TL;DRs, I will always make sure to read what you write 😄

hugokallstrom: Lets be pragmatic with this, I certainly don't think we should create interfaces for every class - but there are situations where it can be beneficial. 

Just the fact the we have an interface in this case tells me as a developer that I should expect this code to change and that there are dependencies that I should not care about. 

This is the biggest reason why (in this case) an interface makes sense, not because of how the language works or because I gain something from a technical perspective. 

I found this comment on SO, which I think summarises my thoughts quite well: 
"The decision to create an interface should not be based on the number of implementing classes, but rather on the number of different ways that object is used."

PR #885, opened 2022-02-07
APP-1359: Show error messages in embark

hugokallstrom:
We can now show error messages like this: 

image

Before, since the passage did not have an action, we showed the "update app" passage.

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/passages/noaction/NoActionFragment.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/passages/noaction/NoActionFragment.kt
@@ -1,0 +1,44 @@
+package com.hedvig.app.feature.embark.passages.noaction
+
+import android.os.Bundle
+import android.view.View
+import androidx.core.os.bundleOf
+import androidx.core.view.doOnNextLayout
+import androidx.fragment.app.Fragment
+import com.hedvig.app.R
+import com.hedvig.app.databinding.FragmentEmbarkNoActionBinding
+import com.hedvig.app.feature.embark.passages.MessageAdapter
+import com.zhuinden.fragmentviewbindingdelegatekt.viewBinding
+import e
+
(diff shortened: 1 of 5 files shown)


StylianosGakis (approved):

app/src/main/java/com/hedvig/app/feature/embark/passages/noaction/NoActionFragment.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.core.os.bundleOf
+import androidx.core.view.doOnNextLayout
+import androidx.fragment.app.Fragment
+import com.hedvig.app.R
+import com.hedvig.app.databinding.FragmentEmbarkNoActionBinding
+import com.hedvig.app.feature.embark.passages.MessageAdapter
+import com.zhuinden.fragmentviewbindingdelegatekt.viewBinding
+import e
+
+class NoActionFragment : Fragment(R.layout.fragment_embark_no_action) {
```
hedvigoscar, line 14: I assume you selected a View-based solution on account of being able to re-use `MessageAdapter`?

hedvigoscar (approved):
lgtm!

app/src/main/java/com/hedvig/app/feature/embark/passages/noaction/NoActionFragment.kt
hugokallstrom: Yeah, and because its specified as a passage in angel with messages, like:
```
<Message>
  {NO_SIGNUP_ERROR_SIZE_MESSAGE_1}
</Message>

<Message>
  {NO_SIGNUP_ERROR_SIZE_MESSAGE_2}
</Message>
```

I think it makes sense to render it as other passages on the client, just without an action

hedvigoscar: Agreed.

PR #886, opened 2022-02-07
APP-1255: Handle constant variable

hugokallstrom:


hedvigoscar (approved):
Perhaps a bit silly, but maybe we should add a test for this functionality?

Implementation looks good to me.

hugokallstrom:
> Perhaps a bit silly, but maybe we should add a test for this functionality?
> 
> Implementation looks good to me.

Yeah thats a good idea tbh. I'll do that.

hugokallstrom:
Tests added together with correct implementation :)

PR #888, opened 2022-02-08
APP-1291: Denmark address autocomplete

StylianosGakis:
The embark fragment (EmbarkAddressAutoCompleteFragment.kt) contains the button for launching an activity and registering for a result (using the activity result APIs), then given the result, either continues with a null result making embark go in the unhappy path of manually adding the address, or makes the button clickable and continues on the happy path.

AddressAutoCompleteActivity is the activity which does the rest, and sets the result on completion.

This requires some further improvements but I am at least trying to salvage an already humongous PR by making it the review a tiny bit easier and then the fix PRs will be easier to follow.
Current known problems that will be addressed in separate PRs:

- The cursor doesn't move when a new input is auto-filled and it stays at the position it was at. Also true when opening the screen again with a pre-filled address (for example when having navigated back to this screen).
- When selecting an address, there is not "smartness" about it to auto-select and exit the screen when the option that was selected is the only valid option and no further details should be selected. Now the address is then isolated and presented as a single option and the user needs to click it again to select it. (Note: This also in some less frequent occasions make some addresses unclickable, as the API still responds with more results for the same query string, never narrowing the list down to 1 item therefore never making it selectable. This is a serious bug and we wouldn't be able to ship with it)

https://user-images.githubusercontent.com/44558292/152984462-542bc2e9-7c15-48a2-9078-a64c6cdf4287.mp4

With the bugs fixed:

https://user-images.githubusercontent.com/44558292/153898054-0ce7188c-54d6-48fe-9a66-8861dc6d5968.mp4


app/src/main/java/com/hedvig/app/feature/addressautocompletion/activityresult/FetchDanishAddressAutoCompleteContractHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.activity.result.ActivityResultRegistry
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import com.hedvig.app.feature.addressautocompletion.model.DanishAddress
+
+/**
+ * Must remember to register as a lifecycle observer in the `onCreate` method of the activity.
+ * More information: https://developer.android.com/training/basics/intents/result#separate
+ */
+class FetchDanishAddressAutoCompleteContractHandler(
```
StylianosGakis, line 13: This along with the FetchDanishAddressContract below contain all the Activity result API classes to make it work

app/src/main/java/com/hedvig/app/feature/addressautocompletion/activityresult/FetchDanishAddressContract.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.hedvig.app.feature.addressautocompletion.ui.AddressAutoCompleteActivity
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Input: DanishAddress? The initial address that will be used as the initial text for autocompletion.
+ * Output: FetchDanishAddressContractResult The result potentially containing the autocomplete address result.
+ */
+class FetchDanishAddressContract : ActivityResultContract<DanishAddress?, FetchDanishAddressContractResult>() {
+    override fun createIntent(context: Context, input: DanishAddress?): Intent {
+        return AddressAutoCompleteActivity.newInstance(context, input)
```
StylianosGakis, line 17: The contract itself knows how to start the activity it needs to start for its result

app/src/main/java/com/hedvig/app/feature/addressautocompletion/data/GetDanishAddressAutoCompletionUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+                AddressAutoCompleteResults(queryData.autoCompleteAddress.map(DanishAddress::fromDto))
+            }
+    }
+}
+
+data class AddressAutoCompleteResults(
+    val resultList: List<DanishAddress>,
+)
+
+private fun DanishAddressInput.toAddressAutocompleteType(): AddressAutocompleteType {
```
StylianosGakis, line 36: The API needs a different type depending on which stage of the querying we are currently in. This method maps the current state of the Address we're looking for, to the type that needs to be queried

app/src/main/java/com/hedvig/app/feature/addressautocompletion/model/DanishAddress.kt
```diff
@@ -0,0 +1,88 @@
...
+            appendIfNotNull(", ")
+            appendIfNotNull(floor)
+            appendIfNotNull(". ")
+            appendIfNotNull(apartment)
+            appendIfNotNull(", ")
+            appendIfNotNull(postalCode)
+            appendIfNotNull(" ")
+            appendIfNotNull(city)
+        }
+    }
```
StylianosGakis, line 67: The API accepts some weird inputs like having the fields missing but the commas/dots existing, so this looks weird but it is how it works.
An input like "Wadstedsvej 1, . ,  Tisvildeleje" for example is perfectly valid

app/src/main/java/com/hedvig/app/feature/addressautocompletion/model/DanishAddressInput.kt
```diff
@@ -0,0 +1,42 @@
+package com.hedvig.app.feature.addressautocompletion.model
+
+class DanishAddressInput private constructor(
+    val rawText: String,
+    val selectedDanishAddress: DanishAddress? = null,
+) {
```
StylianosGakis, line 6: A class to make it easier to hold the input data. Since we need the member to be able to free-form input stuff, but also keep information about when they may have clicked on some address which is stored in the `selectedDanishAddress`

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    private fun finishWithResult(result: FetchDanishAddressContractResult) {
+        setResult(
+            FetchDanishAddressContract.RESULT_CODE,
+            Intent().putExtra(FetchDanishAddressContract.FETCH_DANISH_ADDRESS_CONTRACT_RESULT_KEY, result)
+        )
+        finish()
+    }
```
StylianosGakis, line 56: Result to fulfill the activity result contract

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+            selectAddress = selectAddress,
+            cantFindAddress = cantFindAddress,
+            modifier = Modifier.padding(paddingValues),
+        )
+    }
+}
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+private fun AddressInput(
```
StylianosGakis, line 92: The input area, under the TopAppBar but above the result list

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+        }
+        .onStart { emit(emptyList()) }
+
+    private val checkSelectionChannel: Channel<Pair<DanishAddress, List<DanishAddress>>> = Channel(Channel.UNLIMITED)
+    val resultEvent = checkSelectionChannel.receiveAsFlow()
+        .mapLatest { (selectedAddress, addressList) ->
+            val address = getFinalResultOrNull(selectedAddress, addressList) ?: return@mapLatest null
+            AddressAutoCompleteEvent.Selection(address)
+        }
+        .filterNotNull()
```
StylianosGakis, line 51: This triggers on every address selection to potentially send the event of being done if that's the case.
Currently this does have a fault where it requires an extra click to select the address again, but I will address this in another PR.

app/src/main/java/com/hedvig/app/feature/embark/passages/addressautocomplete/EmbarkAddressAutoCompleteFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+                    is FetchDanishAddressContractResult.Selected -> {
+                        viewModel.updateAddressSelected(result.address)
+                    }
+                    else -> {
+                        // no-op
+                    }
+                }
+            }
+        )
+        lifecycle.addObserver(fetchDanishAddressAutoCompleteContractHandler)
```
StylianosGakis, line 73: This is all that's required on the call site to setup the activity result, along with starting the process in line 101:103.

app/src/main/java/com/hedvig/app/feature/settings/Market.kt
```diff
@@ -82,7 +83,11 @@ enum class Market {
             context.startActivity(ChoosePlanActivity.newInstance(context))
         }
         DK -> {
-            context.startActivity(WebOnboardingActivity.newInstance(context))
+            if (isDebug()) { // TODO NOT COMMIT THIS use for testing the branch
+                context.startActivity(ChoosePlanActivity.newInstance(context))
+            } else {
+                context.startActivity(WebOnboardingActivity.newInstance(context))
+            }
```
StylianosGakis, line 90: This was a fast way I did to be able to test this on device, I guess we can make this a feature flag instead.

app/src/main/java/com/hedvig/app/ui/compose/composables/appbar/TopAppBar.kt
```diff
@@ -0,0 +1,92 @@
+package com.hedvig.app.ui.compose.composables.appbar
```
StylianosGakis, line 1: Nothing changed in this file. Just made some overloads to easily have a top app bar with a close button and a back button. They function the same

build.gradle.kts
```diff
@@ -11,6 +11,12 @@ subprojects {
     plugins.withType<com.android.build.gradle.BasePlugin> {
         project.apply(plugin = "org.gradle.android.cache-fix")
     }
+    tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+        kotlinOptions.freeCompilerArgs += listOf(
+            "-Xopt-in=kotlin.RequiresOptIn",
+            "-Xopt-in=kotlinx.coroutines.ExperimentalCoroutinesApi",
+        )
+    }
```
StylianosGakis, line 19: It was annoying to keep seeing these errors everywhere, but it seems like the IDE still shows these regardless of if we specify this on the root build kts file 🤷‍♂️

app/src/main/res/layout/fragment_embark_address_auto_complete_action.xml
```diff
@@ -15,54 +16,29 @@
...
-    <com.google.android.material.card.MaterialCardView
-        android:id="@+id/input_layout"
+    <androidx.compose.ui.platform.ComposeView
+        android:id="@+id/input_card"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginStart="@dimen/base_margin_double"
         android:layout_marginEnd="@dimen/base_margin_double"
         android:layout_marginBottom="@dimen/base_margin_double"
+        android:transitionGroup="true"
```
StylianosGakis, line 36: This was a very tricky thing. When under a material transition, ComposeView items just make the app crash as reported here as well https://issuetracker.google.com/issues/206947893 This fixes it.

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+    }
+    val closeKeyboard: () -> Unit = {
+        keyboardController?.hide()
+        focusRequester.freeFocus()
+    }
+    Scaffold(
+        topBar = {
+            Column {
+                CenterAlignedTopAppBar(
+                    title = "Address",
```
hugokallstrom, line 77: Use string res

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    val viewState: StateFlow<AddressAutoCompleteViewState> =
+        combine(currentInput, queryResults) { input, results ->
```
hugokallstrom, line 54: Are we handling errors here from queryResults?

hugokallstrom (commented):
Very, very good work Stelios!

app/src/main/java/com/hedvig/app/feature/embark/passages/addressautocomplete/EmbarkAddressAutoCompleteFragment.kt
hugokallstrom: So good

app/src/main/java/com/hedvig/app/feature/settings/Market.kt
hugokallstrom: Remember to remove this before we release :)

app/src/main/java/com/hedvig/app/feature/embark/passages/addressautocomplete/EmbarkAddressAutoCompleteFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+            FetchDanishAddressContractResult.CantFind -> {
+                lifecycleScope.launch {
+                    saveAndAnimate(null)
+                    embarkViewModel.submitAction(data.link)
+                }
+            }
+            is FetchDanishAddressContractResult.Selected -> {
+                viewModel.updateAddressSelected(result.address)
+            }
+            else -> {
```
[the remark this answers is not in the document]
StylianosGakis: Yes this is something I should stop doing, I agree that it's better to not use else to make our lives easier in the future

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteViewModel.kt
StylianosGakis: What we're doing is that we're simply showing no results (since we get none) and the "I can't find my address" red option shows.
Thing is, there is no design on how to handle when there is no network access in this screen, and this would be the only case where there would be an error from "queryResults". I think this is reasonable if the other clients are doing the same. In fact the web is doing the same, but not showing the "Can't find my address" option at all.

hugokallstrom: I see, that's fine. We should generally be better at handling errors though - no internet connection is not the only thing that can go wrong here. The API can change, headers can break etc. Showing the member what has gone wrong is almost always better UX than not showing anything at all IMO!

app/src/main/java/com/hedvig/app/feature/settings/Market.kt
StylianosGakis: Is this OK https://github.com/HedvigInsurance/android/pull/888/commits/0b017c6f03ec6aa8e95a683a954b1fe5acf0cfd0 for now?
Until when the QA is done

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteViewModel.kt
StylianosGakis: Yeah, do you want to bring this up sometime? Whenever there's places like this where error-handling isn't obvious and I can't come up with something on the fly myself, I've learned to just not do it because otherwise I spend way too much time and it ends up looking hella bad (I am horrible at designing stuff 😂).

So if going forward, if we want to approach it in a way where we bring it up early and make something about it with our designers before we actually implement it then we'd avoid situations like this. It's also been released like this in iOS for ~8 months now so I guess it's been accepted to be that way.

PR #889, opened 2022-02-08
APP-1341: Track using hAnalytics

hedvigoscar:
Each commit is more or less one track - for easier reviewing, you may review
commit-by-commit.

Omitted tracks: `app_shutdown` - we are not able to track this event on android,
as far as I know.


app/src/main/java/com/hedvig/app/feature/home/ui/HomeAdapter.kt
```diff
@@ -107,6 +114,9 @@ class HomeAdapter(
         if (holder is ViewHolder.ClaimStatus) {
             holder.composeView.disposeComposition()
         }
+        if (holder is ViewHolder.InfoCard) {
+            holder.composeView.disposeComposition()
+        }
```
StylianosGakis, line 119: In other adapters like the `OfferAdapter` we're using a function that looks like this:
```kotlin
override fun onViewRecycled(holder: ViewHolder) {
    val itemView = holder.itemView
    if (itemView is ComposeView) {
        itemView.disposeComposition()
    }
}
```

Is it something that could be used universally in such situations maybe? To avoid having to manually add all cases when we add more composable ViewHolders

app/src/main/java/com/hedvig/app/feature/home/ui/HomeAdapter.kt
```diff
@@ -314,25 +328,36 @@ class HomeAdapter(
...
+                    HedvigTheme {
+                        ConnectPayinCard(
+                            onActionClick = {
+                                tracker.addPaymentMethod()
+                                marketManager.market?.connectPayin(composeView.context)
+                                    ?.let { composeView.context.startActivity(it) }
+                            },
+                            onShown = onPaymentCardShown,
+                        )
+                    }
```
StylianosGakis, line 360: 🔥

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCard.kt
```diff
@@ -24,7 +25,11 @@ fun ClaimStatusCard(
     uiState: ClaimStatusCardUiState,
     modifier: Modifier = Modifier,
     isClickable: Boolean = false,
+    onClaimCardShown: (String) -> Unit,
 ) {
+    LaunchedEffect(uiState) {
+        onClaimCardShown(uiState.id)
+    }
```
StylianosGakis, line 32: This will now send a new event in case some part of the UI changes, like if the UI is pulled to refresh, and the status changed from in progress to paid for example. Is this something we want to track with `hAnalytics.claimCardVisible(claimId, claim.status.rawValue)`?
Intuitively, I am thinking that the "card shown" event should just have a `Unit` key in LaunchedEffect.
[iOS](https://github.com/HedvigInsurance/ugglan/blob/34aef0933c8e13766adf0f1bde70098c3f8b6b9b/Projects/Home/Sources/Claims/ClaimStatus.swift#L20-L25) seem to trigger this on `trackOnAppear` which triggers once [when the view first appears](https://developer.apple.com/documentation/SwiftUI/AnyView/onAppear(perform:)) so we may want to do the same or tell them to ask it as well?

app/src/main/java/com/hedvig/app/feature/home/ui/connectpayincard/ConnectPayinCard.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.hedvig.app.ui.compose.theme.hedvigContentColorFor
+import com.hedvig.app.ui.compose.theme.warning
+
+@Composable
+fun ConnectPayinCard(
+    onActionClick: () -> Unit,
+    onShown: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    LaunchedEffect(key1 = Unit) {
```
StylianosGakis, line 35: nit:
```suggestion
    LaunchedEffect(Unit) {
```

app/src/main/java/com/hedvig/app/feature/home/ui/connectpayincard/ConnectPayinCard.kt
```diff
@@ -0,0 +1,98 @@
...
+            Spacer(Modifier.height(16.dp))
+            Divider()
+            TextButton(
+                onClick = onActionClick,
+                modifier = Modifier
+                    .padding(
+                        top = 4.dp,
+                        end = 8.dp,
+                        bottom = 4.dp,
+                    )
```
StylianosGakis, line 78: Not suggesting to change this, just a nice alternative that I like to do when you got to declare padding only for 3 sides:

```kotlin
.padding(vertical = 4.dp)
.padding(end = 8.dp)
```
To indicate a bit better that the vertical spacings should probably change together if they were to change and avoid changing only 1 of them by accident.

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
```diff
@@ -78,6 +80,13 @@ class LoggedInViewModelImpl(
     }
 
     override fun onTabVisited(tab: LoggedInTabs) {
+        when (tab) {
+            LoggedInTabs.HOME -> hAnalytics.screenViewHome()
+            LoggedInTabs.INSURANCE -> hAnalytics.screenViewInsurances()
+            LoggedInTabs.KEY_GEAR -> {}
```
StylianosGakis, line 86: Poor KEY_GEAR. I'm still salty it's breaking our .graphqls file 😂 https://github.com/HedvigInsurance/key-gear/pull/39

StylianosGakis (approved):
wow that was a lot of work 😅

app/src/main/java/com/hedvig/app/feature/chat/ui/ChatActivity.kt
```diff
@@ -376,17 +376,12 @@ class ChatActivity : BaseActivity(R.layout.activity_chat) {
...
-        gifPickerBottomSheet.show(
-            supportFragmentManager,
-            GifPickerBottomSheet.TAG
-        )
+        GifPickerBottomSheet
+            .newInstance(isKeyboardShown)
+            .show(
+                supportFragmentManager,
+                GifPickerBottomSheet.TAG
+            )
```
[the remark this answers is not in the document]
hedvigoscar: It's intentional - I moved this behaviour to the `ChatViewModel` instead, making the function obsolete.

app/src/main/java/com/hedvig/app/feature/claimdetail/ui/ClaimDetailViewModel.kt
```diff
@@ -27,7 +29,19 @@ class ClaimDetailViewModel(
     private val claimId: String,
     private val chatRepository: ChatRepository,
     private val getClaimDetailUiStateFlowUseCase: GetClaimDetailUiStateFlowUseCase,
+    private val hAnalytics: HAnalytics,
 ) : ViewModel() {
+    init {
+        viewModelScope.launch {
+            when (val result = getClaimDetailUiStateFlowUseCase.invoke(claimId).first()) {
```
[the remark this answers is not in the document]
hedvigoscar: Good point!

app/src/main/java/com/hedvig/app/feature/claimdetail/ui/ClaimDetailViewModel.kt
```diff
@@ -27,7 +29,19 @@ class ClaimDetailViewModel(
...
+                is Either.Left -> {}
+                is Either.Right -> {
+                    hAnalytics.screenViewClaimsStatusDetail(claimId, result.value.claimStatus.rawValue)
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
hedvigoscar: Much better!

app/src/main/java/com/hedvig/app/feature/home/ui/HomeAdapter.kt
hedvigoscar: Agreed!

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCard.kt
hedvigoscar: Good point, perhaps we could key it to `uiState.id`? That is after all the relevant part 🤔

app/src/main/java/com/hedvig/app/feature/home/ui/connectpayincard/ConnectPayinCard.kt
hedvigoscar: Agree fully

app/src/main/java/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
hedvigoscar: We really should retire it 😞

StylianosGakis: Yes pleaseeee 😂

app/src/main/java/com/hedvig/app/feature/home/ui/claimstatus/composables/ClaimStatusCard.kt
StylianosGakis: Yeah that's possible as well.
As I am thinking about it now, I just wonder now how would this function since the cards are all shown in a row of items using a pager, so they're lazily shown 🤔.
If a new claim is inserted into the row, and their index is moved around, will the existing items get fed the new data, aka change the ID aka send the event again?
I am afraid the more I think about it the less I am certain about how it actually functions.
Unless you have a better understanding how the lazy list would work in these scenarios, maybe just using id as a key is good enough. I think no matter what we do this event trigger a bit more often than what we'd probably prefer it to.

app/src/main/java/com/hedvig/app/feature/chat/ui/ChatActivity.kt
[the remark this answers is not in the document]
StylianosGakis: Oops, missed that 😊

PR #893, opened 2022-02-11
APP-1371: dk autocomplete cursor fix

StylianosGakis:
Dealing with the cursor itself is a tiny bit weird. In my solution, I am somewhat breaking the "Single Source of Truth" pattern, but making the address internally hold the input data while also pushing it to the ViewModel on change.
This was the approach I took since whenever I was sending the data to the VM and it was coming back down to the composable, there was no way to know if that action was done by the user entering letters or clicking on an address suggestion. Since there was no distinction between the two, I couldn't find a way to know when to move the cursor at the end of the word or not. We wanted this behavior as when we click an address we want the cursor to go to the end of the word to allow for further modification. If we do this on *every* change, this then means that if the user edits the word at the middle of it, on pressing something the cursor would jump at the end of the word aka: super bad experience!

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
+++ b/app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
@@ -25,7 +25,10 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
```diff
@@ -129,10 +134,31 @@ private fun AddressInput(
...
         ) {
+            var textFieldValue by remember {
+                val text = viewState.input.rawText
+                mutableStateOf(
+                    TextFieldValue(
+                        text = text,
+                        selection = TextRange(text.length)
+                    )
+                )
+            }
```
StylianosGakis, line 145: Holding the data internally. Important detail here is that the `remember` has no key, therefore the TextFieldValue gets populated once with the initial data that comes from the viewModel.

app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteScreen.kt
```diff
@@ -129,10 +134,31 @@ private fun AddressInput(
...
+            LaunchedEffect(viewState.input.selectedDanishAddress) {
+                if (viewState.input.selectedDanishAddress == null) return@LaunchedEffect
+                textFieldValue = textFieldValue.copy(
+                    text = viewState.input.rawText,
+                    selection = TextRange(viewState.input.rawText.length),
+                )
+            }
```
StylianosGakis, line 152: This is the functionality that moves the cursor to the end of the word every time an address is selected.

hedvigoscar (approved):
lgtm

PR #897, opened 2022-02-14
Preparative refactoring before starting the notification analytics events implementation

hedvigoscar:


app/src/main/java/com/hedvig/app/service/push/PushNotificationService.kt
```diff
@@ -7,38 +7,28 @@ import androidx.work.OneTimeWorkRequest
...
+import com.hedvig.app.util.extensions.injectAll
 import org.koin.android.ext.android.inject
 import java.util.concurrent.TimeUnit
 
 class PushNotificationService : FirebaseMessagingService() {
-    private val paymentNotificationManager: PaymentNotificationManager by inject()
-    private val crossSellNotificationManager: CrossSellNotificationManager by inject()
     private val marketManager: MarketManager by inject()
 
+    private val notificationSenders by injectAll<NotificationSender>()
```
hedvigoscar, line 19: A nifty trick, I think 😎

app/src/main/java/com/hedvig/app/service/push/PushNotificationService.kt
```diff
@@ -55,53 +45,21 @@ class PushNotificationService : FirebaseMessagingService() {
...
-                crossSellNotificationManager
-                    .sendCrossSellNotification(this, remoteMessage)
-            NOTIFICATION_TYPE_REFERRALS_CAMPAIGN ->
-                ReferralsNotificationManager
-                    .sendReferralCampaignNotification(this, remoteMessage)
-            else ->
-                ChatNotificationManager
-                    .sendDefaultNotification(this, remoteMessage)
+        notificationSenders.forEach { sender ->
+            if (sender.handlesNotificationType(remoteMessage.data[NOTIFICATION_TYPE_KEY])) {
```
hedvigoscar, line 49: I decided to go for the loop short-circuiting here, as you never want more than one notification to be shown from one notification-event (as far as I could think of, at least)

app/src/main/java/com/hedvig/app/service/push/PushNotificationService.kt
StylianosGakis: What's the type of this?

app/src/main/java/com/hedvig/app/service/push/PushNotificationService.kt
StylianosGakis: Yeah makes total sense. Should it maybe be a chain of .firstNotNullOfOrNull()?.sendNotification instead to make the short-circuiting feature of this a bit more explicit?

Something like (wrote outside of IDE):
notificationSenders.firstNotNullOfOrNull { sender ->
  sender.handlesNotificationType(remoteMessage.data[NOTIFICATION_TYPE_KEY])
}.sendNotification(remoteMessage)

StylianosGakis (approved):

PR #898, opened 2022-02-15
APP-1381: Notification Analytics-Events

hedvigoscar:
- Preparative refactoring before starting the notification analytics events implementation (#897)
- Implement `notificationPermission`
- Implement `notificationOpened`

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/tracking/LaunchTrackingInitializer.kt
+++ b/app/src/main/java/com/hedvig/app/feature/tracking/LaunchTrackingInitializer.kt
@@ -1,0 +1,20 @@
+package com.hedvig.app.feature.tracking
+
+import android.content.Context
+import androidx.core.app.NotificationManagerCompat
+import androidx.startup.Initializer
+import com.hedvig.app.feature.di.KoinInitializer
+import com.hedvig.hanalytics.HAnalytics
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+class LaunchTrackingInitializer : Initializer<Unit>, KoinComponent {
+    private val hAnalytics: HAnalytics by inject()
+    override fun create(context: Context) {
+        hAnalytics.notificationPermission(
+            NotificationManagerCompat.from(context).areNotificationsEnabled()
+        )
+    }
+
+    override fun dependencies() = listOf(KoinInitializer::class.java)
+}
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/hedvig/app/feature/tracking/NotificationOpenedTrackingActivity.kt
```diff
@@ -0,0 +1,28 @@
+package com.hedvig.app.feature.tracking
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import com.hedvig.hanalytics.HAnalytics
+import org.koin.android.ext.android.inject
+
+class NotificationOpenedTrackingActivity : AppCompatActivity() {
```
hedvigoscar, line 10: This activity implements the technique called 'reverse activity trampoline' in order to track notification opens in a way that should not be too disruptive for the user.

More reading here: https://onesignal.com/blog/android-12-notification-open-changes/

app/src/main/java/com/hedvig/app/feature/tracking/LaunchTrackingInitializer.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import android.content.Context
+import androidx.core.app.NotificationManagerCompat
+import androidx.startup.Initializer
+import com.hedvig.app.feature.di.KoinInitializer
+import com.hedvig.hanalytics.HAnalytics
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+class LaunchTrackingInitializer : Initializer<Unit>, KoinComponent {
```
hedvigoscar, line 11: I decided to go with the Jetpack App Startup for this type of 'track some stuff on launch'-events - it seemed like a good fit.

app/src/main/java/com/hedvig/app/feature/tracking/NotificationOpenedTrackingActivity.kt
hugokallstrom: This wont show as a activity transition since we are not creating a view, right?

hugokallstrom (approved):
LGTM

app/src/main/java/com/hedvig/app/feature/tracking/NotificationOpenedTrackingActivity.kt
hedvigoscar: Not as far as I could see - the transition was basically equal to what was in place before.

PR #902, opened 2022-02-15
Debug feature manager

hugokallstrom:
image


hedvigoscar:
Agree with @StylianosGakis here, I think we should move this to the engineering screen.

StylianosGakis:
> Agree with @StylianosGakis here, I think we should move this to the engineering screen.

Hmm I would say it should be there *in addition* to the profile screen. Let's make their lives easier, no reason for this to be hidden in engineering mode exclusively.

hedvigoscar:
What's the value-add of having it in two places? 🤔 I'm a bit wary of the risk of showing engineering-tooling on the production-screens for end-users that shouldn't see it.

StylianosGakis:
> What's the value-add of having it in two places? 🤔 I'm a bit wary of the risk of showing engineering-tooling on the production-screens for end-users that shouldn't see it.

Ok we can hide it and get feedback from them if it's okay

hugokallstrom:
Added different implementations depending on build variant. This way we make sure to not have any feature flag activity code in prod.

PR #909, opened 2022-02-22
APP-1443: Home screen item builder refactor

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    private fun HomeQuery.Data.isActive() = contracts.any {
+        it.status.asActiveStatus != null ||
+            it.status.asTerminatedTodayStatus != null ||
+            it.status.asTerminatedInFutureStatus != null
+    }
+
+    private fun HomeQuery.Data.isTerminated() = contracts.all { it.status.asTerminatedStatus != null }
+
+    private fun HomeQuery.Data.isSwitching() = contracts.any {
```
hugokallstrom, line 160: New logic for showing switching here

app/src/main/java/com/hedvig/app/feature/home/data/GetHomeUseCase.kt
```diff
@@ -14,11 +15,10 @@ class GetHomeUseCase(
     private val localeManager: LocaleManager,
 ) {
 
-    suspend operator fun invoke(forceReload: Boolean): HomeResult {
-        return when (val response = createHomeCall(forceReload = forceReload).safeQuery()) {
-            is QueryResult.Error -> HomeResult.Error(response.message)
-            is QueryResult.Success -> HomeResult.Home(response.data)
-        }
+    suspend operator fun invoke(forceReload: Boolean): Either<QueryResult.Error, HomeQuery.Data> {
```
StylianosGakis, line 18: Nice 👍

app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
```diff
@@ -0,0 +1,163 @@
...
+import com.hedvig.app.feature.claims.ui.commonclaim.CommonClaimsData
+import com.hedvig.app.feature.claims.ui.commonclaim.EmergencyData
+import com.hedvig.app.feature.home.ui.claimstatus.data.ClaimStatusCardUiState
+import com.hedvig.app.util.featureflags.Feature
+import com.hedvig.app.util.featureflags.FeatureManager
+
+@OptIn(ExperimentalStdlibApi::class)
+class HomeItemsBuilder(
+    private val featureManager: FeatureManager
+) {
```
StylianosGakis, line 17: A great idea! This did end up being quite a lot, with a lot of mixed cases. This feels like it will make it easier to work with going forward!

app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun buildItems(
+        homeData: HomeQuery.Data
+    ): List<HomeModel> = when {
+        homeData.isActive() -> buildActiveItems(homeData)
+        (homeData.isPending() || homeData.isActiveInFuture()) && homeData.isSwitching() -> buildSwitchingItems(homeData)
+        homeData.isPending() -> buildPendingItems(homeData)
+        homeData.isActiveInFuture() -> buildActiveInFutureItems(homeData)
+        homeData.isTerminated() -> buildTerminatedItems(homeData)
+        else -> listOf(HomeModel.Error)
+    }
```
StylianosGakis, line 28: This way we have to make sure that since those are mutually exclusive, and only one of them runs that we have to specify on each case all the items that it needs, and not for example having this one long method that simply adds something to the long existing list. Feels like a win in the long run, where we need to be aware that when we make changes, we take care of all these mutually exclusive cases!

app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
```diff
@@ -0,0 +1,163 @@
...
+@OptIn(ExperimentalStdlibApi::class)
+class HomeItemsBuilder(
+    private val featureManager: FeatureManager
+) {
+
+    fun buildItems(
+        homeData: HomeQuery.Data
+    ): List<HomeModel> = when {
+        homeData.isActive() -> buildActiveItems(homeData)
+        (homeData.isPending() || homeData.isActiveInFuture()) && homeData.isSwitching() -> buildSwitchingItems(homeData)
```
StylianosGakis, line 23: Super small nit, but could we flip the things around the && to put the isSwitching at the front to indicate that this is predominantly why this case exists? At first I read the isPending and thought  how can we have this here and in the line below, but it's all because the isSwitching is the most important thing in this expression.
Again you don't need to do this, but it got me confused initially. Maybe it wouldn't inside the IDE where I could've seen the brackets easier

StylianosGakis (approved):
Suuuuper nice cleanup on the builder, I really like how it looks like at first glance! I feel like this will help us quite a lot going forward

app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
hugokallstrom: Fixed!

PR #910, opened 2022-02-23
APP-1261: Quote cart offer

hugokallstrom:


app/src/engineering/java/com/hedvig/app/feature/offer/MockOfferViewModel.kt
```diff
@@ -87,40 +77,19 @@ class MockOfferViewModel : OfferViewModel() {
                     _viewState.value = ViewState.Error
                     return@launch
                 }
-                val topOfferItems = OfferItemsBuilder.createTopOfferItems(
```
hugokallstrom, line 90: Need to create new mock data here

app/src/main/java/com/hedvig/app/feature/offer/OfferRepository.kt
```diff
@@ -8,38 +8,60 @@ import com.apollographql.apollo.coroutines.await
...
+    private val featureManager: FeatureManager,
 ) {
     fun offerQuery(ids: List<String>) = OfferQuery(localeManager.defaultLocale(), ids)
 
-    fun offer(ids: List<String>) = apolloClient
-        .query(offerQuery(ids))
-        .watcher()
-        .toFlow()
-        .map(::toDataOrError)
+    fun offer(ids: List<String>): Flow<OfferResult> {
```
hugokallstrom, line 34: This function now returns our own data model, instead of generated graphql classes. This means we can map both queries to the same model, and let a feature flag decide which query/subscription to use

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -321,19 +323,31 @@ class OfferViewModelImpl(
...
+                    data.quoteBundle.quotes.first().insuranceTerms
+            } else {
+                emptyList()
+            },
+            insurableLimitsItems = if (data.quoteBundle.quotes.size == 1) {
+                listOf(InsurableLimitItem.Header.Details) +
+                    data.quoteBundle.quotes.first().insurableLimits
+            } else {
+                emptyList()
+            },
```
StylianosGakis, line 347: Maybe it was nice that this if size == 1 check was hidden from the VM inside the builder? Not very important but maybe it's a bit nicer :D

StylianosGakis (approved):
Quite a lot of files changed, probably due to the name change to ..Item and the introduction of the fragment in the graphql query. I hope I didn't miss anything but it's looking awesome overall!

apollo/src/main/graphql/com/hedvig/android/owldroid/graphql/quoteCartSubscription.graphql
```diff
@@ -0,0 +1,34 @@
...
+        }
+        ... on IndefinitePercentageDiscount {
+          indefinitePercentageDiscount: percentageDiscount
+        }
+        ... on PercentageDiscountMonths {
+          monthsPercentageDiscount: percentageDiscount
+          monthsQuantity: quantity
+        }
+        ... on NoDiscount {
+          valueNoDiscount: _
```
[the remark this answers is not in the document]
hugokallstrom: This is just a dummy field, so that we can switch on NoDiscount

PR #913, opened 2022-03-01
Use new endpoint for active payment methods in order to show information from third party providers.

hugokallstrom:
In order to show Vipps!


StylianosGakis (approved):
V2 is approved! :D

PR #917, opened 2022-03-03
APP-1263: Edit quote

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/checkout/EditQuotesUseCase.kt
```diff
@@ -0,0 +1,123 @@
...
+            put("locale", localeManager.defaultLocale().rawValue)
+            putJsonObject("payload") {
+                put("ssn", ssn)
+                put("email", email)
+                putJsonObject("data") {
+                    put("ssn", ssn)
+                    put("email", email)
+                }
+            }
+        }.toString()
```
StylianosGakis, line 71: Ah this is such a nice and simple way to do this, awesome!

app/src/main/java/com/hedvig/app/feature/checkout/EditQuotesUseCase.kt
```diff
@@ -0,0 +1,123 @@
...
+                        .getString("code")
+                } catch (exception: JSONException) {
+                    null
+                }
+
+                if (errorCode != null) {
+                    Either.Left(Error(errorCode))
+                } else {
+                    Either.Right(Success)
+                }
```
StylianosGakis, line 95: I think these can also be written as `Error(errorCode).left()` and `Success.right()` respectively. I usually like avoiding having to write`Either.Right(` for example. Just something you may want to know if you haven't stumbled upon it yet.

app/src/main/java/com/hedvig/app/feature/checkout/EditQuotesUseCase.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private suspend fun editQuoteWithEmailAndSSN(
+        quoteId: String,
+        ssn: String,
+        email: String
+    ): Either<Error, Success> = apolloClient.mutate(EditMailAndSSNMutation(quoteId, ssn, email))
+        .safeQuery()
+        .toEither()
+        .mapLeft { Error(it.message) }
```
StylianosGakis, line 106: We have a `toEither` version that looks like this:
```kotlin
inline fun <ErrorType> toEither(ifEmpty: (message: String?) -> ErrorType): Either<ErrorType, T> = when (this) {
    is Error -> Either.Left(ifEmpty(message))
    is Success -> Either.Right(this.data)
}
```
which could potentially make this instead be 
```kotlin
.toEither { Error(it) }
```

This is to replicate the behavior that exists in the arrow library on the `Option` type:
```kotlin
public inline fun <L> toEither(ifEmpty: () -> L): Either<L, A>
```
Where they also have a toEither where you can in the same function do the mapping of what your error type would be, without having to do a `mapLeft` afterwards. Plus in our case you get access to message directly as you may want to use it in the new error type (like you do in this case with `it.message`)

app/src/main/java/com/hedvig/app/feature/offer/OfferRepository.kt
```diff
@@ -33,20 +32,26 @@ class OfferRepository(
...
-    fun offer(ids: List<String>): SharedFlow<OfferResult> {
+    fun offer(ids: List<String>): Flow<OfferResult> {
         return if (featureManager.isFeatureEnabled(Feature.QUOTE_CART)) {
             val subscription = QuoteCartSubscription(localeManager.defaultLocale(), ids.first())
             apolloClient.subscribe(subscription)
                 .safeSubscription()
                 .map { it.toResult() }
-                .shareIn(CoroutineScope(SupervisorJob()), started = SharingStarted.Lazily)
+                .onStart {
+                    val query = QuoteCartQuery(localeManager.defaultLocale(), ids.first())
```
StylianosGakis, line 42: This isn't new to this PR I guess, but doing .first() will simply throw if the list is empty which feels a bit scary. Are we like 100% we're never doing that mistake?
Would it be worth it to use a `NonEmptyList` on the `ids` parameter potentially?

StylianosGakis (approved):
Looking great!

app/src/main/java/com/hedvig/app/feature/checkout/EditQuotesUseCase.kt
hugokallstrom: Awesome

PR #918, opened 2022-03-04
APP-1265: Quote cart access token

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -374,32 +384,30 @@ class OfferViewModelImpl(
...
-                                result.quote.insurableLimits,
-                                result.quote.insuranceTerms
+                                it.displayName,
+                                it.perils.map { PerilItem.Peril(it) },
+                                it.insurableLimits,
+                                it.insuranceTerms
                             )
                         )
                     )
                 }
```
StylianosGakis, line 410: Hmm I think this one is a bit different from how we use Either in the other cases.
Specifically previously if we wanted to simply do something on left and on right we used fold to do this, but here we have a mapLeft and a map.
This is something I'm still trying to figure out in general how to consistently do things in a certain way, but I have some comments.
Here if we just say we won't do fold() on, the mapLeft feels like it should be a `tapLeft` instead, since with this after the line of mapLeft we changed the type of Either from `Either<A, B>` to `Either<Unit, B>` since we mapped the left to the result of `offerAndLoginStatus.value = OfferAndLoginStatus.Error` which is a variable assignment which I assume would be `Unit`. What we probably were going for was the idea of just "do something if it's left" which `tap|tapLeft` exists for afaik.
Same exact idea with `map`, where we were looking for just doing `tap` since we again don't then need to furhter change the type from `Either<Unit, B>` to `Either<Unit, ChannelResult<Unit>>` which we are here since map will change the type with the last result in the lambda which here is the result of trySend which is the type I said above.

Because this is something that I've been thinking in general I asked [this](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1646391903231499) to potentially get some input from the Arrow community as well. Maybe we can se how people in general like to do this.

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -374,32 +384,30 @@ class OfferViewModelImpl(
...
                         Event.OpenQuoteDetails(
                             QuoteDetailItems(
-                                result.quote.displayName,
-                                result.quote.perils.map { PerilItem.Peril(it) },
-                                result.quote.insurableLimits,
-                                result.quote.insuranceTerms
+                                it.displayName,
+                                it.perils.map { PerilItem.Peril(it) },
+                                it.insurableLimits,
+                                it.insuranceTerms
```
StylianosGakis, line 406: These would still be nice to be named `quote` instead of `it`. I personally really like avoiding `it` if it's not a trivial one-liner

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetPostSignDependenciesUseCase.kt
```diff
@@ -1,35 +1,30 @@
...
+    data class Result(
+        val postSignScreen: PostSignScreen,
+        val displayName: String,
+    )
 
-        object Error : Result()
-    }
+    suspend operator fun invoke(quoteIds: List<String>): Either<ErrorMessage, Result> = offerRepository
+        .offer(NonEmptyList.fromListUnsafe(quoteIds))
+        .firstOrNull()
```
StylianosGakis, line 22: Hmm, I am pretty sure in this case the firstOrNull() will never be null since if it would be null we'd be getting a runtime crash since fromListUnsafe throws when the list passed into it is empty.

I think it's be best if we either make the quoteIds also a NEL and use the safe alternatives when creating it. Or if we want to keep the function signature the same, we could replace this with something safe like:
```kotlin
suspend operator fun invoke(quoteIds: List<String>): Either<ErrorMessage, Result> {
    if (quoteIds.isEmpty()) {
        return ErrorMessage().left()
    }
    return offerRepository
        .offer(NonEmptyList.fromListUnsafe(quoteIds))
        .firstOrNull()
        ?.map { it.toResult() }
        ?: ErrorMessage().left()
}
```
Like you did in some other place in the code.

After the answer in Arrow slack I can see this `either{}` exists too, which could make something like this possible as well:

```kotlin
suspend operator fun invoke(quoteIds: List<String>): Either<ErrorMessage, Result> {
    return either {
        val ids = NonEmptyList.fromList(quoteIds).toEither { ErrorMessage() }.bind()
        val result = offerRepository
            .offer(ids)
            .firstOrNull()
            ?.map { it.toResult() }
            ?: ErrorMessage().left()
        result.bind()
    }
}
```

Interesting, but probably not for this use case. But interesting nonetheless.

app/src/main/java/com/hedvig/app/feature/offer/usecase/getquote/GetQuoteIdsUseCase.kt
```diff
@@ -17,41 +17,31 @@ class GetQuoteIdsUseCase(
...
-                    ifTrue = { QuoteIds(listOf(it.lastQuoteOfMember.asCompleteQuote!!.id)) }
-                )
-            }
-    }
+    private suspend fun getIdsFromLastQuoteOfMember(): Either<ErrorMessage, QuoteIds> = apolloClient
+        .query(LastQuoteIdQuery())
+        .safeQuery()
+        .toEither { ErrorMessage(it) }
+        .map {
+            QuoteIds(listOf(it.lastQuoteOfMember.asCompleteQuote!!.id))
```
StylianosGakis, line 45: Are we like sure sure that this isn't null? Before we were checking it in the `Either.conditionally` but we're not doing this anymore. This isn't safe is it?

app/src/main/java/com/hedvig/app/feature/offer/usecase/getquote/GetQuotesUseCase.kt
```diff
@@ -17,34 +19,18 @@ class GetQuotesUseCase(
...
-                }
-                quoteIds.isNotEmpty() -> emitAll(getOffer(NonEmptyList.fromListUnsafe(quoteIds)))
-                else -> {
-                    when (val result = getQuoteIdsUseCase.invoke(null)) {
-                        is Either.Left -> emit(Result.Error(result.value.message))
-                        is Either.Right -> emitAll(getOffer(NonEmptyList.fromListUnsafe(result.value.ids)))
-                    }
+                featureManager.isFeatureEnabled(Feature.QUOTE_CART) -> quoteCartId?.let {
+                    emitAll(offerRepository.offer(nonEmptyListOf(quoteCartId)))
+                } ?: ErrorMessage("No quote id found").left()
```
StylianosGakis, line 28: Maybe keep the braces here? Maybe it's just GitHub but it's quite hard to read imo

StylianosGakis (approved):
Quite a lot of nits here and there, but it'd be nice to address them, or at least have a discussion about it if you disagree with anything :D

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetPostSignDependenciesUseCase.kt
nomisRev: Hey @StylianosGakis,
Another small suggestion here is that you can use `ensureNotNull`.
This eliminates the need for `map`, and lifting `ErrorMessage()` into `left()`.

```kotlin
suspend operator fun invoke(quoteIds: List<String>): Either<ErrorMessage, Result> = either {
        val ids = NonEmptyList.fromList(quoteIds).toEither { ErrorMessage() }.bind()
        val result = offerRepository
            .offer(ids)
            .firstOrNull()
            ?.bind()?.toResult()
         ensureNotNull(result) { ErrorMessage() }
}
```

app/src/main/java/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -234,21 +233,21 @@ class OfferViewModelImpl(
             initialValue = ViewState.Loading,
         )
 
-    private suspend fun loadQuoteIds() {
-        return when (val result = getQuoteIdsUseCase.invoke(quoteCartId)) {
-            is Either.Left -> offerAndLoginStatus.value = OfferAndLoginStatus.Error
-            is Either.Right -> {
-                hAnalytics.screenViewOffer(result.value.ids)
-                quoteIds = result.value.ids
+    private suspend fun loadQuoteIds() = getQuoteIdsUseCase.invoke(quoteCartId)
```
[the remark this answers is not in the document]
hugokallstrom: Yeah you're right :)

PR #932, opened 2022-03-14
Fix variable extraction

hugokallstrom:


StylianosGakis (approved):
Wow this is subtle. Is this using variables from the wrong GraphQL fragment or what? Are we sure about the "!!" this time?

hugokallstrom:
> Wow this is subtle. Is this using variables from the wrong GraphQL fragment or what? Are we sure about the "!!" this time?

Its referencing the multi variable. We are null checking it above in the when statement. Before we used the variables in the root fragment.

PR #933, opened 2022-03-14
Make KeepScreenOnRequest a "side effect"-like composable

StylianosGakis:
After some great input here https://kotlinlang.slack.com/archives/CJLTWPH7S/p1647259570691099?thread_ts=1646998371.832069&cid=CJLTWPH7S I think this is a better approach to this.
A good opportunity for Oscar take a look at this too

Also a chance to improve this function's name? I don't *love* it, but I can't think of something nicer @hedvigoscar

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderScreen.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderScreen.kt
@@ -28,7 +28,7 @@ import com.hedvig.app.R
 import com.hedvig.app.ui.compose.composables.buttons.LargeContainedButton
 import com.hedvig.app.ui.compose.composables.buttons.LargeTextButton
 import com.hedvig.app.ui.compose.theme.HedvigTheme
-import com.hedvig.app.util.compose.KeepScreenOn
+import com.hedvig.app.util.compose.ScreenOnFlag
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
@@ -133,7 +133,7 @@ fun Recording(
     clock: Clock,
     modifier: Modifier = Modifier,
 ) {
-    KeepScreenOn()
+    ScreenOnFlag()
     Column(
         horizontalAlignment = Alignment.CenterHorizontally,
         modifier = modifier.fillMaxWidth(),
--- a/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderScreen.kt
+++ b/app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderScreen.kt
@@ -28,7 +28,7 @@ import com.hedvig.app.R
 import com.hedvig.app.ui.compose.composables.buttons.LargeContainedButton
 import com.hedvig.app.ui.compose.composables.buttons.LargeTextButton
 import com.hedvig.app.ui.compose.theme.HedvigTheme
-import com.hedvig.app.util.compose.KeepScreenOnRequest
+import com.hedvig.app.util.compose.KeepScreenOn
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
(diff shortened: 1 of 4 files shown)


hedvigoscar (approved):
lgtm!

hedvigoscar:
name could probably be `KeepScreenOn()`, short and sweet 😄

hedvigoscar:
Come to think of it, perhaps we should indicate that it's not a regular `@Composable` that draws some UI? 🤔 

I could imagine having it as a lower-case named function being a good idea, something like this:

```kotlin
@Composable fun WantsToKeepScreenOn() {
    keepScreenOn()
    // all of the actual UI-stuff, like rows and so on
}

// Lowercased to indicate that this doesn't draw any UI
@Composable fun keepScreenOn() {
    // same implementation as current `KeepScreenOnRequest()`
}
```

StylianosGakis:
> Come to think of it, perhaps we should indicate that it's not a regular `@Composable` that draws some UI? 🤔
> 
> I could imagine having it as a lower-case named function being a good idea, something like this:
> 
> ```kotlin
> @Composable fun WantsToKeepScreenOn() {
>     keepScreenOn()
>     // all of the actual UI-stuff, like rows and so on
> }
> 
> // Lowercased to indicate that this doesn't draw any UI
> @Composable fun keepScreenOn() {
>     // same implementation as current `KeepScreenOnRequest()`
> }
> ```

Yeah I've thought of that too, however I think the precedent is set with functions like [BackHandler](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#BackHandler(kotlin.Boolean,kotlin.Function0)) which do exactly this. They're not emitting UI but they're not returning a value either. It's practically closer to what `SideEffect` or `LaunchedEffect` is, hence the PascalCase.

Reading the compose API guidelines, it explicitly states that function returning composables are the ones that should be camelCase [here](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#BackHandler(kotlin.Boolean,kotlin.Function0)). With that said, they also [suggest](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#why-2) that composables should be nouns, which is what `BackHandler` follows, which made me reluctant to use the verb `KeepScreenOn` but still can't find a nice noun for it. I've thought of things like `WithScreenOn` which unfortunately feels like it'd be a scoping function "from kotlin with()" or like `ScreenOn` which isn't descriptive at all.

Naming is hard 😅

hedvigoscar:
> Yeah I've thought of that too, however I think the precedent is set with functions like [BackHandler](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#BackHandler(kotlin.Boolean,kotlin.Function0)) which do exactly this. They're not emitting UI but they're not returning a value either. It's practically closer to what `SideEffect` or `LaunchedEffect` is, hence the PascalCase.

Ah, precedent. Good. Let's use `PascalCase`.
 
> Reading the compose API guidelines, it explicitly states that function returning composables are the ones that should be camelCase [here](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#BackHandler(kotlin.Boolean,kotlin.Function0)). With that said, they also [suggest](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md?rgh-link-date=2022-03-14T13%3A31%3A55Z#why-2) that composables should be nouns, which is what `BackHandler` follows, which made me reluctant to use the verb `KeepScreenOn` but still can't find a nice noun for it. I've thought of things like `WithScreenOn` which unfortunately feels like it'd be a scoping function "from kotlin with()" or like `ScreenOn` which isn't descriptive at all.

That does complicate it slightly.

To me it feels like some kind of token, marker or semaphore. `Semaphore` would be inappropriate given that it already has some special meaning in synchronization, which leaves `Token` or `Marker` as viable options, perhaps?

StylianosGakis:
I'm leaning towards breaking this "Should" here and keep it a verb to keep the name descriptive but non-conventional, as opposed to conventional but confusing. Unless you say let's use "x" which I would happily accept if you have 1 alternative that you like

hedvigoscar:
How about `ScreenOnFlag` or `KeepScreenOnFlag`?

As long as at least one flag is in composition, the screen gets kept on.

StylianosGakis:
ScreenOnFlag wins as I believe it fulfills all the criteria while also being descriptive enough, thanks for that!

PR #934, opened 2022-03-16
APP-1495: Migrate moving flow flag

StylianosGakis:
I figured this one can now be deleted completely.


hedvigoscar:
I think we may want to keep it, given that it's present in `hAnalytics` 😉

StylianosGakis:
Aha, but why is it present in hAnalytics? What are we gonna be using this feature flag for?

StylianosGakis:
Yeah, let me do that

PR #935, opened 2022-03-17
feature/hanalytics feature flag provider

hugokallstrom:


hugokallstrom:
> Interesting!
> 
> Some thoughts came up while reading through the code:
> 
> * How would this implementation treat the case where the return type of the feature flag was non-booleanic?
> * Where would feature flag-invalidation come into play in this implementation?

Good questions! We talked about this before and after some thinking this is my conclusions: 

What would make most sense is to not treat those cases as feature flags at all - but rather have other interfaces like `PaymentTypeProvider` or `GetLoginMethodUseCase`. We can build different interfaces on top of HAnalytics, and expose eg. events, feature flags or other functionality (like what payment type to use) depending on the use case. 

We could add a suspending function for invalidating featureflags in `FeatureManager`, and call that in places where we need to invalidate - eg. login.

PR #936, opened 2022-03-17
Some changes for the PR, explained further in the comment

StylianosGakis:
Some juggling with HAnalytics implementation

Extracts all the network-related functionality to a retrofit-like 
interface called HAnalyticsService which can take care of all the 
networking logic like constructing the OkHttp request, adding the 
headers and so on. Leaves a simple interface of the functions that we 
can access from the REST API provided to us from the hAnalytics backend 
service.

This allows our implementation of the HAnalytics interface be a simple 
delegation to one class which manages what happens when we want to send 
an HAnalyticsEvent and one that handles experiments.

ExperimentManager does exactly that, make use of the REST API interface,
keeps a local map of the loaded experiments, and handles all the 
necessary concurrency protection that is needed using a Mutex.
 
SendHAnalyticsEventUseCase exists as the de-facto way to send an 
HAnalytics event as it makes sure that all the registered sinks will 
receive each event that is sent.


app/src/engineering/java/com/hedvig/app/feature/tracking/Module.kt
```diff
@@ -1,12 +1,15 @@
...
 import org.koin.dsl.module
 
 val trackingLogModule = module {
     viewModel { TrackingLogViewModel(get()) }
-    single { EngineeringTracker() } bind HAnalyticsSink::class
-    single { DebugLogTrackerSink() } bind HAnalyticsSink::class
-    single { ShakeTrackerSink() } bind HAnalyticsSink::class
+    single<EngineeringTrackerSink> { EngineeringTrackerSink() } bind HAnalyticsSink::class
+    single<DebugLogTrackerSink> { DebugLogTrackerSink() } bind HAnalyticsSink::class
+    single<ShakeTrackerSink> { ShakeTrackerSink() } bind HAnalyticsSink::class
```
hedvigoscar, line 14: Why do we need explicit type arguments here? 🤔

gradle/libs.versions.toml
```diff
@@ -204,7 +203,6 @@ moneta = { module = "org.javamoney:moneta", version.ref = "moneta" }
 
 okhttp-loggingInterceptor = { module = "com.squareup.okhttp3:logging-interceptor", version.ref = "okhttp" }
 okhttp-mockWebServer = { module = "com.squareup.okhttp3:mockwebserver", version.ref = "okhttp" }
-okhttp-coroutines = { module = "ru.gildor.coroutines:kotlin-coroutines-okhttp", version.ref = "okhttpCoroutines" }
```
hedvigoscar, line 207: Why did we stop using this? 🤔

app/src/main/java/com/hedvig/app/feature/hanalytics/ExperimentsInitializer.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlinx.coroutines.launch
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+class ExperimentsInitializer : Initializer<Unit>, KoinComponent {
+    private val experimentManager: ExperimentManager by inject()
+
+    override fun create(context: Context) {
+        CoroutineScope(Dispatchers.IO).launch {
+            experimentManager.preloadExperiments()
```
hedvigoscar, line 17: What happens if this call fires before a token has been acquired?

app/src/main/java/com/hedvig/app/util/featureflags/Feature.kt
```diff
@@ -4,7 +4,8 @@ enum class Feature(
     val key: String,
     val title: String,
     val explanation: String,
-    val enabledByDefault: Boolean = false
+    val enabledByDefault: Boolean = false,
+    val isFromHAnalytics: Boolean = false,
```
hedvigoscar, line 8: Is this property used? 🤔

StylianosGakis: No my bad!

gradle/libs.versions.toml
StylianosGakis: It was messing up the IDE hints since this was pulling in an older version of OkHttp and then we were getting wrong errors when we were accessing the body of a okhttp Response. After removing this it now refers to the correct functions without wrong IDE hints.
The fact that everything was still working was just because the function names were changed in a backwards compatible way, but the type hints we were getting all this time were actually wrong. Particularly, we were getting `.body` as a `String` but it is actually a `ResponseBody` which just now I realized that then when we do `.string()` to it, it under the hood uses it properly, like doing `use {}` on it since it's `Closeable` and all that stuff.
The library is a 4-5 year old one anyway which isn't getting updated anyway, and it only provided one function, so I simply pulled that in to our code-base instead, I thought the tradeoff seemed logical to improve our IDE experience

hedvigoscar: Makes sense!

app/src/engineering/java/com/hedvig/app/feature/tracking/Module.kt
StylianosGakis: Yeah we don't *need* them, but I've grown to really like being explicit with Koin and what kind of types the `singles` provide.
I remember talking about this before here https://github.com/HedvigInsurance/android/pull/891#discussion_r803653451 and I feel like if we did an explicit paramter for every single thing we provide with Koin it would make it easier for our eyes. Maybe you didn't see that comment though, because I haven't heard your opinion on it :D

hedvigoscar: Fair enough 🙂

hugokallstrom (approved):
LGTM

app/src/main/java/com/hedvig/app/feature/hanalytics/ExperimentsInitializer.kt
StylianosGakis: Removed this Initializer because of exactly this problem, thanks for the heads up!

PR #938, opened 2022-03-22
Feature/hanalytics with individual providers

StylianosGakis:
An idea of how it'd work to have the debug build only be run by local providers and the real and app distribution build rely on unleashed.


hugokallstrom (commented):
Like it!

hugokallstrom (approved):

PR #940, opened 2022-03-28
APP-1436: FeatureManager using hAnalytics

StylianosGakis:
Draft just to have the PR open and have the CI run on changes


app/src/main/java/com/hedvig/app/feature/profile/data/ObserveProfileUiStateUseCase.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.hedvig.app.util.apollo.QueryResult
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.mapLatest
+
+class ObserveProfileUiStateUseCase(
+    private val profileRepository: ProfileRepository,
+    private val profileQueryDataToProfileUiStateMapper: ProfileQueryDataToProfileUiStateMapper,
+) {
+    suspend fun invoke(): Flow<Either<QueryResult.Error, ProfileUiState>> {
```
hugokallstrom, line 16: Map to `ErrorMessage` here instead of `QueryResult.Error`

app/src/main/java/com/hedvig/app/feature/profile/ui/tab/ProfileUiState.kt
```diff
@@ -0,0 +1,77 @@
+package com.hedvig.app.feature.profile.ui.tab
+
+import androidx.annotation.StringRes
+import com.hedvig.android.owldroid.fragment.CashbackFragment
+import com.hedvig.android.owldroid.graphql.ProfileQuery
+
+data class ProfileUiState(
```
hugokallstrom, line 7: I think we should see these as domain objects instead of ui state objects. Repositories or use cases map from network to domain, and then viewmodels can create and hold state based on these domain classes. Having immutable domain classes like "Profile", "Cashback" etc. is perfectly fine imo, no need to postfix with UiState :)

hugokallstrom (approved):
Awesome work

app/src/main/java/com/hedvig/app/feature/profile/ui/tab/ProfileUiState.kt
StylianosGakis: Yeah the "UiState" was what I decided on instead of just `Profile` which would be super confusing of course. Took inspiration from here https://developer.android.com/jetpack/guide/ui-layer#define-ui-state and even more specifically this section https://developer.android.com/jetpack/guide/ui-layer#naming-conventions.
In my mind this would be the "domain" object which drives what's on the UI of this specific screen. I am however not like 100% confident about the naming conventions. I think the way it works is perfectly fine, just not sure about naming it 😅

oscarnylander: Arguably, this `data class` is a ViewModel-level representation (its fields are things like `paymentState: Show|DontShow` - clearly not representing the data at a domain-level 🤔 Hence, I think `UiState`/`ViewState` is appropriate here.

hugokallstrom: True, in this case we use it directly as ui state in the viewmodel. It just feels weird but I guess thats the benefit of graphql - its often possible to go directly from network query -> ui state :)

hugokallstrom: "The ViewModel holds and exposes the state to be consumed by the UI. The UI state is application data transformed by the ViewModel."

`ProfileUiState` could be application data, which doesnt need any transformation to state in the viewmodel. Looking that this:
[https://developer.android.com/topic/libraries/architecture/images/mad-arch-ui-udf.png](url)

StylianosGakis: Right, yeah they do say "transformed by the ViewModel" which in this case is not happening really, it's happening inside the UseCase instead, which I believe is what brings this confusion.
In our case, it's happening inside `ObserveProfileUiStateUseCase` by using `ProfileQueryDataToProfileUiStateMapper`.
We could have the mapper be injected to the ViewModel instead, and have the mapping happen in this chain of operations https://github.com/HedvigInsurance/android/blob/3d3e83be19abba3b4a26f312073738d025b4caa0/app/src/main/java/com/hedvig/app/feature/profile/ui/ProfileViewModel.kt#L50 or something like that.
So if anything, I would say the model is named adequately but is done on the data (or domain if you'd like https://developer.android.com/jetpack/guide/data-layer#architecture) layer which is technically wrong for this sort of transformation.

Right? 😅

StylianosGakis: And that could also make the existence of `ObserveProfileUiStateUseCase` not that necessary as well (maybe keep it to do the mapping to Either<ErrorMessage, ProfileQuery.Data> from what it originally is in the Repository. Otherwise this could be done in the repository itself too if there aren't any clients that would be annoying to adapt).
Then the model could come from the repository as it was, and the VM could then use this mapper to do the conversion to the `UiState` model and then we're all good :D

hugokallstrom: Yeah we could do that, but its fine to keep as it is also :) We just have to think about re-usability with use cases and domain objects where it makes sense.

StylianosGakis: Resolving due to this https://github.com/HedvigInsurance/android/pull/972

app/src/main/java/com/hedvig/app/feature/profile/data/ObserveProfileUiStateUseCase.kt
StylianosGakis: Resolving due to this https://github.com/HedvigInsurance/android/pull/972

PR #943, opened 2022-03-30
APP-1494: migrate key gear flag

StylianosGakis:
Took the opportunity to migrate away from LiveData and bunch the results in a ViewState while at it


hedvigoscar (approved):
lgtm

PR #944, opened 2022-03-31
APP-1262: Quote cart campaign

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/referrals/ui/redeemcode/RedeemCodeViewModel.kt
```diff
@@ -1,28 +1,60 @@
...
         viewModelScope.launch {
-            val response =
-                runCatching { redeemReferralCodeRepository.redeemReferralCode(code) }
-            if (response.isFailure) {
-                response.exceptionOrNull()?.let { e(it) }
-                return@launch
+            val newState = if (featureManager.isFeatureEnabled(Feature.QUOTE_CART)) {
+                editQuoteCart(code)
+            } else {
+                redeemCode(code)
```
StylianosGakis, line 39: I think these functions can just do `_viewState.update { it.copy... }` directly without reporting the new viewState back to the call site. This will also be true if we add more functions (maybe the error state clearing function mentioned above) where those would also need to handle it themselves instead of returning the new viewState somewhere. It'd be consistent across all of them that way.

StylianosGakis, line 39: I think these functions can just do `_viewState.update { it.copy... }` directly without reporting the new viewState back to the call site. This will also be true if we add more functions (maybe the error state clearing function mentioned above) where those would also need to handle it themselves instead of returning the new viewState somewhere. It'd be consistent across all of them that way.

app/src/main/java/com/hedvig/app/feature/referrals/ui/redeemcode/RedeemCodeViewModel.kt
```diff
@@ -1,28 +1,60 @@
...
+        viewState.value.copy(errorMessage = "No quote id found")
+    } else {
+        editCampaignUseCase.addCampaignToQuoteCart(code, quoteCartId)
+            .fold(
+                ifLeft = { error -> _viewState.value.copy(errorMessage = error.message) },
+                ifRight = { id -> _viewState.value.copy(quoteCartId = id) }
+            )
+    }
+
+    private suspend fun redeemCode(code: String) = redeemReferralCodeRepository.redeemReferralCode(code)
```
StylianosGakis, line 55: I think these functions can just do `_viewState.update { it.copy... }` directly without reporting the new viewState back to the call site. This will also be true if we add more functions (maybe the error state clearing function mentioned above) where those would also need to handle it themselves instead of returning the new viewState somewhere. It'd be consistent across all of them that way.

app/src/main/java/com/hedvig/app/feature/referrals/ReferralsReceiverActivity.kt
```diff
@@ -21,7 +24,9 @@ class ReferralsReceiverActivity : BaseActivity(R.layout.referrals_receiver_activ
         super.onCreate(savedInstanceState)
 
         binding.apply {
-            referralViewModel.redeemCodeStatus.observe(this@ReferralsReceiverActivity) { startChat() }
+            referralViewModel.viewState.onEach {
```
StylianosGakis, line 27: Same comment as the other place where we're observing a flow instead of a LiveData

app/src/main/java/com/hedvig/app/HedvigApplication.kt
```diff
@@ -67,7 +70,9 @@ open class HedvigApplication : Application() {
         if (authenticationTokenService.authenticationToken == null) {
             whatsNewRepository.removeNewsForNewUser()
             CoroutineScope(IO).launch {
-                acquireHedvigToken()
+                if (!featureManager.isFeatureEnabled(Feature.QUOTE_CART)) {
```
StylianosGakis, line 73: Hmm what's the interaction like about this where we don't want this token when using quote cart? This is handled differntly in that case I presume? is the `authenticationTokenService` also updated in that approach, or something completely different?

StylianosGakis (commented):
Super nice with the `CreateQuoteCartUseCase.QuoteCartId` going outside of that namespace and we can simply refer to it as QuoteCartId now. I think the concept is clear enough on its own anyway so this feels good. 

Some comments here and there regarding safely collecting flows, and other general nits or things I didn't understand and would like to get some more context about.

app/src/main/java/com/hedvig/app/feature/offer/usecase/EditCampaignUseCase.kt
```diff
@@ -0,0 +1,50 @@
...
+        .mutate(QuoteCartAddCampaignMutation(code, quoteCartId.id))
+        .safeQuery()
+        .toEither()
+        .mapLeft { ErrorMessage(it.message) }
+        .map { it.quoteCart_addCampaign }
+        .flatMap {
+            it.asQuoteCart
+                ?.id
+                ?.let { id -> QuoteCartId(id) }
+                .rightIfNotNull { ErrorMessage(it.asBasicError?.message) }
```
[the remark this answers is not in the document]
hugokallstrom: Yes that would be easier to read. What I like about not having a fragment is that we can query additional fields in eg. `QuoteCartAddCampaign` without changing other queries.

app/src/main/java/com/hedvig/app/feature/offer/OfferRedeemCodeBottomSheet.kt
```diff
@@ -1,10 +1,16 @@
...
 
+import android.os.Bundle
 import com.hedvig.android.owldroid.graphql.RedeemReferralCodeMutation
+import com.hedvig.app.feature.offer.model.QuoteCartId
 import com.hedvig.app.feature.referrals.ui.redeemcode.RedeemCodeBottomSheet
 import org.koin.androidx.viewmodel.ext.android.sharedViewModel
 
 class OfferRedeemCodeBottomSheet : RedeemCodeBottomSheet() {
+
+    override val quoteCartId: QuoteCartId?
```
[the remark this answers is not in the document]
hugokallstrom: Its used by `RedeemCodeBottomSheet`

app/src/main/java/com/hedvig/app/feature/referrals/ui/redeemcode/RedeemCodeViewModel.kt
```diff
@@ -1,28 +1,60 @@
...
+    private val featureManager: FeatureManager,
+    private val editCampaignUseCase: EditCampaignUseCase,
 ) : ViewModel() {
 
-    val redeemCodeStatus: MutableLiveData<RedeemReferralCodeMutation.Data> = MutableLiveData()
+    data class ViewState(
+        val quoteCartId: QuoteCartId? = null,
+        val data: RedeemReferralCodeMutation.Data? = null,
+        val loading: Boolean = false,
+        val errorMessage: String? = null
```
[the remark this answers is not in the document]
hugokallstrom: Great resources, thanks. In this case we dont actually need to dismiss the error, since we will close the bottom sheet when the code is successful.

app/src/main/java/com/hedvig/app/HedvigApplication.kt
hugokallstrom: Good question. I will ask the backenders!

PR #947, opened 2022-04-04
APP-1543: Refactor Marketing Screen

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+        val language = currentState.language ?: throw IllegalStateException("Can't submit, language is null")
+        updateMarketPickerState { it.copy(isLoading = true) }
+        viewModelScope.launch {
+            val result = submitMarketAndLanguagePreferencesUseCase
+                .invoke(language, market)
+            if (result.isLeft()) {
+                updateMarketPickerState { it.copy(isLoading = false) }
+                return@launch
+            }
+            updateApplicationLanguageUseCase.invoke(market, language)
```
hugokallstrom, line 176: Do we need to update here, since we are calling this in `setMarket` and `setLanguage`?

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+        if (!currentState.isValid) {
+            throw IllegalStateException("Can't submit, PickMarket is not valid")
+        }
+        val market = currentState.market ?: throw IllegalStateException("Can't submit, market is null")
+        val language = currentState.language ?: throw IllegalStateException("Can't submit, language is null")
+        updateMarketPickerState { it.copy(isLoading = true) }
+        viewModelScope.launch {
+            val result = submitMarketAndLanguagePreferencesUseCase
+                .invoke(language, market)
+            if (result.isLeft()) {
```
hugokallstrom, line 172: Here you can do 
```
submitMarketAndLanguagePreferencesUseCase
                .invoke(language, market)
                .fold(
                  ifLeft = {
                   updateMarketPickerState { it.copy(isLoading = false) }
                  },
                  ifRight = {
                   updateApplicationLanguageUseCase.invoke(market, language)
                   _state.value = MarketPicked.Loading
                  loadMarketPicked()
                  }
                )
```

Or with switch statement for left + right like in `loadMarketPicker`

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+        val language = state.language ?: return defaultLanguage(market)
+
+        if (!isCompatibleLanguage(language, market)) {
+            return defaultLanguage(market)
+        }
+
+        return language
+    }
+
+    private fun isCompatibleLanguage(language: Language, market: Market) = when (market) {
```
hugokallstrom, line 147: Move this to `MarketExt` maybe?

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -0,0 +1,240 @@
...
+        Market.SE -> language == Language.EN_SE || language == Language.SV_SE
+        Market.NO -> language == Language.EN_NO || language == Language.NB_NO
+        Market.DK -> language == Language.EN_DK || language == Language.DA_DK
+        Market.FR -> language == Language.EN_FR || language == Language.FR_FR
+    }
+
+    private fun defaultLanguage(market: Market) = when (market) {
```
hugokallstrom, line 154: And this

app/src/main/java/com/hedvig/app/feature/marketing/data/GetMarketingBackgroundUseCase.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class GetMarketingBackgroundUseCase(
+    private val apolloClient: ApolloClient,
+    private val localeManager: LocaleManager,
+) {
+    suspend operator fun invoke() = apolloClient
+        .query(MarketingBackgroundQuery(localeManager.defaultLocale().rawValue))
+        .safeQuery()
+        .toEither()
+        .map { it.appMarketingImages.firstOrNull() }
```
hugokallstrom, line 16: We could map this to our own domain model, and perhaps return error if blurhash etc. are null? Then we can skip the safeLets in the viewmodel.

hugokallstrom (approved):
Great rewrite 💯

app/src/main/java/com/hedvig/app/feature/marketing/marketpicked/MarkedPicked.kt
```diff
@@ -0,0 +1,98 @@
+package com.hedvig.app.feature.marketing.marketpicked
```
StylianosGakis, line 1: The file name probably was meant to be MarkerPicked instead of MarkedPicked right :D

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
hedvigoscar: As it stands, we need to do it, in case the user does not explicitly make a selection and uses the values by default provided by the geolocation information.

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
```diff
@@ -0,0 +1,294 @@
...
+) {
+    var sheet by rememberSaveable { mutableStateOf<PickMarketSheet?>(null) }
+    val coroutineScope = rememberCoroutineScope()
+    val modalBottomSheetState = rememberModalBottomSheetState(ModalBottomSheetValue.Hidden)
+
+    val insets = LocalWindowInsets.current
+    val navigationBarHeight = with(LocalDensity.current) { insets.navigationBars.bottom.toDp() }
+    ModalBottomSheetLayout(
+        sheetState = modalBottomSheetState,
+        sheetContent = {
```
StylianosGakis, line 72: Maybe extract the sheet content to another composable? Kind of long with this and the screen content beeing in the same composable

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
```diff
@@ -0,0 +1,294 @@
...
+            )
+            Spacer(Modifier.height(8.dp))
+            Box(
+                modifier = Modifier
+                    .size(width = 32.dp, height = 4.dp)
+                    .align(Alignment.CenterHorizontally)
+                    .background(
+                        color = colorResource(R.color.color_divider),
+                        shape = RoundedCornerShape(20.dp)
+                    )
```
StylianosGakis, line 87: Running this locally I couldn't see this ð Changing it to like Color.Black it was visible. Have you noticed something like that on your local environment too? Might be smth weird on my emulator too somehow

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
hugokallstrom: Ah I see

app/src/main/java/com/hedvig/app/feature/marketing/ui/BackgroundImage.kt
```diff
@@ -0,0 +1,46 @@
...
+import coil.compose.rememberImagePainter
+import coil.size.Scale
+import com.google.accompanist.systemuicontroller.rememberSystemUiController
+import com.hedvig.app.feature.marketing.Background
+import com.hedvig.app.util.compose.rememberBlurHash
+
+@Composable
+fun BackgroundImage(background: Background, content: @Composable BoxScope.() -> Unit) {
+    Box(modifier = Modifier.fillMaxSize()) {
+        (background as? Background.Loaded)?.let { bg ->
```
StylianosGakis: Could we maybe make this an early return if it's not Loaded instead of indenting everything below this? And let the compiler then infer that it's Background.Loaded since we're returning early if it's not.

app/src/main/java/com/hedvig/app/feature/marketing/ui/BackgroundImage.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.ContentScale
+import coil.compose.rememberImagePainter
+import coil.size.Scale
+import com.google.accompanist.systemuicontroller.rememberSystemUiController
+import com.hedvig.app.feature.marketing.Background
+import com.hedvig.app.util.compose.rememberBlurHash
+
+@Composable
+fun BackgroundImage(background: Background, content: @Composable BoxScope.() -> Unit) {
```
StylianosGakis, line 20: Is this the API we want to expose for the background image? The way this will work now is that if the background is still loading, or takes a long time to load, or fails to load, the rest of the screen will never draw on the screen.
What if where we fetch the image from fails but we still want to have the rest of the screen to work just fine possibly with a placeholder background or some plain color or whatever.
I'd imagine the background image and the front content to be in the same Box scope, but have the be individually drawing on the screen. What do you think? Sort of one laid out on top of the other.

StylianosGakis (approved):
A lot of view code deleted, feels so nice :D 
Some thoughts here and there, plus maybe we can consider how we can make the flickering not happen somehow. I think it's worth spending some thought on it, as it's quite intense with the light changes when it happens.
But that can happen on another PR/later of course

app/src/main/java/com/hedvig/app/feature/marketing/ui/BackgroundImage.kt
hedvigoscar: Definitely not - this was a mistake on my part. The intended behaviour was to show the content, regardless of if the background image loads or not.

hedvigoscar: If the background image fails to load, we still have a default background color showing, which I believe should be sufficient.

app/src/main/java/com/hedvig/app/feature/marketing/ui/BackgroundImage.kt
hedvigoscar: With the rewritten code, this is no longer possible 😄

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
hedvigoscar: I did consider this, however, I felt like passing all of the dependencies to this composable would be almost as long as the composable itself, making it slightly pointless to extract it 😞

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
StylianosGakis: Yeah I was running on API 30
It seems to work just fine for dark mode though
image
image

Not sure what is going on. Maybe something to do with the theme picking up the wrong thing?
Not sure how it'd ever work on some API and not on some other API after looking at the code. Only thing I can imagine is if it came from a composable function instead with it checking if the system is in dark theme instead if might be a bit more reliable that using `colorResource` with an xml color?

To test this I tried adding this inside `com.hedvig.app.ui.compose.theme.Colors.kt`
```kotlin
val Colors.divider: Color
    @Composable
    @ReadOnlyComposable
    get() = primary.copy(alpha = 0.12f)
```

And using it still doesn't fix it.

But I also see that the radio buttons are also hidden in the light mode. Something must be going wrong in general with light mode and the colors that are picked up.

StylianosGakis: Hmm right I see we're doing this theme overlay with `ThemeOverlay.Hedvig.MarketingActivity` in the `MarketingActivity` so it works, it's just that the primary color is White therefore it's just not showing with the white background. Same for the radio buttons.

hedvigoscar: Damn. Yeah, I see it now. Thanks for the feedback - I'll have to rethink that thing, I guess 😞

PR #952, opened 2022-04-13
Fix issues with market picker

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
```diff
@@ -47,13 +49,29 @@ import com.hedvig.app.feature.settings.Market
...
+    val bool by produceState(false) {
+        while (isActive) {
+            value = !value
+            delay(1000.milliseconds)
+        }
+    }
+    val f = animateFloatAsState(if (bool) 0f else 1f)
+    println(f.value)
+}
+
```
StylianosGakis, line 74: Oh we did kept this, probably by accident right?

Personally, I'd love to have this handy in the codebase, if placed in an appropriate package and not used from production code, because when doing UI work I might want to use it real quick. Just like with another function that I pretty much always add while working, and delete before commit which is:
```kotlin
fun Modifier.debugBorder(color: Color = Color.Red, width: Dp = 1.dp): Modifier {
    return this.then(Modifier.border(width, color))
}
```
So as I said in the office as well, I'd not oppose having a place for such non-production functions, but otherwise just delete this :D

StylianosGakis (approved):
Looking good :)

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
hedvigoscar: it was accidental, yeah 😬

PR #955, opened 2022-04-14
APP-1478: New fields for cross sell and moving flow stories

hugokallstrom:


StylianosGakis (approved):
Those are from what I understand placeholders for something that there is to come in the future right? The `quoteCartEnabled` checks that for now return the same thing.

hugokallstrom:
> Those are from what I understand placeholders for something that there is to come in the future right? The `quoteCartEnabled` checks that for now return the same thing.

Yes this was supposed to be a draft. I'll use the new fields when they are available in the schema.

StylianosGakis:
Is this targeted at https://github.com/HedvigInsurance/android/pull/963 originally? Could we change the PR target to make the diff a bit more clear to go through?

hugokallstrom:
> Is this targeted at #963 originally? Could we change the PR target to make the diff a bit more clear to go through?

Sorry, fixed

PR #956, opened 2022-04-14
APP-1497: Experiment post onboarding show payment step

StylianosGakis:
AKA Connect payment at sign.


hugokallstrom (approved):
LGTM

app/src/main/java/com/hedvig/app/common/Mapper.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.app.common
+
+fun interface Mapper<From, To> {
```
hedvigoscar, line 3: What purpose does this interface serve? 🤔

hedvigoscar (commented):
One question, everything else looks good 🙂

app/src/main/java/com/hedvig/app/common/Mapper.kt
StylianosGakis: So we could have the two (and however many more in the future) mapper classes without the interface as well. It is true that we never use them as the `Mapper` type.
The purpose it serves now is that you can start with a class and name it whatever. Make it extend this interface, and type the two types it's mapping to/from, and then you get the convenience of it auto-generating the function that needs to be implemented.
On top of that, it serves as a way to at a glance look at that class and see that it implements this interface and the types it has chosen and then you don't have to look at anything else to see its purpose. If not, you'd have to rely on the class itself having a "Mapper" suffix (which it will anyway tbf) to indicate this, plus you'd have to rely on the developer to name the function "map" instead from forcing it with this contract to avoid having different mappers looking different on the call-site.
Idk, I personally like such rails when we're doing things like this, but no, we don't *need* to have this interface to exist for correctness or any other benefit atm 🤷‍♂️

hedvigoscar: I can't say that I understand the value of it, but I guess it doesn't hurt 🤷

PR #957, opened 2022-04-14
Add version build number in the about app screen

StylianosGakis:
Since we use this step https://github.com/bitrise-steplib/steps-change-android-versioncode-and-versionname it should work to simply add it on that screen. I didn't know we do that before, so I thought this wouldn't just work :D


hedvigoscar (approved):
lgtm

PR #958, opened 2022-04-19
APP-1496: Integrate `paymentType`-flag

hedvigoscar:


app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel.kt
```diff
@@ -17,34 +21,36 @@ import kotlinx.coroutines.launch
...
                 .payment()
-                .onEach { _paymentData.postValue(it.data) }
+                .onEach { _paymentData.value = it.data }
                 .catch { e(it) }
                 .launchIn(this)
 
             payinStatusRepository
                 .payinStatusFlow()
-                .onEach { _payinStatusData.postValue(it.data) }
+                .onEach { _payinStatusData.value = Pair(it.data, featureManager.getPaymentType()) }
```
StylianosGakis, line 53: This will now ask the featureManager on each new value that `payingStatusFlow` returns, is this intended? Would we maybe want the result of the feature manager function to be called just once and saved maybe even in its own StateFlow or something like that?

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel.kt
```diff
@@ -17,34 +21,36 @@ import kotlinx.coroutines.launch
 abstract class PaymentViewModel(
     hAnalytics: HAnalytics,
 ) : ViewModel() {
-    protected val _paymentData = MutableLiveData<PaymentQuery.Data>()
-    protected val _payinStatusData = MutableLiveData<PayinStatusQuery.Data>()
-    val data = combineTuple(_paymentData, _payinStatusData)
+    protected val _paymentData = MutableStateFlow<PaymentQuery.Data?>(null)
+    protected val _payinStatusData = MutableStateFlow<Pair<PayinStatusQuery.Data?, PaymentType>?>(null)
+    val data: StateFlow<Pair<PaymentQuery.Data?, Pair<PayinStatusQuery.Data?, PaymentType>?>> =
+        combineState(_paymentData, _payinStatusData, viewModelScope)
```
StylianosGakis, line 27: combineState turns everything into nullables right. Can we not use this instead https://github.com/Kotlin/kotlinx.coroutines/blob/52cbf01c63c0737ee745ff08568f26385271a9ab/kotlinx-coroutines-core/common/src/flow/operators/Zip.kt#L51-L52 and handle the null cases on the ViewModel with a local `ViewState` (sealed?) object instead of in the call site? Losing the nullability information and making it always nullable doesn't sound that much fun.

Something like this https://github.com/HedvigInsurance/android/blob/48e4e9f44ec5484ed830b0b259e6b0f468b00bc7/app/src/main/java/com/hedvig/app/feature/addressautocompletion/ui/AddressAutoCompleteViewModel.kt#L82-L95 where all the inputs are combined into a ViewState.

StylianosGakis (approved):
Just some thoughts about `combineState`. I think I'd rather we just use the existing combine() function and be explicit in each case with the initial state using `stateIn`. This will encourage more well-thought initial states instead of having everything as nullable and forwarding the responsibility to handle all of this on the UI.

But besides that, if we don't want to do that at all, or don't want to do it now, it's looking good to merge :D

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel.kt
hedvigoscar: This is indeed intended. Calls to `FeatureManager` should generally be pretty cheap to make, making this a low-priority concern, and if we introduce a caching layer in the `ViewModel` here, we risk cache staleness on the `ViewModel`-level.

app/src/main/java/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel.kt
hedvigoscar: I was mostly just trying to mimic exactly the utility that was in place before (`combineTuple`), but I think I agree - being explicit is likely better here.

PR #960, opened 2022-04-19
APP-1501: Experiment show charity

StylianosGakis:
Took the opportunity to also do a removal of using a generated GraphQL model in the UI layer by introducing ProfileUiState.
This let me move the business logic from inside the fragment to inside the mapper.

Won't merge until the qasa hAnalytics interface is decided upon
`Feature.SHOW_CHARITY -> false // TODO hAnalytics.qasa()`


app/src/main/java/com/hedvig/app/feature/profile/data/ObserveProfileUiStateUseCase.kt
```diff
@@ -0,0 +1,24 @@
...
+) {
+    suspend fun invoke(): Flow<Either<QueryResult.Error, ProfileUiState>> {
+        return profileRepository
+            .profile()
+            .map(QueryResult<ProfileQuery.Data>::toEither)
+            .mapLatest { either ->
+                either.map { profileQueryDataToProfileUiStateMapper.map(it) }
+            }
+    }
+}
```
StylianosGakis, line 24: So about this one. The only reason I added this to be between the VM and the Repository is so that I wouldn't make the Repository have to do the dto -> domain model mapping, but have it serve the data coming from `profile()` as it was before, and have the mapping being done here.
With that said, I don't know what's the "best" way to do this, or simply what it is that we prefer in such cases, so I am completely open to changing this if you have some opinions.
We could change it so that the repository changes the `profile()` function to return a flow of the mapped type directly, so `ProfileUiState` instead of `ProfileQuery.Data`. I am open to either way.

app/src/main/java/com/hedvig/app/feature/profile/ui/ProfileViewModelImpl.kt
```diff
@@ -57,23 +87,8 @@ class ProfileViewModelImpl(
...
-                    response.data?.let { _data.value = ViewState.Success(it) }
-                }
-                .catch { exception ->
-                    e(exception)
-                    _data.value = ViewState.Error
-                }
-                .launchIn(this)
-        }
+    override fun reload() {
+        observeProfileRetryChannel.retry()
```
StylianosGakis, line 91: Changed this since as I see the old `load` function, if it were to be called twice (which it does if the load function is called which was called from the UI when the reload button was clicked), it'd actually launch 2 separate coroutines observing the `profile()`. This makes it only observe once, cancelling the previous collection if it was still active.

app/src/main/java/com/hedvig/app/feature/profile/ui/charity/CharityAdapter.kt
```diff
@@ -40,18 +40,18 @@ class CharityAdapter(
...
         val paragraph: TextView = binding.cashbackOptionParagraph
         val button: Button = binding.cashbackSelect
-        fun bind(item: ProfileQuery.CashbackOption, clickListener: (String) -> Unit) {
+
+        fun bind(charityOption: CharityOption, clickListener: (id: String) -> Unit) {
             binding.apply {
-                cashbackOptionTitle.text = item.name
-                cashbackOptionParagraph.text = item.paragraph
+                cashbackOptionTitle.text = charityOption.name
+                cashbackOptionParagraph.text = charityOption.description
```
StylianosGakis, line 47: This change was due to the type `Cashback` having a deprecated field.
`paragraph: String @deprecated(reason: "use description instead")`

app/src/main/java/com/hedvig/app/feature/profile/ui/tab/ProfileFragment.kt
```diff
@@ -151,16 +89,72 @@ class ProfileFragment : Fragment(R.layout.profile_fragment) {
...
+                    onClick = {
+                        startActivity(Intent(requireContext(), MyInfoActivity::class.java))
+                    }
+                )
+            )
+            when (val charityState = profileUiState.charityState) {
+                CharityState.DontShow -> {}
+                CharityState.NoneSelected -> add(buildCharityRowItem())
+                is CharityState.Selected -> add(buildCharityRowItem(charityState.charityName))
+            }
```
StylianosGakis, line 109: On this build function this is the change where the charity isn't shown when not needed when the mapper function determines that the fflag is off.

app/src/main/java/com/hedvig/app/feature/profile/ui/tab/ProfileQueryDataToProfileUiStateMapper.kt
```diff
@@ -0,0 +1,50 @@
...
+) : Mapper<ProfileQuery.Data, ProfileUiState> {
+
+    override suspend fun map(from: ProfileQuery.Data): ProfileUiState {
+        val cashbackFragment = from.cashback?.fragments?.cashbackFragment
+        val charityState = run {
+            if (featureManager.isFeatureEnabled(Feature.SHOW_CHARITY).not()) return@run CharityState.DontShow
+            val charityName = cashbackFragment?.name
+            if (charityName != null) return@run CharityState.Selected(charityName)
+            CharityState.NoneSelected
+        }
```
StylianosGakis, line 26: Is this use of `run` a bit weird? This was the only simple way I found of wanting to do a simple val assignment but with some logic in there. I am thinking it may be weird because I've never had to use this before, but maybe there simply wasn't a case where I needed it 🤷‍♂️

app/src/main/java/com/hedvig/app/feature/profile/ui/charity/CharityAdapter.kt
hedvigoscar: Did you verify the correctness of this change? 😬 Sometimes these deprecations are not actually proper replacements, and hence they have to be vetted for every use-case.

StylianosGakis: Oh, okay I will double check this then, I didn't know we'd deprecate something with the message "use X instead" and not actually have it be a replacement :D
From running the app and looking at differences it looked fine, but I will admit I wasn't expecting this really to be a problem at all so I didn't pay 100% attention. I shall check again 😅

StylianosGakis: Yeah just checked in the graphql playground with all locales for all markets and they all respond with the exact same description and paragraph field.

app/src/main/java/com/hedvig/app/util/featureflags/flags/HAnalyticsFeatureFlagProvider.kt
```diff
@@ -9,13 +9,14 @@ class HAnalyticsFeatureFlagProvider(
...
         Feature.EXTERNAL_DATA_COLLECTION -> hAnalytics.allowExternalDataCollection()
         Feature.FRANCE_MARKET -> hAnalytics.frenchMarket()
         Feature.KEY_GEAR -> hAnalytics.keyGear()
         Feature.MOVING_FLOW -> hAnalytics.movingFlow()
-        Feature.CONNECT_PAYMENT_AT_SIGN -> hAnalytics.postOnboardingShowPaymentStep()
-        Feature.UPDATE_NECESSARY -> hAnalytics.updateNecessary()
-        Feature.REFERRAL_CAMPAIGN -> hAnalytics.foreverFebruaryCampaign()
         Feature.QUOTE_CART -> hAnalytics.useQuoteCart()
+        Feature.REFERRAL_CAMPAIGN -> hAnalytics.foreverFebruaryCampaign()
+        Feature.SHOW_CHARITY -> false // TODO hAnalytics.qasa()
```
hedvigoscar, line 19:
```suggestion
        Feature.SHOW_CHARITY -> TODO()
```

This way, we won't accidentally release this code, as there will be a certain crash on the profile tab.

hedvigoscar (approved):

app/src/main/java/com/hedvig/app/feature/profile/ui/tab/ProfileQueryDataToProfileUiStateMapper.kt
hedvigoscar: I think it's sensible.

app/src/main/java/com/hedvig/app/feature/profile/ui/ProfileViewModelImpl.kt
hedvigoscar: Neat!

app/src/main/java/com/hedvig/app/feature/profile/data/ObserveProfileUiStateUseCase.kt
hedvigoscar: Looks sensible to me.

app/src/main/java/com/hedvig/app/util/featureflags/flags/HAnalyticsFeatureFlagProvider.kt
StylianosGakis: Yeah very good idea. Will fix right away!

app/src/engineering/java/com/hedvig/app/feature/profile/MockProfileViewModel.kt
```diff
@@ -1,24 +1,37 @@
...
+import com.hedvig.app.feature.profile.ui.tab.ProfileQueryDataToProfileUiStateMapper
 import com.hedvig.app.testdata.feature.profile.PROFILE_DATA
 import com.hedvig.app.util.LiveEvent
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
 
-class MockProfileViewModel : ProfileViewModel() {
+class MockProfileViewModel(
```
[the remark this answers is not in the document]
StylianosGakis: Yeah good idea! Addressed with this commit https://github.com/HedvigInsurance/android/pull/960/commits/c3131a0e1c1c8e0ccde7ca018b384c5060ee1793

PR #961, opened 2022-04-19
APP-1499: Feature flag show forever tab

StylianosGakis:
The big change is that since now both key_gear and referrals are decided by hAnalytics, we no longer query the member at all on the loggedIn query.
This made is so that a ton of tests are touched by this PR. For a cleaner review it might be worth looking at all the commits prior to the last two which are the ones that touch the tests and mock data builders.

Also I guess I won't merge this until the hAnalytics interface is decided upon, to not have this accidentally reach production without the feature flag being setup properly.


hedvigoscar:
Is this a duplicate of #959?

StylianosGakis:
> Is this a duplicate of #959?

Yeah that was a big oopsie by me. We will have to merge one of them probably though. It’s basically the same diff with some extra stuff.

- Just changed the test builder object for that response to be one, since it doesn’t contain the feature response anymore which made a big diff inside the tests.
- Did some cleanup by removing the shouldOverrideFeatureFlags function
- Made the LoggedInViewModel actually respond with a viewState after it has received a response from all the 4 different sources by turning them into flows without initial value. So now it only responds with a viewState when all 4 flows have something to offer. With the MutableStateFlow implementation, they all start with an initial value which isn't necessarily true, and therefore can make the ViewState emit wrong values back to back until all of the values are initialized to their proper result.

If it's ok I think we should probably merge this one due to just doing the extra cleanup work around all this stuff while I was doing this. If there's something that you don't like though we can also merge the other one. Totally my bad for making this situation like this though :D Not my intention 😅

oscarnylander:
I agree.

Let's make this a draft and have it return `TODO()` for `Feature.REFERRALS`, and I'll close the other one.

oscarnylander (approved):

PR #962, opened 2022-04-20
APP-1500: Add Cross-Sell feature flag

oscarnylander:
Draft until we have implemented the upstream feature flag.


StylianosGakis (approved):
lgtm!

The build fail is from the gradient enum which now has an extra entry and the exhaustive when is failing. Could rebase this on top of 1501 and make this PR point there instead which would fix it.

PR #963, opened 2022-04-20
APP-1577: Offer bundle toggling

hugokallstrom:
Query possibleVariations and show selected bundle variation in offer screen. Remove quote cart feature flag.


app/src/main/java/com/hedvig/app/feature/offer/ui/composable/VariantButton.kt
```diff
@@ -0,0 +1,90 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.hedvig.app.ui.compose.theme.HedvigTheme
+import com.hedvig.app.ui.compose.theme.hedvigBlack
+import com.hedvig.app.ui.compose.theme.hedvigBlack12percent
+
+@Composable
+fun VariantButton(
```
StylianosGakis, line 28: Looking at the preview, the radio button seems to be a bit too high maybe? Is there a FIgma file for this, I couldn't find one. Maybe we can fix it up a bit before merging if my eyes are not deceiving me.

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetBundleVariantUseCase.kt
```diff
@@ -0,0 +1,34 @@
...
+                val bundleVariant = offerModel.variants
+                    .takeIf { variantId.isNotEmpty() }
+                    ?.find { it.id == variantId }
+                    ?: offerModel.variants.first()
+
+                Pair(offerModel, bundleVariant)
+            }
+        }
+
+    fun invoke(variantId: String) {
```
StylianosGakis, line 31: Hmm I think the class name doesn't give enough context in this case to know what invoke would mean. Could we name this something like `selectVariant` instead?

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetBundleVariantUseCase.kt
```diff
@@ -0,0 +1,34 @@
...
+import arrow.core.Either
+import com.hedvig.app.feature.offer.OfferRepository
+import com.hedvig.app.feature.offer.model.OfferModel
+import com.hedvig.app.feature.offer.model.QuoteBundleVariant
+import com.hedvig.app.util.ErrorMessage
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.combine
+
+class GetBundleVariantUseCase(
```
StylianosGakis, line 12: Not super comfortable having a UseCase hold state like this instead of being a pure function with simple consistent parameters in/ results out concept. Especially in this case where we're using this UseCase inside `CheckoutViewModel` only to observe the result, meaning that we're relying on the same usecase to be provided to that and to `OfferViewModel` which it does happen in our case since we're providing it with a koin `single` but still not sure if we want to rely on such behaviors with UseCases.
I could be wrong too though :D

StylianosGakis (commented):
Hey was a bit hard to go through everything, I think it'd be nice for Oscar to take a look at this as well.

Most of what I am seeing makes sense and looks good. I posted the comments I had here and there. Also curious as I stated in one of my questions if this PR also implies that the quote cart feature flag is not gonna be used anymore, or if I just missed how this still works with the non quote cart ways still.

app/src/main/java/com/hedvig/app/feature/offer/usecase/GetBundleVariantUseCase.kt
hugokallstrom: What do you suggest? Move to OfferViewModel?

StylianosGakis: Hmm I thought about it a bit, and since this is used from the CheckoutViewModel it wouldn't work.
If anything, it'd have to be raised one layer higher, potentially saved in DataStore or something like that which would be a reliable storage.
However, what that would do is have the same practical result as this StateFlow here, which is like a form of storage, just in-memory. I'll make a PR with a suggestion, but this shouldn't block this merge imo.

hugokallstrom: I dont think datastore is the right use case for this tbh. I feel datastore should be app-wide storage where as this is bound to the offer screen lifecycle only.

StylianosGakis: Right, but we're using this inside `CheckoutViewModel` as well, which escapes that screens scope, doesn't it?

hugokallstrom: True, its still bound to the "checkout-lifecycle" though. Datastore should be for data we want to keep between app lifecycles. We would in this case also need to "reset" this selected variant to null or empty after we leave the offer flow.

StylianosGakis: Yeah, with the current impl it will also persist since the UseCase is a single one persisted in the app as well, even if you leave this flow. But I guess it will reset if the app is killed, which wouldn't happen with a more persistent storage.
To avoid wasting time on this I think we should just merge it and see if we need to do any changes to it in the future.

hugokallstrom: I renamed this use case, hopefully it makes more sense

PR #964, opened 2022-04-20
APP-1575: Make embark continue when selecting "other" as insurer

StylianosGakis:
This screen has the same UI and works super similar with the `ExternalInsurerFragment`. Took inspiration from how that one handles this case here https://github.com/HedvigInsurance/android/blob/ee3d88a1ce79a4f39fdbd2320eec367b9b91461e/app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/ExternalInsurerFragment.kt#L162 .

Now it continues without adding anything to the value store, as I assume is the correct approach to this problem.
I do however wonder if this ever worked in the first place, I did not find any code in the history that would indicate so to try and revert to that instead. I might've missed smth though.


hugokallstrom (approved):
LGTM!

PR #967, opened 2022-04-21
Variant cards UI tweaks

StylianosGakis:
Some changes to do the following:

- Allow the card to expand vertically when it needs more space instead of hiding the text
- Make the radio button go away from the material fixed size specs and allow it to be bigger so that it matches the design
- Add a layout which allows two texts that are side to side to not squish each other or to have to use fixed widths to avoid this interaction. Instead this layout lets each text take the minimum space they need as possible, before expanding to take on the rest of the space if it's available.

Please if you can find a better name than `HorizontalTextsWithMaximumSpaceTaken` suggest it 😅

image


hugokallstrom (approved):
👏

PR #971, opened 2022-04-25
APP-1554: Last embark passage not commiting student status

StylianosGakis:


StylianosGakis:
This along with this PR https://github.com/HedvigInsurance/embark/pull/896 in embark should fix this problem, we can also double check with QA

StylianosGakis:
With that said, I am doing some things inside EmbarkViewModel that might need some cleanup.

Specifically the preProcessResponseWithCommittedValueStore function, and how it's only called from the SelectActionFragment for now. Is this some functionality that we maybe want to be true no matter when it's called? Maybe even rename it back to `preProcessResponse` and have the previous one be private, so that one may only be able to call this and have it work on the committed valueStore?

Actually that might be a good idea, but we'd need to QA this even harder. This is the proposed PR https://github.com/HedvigInsurance/android/pull/974.
Or maybe someone has a better idea or reasons why my approach isn't the best idea.

hugokallstrom (approved):
LGTM

PR #973, opened 2022-04-25
Hotfix/6.5.6

StylianosGakis:
Change font usage to match the new requirements.

Specifically:
Don't use ".._small" font anywhere but explicitly in the big title in the Home page.
Update Headline4 to the new size/line height as specified in the specs


oscarnylander (approved):
lgtm

PR #977, opened 2022-04-28
Make offer page header and variants not flicker

StylianosGakis:
Just let the reyclerView know they're the same item because on top of flickering, it was also scrolling down when selecting another variant, making the experience not ideal


hugokallstrom (approved):
LGTM

PR #978, opened 2022-04-28
Make VariableExtractor handle empty multi action items

StylianosGakis:
House was failing to do the query if there were no extra buildings specified. This fixes it.


hugokallstrom (approved):
LGTM

PR #979, opened 2022-04-29
APP-1615: Query textaction subtitle

hugokallstrom:
To keep scope down I decided to just set this subtitle as a regular hint. Hints will be overwritten by subtitles though, thats a limitation we can live with for now.

Design for how it should look like is here: https://www.figma.com/file/8zGhkNSDxyWSKDylTJ2mnq/Car-Insurance-iOS?node-id=30%3A4175


StylianosGakis (approved):
Looks reasonable to me yeah. Is there some design of how this *should* be in the future, or has the App implementation not been considered for this yet?

PR #983, opened 2022-05-03
Add new development screen used to test embark stories

hugokallstrom:


app/src/engineering/java/com/hedvig/app/DevelopmentActivity.kt
```diff
@@ -40,6 +43,10 @@ class DevelopmentActivity : AppCompatActivity(R.layout.activity_development) {
             it.submitList(
                 listOf(
                     DevelopmentScreenAdapter.DevelopmentScreenItem.Header,
+                    DevelopmentScreenAdapter.DevelopmentScreenItem.Row("Embark Story Tester") {
+                        loadKoinModules(embarkStoryTesterModule)
```
oscarnylander, line 47: This will crash if this row is pressed twice.

I suggest adding this loading to `EmbarkStoryTesterActivity::onCreate`, and pairing it with an `unloadKoinModules` in `EmbarkStoryTesterActivity::onDestroy`.

app/src/engineering/java/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+        model.onStorySelected(null)
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContent {
+            val viewState by model.viewState.collectAsState()
+
+            viewState.selectedStoryName?.let {
```
oscarnylander, line 65: This will potentially trigger multiple times on configuration change, right? 🤔 I think this value might need to be cleared after the 'event' has been consumed, in order to mitigate that issue.

app/src/main/AndroidManifest.xml
```diff
@@ -284,6 +284,7 @@
             android:name=".feature.embark.ui.EmbarkActivity"
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
             android:launchMode="singleTask"
+            android:exported="true"
```
oscarnylander, line 287: Does this activity need to be exported? 🤔

oscarnylander (commented):
Cool!

app/src/engineering/java/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
StylianosGakis: Yeah that is correct, doing such a side effect from inside Compose can very easily trigger more than once if this function recomposes again, and there may be various reasons why it may recompose.
I'd say something like this should work fine:
```kotlin
LaunchedEffect(viewState.selectedStoryName) {
    val selectedStoryName = viewState.selectedStoryName ?: return@LaunchedEffect
    model.onStorySelected(null)
    startActivity(
        EmbarkActivity.newInstance(
            this@EmbarkStoryTesterActivity,
            selectedStoryName,
            selectedStoryName,
        )
    )
}
```
This makes sure it only runs every time `selectedStoryName` changes, and it does nothing when it's null, and it clears the value before it launches the activity so that if you come back it doesn't trigger it again by any chance.

Should work I think, didn't run it myself

app/src/engineering/java/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+}
+
+class EmbarkStoryTesterViewModel(
+    private val getBundlesUseCase: GetBundlesUseCase,
+    private val marketManager: MarketManager,
+    private val authenticationTokenService: AuthenticationTokenService,
+) : ViewModel() {
+
+    data class ViewState(
+        val loading: Boolean = true,
```
StylianosGakis, line 204: I think this is unused right? Can just remove it since it's just dev tooling if we don't need it in this screen

app/src/engineering/java/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
hugokallstrom: Thanks

PR #987, opened 2022-05-05
APP-1651: Make insurance query not show stale data after approving a moving flow

StylianosGakis:
Judging from the logs:
`2022-05-05T08:12:41.157000+00:00 com.hedvig.test.app: 23377:23474 D/a: Cache HIT for operation InsuranceQuery`

The cache is hit when it should actually try to fetch from the network.

Tested this change by using the original code:
- signing a new member
- approved the accident cross sell
- refreshing the page and seeing the cache being hit despite the giraffe query returning both contracts at this point

After this commit, did this:
- unsigned old member and started fresh
- signed a new member
- approved the accident cross sell
- now the page even before refreshing properly fetches the new data that the screens should be showing.

### DIFF
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -656,7 +656,7 @@ val useCaseModule = module {
     single { StartDanishAuthUseCase(get()) }
     single { StartNorwegianAuthUseCase(get()) }
     single { SubscribeToAuthStatusUseCase(get()) }
-    single { SignQuotesUseCase(get()) }
+    single { SignQuotesUseCase(get(), get()) }
     single { LogoutUseCase(get(), get(), get(), get(), get(), get(), get()) }
     single { GetContractsUseCase(get(), get()) }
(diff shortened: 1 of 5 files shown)


hugokallstrom (approved):
LGTM

oscarnylander:
I'm curious - why do we need to do this?

oscarnylander:
So, if I understand it correctly, the issue is that the cache is stale once the moving flow has been completed?

Wouldn't the correct solution then be to issue a cache invalidation when the moving flow has been completed? 🤔

StylianosGakis:
> So, if I understand it correctly, the issue is that the cache is stale once the moving flow has been completed?
> 
> Wouldn't the correct solution then be to issue a cache invalidation when the moving flow has been completed? 🤔

Yeah that is indeed correct, I completely forgot that we can actually clear the cache like that. Let me look into how that could be done instead 👀

StylianosGakis:
Please take a peek at the new commit. I see we had a `...andClearCache` function which didn't actually clear the cache any more. And if it does, it solves this problem of ours actually as I see it. Removed the network_only part of the insurance query too

oscarnylander (approved):
lgtm

PR #988, opened 2022-05-06
Fix/flipped connect payment at sign flag

StylianosGakis:
Context: https://hedviginsurance.slack.com/archives/C033ALASHL6/p1651843005804419?thread_ts=1651830564.242919&cid=C033ALASHL6

Apparently we understood this flag flipped. It was the opposite of what we had in mind on Android. This flips it basically and tries to help with understanding the context with the new flag description.


StylianosGakis:
Wait, did only 1 check run? How did that happen 🤔

oscarnylander:
Weird 🤔 perhaps try sending an empty commit onto it and see if that helps

PR #990, opened 2022-05-09
Simplify GetAddressChangeStoryIdUseCase with `eager {}`

StylianosGakis:
Specifically changes it so that `arrow.core.Either.Right` isn't used for the unhappy path. Uses `either` to short-circuit in the case of reaching the unhappy path and have the logic happen sequentially from top to down.

### DIFF
--- a/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetAddressChangeStoryIdUseCase.kt
+++ b/app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetAddressChangeStoryIdUseCase.kt
@@ -1,10 +1,7 @@
 package com.hedvig.app.feature.home.ui.changeaddress
 
 import arrow.core.Either
-import arrow.core.computations.either
-import arrow.core.computations.ensureNotNull
 import arrow.core.getOrHandle
-import arrow.core.identity
 import com.apollographql.apollo.ApolloClient
 import com.hedvig.android.owldroid.graphql.ActiveContractBundlesQuery
 import com.hedvig.app.feature.embark.QUOTE_CART_ID_KEY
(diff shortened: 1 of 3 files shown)


hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetAddressChangeStoryIdUseCase.kt
```diff
@@ -1,57 +1,60 @@
...
+            )
     }
 
     sealed class SelfChangeEligibilityResult {
         data class Eligible(val embarkStoryId: String) : SelfChangeEligibilityResult()
         object Blocked : SelfChangeEligibilityResult()
         data class Error(val message: String?) : SelfChangeEligibilityResult()
     }
+
+    private sealed interface SelfChangeFailureType {
```
StylianosGakis, line 46: This was basically needed since we don't have Union types and I have to provide something for the `Either` in line 19. I did this to preserve that the invoke function returns a SelfChangeEligibilityResult instead of an Either to not have to change all the call sites as well.
We could even make `invoke` directly respond with an either of <Blocked | Error, Eligible> to not have to have both SelfChangeFailureType and SelfChangeEligibilityResult exist.

app/src/main/java/com/hedvig/app/feature/home/ui/changeaddress/GetAddressChangeStoryIdUseCase.kt
```diff
@@ -1,50 +1,40 @@
 package com.hedvig.app.feature.home.ui.changeaddress
 
 import arrow.core.Either
-import arrow.core.flatMap
+import arrow.core.computations.either
```
oscarnylander, line 4: Don't forget to run `ktlint -F` 🙂

oscarnylander (approved):
lgtm

PR #992, opened 2022-05-09
APP-1184: Disable moving flow

hugokallstrom:
Related to [this](https://hedvig.atlassian.net/jira/software/c/projects/APP/boards/6?modal=detail&selectedIssue=APP-1613&quickFilter=51) ticket. We will use a new mutation for moving flow with quote cart. 

Moving flow can be disabled while that mutation is being implemented, in order to release faster.


app/src/main/java/com/hedvig/app/feature/home/model/HomeItemsBuilder.kt
```diff
@@ -50,10 +49,9 @@ class HomeItemsBuilder(
                 ).toTypedArray()
             )
         )
-        if (featureManager.isFeatureEnabled(Feature.MOVING_FLOW)) {
```
hugokallstrom, line 53: Now we will always show the moving flow button - but take the user to the chat instead of the embark flow.

StylianosGakis (approved):
Nice!

PR #995, opened 2022-05-10
APP-1650: Make ValueStore's Stack also keep history

StylianosGakis:
So the problem was as such:

The way value store worked before this commit, "stage" is a map which on commit gets put into the storedValues, and gets cleared. And on going back, nothing happens to it. Therefore, when going back and forwards whatever might've been in "stage" is gone.
This is why the `withCommittedValues` for example had to save and restore the "stage" manually to preserve the store's state when going forwards and then backwards

The way it works after the commit is that the stage is also a Stack<HashMap<String, String>> which remembers the way it was on each step, allowing the state to restore back to what it was before when going back, including the value not yet committed.

Why?

In bug [1650](https://hedvig.atlassian.net/browse/APP-1650), we had the problem where going into the embark flow, a quoteCartId was queried and put into the valueStore, therefore existing inside the stage. Then in that specific case, a number over 6 was put for co-insured, going into an error screen, at the same time committing the quoteCartId. Then when going back to add another number, since we popped the backstack, the head of the storedValues is deleted and with it the quoteCartId, leading to nothing working after the fact.

I'm not sure if this would break something else right now, if it does we might want to take another approach to solve this problem. On top of that, if you're worried about some other interaction with the value store please feel free to add some tests in ValueStoreImplTest


hugokallstrom (approved):
LGTM

StylianosGakis:
I just thought of one case where this can be problematic.

We're on a step, and we do an action that sets let's say a "isStudent = true" on the store, then we proceed and commit.
Then we rollback, and we have "isStudent = true" on the stage.
This time, we click another option, which sets some other key though, which isn't "isStudent". (If we set isStudent again, to something like "false" it would be fine since we're overriding it)
In this case, both of them will be inside stage and when committed they will be persisted.
But in that path of embark we may not want that key to be set like that.

I am not aware of any cases where we are relying on such behavior, but I can definitely imagine that we do. I think it's best we discard this thought afteral 🤷‍♂️

PR #996, opened 2022-05-11
APP-1661: Generic auth login

hugokallstrom:
Add button in QR code dialog which opens generic auth flow. This will only work for Qasa members, but we should be able to activate it for all members without doing any changes in the clients.

### DIFF
--- a/app/src/main/res/layout/dialog_authenticate.xml
+++ b/app/src/main/res/layout/dialog_authenticate.xml
@@ -50,6 +50,6 @@
         android:id="@+id/login"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:text="Login with email" />
+        android:text="@string/bankid_missing_login.email_button" />
 
 </LinearLayout>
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/dialog_authenticate.xml
```diff
@@ -46,4 +46,10 @@
         android:contentDescription="@null"
         app:srcCompat="@drawable/ic_bank_id" />
 
+    <Button
+        android:id="@+id/login"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Login with email" />
```
oscarnylander, line 53: Needs string resource

oscarnylander (changes requested):

app/src/main/res/layout/dialog_authenticate.xml
hugokallstrom: 🙈

hugokallstrom: Fixed 👍

PR #1004, opened 2022-05-13
Chore: Update a bunch of dependencies

oscarnylander:
To be clear, this is not all dependencies. Still a bit to go. These are mainly
just the easy ones to update at this point. Integration tests seem to be running
smoothly so I'm considering these safe.


build.gradle.kts
```diff
@@ -1,10 +1,11 @@
...
 plugins {
-    id("com.dipien.releaseshub.gradle.plugin") version "2.0.2"
     id("net.rdrei.android.buildtimetracker") version "0.11.0"
     id("com.github.konifar.gradle.unused-resources-remover") version "0.3.3"
-    id("org.gradle.android.cache-fix") version "2.4.4" apply false
-    id("com.osacky.doctor") version "0.7.3"
+    id("org.gradle.android.cache-fix") version "2.5.3" apply false
+    id("com.osacky.doctor") version "0.8.0"
+    id("com.github.ben-manes.versions") version "0.41.0"
+    id("nl.littlerobots.version-catalog-update") version "0.3.1"
```
StylianosGakis, line 8: wbu these, are they something we want to use in the end, or are they left from the testing you were doing?

gradle/libs.versions.toml
```diff
@@ -20,44 +20,44 @@ androidx-compose = "1.1.1"
...
-jsonTest = "20210307"
+insetter = "0.6.1"
+jsonTest = "20220320"
 kaspresso = "1.4.1"
-koin = "3.1.2"
-leakCanary = "2.7"
-lottie = "4.2.0"
+koin = "3.2.0"
+leakCanary = "2.9.1"
+lottie = "5.1.1"
```
StylianosGakis, line 60: Where do we use this actually? I think there's definitely some room of removing some deps here and there if it's stuff we no longer use.

StylianosGakis (approved):

gradle/libs.versions.toml
oscarnylander: We use `lottie` to display the audio recorder-action in the chat. Which to be fair is no longer used, I guess we can probably cut that out 😄

build.gradle.kts
oscarnylander: `com.github.ben-manes.versions` will probably get used, but perhaps only in the scope of a Github Action - TBD. The other one is left in from testing.

gradle/libs.versions.toml
oscarnylander: I'm going to remove `lottie` as a separate PR, I think, in order to manage the scope of this PR.

PR #1009, opened 2022-05-16
Fix runtime Koin crash when QuoteCartId is null for a non-null field

StylianosGakis:
Addresses APP-1608 too.

### DIFF
--- a/app/src/engineering/java/com/hedvig/app/feature/offer/OfferMockActivity.kt
+++ b/app/src/engineering/java/com/hedvig/app/feature/offer/OfferMockActivity.kt
@@ -3,6 +3,7 @@ package com.hedvig.app.feature.offer
 import androidx.lifecycle.lifecycleScope
 import com.hedvig.app.MockActivity
 import com.hedvig.app.feature.offer.MockOfferViewModel.Companion.OfferMockData
+import com.hedvig.app.feature.offer.model.QuoteCartId
 import com.hedvig.app.feature.offer.ui.OfferActivity
 import com.hedvig.app.genericDevelopmentAdapter
 import com.hedvig.app.offerModule
@@ -39,6 +40,8 @@ class OfferMockActivity : MockActivity() {
         }
     )
 
+    private val fakeQuoteCartId = QuoteCartId("") // MockOfferViewModel doesn't actually use this
+
     override fun adapter() = genericDevelopmentAdapter {
         header("Offer Screen")
         clickableItem("Swedish Apartment") {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/hedvig/app/authenticate/LoginStatusService.kt
```diff
@@ -47,9 +46,14 @@ class SharedPreferencesLoginStatusService(
...
-            isViewingOffer -> LoginStatus.InOffer(
-                quoteCartId = sharedPreferences.getString("quoteCartId", null)?.let { QuoteCartId(it) }
-            )
+            isViewingOffer -> {
+                val quoteCartId = sharedPreferences.getString("quoteCartId", null)?.let(::QuoteCartId)
+                if (quoteCartId != null) {
+                    LoginStatus.InOffer(quoteCartId)
+                } else {
+                    LoginStatus.Onboarding
+                }
```
StylianosGakis, line 55: This is the most meaningful change, since in the SplashScreen, if this returns Onboarding it won't try to launch the Offer screen with a null QuoteCartId

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -158,8 +158,12 @@ abstract class EmbarkViewModel(
                 //  meaning that the old values were returned from getList/get.
                 valueStore.withCommittedVersion {
                     val quoteCartId = this.get(QUOTE_CART_ID_KEY)?.let { QuoteCartId(it) }
-                    val selectedContractTypes = nextPassage.getSelectedContractTypes(this)
-                    _events.trySend(Event.Offer(quoteCartId, selectedContractTypes))
+                    if (quoteCartId == null) {
+                        _events.trySend(Event.Error())
+                    } else {
```
StylianosGakis, line 163: If we've reached the point where `key != null` aka `nextPassage?.getOfferKeyOrNull(valueStore)` returned a non-null result, then we must have a QuoteCartId in the store. If not, as I understand it there's a big problem anyway, therefore we can error here. Would love to know if this understanding of mine is wrong.

app/src/main/java/com/hedvig/app/feature/embark/EmbarkViewModel.kt
```diff
@@ -260,11 +264,15 @@ abstract class EmbarkViewModel(
...
-        )
+        val quoteCartId = valueStore.get(QUOTE_CART_ID_KEY)?.let(::QuoteCartId)
+        return if (quoteCartId != null) {
+            Event.Offer(
+                quoteCartId = quoteCartId,
+                selectedContractTypes = emptyList(),
+            )
+        } else {
+            Event.Error()
+        }
```
StylianosGakis, line 275: Again here, if we've reached the point that we call `createOfferEvent` and we don't have a QuoteCartId it seems like quite a bigger problem. Since we're using only that for now, we can't have an offer without an ID. So safe to error here as well afaik.

app/src/main/java/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -70,9 +70,10 @@ class OfferActivity : BaseActivity(R.layout.activity_offer) {
     private lateinit var concatAdapter: ConcatAdapter
     override val screenName = "offer"
 
-    private val quoteCartId: QuoteCartId?
+    private val quoteCartId: QuoteCartId
         get() = intent.getParcelableExtra(QUOTE_CART_ID)
             ?: intent.getStringExtra(QUOTE_CART_ID)?.let { QuoteCartId(it) }
+            ?: error("A quote cart ID must be passed into OfferActivity")
```
StylianosGakis, line 76: Safe since we're passing it as a non-null variable in the `newInstance`

app/src/engineering/java/com/hedvig/app/feature/offer/OfferMockActivity.kt
```diff
@@ -39,49 +40,51 @@ class OfferMockActivity : MockActivity() {
         }
     )
 
+    private val fakeQuoteCartId = QuoteCartId("") // MockOfferViewModel doesn't actually use this
+
     override fun adapter() = genericDevelopmentAdapter {
```
oscarnylander, line 45: I thought we killed this thing 🤔 I think we should just remove it tbh

oscarnylander (approved):
lgtm

app/src/engineering/java/com/hedvig/app/feature/offer/OfferMockActivity.kt
StylianosGakis: Yeah this is just the state of the release version for now which is pre-merge of the hanalytics stuff which came with such refactors. Which is also the reason why I didn't spend more time other than making the compiler happy. When I merge this back into develop, since this class is gone this change will also be gone. I just didn't want to make such a big diff for a hotfix for a thing that we've fixed already anyway :D

PR #1035, opened 2022-05-17
Chore: Migrate plugins to Version Catalog

oscarnylander:
Unfortunately, due to us having some extensions in `buildSrc`,
we're unable to port all of them. Sad times.


build.gradle.kts
```diff
@@ -1,9 +1,9 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id("net.rdrei.android.buildtimetracker") version "0.11.0"
-    id("com.github.konifar.gradle.unused-resources-remover") version "0.3.3"
-    id("org.gradle.android.cache-fix") version "2.5.3" apply false
-    id("com.osacky.doctor") version "0.8.0"
+    alias(libs.plugins.buildTimeTracker)
+    alias(libs.plugins.unusedResourcesRemover)
+    alias(libs.plugins.cacheFix) apply false
+    alias(libs.plugins.doctor)
```
StylianosGakis, line 6: What an improvement, this is amazing!

build.gradle.kts
```diff
@@ -28,12 +28,7 @@ buildscript {
 
     dependencies {
         classpath(libs.android.gradlePlugin)
-        classpath(libs.apollo.gradlePlugin)
-        classpath(libs.crashlytics.gradlePlugin)
-        classpath(libs.googleServices.gradlePlugin)
         classpath(libs.kotlin.gradlePlugin)
-        classpath(libs.licenses.gradlePlugin)
-        classpath(libs.lokalise.gradlePlugin)
```
StylianosGakis, line 36: Aha so they don'tneed to be specified here whatsoever? Does it all happen automagically from the toml file?
Does it have something to do with the thing you added in buildSrc/serttings.gradle.kts or not?

oscarnylander:
We could duplicate what the extension abstracts right now, but it would probably not be an improvement in all honesty.

I'm not aware of any means of doing the same thing, barring perhaps creating our own plugin to do it, but that would be more complex, I think.

build.gradle.kts
StylianosGakis: Right, the same way it would work if we didn't have them on the classpath but on the call site we defined them with the name AND the version. What this classpath was doing was having it not have to call the version right? but with the `alias` thing, it does give it the correct version as well, which is also consistent since it comes from the toml. So no need for the classpath solution.
I think I understand it now, thank you!

StylianosGakis:
> We could duplicate what the extension abstracts right now, but it would probably not be an improvement in all honesty.
> 
> I'm not aware of any means of doing the same thing, barring perhaps creating our own plugin to do it, but that would be more complex, I think.

Yeah that's totally fair. I wasn't thinking of duplicating this in all call sites, but to have another approach which does the same without the limitations of the buildSrc. If there is no such way, or at least not an obvious one we can skip it for now and potentially revisit this if someone happens to learn more about this.

PR #1042, opened 2022-05-23
APP-1730: Remove unused query field to fix car mutation

hugokallstrom:
This was never used, and gave an error if not present in PCMS. Talked to Sam and they dont use it on iOS either.


StylianosGakis (approved):
We were just not using this at all I guess then? lgtm!

PR #1047, opened 2022-05-27
APP-1741: Hidden embark links

hugokallstrom:


StylianosGakis (approved):
Not sure what "Hidden" means for embark stories, but lgtm with my understanding :D

hugokallstrom:
> Not sure what "Hidden" means for embark stories, but lgtm with my understanding :D

It means that those passages should not be included when calculating the depth of the story. This is because we want the progress bar to be at 100% when going to the offer, even though we have an error passage after.

PR #1066, opened 2022-06-02
APP-1290: Allow submission using IME for Insurely form

oscarnylander:


app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -77,6 +79,7 @@ class RetrievePriceInfoActivity : BaseActivity() {
                     }
                 ) {
                     RetrievePriceScreen(
+                        modifier = Modifier.padding(it),
```
StylianosGakis, line 82: nit: can we name `it` `paddingValues`? Quite unclear what `it` is here otherwise unless you see that there is a `Scaffold` above and remember that it provides paddingValues as a parameter to this lambda.

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
```diff
@@ -107,6 +110,7 @@ class RetrievePriceInfoActivity : BaseActivity() {
 
 @Composable
 fun RetrievePriceScreen(
+    modifier: Modifier = Modifier,
```
StylianosGakis, line 113: Can we make modifier be the first optional parameter in the composable as the compose code style guidelines suggest? https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter First sentence in this section.
This would also mean changing onContinue to come first, then modifier, then the viewModel which can be omitted. Or maybe removing the default `= viewModel()` since we never call this function without providing it a viewModel as we probably should anyway, and then putting modifier last.

StylianosGakis (approved):
A couple of details, otherwise lgtm!

app/src/main/java/com/hedvig/app/feature/embark/passages/externalinsurer/retrieveprice/RetrievePriceActivity.kt
oscarnylander: Yes

PR #1068, opened 2022-06-03
Remove apollo-coroutines-support and apollo-android-support

StylianosGakis:
Will resolve the conflicts when we need to merge. Hard to keep them up to date all the time since things get merged to develop all the time.


apollo/README.md
```diff
@@ -0,0 +1,7 @@
+To download and generate all models do:
+
+./gradlew :apollo:downloadGiraffeApolloSchemaFromIntrospection
+
+Search for a text: "When we've deleted an" and then delete the escaped comment block: `/"""` to make the generation work
```
StylianosGakis, line 5: This is actually fixed in a latest version, it parses code blocks inside comments just fine

oscarnylander: Should we remove this comment, in that case? 😬

StylianosGakis: Yeah I just remembered that I wrote this back then. I will fix it but I wanted to keep the numbers 1 to 14 rebased on top of each other to make merging a no-brainer. It was faster for me to write this comment to remind myself to fix it as well :D 
Also good to note about this https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1654242074977499.

oscarnylander (approved):
Beautiful!

PR #1071, opened 2022-06-03
Apollo3 4 Custom scalar adapters

StylianosGakis:


apollo/build.gradle.kts
```diff
@@ -31,14 +31,14 @@ apollo {
...
         mapScalarToKotlinString("URL")
-        mapScalar("LocalDate", "java.time.LocalDate")
-        mapScalarToUpload("Upload")
-        mapScalar("PaymentMethodsResponse", "com.adyen.checkout.components.model.PaymentMethodsApiResponse")
         mapScalarToKotlinString("CheckoutPaymentsAction")
         mapScalarToKotlinString("CheckoutPaymentAction")
-        mapScalar("JSONString", "org.json.JSONObject")
-        mapScalar("Instant", "java.time.Instant")
+        mapScalarToUpload("Upload")
+        mapScalar("LocalDate", "java.time.LocalDate", "com.apollographql.apollo3.adapter.JavaLocalDateTimeAdapter")
```
oscarnylander, line 37: Is this one as promiscuous as the one we implemented? If not, we may run into problems here.

app/src/main/java/com/hedvig/app/util/apollo/PaymentMethodsApiResponseAdapter.kt
```diff
@@ -6,6 +6,8 @@ import com.apollographql.apollo3.api.CustomTypeValue
 import com.hedvig.app.util.toJsonObject
 import org.json.JSONObject
 
+// TODO: Not sure why we're doing the LinkedHashMap check and how to migrate this to apollo3 Adapter. We can use the
```
oscarnylander, line 9: Good question - not sure. But given that Apollo 3 comes with different type guarantees, I guess this should be just fine to convert to the new API.

oscarnylander (approved):
lgtm

apollo/build.gradle.kts
StylianosGakis: Right, I've been meaning to ask you about this. I wasn't quite sure what `promiscuous` meant in this context. I could re-create the same functionality in a new adapter though, was the idea that we were not interested at all in all the content of the string after the YYYY-MM-DD ?
So reading a string like `2011-01-02T05:19` as just `2011-01-02`.
If we're relying on that behavior somewhere somehow it should be easy to just bring that custom adapter back in.

app/src/main/java/com/hedvig/app/util/apollo/PaymentMethodsApiResponseAdapter.kt
StylianosGakis: Alright, I'll keep this in mind and try with and without migrating this and see how the app functions in cases where it parses this.
Also as I see that Hugo was the one who touched this part last I will consult him too to see if there was something we need to be wary of in this process.

apollo/build.gradle.kts
oscarnylander: image

In this case, it definitely refers to the second point 😉 

It basically means that it accepts a very loose definition of a `LocalDate` - even if there's time information or something else, it will just truncate that into a `LocalDate`.

StylianosGakis: Hahaha til. And I was *really* wondering how the hell an adapter can be promiscuous 😂

Aha is the use case basically to not fail to parse this exact case?
https://pl.kotl.in/wAn0op3R6
I can totally see it then. I will make the new adapter! Thanks for pointing it out! What an eagle eye spotting that 👀

oscarnylander: Moreso like this: https://pl.kotl.in/otOfxjDkh

StylianosGakis: Yeah I was referring to how this breaks if there is also hour and minute etc information (which there is coming from the backend as I understand), that's what I tried to convey with my example as well :D

PR #1074, opened 2022-06-03
Apollo3 7 Optional values

StylianosGakis:
The magic happens in the apollo {} config in:
More context: https://www.apollographql.com/docs/kotlin/migration/3.0#non-optional-variables-generation


oscarnylander (approved):
lgtm

PR #1076, opened 2022-06-03
Apollo3 9 Migrate SunsettingInterceptor

StylianosGakis:


app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -294,9 +296,15 @@ val applicationModule = module {
         }
         builder.build()
     }
-    single { SunsettingInterceptor.Factory(get()) } bind ApolloInterceptorFactory::class
+    single<SunsettingInterceptor> { SunsettingInterceptor(get()) } bind ApolloInterceptor::class
     single<ApolloClient> {
         val builder: ApolloClient.Builder = ApolloClient.Builder()
+        if (isDebug()) {
+            builder.logCacheMisses { log: String ->
```
oscarnylander, line 303: Do we want this? 😬

oscarnylander (approved):
lgtm

app/src/main/java/com/hedvig/app/ApplicationModule.kt
oscarnylander: I see.

I think the cache miss logs kind of are just white noise though, so if we're not getting any other logs from this, I vote that we cut it 😬

PR #1077, opened 2022-06-03
Apollo3 10 Migrate to new way of reconnecting subscriptions

StylianosGakis:


app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -306,18 +304,14 @@ val applicationModule = module {
         }
         builder
             .serverUrl(get<HedvigApplication>().graphqlUrl)
+            .webSocketServerUrl(get<HedvigApplication>().graphqlSubscriptionUrl)
             .okHttpClient(get<OkHttpClient>())
```
oscarnylander, line 308: I'm going to assume that the connection params come in through another PR, otherwise this would break stuff. 😰

oscarnylander (approved):
lgtm

app/src/main/java/com/hedvig/app/ApplicationModule.kt
StylianosGakis: Are you referring to [this](https://github.com/HedvigInsurance/android/blob/d80a02973ab134b9f3e9ef3e69beb3bec2cb0983/app/src/main/java/com/hedvig/app/ApplicationModule.kt#L307-L311) in our current 2.x implementation?

oscarnylander: I am indeed

StylianosGakis: I'll have to double check that yeah. I think what I had in mind is that it would reuse the authentication interceptor from here by default, but I could definitely be wrong. And there's a chance I hadn't tested this yet. Thank you for pointing it out!

StylianosGakis: It's now part of the migration docs. It did indeed need an extra configuration for connection parameters
https://github.com/apollographql/apollo-kotlin/pull/4188/files

PR #1078, opened 2022-06-03
Apollo3 11 Migrate FileUpload

StylianosGakis:


app/src/main/java/com/hedvig/app/feature/chat/data/ChatRepository.kt
```diff
@@ -134,7 +135,10 @@ class ChatRepository(
         mimeType: String,
     ): ApolloResponse<UploadFileMutation.Data> {
         val uploadFileMutation = UploadFileMutation(
-            file = FileUpload(mimeType, path)
+            file = DefaultUpload.Builder()
```
oscarnylander, line 138: This new interface is really good, and we should probably stop doing this by `path` in the future, as we have to do some really arcane stuff to figure out the file paths.

No need to do it in the scope of this PR, though.

oscarnylander (approved):

app/src/main/java/com/hedvig/app/feature/chat/data/ChatRepository.kt
StylianosGakis: Yeah that's a good idea. We should keep track of potential other stuff like this that we'd want to do after the big merge is done.

PR #1079, opened 2022-06-03
apollo3-12-typename_and_import_fixes

StylianosGakis:


app/src/main/java/com/hedvig/app/util/apollo/SafeQuery.kt
```diff
@@ -50,6 +50,7 @@ fun <D : Subscription.Data> ApolloCall<D>.safeSubscription(): Flow<QueryResult<D
 
 fun <D : Query.Data> ApolloCall<D>.safeWatch(): Flow<QueryResult<D>> {
     return try {
+        watch(null)
```
oscarnylander, line 53: What's going on here?

app/src/main/java/com/hedvig/app/feature/marketpicker/LanguageRepository.kt
```diff
@@ -1,29 +1,25 @@
...
+import com.hedvig.android.owldroid.graphql.type.Locale
 import e
 import i
 
 class LanguageRepository(
     private val apolloClient: ApolloClient,
 ) {
 
-    fun uploadLanguage(acceptLanguage: String, locale: Locale) {
+    suspend fun uploadLanguage(acceptLanguage: String, locale: Locale) {
```
oscarnylander, line 13: This mutation currently discards any errors (because it's really not that important), but the new implementation doesn't seem to mirror this fact (unless `execute` has guarantees that it doesn't throw exceptions?)

oscarnylander (approved):
lgtm

app/src/main/java/com/hedvig/app/util/apollo/SafeQuery.kt
StylianosGakis: A mistake is going on here 🙊 I thought I removed it but alas I had not.
I was playing around with the stuff I was looking through in [this](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1652857651123899?thread_ts=1652794902.304599&cid=C01A6KM1SBZ) discussion and forgot to remove it :D

PR #1093, opened 2022-06-17
Send `receivedQuotes` analytics event

StylianosGakis:
The place where we know we're getting a quote is in the `QuoteCartQuery` query which is done inside OfferRepository. However, we're calling that function a *lot*, so this event triggers way too much. One thing I did to at least not trigger this event on a loop, was make the SwedishBankId dialog use a new query which only asks for the Checkout status.

A lot of it has to do with the fact that the `offerState` inside `OfferViewModelImpl` is a cold flow, and whenever we're doing .first() on it it starts a new flow which calls `ObserveOfferStateUseCase::observeOfferState` which calls `queryAndEmitOffer` which does the query again.

I think when we needed to add the varianted offers the code evolved in a way that is a bit hard to reason about here. I think revising this a bit and changing the OfferRepository to expose a simple interface of `getOffer(quoteCartId)` which gets an already fetched offer if present, otherwise fetches the offer before returning and a `clearOffer()` will do wonders in avoiding calling `queryAndEmitOffer` in 8 places when all those places don't need a new offer, but to look at the existing one if it already exists.

There's not problem of staleness afaik here since all data should be the same given the same QuoteCartId. The only mutable state is the CheckoutState, which is why I introduced this new query which returns just that. Since we want to know about that more often, it also saves some networking time to ask just for that instead of the entire query just to learn about the Checkout state.

With all that said, this does add the event here, and the only problem is that it will still trigger the even more than once. Not every 500 milliseconds while waiting for BankID to sign, but still not only once as it should. The rest can be part of another PR I guess.


app/src/main/java/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignDialog.kt
```diff
@@ -58,61 +56,45 @@ class SwedishBankIdSignDialog : DialogFragment() {
...
-        model
-            .events
-            .flowWithLifecycle(viewLifecycle)
-            .onEach { event ->
-                when (event) {
-                    SwedishBankIdSignViewModel.Event.StartBankID -> {
+        var viewState: BankIdSignViewState by mutableStateOf(viewModel.viewState.value)
+        viewLifecycleScope.launch {
+            viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                viewModel.viewState.collect { state ->
```
StylianosGakis, line 63: This preserves the behavior of only querying the CheckoutStatus and reacting to the changes while the app is on the foreground.

app/src/main/java/com/hedvig/app/feature/offer/OfferRepository.kt
```diff
@@ -45,10 +47,16 @@ class OfferRepository(
...
-        ensureNotNull(result.quoteCart.fragments.quoteCartFragment.bundle) {
+        val quoteCartFragment = result.quoteCart.fragments.quoteCartFragment
+        val bundle = quoteCartFragment.bundle
+        ensureNotNull(bundle) {
             ErrorMessage("No quotes in offer, please try again")
         }
+        val receivedQuoteIds = bundle.possibleVariations
+            .flatMap { variation -> variation.bundle.fragments.quoteBundleFragment.quotes }
+            .map { quote -> quote.id }
+        hAnalytics.receivedQuotes(receivedQuoteIds)
```
hugokallstrom, line 58: Could we have logged this higher up, in the viewmodel perhaps? To avoid sending multiple analytics events.

hugokallstrom (approved):
LGTM, love the tests

app/src/main/java/com/hedvig/app/feature/offer/OfferRepository.kt
StylianosGakis: Hmm I've thought of that as well, but that leaves us with only sending the event sometimes when we're doing this network request, which means we have to manually think each time we request for the quotes if we want to log it this time or not. And depending on how this evolves it may mean that we have to then add the event in two or more places and in general wonder if we're under-sending it all the time. So it made sense to me to simply send the event when we're doing the network request. And this makes it so that for the analytics team to get a good grip of what is happening to filter out the absolutely equivalent events (same quote cart ID and same exact list of quote IDs in it). If we revisit this in the future we can look into improving it for sure. If this feels like a blocker to you I can definitely try and re-iterate on it a bit.

hugokallstrom: I see, then this implementation is reasonable.

PR #1096, opened 2022-06-21
APP-1548: Add analytics events for audio recorder

hugokallstrom:


app/src/main/java/com/hedvig/app/feature/embark/passages/audiorecorder/AudioRecorderFragment.kt
```diff
@@ -55,16 +57,35 @@ class AudioRecorderFragment : Fragment() {
...
+                        logWithStoryAndStore(hAnalytics::embarkAudioRecordingRetry)
+                    },
+                    play = {
+                        model.play()
+                        logWithStoryAndStore(hAnalytics::embarkAudioRecordingPlayback)
+                    },
+                    pause = {
+                        model.pause()
+                        logWithStoryAndStore(hAnalytics::embarkAudioRecordingStopped)
+                    },
```
StylianosGakis, line 79: By putting them here, there's only the slight "problem" that if any of these are called at a state that is "illegal" the event would get triggered but the viewmodel function may reject that action, like doing "play" but the audio player isn't ready yet.
Probably the vast majority of the times the event will be correct though, so probably this is fine.

StylianosGakis (approved):
LGTM

PR #1101, opened 2022-06-23
Migrate PaymentMethodsApiResponseAdapter and move adapters inside the apollo module

StylianosGakis:
To make this change a core module was introduced to serve as a place for all common functionality that *all* modules need to share. Inspiration taken for this module extraction from Google samples of multi-module apps like [tivi](https://github.com/chrisbanes/tivi) or [NiA](https://github.com/android/nowinandroid)

Also serves as a first step to further modularizing our app if we feel like doing so.


hugokallstrom (approved):
LGTM!

PR #1109, opened 2022-07-04
Remove buildSrc and use convention plugins

StylianosGakis:
Move our custom JSONObject handling code in there to be shared among the
app module and the apollo module.Heavily inspired from NiA https://github.com/android/nowinandroid
Converts our custom configuration inside buildSrc into convention
plugins ready to be consumed from an app module or a library module.

This enables making new modules easier by applying the appropriate
convention plugin and having most of the common configuration applied
automatically, leaving mostly the module-specific things left to be
implemented on that module's build.gradle.

build-logic/README.md contains some more explanation on this process.

In the process also removed the explicit dependency to kotlin stdlib for
jdk7 because ever since Kotlin 1.4 a stdlib is bundled in the
org.jetbrains.kotlin.android gradle plugin


hugokallstrom (approved):
Really cool, GJ!

PR #1114, opened 2022-07-05
Change into tabs of size 2

StylianosGakis:
Inspiration coming from other codebases like apollo-kotlin and how they
look like with 2 spaces which feels quite comfortable for a language
like Kotlin which indents quite a lot with the multiple lambdas here and
there. A fact only becoming more prominent with the use of Compose.

Also from code style provided for Java and Kotlin by Square
https://github.com/square/java-code-styles/blob/main/configs/codestyles/Square.xml

This is a trial and we might want to go back if we feel like we don't
like how it feels like.


hugokallstrom (commented):
LGTM!

PR #1117, opened 2022-07-08
APP-872: Composable error screen

hugokallstrom:
Removed xml error view and replaced with composable view.


app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearFragment.kt
```diff
@@ -96,12 +93,12 @@ class KeyGearFragment : Fragment(R.layout.fragment_key_gear) {
...
-              errorContainer.root.isVisible = true
-              contentContainer.isVisible = false
+              error.show()
+              contentContainer.remove()
             }
             is KeyGearViewModel.ViewState.Success -> {
-              errorContainer.root.isVisible = false
-              contentContainer.isVisible = true
+              error.hide()
+              contentContainer.remove()
```
StylianosGakis, line 101: Probably a `show()` here right :D
And anyway I really wonder why we even still have all the key-gear code in the codebase, I think it's pretty clear at this point we're not gonna revive it, and if we do it'd receive a complete re-design.

app/src/main/java/com/hedvig/app/feature/onboarding/ui/OnboardingAdapter.kt
```diff
@@ -30,21 +33,21 @@ import com.hedvig.app.util.invalidData
 
 class OnboardingAdapter(
   val viewModel: ChoosePlanViewModel,
-  val marketProvider: MarketManager,
+  private val marketProvider: MarketManager,
 ) :
```
StylianosGakis, line 37: Nice catch with the private, I guess the viweModel could also be private right?

app/src/main/java/com/hedvig/app/ui/compose/composables/screens/GenericErrorScreen.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import com.hedvig.app.R
+import com.hedvig.app.ui.compose.theme.HedvigTheme
+
+class GenericErrorScreenView @JvmOverloads constructor(
+  context: Context,
+  attrs: AttributeSet? = null,
+  defStyle: Int = 0
+) : AbstractComposeView(context, attrs, defStyle) {
```
StylianosGakis, line 23: ktlint is probably gonna cry about this missing trailing comma after `defStyle: Int = 0`  :D

app/src/main/java/com/hedvig/app/ui/compose/composables/screens/GenericInfoScreen.kt
```diff
@@ -0,0 +1,97 @@
...
+    Column(modifier = Modifier.padding(16.dp)) {
+      Icon(
+        painter = painterResource(id = icon),
+        contentDescription = null,
+        Modifier.padding(top = 80.dp).height(25.dp).width(25.dp),
+      )
+      Text(
+        text = title,
+        style = HedvigTypography.h5,
+        modifier = Modifier.padding(top = 16.dp),
```
StylianosGakis, line 52: This again is not *that* important and mostly a stylistic choice, but it could be that instead of having each Text specify its own custom padding to fit the design, this could instead be `Spacer` objects in-between those texts like `Spacer(Modifier.height(16.dp))`
The reason I usually say it is so that we can have an easier mental model that "here is a Column, it has a text, some space, a text, more space, and a text" instead of "here's a column with 3 texts, and let's look into their modifiers to figure out their placement rules".
Also comes with the exctra benefit that if some of the text moves around or gets deleted in some future design, the modifier that specified the spacings isn't gone with it, but it still lives in the Spacer. And then if you get a Spacer next to another Spacer you *know* it's time to either remove one of them or merge them. With the modifiers that's not that obvious.

app/src/main/java/com/hedvig/app/ui/compose/composables/screens/GenericInfoScreen.kt
```diff
@@ -0,0 +1,97 @@
...
+) {
+  Surface(
+    modifier = modifier.fillMaxSize(),
+    color = MaterialTheme.colors.onPrimary,
+  ) {
+    Column(modifier = Modifier.padding(16.dp)) {
+      Icon(
+        painter = painterResource(id = icon),
+        contentDescription = null,
+        Modifier.padding(top = 80.dp).height(25.dp).width(25.dp),
```
StylianosGakis, line 47: nit: height/width could be 1 modifier of `.size(25.dp)`

core-common/src/test/kotlin/com/hedvig/android/core/JSONObjectTest.kt
```diff
@@ -8,115 +8,115 @@ import org.json.JSONObject
 import org.junit.Test
 
 class JSONObjectTest {
```
StylianosGakis, line 10: Oops did I fail to format this somehow? That's interesting, gotta check if there's some issue with which files ktlint is looking at to format.

StylianosGakis:
Oh also it seems that a bunch of ui tests had been using some of the references to the retry view which now doesn't exist anymore. I guess they need some touches there as well.

core-common/src/test/kotlin/com/hedvig/android/core/JSONObjectTest.kt
StylianosGakis: Addressed here https://github.com/HedvigInsurance/android/pull/1118

app/src/main/java/com/hedvig/app/feature/keygear/ui/tab/KeyGearFragment.kt
hugokallstrom: I think we have it enabled for some members still. But lets check with Sara

app/src/main/java/com/hedvig/app/ui/compose/composables/screens/GenericInfoScreen.kt
hugokallstrom: Yes that is a very good point.

PR #1119, opened 2022-07-08
APP-1818: Link to web onboarding

hugokallstrom:
Also spent some time refactoring marketing view model. So nice to have unit tests in place, made it very easy to refactor.


app/src/main/java/com/hedvig/app/feature/marketing/data/GetInitialMarketPickerValuesUseCase.kt
```diff
@@ -39,6 +39,6 @@ class GetInitialMarketPickerValuesUseCase(
       .fold({ null }, ::identity)
       ?: Market.SE
 
-    return Pair(marketOrDefault, null)
+    return Pair(marketOrDefault, marketOrDefault.defaultLanguage())
```
StylianosGakis, line 42: I wonder if we relied on this null behavior somewhere in a way that was meaningful. If not I think this leaves this function never returning null for language, so the signature can change to be a non-null Language then

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -167,37 +119,17 @@ class MarketingViewModel(
...
-) : ViewState
-
-sealed interface MarketPicked : ViewState {
-  object Loading : MarketPicked
-
-  data class Loaded(
-    val selectedMarket: Market,
-    val loginMethod: LoginMethod,
-  ) : MarketPicked
+data class ViewState(
```
StylianosGakis, line 124: Since this class is outside of the namespace of `MarketingViewModel` (which I agree with, I like them as top level) I think it's a good idea to name it `MarketingViewState`|`MarketingUiState` instead. Otherwise this will pollute the namespace whenever writing ViewState anywhere in the project this will be popping up as a suggestion.

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
```diff
@@ -129,8 +133,8 @@ fun PickMarketScreen(
             }
           },
           header = stringResource(R.string.market_language_screen_market_label),
-          label = data.market?.label?.let { stringResource(it) },
-          enabled = true,
+          label = selectedMarket?.label?.let { stringResource(it) },
+          enabled = enabled,
```
StylianosGakis, line 137: This was always true before, now it depends on if market and lanaguge are both not null as I think that's where this enabled comes from.
From the PR files maybe it's a bit hard for me to grasp how this entire flow goes, but is this expected behavior? I guess it was always true before to let the user pick a market if it ended up being null somehow (not sure how)

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
```diff
@@ -176,14 +180,14 @@ fun BottomSheetHandle(modifier: Modifier = Modifier) {
...
       ),
-
   )
 }
 
 @Composable
 fun PickMarketSheetContent(
   onSelectMarket: (Market) -> Unit,
-  data: PickMarket,
+  selectedMarket: Market?,
```
StylianosGakis, line 189: I like this thing you're doing with composables only receiving the thing they need instead of a bigger object which may have things the composable doesn't need. I think in general it is a good idea, expect for the "top level" if that makes sense composables which do need the whole object.

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -167,37 +119,17 @@ class MarketingViewModel(
...
+  val market: Market? = null,
+  val language: Language? = null,
+  val availableMarkets: List<Market> = emptyList(),
+  val selectedMarket: Market? = null,
+  val loginMethod: LoginMethod? = null,
+  val isLoading: Boolean = true,
+) {
+  fun canSetMarketAndLanguage() = market != null && language != null
```
StylianosGakis, line 132: I love the win we get here with this basically being a computed property instead of manually having to check for it when changing the object. Super nice!

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -56,108 +55,61 @@ class MarketingViewModel(
...
-      throw IllegalStateException("Can't submit, PickMarket is not valid")
-    }
-    val market = currentState.market ?: throw IllegalStateException("Can't submit, market is null")
-    val language = currentState.language ?: throw IllegalStateException("Can't submit, language is null")
-    updateMarketPickerState { it.copy(isLoading = true) }
     viewModelScope.launch {
-      val result = submitMarketAndLanguagePreferencesUseCase
-        .invoke(language, market)
-      when (result) {
+      val market = _state.value.market ?: throw IllegalArgumentException("Market null")
```
StylianosGakis, line 90: nit: I like replacing this throw.. with `?: error("Market null")` which is more concise and actually throws the semantically correct exception of `IllegalStateException` which is the case the majority of the times. IllegalArgument would fit more for arguments passed into this function, but in the current case it's the state which is in a "non correct" state for this function to be called.
Same for the line below of course.

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
```diff
@@ -56,108 +55,61 @@ class MarketingViewModel(
...
           updateApplicationLanguageUseCase.invoke(market, language)
           featureManager.invalidateExperiments()
-          _state.value = MarketPicked.Loading
-          loadMarketPicked()
+          _state.update {
+            it.copy(
+              selectedMarket = market,
+              isLoading = false,
+            )
+          }
```
StylianosGakis, line 105: This was setting the state to loading, then called`featureManager.getLoginMethod()` inside the `loadMarketPicked` each time and then set that as a result. 
Now we're only changing the loginMethod at init{} as I see it.
However I see that here when we've submitted a new market and langauge we're doing a `featureManager.invalidateExperiments()` which means that this may in fact return a new value from that point on.
Unless I am missing something we'd need to do the same thing as before again on the Either.Right case. What do you think?

app/src/main/java/com/hedvig/app/util/featureflags/loginmethod/LoginMethodProvider.kt
```diff
@@ -3,5 +3,5 @@ package com.hedvig.app.util.featureflags.loginmethod
 import com.hedvig.hanalytics.LoginMethod
 
 interface LoginMethodProvider {
-  suspend fun getLoginMethod(): LoginMethod
+  suspend fun getLoginMethod(): LoginMethod?
```
StylianosGakis, line 6: Why was thing change necessary? Just so I understand what has changed (or what was wrong before)

app/src/test/java/com/hedvig/app/feature/marketing/ui/MarketingViewModelTest.kt
```diff
@@ -33,8 +30,8 @@ class MarketingViewModelTest {
   @get:Rule
   val mainCoroutineRule = MainCoroutineRule()
 
-  private fun sut(
-    marketManager: MarketManager = mockk(relaxed = true),
+  private fun createMarketingViewModel(
```
StylianosGakis, line 33: Agree with steering away from `sut`. In hindsight I don't think I like it either, despite the fact that I wrote that in the first place.

StylianosGakis (approved):
Some super nice changes in there, great job. I had a couple concerns/thoughts in a couple of places please do help me understand if I am wrong in any of them.
Also don't forget `./gradlew ktlintCheck` | `./gradlew ktlintFormat` ^_^

app/src/main/java/com/hedvig/app/feature/marketing/pickmarket/PickMarketScreen.kt
hugokallstrom: Good catch, error on my part. This should always be enabled

app/src/main/java/com/hedvig/app/feature/marketing/MarketingViewModel.kt
hugokallstrom: Good catch. Fixing

app/src/main/java/com/hedvig/app/util/featureflags/loginmethod/LoginMethodProvider.kt
hugokallstrom: Its because `DevLoginMethodProvider` checks `marketManager`, and we have cases were the market could be null (eg on first install). This was not the case before though, perhaps we initialised the market somewhere else. I defaulted to SE login method for the dev provider now instead of throwing, so we can still return non nullable login method. Good catch again :)

PR #1132, opened 2022-07-18
Add generateProjectDependencyGraph gradle task

StylianosGakis:
Adds a task which generates a PNG of our module structure.

Can be used to aid our efforts of modularization and making sure we got a clear view of the bigger picture.

Can be compared with bigger graphs like the one [here](https://miro.medium.com/max/1400/1*fg571ukn-dDaEV7FHAyd4A.png) from [this post](https://medium.com/androiddevelopers/jetpack-compose-before-and-after-8b43ba0b7d4f) to see how it could look like in the future


hugokallstrom (approved):
Nice!

PR #1135, opened 2022-07-20
Change auth subscription logic to not error out the screen on a failed response

StylianosGakis:
Move auth subscription logic into SubscribeToAuthSuccessUseCase and turn
it into a flow which keeps the graphql subscription alive to check
on the auth status while also restarting said subscription when the auth
fails or there is some sort of network error.
This is to ensure that no matter what we are informed about a successful
authentication.

Context on the bug is here https://hedvig.atlassian.net/browse/APP-1827
My idea of fixing it is that if the subscription gives FAILED at any point, we don't need to error out the screen, but we can still keep the subscription alive because after failing we can still go on with the flow, give the otp and password and *then* have the subscription return a successful auth status.


app/src/main/java/com/hedvig/app/authenticate/LoginDialog.kt
```diff
@@ -29,21 +29,23 @@ class LoginDialog : AuthenticateDialog() {
...
-      dismissAllowingStateLoss()
-      startActivity(
-        LoggedInActivity.newInstance(requireContext(), withoutHistory = true),
-      )
+  private fun bindNewStatus(state: AuthState) {
+    when (state) {
+      AuthState.INITIATED -> binding.authTitle.setText(R.string.BANK_ID_AUTH_TITLE_INITIATED)
+      AuthState.IN_PROGRESS -> binding.authTitle.setText(R.string.BANK_ID_LOG_IN_TITLE_IN_PROGRESS)
+      is AuthState.UNKNOWN__,
+      AuthState.FAILED,
```
hugokallstrom, line 37: Are we returning failed state anywhere now? Or will this status always be successful? Do we want some kind of time out in order to not get an infinite loading screen if we never get success?

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/authenticate/LoginDialog.kt
StylianosGakis: That's the issue though, we *are* getting a failed state but way too early as seen in the video in Jira. There's basically no case where we know that we got a Failed state and it's actually a failed state that we do want to show as an error from this subscription. I could not think of a better approach which doesn't also fail on the normal flow where the user simply might take too much time to fill in their data and backend randomly returns a failed state.

StylianosGakis: The moment you see there that we get an error screen is when backend returned failed. But as seen in that video, the user didn't even do anything, just went inside the screen to fill in their data and then got kicked to an error screen.

hugokallstrom: I see, so we are getting a failed state - but we cant trust it? Is this something that should be fixed on the backend instead?

StylianosGakis: Could be yeah, I just couldn't find myself where this logic exists to try and see what may be going wrong. Let me try and think bout this a bit more actually.

PR #1137, opened 2022-07-20
Introduce core-ui and core-design-system modules

StylianosGakis:
core-ui exists as a common place to put composables that need to cross
the boundaries of modules that mustn't depend on each other.

core-design-system serves as the module which contains everything
regarding the design system of the app.

More info for both exist in their respective README files


hugokallstrom (approved):
Super! If I understand correctly, core-design-system should be the basic building blocks like buttons, dialogs etc. core-ui can be full screens or part of screens?

StylianosGakis:
Yes that's exactly it for core-design-system, for stuff that are standard in our app but not vanilla material design. The buttons I put in there are the most obvious choice for example.
For core-ui it should serve as a place to put some components that we do want to be re-used across different parts of the app but are not part of the design-system. I don't envision those to be full screens necessarily.
For a scenario, let's say we have a feature module called feature-x and in there we have some specialized card composable. Then later we get another feature in the app, called feature-y which also needs to have access to that composable. Feature modules can't depend on each other and therefore can't access each others code, so that composable would have to be moved to a place where both of those can see, which would be exactly core-ui.

For a practical example, we can look at [NowInAndroid](https://github.com/android/nowinandroid) app and their core-ui module where for example the [NewsResourceCard](https://github.com/android/nowinandroid/blob/main/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt) lives in there because it's a component which is used in three different screens that exist in different feature modules.

hugokallstrom:
> Yes that's exactly it for core-design-system, for stuff that are standard in our app but not vanilla material design. The buttons I put in there are the most obvious choice for example. For core-ui it should serve as a place to put some components that we do want to be re-used across different parts of the app but are not part of the design-system. I don't envision those to be full screens necessarily. For a scenario, let's say we have a feature module called feature-x and in there we have some specialized card composable. Then later we get another feature in the app, called feature-y which also needs to have access to that composable. Feature modules can't depend on each other and therefore can't access each others code, so that composable would have to be moved to a place where both of those can see, which would be exactly core-ui.
> 
> For a practical example, we can look at [NowInAndroid](https://github.com/android/nowinandroid) app and their core-ui module where for example the [NewsResourceCard](https://github.com/android/nowinandroid/blob/main/core-ui/src/main/java/com/google/samples/apps/nowinandroid/core/ui/NewsResourceCard.kt) lives in there because it's a component which is used in three different screens that exist in different feature modules.

I understand, makes sense!

PR #1138, opened 2022-07-21
APP-1835: Progress composable

hugokallstrom:


app/src/main/java/com/hedvig/app/ui/compose/composables/BlurredFullScreenProgressOverlay.kt
```diff
@@ -0,0 +1,155 @@
...
+      drawCircle(progressBlue, radius = radius.toPx())
+    }
+
+    translate(
+      left = min(translationX2.toPx(), screenWidth),
+      top = min(translationY2.toPx(), screenHeight),
+    ) {
+      drawCircle(progressYellow, radius = radius2.toPx())
+    }
+  }
```
StylianosGakis, line 142: Instead of having to do a min() here to make sure it's within bounds, maybe we could make sure that the translation values don't go outside in the first place. That would also help in the case that let's say half of the value is outside of the screen, and the other half is inside, for the entire process when the value is outside of the screen's bounds it would look like it's not animating since it'd be forced to be inside the bounds.
Also the min only works if we're going outside the bounds on the right side, what if the value would go to the minus, then the min wouldn't do anything right?

I may be missing smth with the numbers here of course, but these are just some thoughts from looking at the code.

StylianosGakis (approved):
Ran this on my phone as well and damn it's looking nice 😅 I love how the blur looks like, especially when the two circles go over each other. Eye candy 🍬

StylianosGakis:
Oh and and run a `./gradlew ktlintFormat` there must be smth off somewhere

app/src/main/java/com/hedvig/app/ui/compose/composables/BlurredFullScreenProgressOverlay.kt
```diff
@@ -0,0 +1,155 @@
...
+}
+
+@Composable
+private fun AnimatedCircles() {
+  val transition = rememberInfiniteTransition()
+  val configuration = LocalConfiguration.current
+
+  val translationX by transition.animateValue(
+    initialValue = Random.nextInt(-200..-100).dp,
```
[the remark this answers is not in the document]
hugokallstrom: Super!

app/src/main/java/com/hedvig/app/ui/compose/composables/BlurredFullScreenProgressOverlay.kt
hugokallstrom: Could not really find a better alternative to doing it like this. Let's revisit if we find the animation looks off on other devices.

PR #1139, opened 2022-07-22
Introduce core-resources to hold on to all string.xml files used across the app

StylianosGakis:
Practically, after this is merged since we're gitignoring the string.xml files from the app module, the developer will need to delete all ignored string.xml files manually from the app module to properly make use of this change.

Read the new README.md for even more context


hugokallstrom (approved):
Really good!!

PR #1140, opened 2022-07-25
Add ClaimDetail, AppStateInformation in core-design-system, core-ui

StylianosGakis:
Most notable changes at
core-design-system/src/main/java/com/hedvig/android/core/designsystem/component/information/AppStateInformation.kt
core-ui/src/main/java/com/hedvig/android/core/ui/genericinfo/GenericErrorScreen.kt
core-ui/src/main/java/com/hedvig/android/core/ui/genericinfo/GenericInfoScreen.kt

Create AppStateInformation as a 1:1 mapping of the figma file into the core-design-system module.
Move the info screen building blocks into core-ui and make use of them whenever possible.

Also makes sure the paddings are reasonable for each call site of them.


hugokallstrom (approved):
LGTM! 💯

PR #1141, opened 2022-07-26
Add lokalise gradle configuration to fix french translations

StylianosGakis:
`:core-resources:lintDebug` is failing on CI due to some things not
being translated properly. Specifically French requires "many" to be
specified for plural strings, something that we don't currently do.
Along with having to have a placeholder for plural strings that would
look different depending on the quantity specified. This is also
something we're not doing but this is on purpose, so we're disabling
this check with a tools:ignore.


hugokallstrom (approved):
LGTM! 🤠

PR #1144, opened 2022-07-27
APP-1704: Generic auth error handling

hugokallstrom:
Simplify view model and add tests.


app/src/main/java/com/hedvig/app/feature/embark/util/MaskType.kt
```diff
@@ -53,7 +53,7 @@ enum class MaskType {
   EMAIL {
     override fun mask(text: String) = text
     override fun unMask(text: String) = text
-    override fun isValid(text: String): Boolean = EMAIL_REGEX.matcher(text).find()
+    override fun isValid(text: String): Boolean = PatternsCompat.EMAIL_ADDRESS.matcher(text).matches()
```
StylianosGakis, line 56: Nice didn't know built-in solutions exist! Sounds like a good idea to use this

app/src/main/java/com/hedvig/app/feature/genericauth/CreateOtpAttemptUseCaseImpl.kt
```diff
@@ -6,28 +6,28 @@ import com.hedvig.app.util.apollo.QueryResult
...
-      Result.Error
+      CreateOtpResult.Error
     }
   }
 }
+
+sealed class CreateOtpResult {
+  data class Success(val id: String) : CreateOtpResult()
+  object Error : CreateOtpResult()
+}
```
StylianosGakis, line 33: +1 for the extraction, I like this style more! *Maybe* name this CreateOtpAttemptResult? I think it's good to keep the name matching and adding the Result/UseCase suffix to it to clearly indicate that they go together.

Also maybe go with a sealed interface instead? In general I like to default to interface unless we *need* it to be an abstract class for some reason.

```suggestion
sealed interface CreateOtpResult {
  data class Success(val id: String) : CreateOtpResult
  object Error : CreateOtpResult
}
```

app/src/main/java/com/hedvig/app/feature/genericauth/EmailInputScreen.kt
```diff
@@ -100,7 +96,9 @@ fun EmailInputScreen(
                 contentDescription = null,
                 colorFilter = ColorFilter.tint(color = MaterialTheme.colors.error),
               )
-            } else {
+            } else if (loading) {
+              CircularProgressIndicator(modifier = Modifier.size(24.dp))
+            } else if (inputValue.isNotBlank()) {
```
StylianosGakis, line 101: Nice addition!

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthActivity.kt
```diff
@@ -26,24 +23,20 @@ class GenericAuthActivity : BaseActivity() {
...
-          model.eventsFlow.collect { event ->
-            when (event) {
-              is GenericAuthViewModel.Event.SubmitEmailSuccess -> startOtpInputActivity(event)
-            }
-          }
-        }
+      viewState.otpId?.let {
+        model.onStartOtpInput()
+        startOtpInputActivity(otpId = it, email = viewState.input)
       }
```
StylianosGakis, line 29: A slight concern about doing a start activity side effect in composition.
I think we need to replace this with something like:
```suggestion
      LaunchedEffect(viewState.otpId) {
          val otpId = viewState.otpId ?: return@LaunchedEffect
          model.onStartOtpInput()
          startOtpInputActivity(otpId = otpId, email = viewState.input)
      }
```
That way we're not running a startActivity in composition which sounds a bit scary, I am not sure if there will ever be a case where two of them will run before model.onStartOtpInput is run and finished.

Also a good section to read https://developer.android.com/jetpack/compose/performance#avoid-backwards
Here we're reading viewState.otpId in composition, and then we're calling onStartOtpInput which changes the otpId since it changes the state in the VM. And we're not doing this inside any sort of effect handler like LaunchedEffect.

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
```diff
@@ -2,39 +2,24 @@ package com.hedvig.app.feature.genericauth
...
 
 class GenericAuthViewModel(
   private val createOtpAttemptUseCase: CreateOtpAttemptUseCase,
 ) : ViewModel() {
-  private val eventChannel = Channel<Event>(Channel.UNLIMITED)
-  val eventsFlow = eventChannel.receiveAsFlow()
 
-  sealed class Event {
-    data class SubmitEmailSuccess(val id: String, val credential: String) : Event()
-  }
```
StylianosGakis, line 21: Kill them events 🔪

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
```diff
@@ -44,56 +29,56 @@ class GenericAuthViewModel(
...
-    viewModelScope.launch {
-      // TODO: Set Loading-state, once one exists
-      val email = viewState.value.input
-      when (val result = createOtpAttemptUseCase.invoke(email)) {
-        is CreateOtpAttemptUseCase.Result.Success -> {
-          eventChannel.trySend(Event.SubmitEmailSuccess(result.id, email))
-        }
-        CreateOtpAttemptUseCase.Result.Error -> {
-          _viewState.update { it.copy(error = ViewState.TextFieldError.NETWORK_ERROR) }
+    with(_viewState) {
```
StylianosGakis, line 45: Hmm maybe I'm wrong on this one, but I find it kinda hard to follow through the code when we wrap things like this in a with() and use it as an implicit `this`.
I like `with` when we use it to bring it into scope since we're using things that need this thing to be a `this` to make extension functions work, like we did [here](https://github.com/HedvigInsurance/android/blob/0b31c97c8cf78d22abe7e0cee1ee0fcad6c2f6fe/app/src/main/java/com/hedvig/app/feature/marketing/marketpicked/MarketPickedScreen.kt#L39-L40) since .toDp() needs to be in the context of LocalDensity to work.

If you disagree I am 100% fine leaving it as it is, just sharing my opinion here 😊

app/src/main/java/com/hedvig/app/util/Validation.kt
```diff
@@ -1,15 +1,17 @@
...
 import androidx.annotation.StringRes
+import androidx.core.util.PatternsCompat
 import com.hedvig.app.R
 
 object Regexes {
-  val emailRegex = Regex("^\\S+@\\S+\$")
   val phoneNumberRegex = Regex("([+]*[0-9]+[+. -]*)")
 }
 
+fun String.isValidEmail() = isBlank() && PatternsCompat.EMAIL_ADDRESS.matcher(this).matches()
```
StylianosGakis, line 11: One concern for extension functions like these.
This means that in every single part of the code base, whenever we have a String type and we do `.` on it, we're gonna be suggested this as a possible thing we want to use.
And the ratio of when we do `.` on a String and when we specifically have that string be an email and we might be interested in this function is quite low. In a sense this will be there but not used 95%+ (I just made that number up) of the time.

With all that said, we can go ahead and keep this if you want or we don't come up with a good alternative, but it's an interesting thing to keep in mind in general imo 😊

app/src/test/java/com/hedvig/app/feature/genericauth/GenericAuthViewModelTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.hedvig.app.util.coroutines.MainCoroutineRule
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.test.advanceTimeBy
+import kotlinx.coroutines.test.advanceUntilIdle
+import kotlinx.coroutines.test.runTest
+import org.junit.Rule
+import org.junit.Test
+import kotlin.time.Duration.Companion.milliseconds
+
+class GenericAuthViewModelTest {
```
StylianosGakis, line 14: Nice! Thanks for adding these!

StylianosGakis (approved):
Super nice PR, had some thoughts here and there but aside from them this is a great improvement overall imo, great work!

app/src/main/java/com/hedvig/app/feature/genericauth/GenericAuthViewModel.kt
hugokallstrom: I see your point. Changing 👍

app/src/main/java/com/hedvig/app/util/Validation.kt
hugokallstrom: Good point!

PR #1145, opened 2022-07-27
APP-1789: Clear cache on language change

hugokallstrom:
Also another commit with some additional error handling.

### DIFF
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -17,7 +17,6 @@ import coil.decode.SvgDecoder
 import com.apollographql.apollo3.ApolloClient
 import com.apollographql.apollo3.cache.normalized.api.MemoryCacheFactory
 import com.apollographql.apollo3.cache.normalized.api.NormalizedCacheFactory
-import com.apollographql.apollo3.cache.normalized.apolloStore
 import com.apollographql.apollo3.cache.normalized.normalizedCache
 import com.apollographql.apollo3.interceptor.ApolloInterceptor
 import com.apollographql.apollo3.network.okHttpClient
@@ -214,9 +213,9 @@ import com.hedvig.app.service.push.senders.NotificationSender
 import com.hedvig.app.service.push.senders.PaymentNotificationSender
 import com.hedvig.app.service.push.senders.ReferralsNotificationSender
 import com.hedvig.app.util.LocaleManager
-import com.hedvig.app.util.apollo.CacheManager
 import com.hedvig.app.util.apollo.DeviceIdInterceptor
 import com.hedvig.app.util.apollo.GraphQLQueryHandler
+import com.hedvig.app.util.apollo.NetworkCacheManager
 import com.hedvig.app.util.apollo.ReopenSubscriptionException
 import com.hedvig.app.util.apollo.SunsettingInterceptor
 import com.hedvig.app.util.featureflags.ClearHAnalyticsExperimentsCacheUseCase
@@ -380,7 +379,7 @@ val viewModelModule = module {
       repository = get(),
       localeBroadcastManager = get(),
       hAnalytics = get(),
-      clearCache = { get<ApolloClient>().apolloStore.clearAll() },
+      cacheManager = get(),
     )
   }
   viewModel { DatePickerViewModel() }
@@ -713,7 +712,7 @@ val useCaseModule = module {
 }
 
 val cacheManagerModule = module {
-  single { CacheManager(get()) }
+  single { NetworkCacheManager(get()) }
 }
 
 val pushTokenManagerModule = module {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/hedvig/app/ApplicationModule.kt
```diff
@@ -374,7 +375,14 @@ val viewModelModule = module {
   viewModel { (quoteCartId: QuoteCartId?) -> RedeemCodeViewModel(quoteCartId, get(), get()) }
   viewModel { UserViewModel(get(), get(), get(), get(), get(), get()) }
   viewModel { WelcomeViewModel(get()) }
-  viewModel { SettingsViewModel(get(), get(), get()) }
+  viewModel {
+    SettingsViewModel(
+      repository = get(),
+      localeBroadcastManager = get(),
+      hAnalytics = get(),
+      clearCache = { get<ApolloClient>().apolloStore.clearAll() },
```
hugokallstrom, line 383: Is it too late in the after noon or is this actually beneficial? With composable we only want to pass lambdas to avoid coupling, cant we do the same with view models?

app/src/main/java/com/hedvig/app/util/apollo/SafeQuery.kt
```diff
@@ -65,7 +65,12 @@ fun <D : Query.Data> ApolloCall<D>.safeWatch(): Flow<QueryResult<D>> {
 private fun <D : Operation.Data> ApolloResponse<D>.toQueryResult(): QueryResult<D> {
   val data = data
   return when {
-    hasErrors() -> QueryResult.Error.QueryError(errors?.first()?.message)
+    hasErrors() -> {
+      val exception1 = errors?.first()?.extensions?.get("exception")
```
hugokallstrom, line 69: Not very pretty, but it works

app/src/main/java/com/hedvig/app/ApplicationModule.kt
StylianosGakis: Hmm so about this I think the reason why it may not be as useful is this.

In compose, we offload this kind of work from the Composable, to the ViewModel, which is meant to be holding such logic.
In this case we would be offloading this logic from the ViewModel to the DI Module definition, which if anything would make this already humongous file even more gigantic 😅
I think in general these modules should be as dumb and straightforward as possible, like you are this Object, and you need these dependencies, so go ahead and take them.

Imagine a if this were normal Dagger and had a @Module and a @Provides and inside the fun annotated with @Provides you'd have logic about what lambda is going what instead of just acting like a dependency injection system and simply injecting the necessary dependencies that the object needs. I think it would look even more weird there, here it looks more normal since it's Koin and all of it is more free-form sort of, if that makes  sense :D 


And besides all the long text above :D 
We have the com.hedvig.app.util.apollo.CacheManager which exists to expose exactly one function called `clearCache`, so here instead you could simply inject the CacheManager and let the ViewModel call `cacheManager.clearCache`.
Which makes me think of course maybe this could be named ApolloCacheManager or something, since CacheManager technically could mean more things.

app/src/main/java/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
```diff
@@ -75,7 +75,7 @@ class InsuranceFragment : Fragment(R.layout.fragment_insurance) {
             ?.also { insuranceViewModel.crossSellActionOpened() }
 
           errorMessage?.let {
-            adapter?.submitList(listOf(InsuranceModel.Header, InsuranceModel.Error))
+            adapter?.submitList(listOf(InsuranceModel.Header, InsuranceModel.Error(it)))
           }
 
           items?.let {
```
StylianosGakis, line 81: In this entire section we make it so that if somehow we get both errorMessage as non-null and items as non-null (not sure if possible) we would override the other thing by submitting a new list again. Could we maybe make this a when statement or something like that so that only one of all of them is run? With order of importance like first check if the errors are not null, show that and so on. Or something like that.
Or is there some other reason we're not doing that here that I may be missing?

app/src/main/java/com/hedvig/app/util/apollo/SafeQuery.kt
StylianosGakis: Yeah but if we *know* we're getting good error messages inside the .exception.body.message then this is completely fine.
If anything I'd replace exception1 -> exception and instead of doing LinkedHashMap<*, *> I'd do Map<*, *> in both cases, but that's it. It's also quite safe in the sense that we're always going for a nullable type for the entire chain. So go for it imo!
We'd just have to check that it does not sometimes give off super janky responses or things we shouldn't show. In general it shouldn't be the case though I think

StylianosGakis (approved):
I think it'd be nice if we instead pass the CacheManager as I described in the comment for the SettingsViewModel change 😊
Looking good with extracting the error message from the gql response!

app/src/main/java/com/hedvig/app/ApplicationModule.kt
hugokallstrom: Yes that makes sense :)

hugokallstrom: I will rename CacheManager to NetworkCacheManager, so that its clear that its used for network cache.

StylianosGakis: I was thinking of ApolloCacheManager since technically we could be referring to the OkHttp network cache or the cache that Coil has with the names `CacheManager` or `NetworkCacheManager`.
Because both of those two are also caching responses from the network and we use them on our project. But we usually don't really touch those so maybe it's fine, but if we do in the future it might get a bit confusing :D

PR #1146, opened 2022-07-28
Catch error when market and payment method is not compatible.

hugokallstrom:


StylianosGakis (approved):
So now I guess this makes it so that the notification will at least open the logged in activity, instead of going to the payin directly, which is much better than crashing of course :D
LGTM!

PR #1148, opened 2022-07-29
Do not try to launch WebOnboarding if not supported

StylianosGakis:
According to the docs
https://developer.android.com/guide/components/intents-common#Browser
And the recent crash (probably from the Play Store monkey testers)
https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/3386af3e8dbbd2a0b6e2609c949ffff3?time=last-thirty-days&versions=7.0.0%20(1658995833);7.0.0%20(1658995590)&sessionEventKey=62E24BFC01E8000125FF85939E5C49DF_1703787032309783907

One should not `.startActivity` on an Intent which hasn't been checked
if it works or not before. If the phone has no supporting app that can
launch this intent, in this case a URL, it will straight up crash the
app.

Current change prints NETWORK_ERROR_ALERT_MESSAGE aka "We couldn't
reach Hedvig right now, are you sure you have an internet connection?"
since we don't have a more appropriate string ready, but at least it
might prompt people to go to the website directly


app/src/main/java/com/hedvig/app/feature/settings/Market.kt
```diff
@@ -61,9 +62,14 @@ enum class Market {
...
-    val url = context.getString(R.string.WEB_BASE_URL) + "/" + webPath + "/new-member"
-    val browserIntent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
-    context.startActivity(browserIntent)
+    val uri = Uri.parse("""${context.getString(R.string.WEB_BASE_URL)}/$webPath/new-member""")
+    val browserIntent = Intent(Intent.ACTION_VIEW, uri)
+    if (browserIntent.resolveActivity(context.packageManager) != null) {
+      context.startActivity(browserIntent)
+    } else {
+      e { "Tried to launch $uri but the phone has nothing to support such an intent." }
+      context.makeToast(hedvig.resources.R.string.NETWORK_ERROR_ALERT_MESSAGE)
```
hugokallstrom, line 71: Can we use a more suitable message here?

hugokallstrom (approved):

app/src/main/java/com/hedvig/app/feature/settings/Market.kt
StylianosGakis: Yeah I could not find something in lokalise that would suit this here. If we want to create a new string then sure we could do that, and then wait for translations

StylianosGakis: Gonna go with
general_unknown_error -> "Unknown Error"
as discussed 👍

PR #1149, opened 2022-07-31
hanalytics module

StylianosGakis:
For hanalytics to be extracted, some more modules had to be introduced for hanalytics to depend on since it can't depend on :app itself.

It's easier to look at the new structure by checking out the repo at this state instead of looking through the diff. There are so many files that have only their import changes and where they are meaning their path is different which is hard to see from the PR view.

The /misc/images/modularization-graph.png also gives a good overview.


hugokallstrom (approved):
Good job!! LGTM

PR #1150, opened 2022-08-01
Testing: Use fake feature manager

StylianosGakis:
I want some feedback regarding the impl of FakeFeatureManager.

Using lambdas is I think a good improvement, so that if we want more control over what it will return depending on other local state we can simply do that inside the lambda. So I am confident about this decision.
However for the decision to error out if some feature is asked from the map but wasn't explicitly set I'd like your opinion.

If we let it default to `false` by doing `getOrDefault()` as we did before, that means that some tests may be using that return value in a way that is crucial for the test to pass. If we error out as I've done it, one *needs* to be explicit about what the return value would need to be for that particular test.
With that said, we do have tests where sometimes a value may be asked from the implementing code, but it may *actually* not be important for the test outcome. For such cases I opted for nextBoolean, or enumValues<T>().random() to indicate that we need a return value to be there, but we do not care about what it will be. With the mockk approach we had before, this was indicated by not declaring it at all, however then mockk was picking the same default all the time, which for some tests we may accidentally be depending on what that default value was.

I don't like implicitly depending on such defaults, whether that comes from mockk picking one, or having us default to `false` for each map entry that we're asking for but haven't previously declared. Because that means that for example for the test:
```kotlin
@Test
fun `when common claims-feature is disabled, should not show common claims`() = runTest {
  val featureManager = mockk<FeatureManager>(relaxed = true)
  coEvery { featureManager.isFeatureEnabled(Feature.COMMON_CLAIMS) } returns false  // <--- emphasis
  val builder = sut(featureManager)

  val result = builder.buildItems(HOME_DATA_ACTIVE)

  assertThat(result).containsNoneOfType<HomeModel.CommonClaim>()
}
```
Could be written as:
```kotlin
@Test
fun `when common claims-feature is disabled, should not show common claims`() = runTest {
  val featureManager = mockk<FeatureManager>(relaxed = true)
  // <---
  val builder = sut(featureManager)

  val result = builder.buildItems(HOME_DATA_ACTIVE)

  assertThat(result).containsNoneOfType<HomeModel.CommonClaim>()
}
```
And it'd still pass since we have the default.

But with the current impl, if we do
```kotlin
@Test
fun `when common claims-feature is disabled, should not show common claims`() = runTest {
  val featureManager: FeatureManager = FakeFeatureManager(
    featureMap = { mapOf(Feature.COMMON_CLAIMS to false) }, // <---
  )
  val builder = HomeItemsBuilder(featureManager)

  val result = builder.buildItems(HOME_DATA_ACTIVE)

  assertThat(result).containsNoneOfType<HomeModel.CommonClaim>()
}
```
It passes.

And if we do
```kotlin
@Test
fun `when common claims-feature is disabled, should not show common claims`() = runTest {
  val featureManager: FeatureManager = FakeFeatureManager(
    featureMap = { mapOf() }, // <---
  )
  val builder = HomeItemsBuilder(featureManager)

  val result = builder.buildItems(HOME_DATA_ACTIVE)

  assertThat(result).containsNoneOfType<HomeModel.CommonClaim>()
}
```
It will actually fail since we didn't specify that `COMMON_CLAIMS` `false`. And I like that it won't pass by using the implicit defaults.

So I see this entire thing as a compromise, and I myself lean more towards the "Be explicit and sometimes over-do stuff" over "Be implicit and sometimes pass tests even by chance".

We could event take it a step further and model the "It needs to be there but I don't care about what it returns" with a function inside the companion object of FakeFetureManager like
```kotlin
companion object {
  fun whateverLoginMethod(): LoginMethod = enumValues<LoginMethod>().random()
  fun whateverPaymentType(): PaymentType = enumValues<PaymentType>().random()
}
```
And then indicate this in tests like
`FakeFeatureManager(paymentType = { FakeFeatureManager.whateverPaymentType() }),`
instead of
`FakeFeatureManager(paymentType = { enumValues<PaymentType>().random() }),`
But I am not sure if this then is overkill :D

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -53,7 +53,7 @@ android {
     maybeCreate("staging")
     maybeCreate("pullrequest")
     named("release") {
-//      signingConfig = signingConfigs.getByName("debug") uncomment to run release build locally
+//      signingConfig = signingConfigs.getByName("debug") // uncomment to run release build locally
       applicationIdSuffix = ".app"
       manifestPlaceholders["firebaseCrashlyticsCollectionEnabled"] = true
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -142,6 +142,7 @@ dependencies {
   implementation(projects.hedvigMarket)
 
   testImplementation(projects.hanalyticsTest)
+  androidTestImplementation(projects.hanalyticsTest)
 
   implementation(projects.hanalyticsEngineeringApi)
   releaseImplementation(projects.hanalyticsEngineeringNoop)
--- a/app/src/androidTest/java/com/hedvig/app/util/FeatureFlagRule.kt
+++ b/app/src/androidTest/java/com/hedvig/app/util/FeatureFlagRule.kt
@@ -3,25 +3,24 @@ package com.hedvig.app.util
 import com.hedvig.android.hanalytics.di.featureManagerModule
 import com.hedvig.android.hanalytics.featureflags.FeatureManager
 import com.hedvig.android.hanalytics.featureflags.flags.Feature
+import com.hedvig.android.hanalytics.test.FakeFeatureManager
 import com.hedvig.hanalytics.PaymentType
-import io.mockk.coEvery
-import io.mockk.mockk
 import org.junit.rules.ExternalResource
 import org.koin.core.context.loadKoinModules
 import org.koin.core.context.unloadKoinModules
 import org.koin.dsl.module
 
+@Suppress("RemoveExplicitTypeArguments")
 class FeatureFlagRule(
   vararg flags: Pair<Feature, Boolean>,
-  private val paymentType: PaymentType? = PaymentType.TRUSTLY,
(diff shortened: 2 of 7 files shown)


hugokallstrom (approved):
I like it. It would be nice to have an extension function to clarify that nextBoolean means we dont care about that feature flag - it looks a bit cryptic now.

StylianosGakis:
> I like it. It would be nice to have an extension function to clarify that nextBoolean means we dont care about that feature flag - it looks a bit cryptic now.

I had left it like this as `nextBoolean` is a Kotlin stdlib function for basically a random Boolean, so if we introduce a new function, if anything it will be more cryptic since the burden of finding the appropriate name falls on us instead of relying on the stdlib for it.

But with that said, I guess I realize that `nextBoolean` itself is not a widely used fun from stdlib, so we're back at it being the more cryptic one :D 

If I don't find a more straightforwardly named function fun from stdlib I'll do what you suggested and introduce smth of our own!

PR #1177, opened 2022-08-15
APP-1879: Migrate insets otp screen

StylianosGakis:
Get started with migrating to the new insets APIs.
tl;dr is that if you want to simply add the paddings to safely draw your items away from the system bars and the IME you can use the modifier `fun Modifier.windowInsetsPadding(insets: WindowInsets): Modifier` which takes in `WindowInsets` and you can decide which one you want from this list [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/androidMain/kotlin/androidx/compose/foundation/layout/WindowInsets.android.kt;l=379-405) and then there's a list of convinience modifiers [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/androidMain/kotlin/androidx/compose/foundation/layout/WindowInsetsPadding.android.kt) that do this for you under the hood.
In our use case we could do either:
`.windowInsetsPadding(WindowInsets.safeDrawing)`
OR:
`.safeDrawingPadding()`
And it's equivalent.

But one may want to do the more verbose solution when needing more control, like insetting for safeDrawing, but only horizontally and bottom. Sample here:
```
.windowInsetsPadding(
  WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal)
    .union(WindowInsets.safeDrawing.only(WindowInsetsSides.Bottom)),
)
```

Refactored the OTP screen as it was all one big composable and made the migration quite hard

Used `.graphicsLayer { rotationZ = angle }` instead of `rotation(angle)` as that means we'll now be doing less recompositions in the way we're using it here.

While at it, also made sure that the error message goes away when editing the input field as the error persisted there forever. Look at the test `updating the input after getting an error should clear the error` for more details.

### DIFF
--- a/app/src/main/java/com/hedvig/app/ApplicationModule.kt
+++ b/app/src/main/java/com/hedvig/app/ApplicationModule.kt
@@ -382,7 +382,7 @@ val viewModelModule = module {
     CrossSellDetailViewModel(crossSell.action, get(), get())
   }
   viewModel { GenericAuthViewModel(get()) }
-  viewModel { (otpId: String, credential: String) ->
+  viewModel<OtpInputViewModel> { (otpId: String, credential: String) ->
     OtpInputViewModel(
       otpId,
       credential,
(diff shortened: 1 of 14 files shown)


hugokallstrom (approved):
LRGTM

PR #1184, opened 2022-08-17
APP-1899: Audio recorder insets

StylianosGakis:
Also snuck in a change inside `EmbarkStoryTesterActivity` to make sure the content is at least inside a clickable area, didn't pay much attention since it's developer only.


hugokallstrom (approved):
👍

PR #1185, opened 2022-08-18
APP-1900: Marketing screen bottom sheet fixes

StylianosGakis:
Remove the insets dependency completely but keep the insets-ui from accompanist as there's no alternative for that yet.

More details on each individual commit.


hugokallstrom (approved):
LGTM

PR #1188, opened 2022-08-22
APP-1831, APP-1595: Add range from today when showing date picker in embark

hugokallstrom:


StylianosGakis (approved):
Do we ever use this calendar picker to let the member pick for example something like when a claim has happened, which would be earlier than today? Not familiar with 100% of the uses we have in embark of this component.
If not then lgtm!

hugokallstrom:
> Do we ever use this calendar picker to let the member pick for example something like when a claim has happened, which would be earlier than today? Not familiar with 100% of the uses we have in embark of this component. If not then lgtm!

No this is only for moving flow. So we can assume for now that we should never allow dates in the past. When we migrate to odyssey we can make sure to specify a range in the back-end.

PR #1189, opened 2022-08-22
Change model to viewmodel

StylianosGakis:
Just a consistency PR, since when we're talking about them we always refer to viewmodels as "viewModel(s)" I think it makes it easier on your brain to see `viewModel` and know what it means.
We do use the term `model` some times in the code base also for a simple class which holds the "model" of an UI, like [here](https://github.com/HedvigInsurance/android/blob/f689a0e2e0122c53bda476a81cdb2d7ef6ec6565/app/src/main/java/com/hedvig/app/feature/checkout/CheckoutViewModel.kt#L85) so figured this change makes sense to differentiate from such usages.


hugokallstrom (approved):
👍

PR #1191, opened 2022-08-22
Error handling/no internet

StylianosGakis:
The user facing changes are that now when one goes into the settings screen and they've got no internet (the network request fails) they will actually get the error screen which lets them retry instead of getting a clean screen which does nothing.
Same for the forever tab because it was getting stuck on the shimmer effect forever since watching the query never finished or failed.


hugokallstrom (approved):
Nice cleanup!

PR #1196, opened 2022-08-25
Remove charity picking

StylianosGakis:
This is due to the new status of charity where there's only 1 so there's no choice.
Instead I removed the old impl and made 1 simple compose-based implementation of the single screen we should be showing in that case.


StylianosGakis:
This is failing everything due to the key-gear removal. The key-gear is stacked on *top* of this though, so this will start passing once that's merged in.
Didn't do it yet however since it'd make this PR harder to review.

oscarnylander:
> This is failing everything due to the key-gear removal. The key-gear is stacked on _top_ of this though, so this will start passing once that's merged in. Didn't do it yet however since it'd make this PR harder to review.

Maybe we could base this PR on that branch in that case? 🤔

feature-charity/src/main/java/com/hedvig/android/feature/charity/CharityActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.compose.runtime.getValue
+import androidx.core.view.WindowCompat
+import coil.ImageLoader
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.feature.charity.ui.CharityScreen
+import org.koin.android.ext.android.get
+import org.koin.androidx.viewmodel.ext.android.getViewModel
+
+@OptIn(ExperimentalMaterial3WindowSizeClassApi::class)
+class CharityActivity : AppCompatActivity() {
```
oscarnylander, line 18: Should this extend `BaseActivity`?

feature-charity/src/main/java/com/hedvig/android/feature/charity/CharityActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    WindowCompat.setDecorFitsSystemWindows(window, false)
+    val imageLoader: ImageLoader = get()
+    val viewModel = getViewModel<CharityViewModel>()
```
oscarnylander, line 24: What's the reason for getting these dependencies with koin in the `onCreate`-body over the more common `inject`-variant?

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityBottomSheet.kt
```diff
@@ -0,0 +1,71 @@
...
+import hedvig.resources.R
+
+@Composable
+internal fun CharityBottomSheet(
+  isShowing: Boolean,
+  closeSheet: () -> Unit,
+) {
+  BackHandler(
+    enabled = isShowing,
+    onBack = { closeSheet() },
```
oscarnylander, line 34:
```suggestion
    onBack = ::closeSheet,
```

(on condition that `onBack: () -> Unit`, of course)

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityBottomSheet.kt
```diff
@@ -0,0 +1,71 @@
...
+      )
+      Spacer(Modifier.height(24.dp))
+    }
+  }
+}
+
+@Preview
+@Composable
+private fun CharityBottomSheetPreview() {
+  CharityBottomSheet(true) {}
```
oscarnylander, line 70: A matter of taste, but I prefer putting the parameters explicitly here, as this makes it look like the function passed in would be a `Composable`, when in reality is isn't 😬

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityBottomSheet.kt
```diff
@@ -0,0 +1,71 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.BottomSheetHandle
+import hedvig.resources.R
+
+@Composable
+internal fun CharityBottomSheet(
+  isShowing: Boolean,
+  closeSheet: () -> Unit,
```
oscarnylander, line 30:
```suggestion
  closeSheet: () -> Unit,
  isShowing: Boolean,
```

This way, there is no way to mix up `closeSheet` being a `Composable` 🙂

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityScreen.kt
```diff
@@ -0,0 +1,272 @@
...
+  goBack: () -> Unit,
+  retry: () -> Unit,
+  imageLoader: ImageLoader,
+  windowSizeClass: WindowSizeClass,
+) {
+  Box(Modifier.fillMaxSize()) {
+    Column {
+      TopAppBarWithBack(
+        onClick = { goBack() },
+        title = "",
```
oscarnylander, line 112: This line seems like an indicator that the API of `TopAppBarWithBack` could maybe be improved 🤔 

Alternatively that it's inappropriate for the usecase.

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityScreen.kt
```diff
@@ -0,0 +1,272 @@
...
+  showSheet: () -> Unit,
+  goBack: () -> Unit,
+  retry: () -> Unit,
+  imageLoader: ImageLoader,
+  windowSizeClass: WindowSizeClass,
+) {
+  Box(Modifier.fillMaxSize()) {
+    Column {
+      TopAppBarWithBack(
+        onClick = { goBack() },
```
oscarnylander, line 111:
```suggestion
        onClick = ::goBack,
```

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 # androidSdkVersions
-compileSdkVersion = "32"
+compileSdkVersion = "33"
```
oscarnylander, line 3: Is this the right PR to introduce these changes? 🤔 There seems to be a lot of version updates in here 😬

StylianosGakis:
That branch is on top of this branch. If I were to do that I'd have to rebase the other branch, remove the commits that are from this branch, potentially deal with conflicts there if some things were on the same changed files, then take this branch and rebase it off of that other branch and have to force push both of them to GitHub.
I think I'd rather not do that if possible and simply merge the key-gear branch here after this can be reviewed and then merge to develop only after CI is green again.

oscarnylander (changes requested):
Nice!

feature-charity/src/main/java/com/hedvig/android/feature/charity/CharityActivity.kt
StylianosGakis: A good opportunity to talk about this.
We don't have access to BaseActivity from other modules since it's inside :app. So either I'll extract it out, or consider what is it that we're doing inside BaseActivity and see if we need it or if we can do it in a different way.

As I see it now, it's there to share the behavior of listening to locale changes. Could I get some historical information of how this works and helps in these situations? I know we have a fixed language picked from the settings, and if we're in for example this screen there's no way to change that language picked without leaving it, going into settings, changing the language, and then navigating back to this screen after we've already changed the language therefore it should just be applied.
I'm probably missing smth since I haven't really worked with Android locales and stuff like that so please do help me out here :D

feature-charity/src/main/java/com/hedvig/android/feature/charity/CharityActivity.kt
StylianosGakis: Reducing the scope in which it's available mostly. If it doesn't need to be accessed globally in this class and can stay contained inside the function I figured it'd better stay that way.
The lazy impl in koin is just doing a `lazy` wrapper around `get()` and the `lazy` comes with an overhead of doing all this fetching by aquring a lock to make sure it's thread safe and that kind of stuff https://github.com/InsertKoinIO/koin/blob/c954e53959cf982725ca33c9680acb650c63da52/android/koin-android/src/main/java/org/koin/android/ext/android/ComponentCallbackExt.kt#L40-L58 
So figured why not get the benefit of removing this locking and the reduced scope, seemed like a win-win to me.

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityBottomSheet.kt
StylianosGakis: Yup totally agree with this one!
Had to use the `onback = closeSheet` syntax since `closeSheet` is itself of type `() -> Unit` and not a function which we would need to do `::` on

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityBottomSheet.kt
StylianosGakis: Yeah I try to be mindful not to put function types as last parameters when we don't actually want the trailing lambda syntax, will fix this too!

feature-charity/src/main/java/com/hedvig/android/feature/charity/ui/CharityScreen.kt
StylianosGakis: Yeah I agree, I still like to call the `TopAppBarWithBack` function here to make sure that we've got the same look, height etc as the other screens, but the API needs to be updated to support an empty title without this "" weird looking call site

gradle/libs.versions.toml
StylianosGakis: The new deps are

androidx-compose-material3-windowSizeClass: To enable not stretching the image super wide when on a wide screen or on landscape mode.

androidx-customview-poolingcontainer: A bug which disables compose previews in feature module requires this dependency.

accompanist-placeholder: A lib with a simple compose Modifier to get a nice loading effect behind the image while it's still loading.

The rest of the changes are just sorting dependencies by name which given enough time just get mixed up a bit.

The 33 compileSdk is needed for the windowSizeClass one, but it's just a bump on the compile sdk, not the targetsdk, so that's a safe thing to do without changing the app's behavior. Plus we get new lint hints and stuff like that which we will need to be aware of for when we eventually target sdk 33. https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd

feature-charity/src/main/java/com/hedvig/android/feature/charity/CharityActivity.kt
oscarnylander: It's the thing that does basically what the new APIs for changing language does - it restarts every activity when the language changes, so that these changes are picked up.

StylianosGakis: So in the context of the CharityActivity screen, is it possible that the language will ever change while this screen is showing?
Is this needed in the first place because we specifically allow people to change their language from inside the app? (something which is not that common I guess)
Is there a way to do this without using this BaseActivity inheritance? And if not, we should still be able to extract this from inside the :app module in order for other modules to have access to it right?

oscarnylander: I'd say it's very unlikely, but one possible path would be:

1. User opens the app
2. User navigates to charity screen
3. User backgrounds the app
4. User opens the app through a chat notification
5. User accesses settings through the chat screen
6. User changes language in this screen
7. User navigates back twice, landing on a now incorrectly localized charity screen

StylianosGakis: Yup makes sense, let's see how it goes with the thing you're trying out with the language APIs then and go from there.
In the meantime I guess I'll have to
1) make BaseActivity available to this module in order to merge it in
2) Take the key gear branch and take it off of the top of this one and merge that into develop to allow us to do other stuff in the meantime while keeping this branch on hold until this is fixed
3) Just merge this in and consider this something that we can fix ASAP afterwards, since the chances that someone encounters this and is bothered by it seem extremely low. And the impact isn't the end of the world either.

If I am to do one, this is the order of the one I want to do the most, to the least 3 -> 1 -> 2 but you can convince me otherwise 😅

PR #1202, opened 2022-08-29
APP-1816: Migrate release builds from Bitrise to Github Actions

oscarnylander:
- Add upload to play store workflow
- Configure Play Store upload-step
- Fix release pre-build script
- Add gradle copy-step
- Split the calls up, maybe that fixes it?
- Add debugging
- Revert "Add debugging"
- smh
- Bump deploy action version

### DIFF
--- a/.github/workflows/upload-to-play-store.yml
+++ b/.github/workflows/upload-to-play-store.yml
@@ -10,8 +10,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - name: Copy CI gradle.properties
+        run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties
       - name: Setup JDK 11
-        uses: actions/setup-java@v3.3.0
+        uses: actions/setup-java@v3.4.1
         with:
           java-version: '11'
           distribution: 'zulu'
--- a/.github/workflows/upload-to-play-store.yml
+++ b/.github/workflows/upload-to-play-store.yml
@@ -2,7 +2,8 @@ name: Build and Publish to Play Store
 on:
   push:
     branches:
-      - master
+      - release/**
+      - hotfix/**
 jobs:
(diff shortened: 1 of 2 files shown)


oscarnylander:
Perhaps! I have no preferred strategy, so I leave it up to you and @hugokallstrom to determine which branches should trigger this action 🙂

StylianosGakis:
That's fair. In my head the process that makes the most sense without going too custom is something like this.

Releases to the play store are built from when things are pushed to "release/vX.X.X" and "hotfix/*" branches.

Not off of master, since we might want to for example release an older state of master, but we wouldn't want to branch off of it, revert some commits, merge back to master (only to release) then branch out again, re-revert and then merge back to master again to have that up to latest. If we release off of hotfix branches we can simply create one, revert that and release right there directly, without doing this dance.
Also not off of develop since that one gets merged in eagerly and most likely we don't want to release the latest develop, but something a bit older than that.

Aside from the plays store releases.
Firebase app tester releases, I'd say every merge into master or develop should be heading there. As that is where QA happens and we for sure want all of those to exist.

As far as the pull-request build we have which happens on pull-requests, I don't think we're using that anymore, at least we haven't in the past (many) months. Also the showkase stuff we have not been using at all either. It still feels useful but we might need to re-imagine how this process would be like, because agains it's something we've never actually used.
We merge into develop more eagerly in general, with the idea that if we want something bigger instead of doing long-lasting branches we'd rather feature flag it and put it inside develop asap, aka testable from Firebase. So the build happening on the PR before it's merged, and being blocked from being merged into develop until the PR review happens *and* the QA happens isn't something that we've enjoyed doing. So I think we can remove that build type and simply not post that release off of slack (or just post the same ones that go to firebase app tester for those who want to test and don't have firebase).

oscarnylander:
Done!

StylianosGakis:
Merge whenever you feel it's ready 🙌

PR #1204, opened 2022-08-31
Change `java` to `kotlin` source set

StylianosGakis:
Now that we're no longer held back from this bug https://github.com/JLLeitschuh/ktlint-gradle/issues/524 since we've moved away from this plugin we can simply revert [this commit](https://github.com/HedvigInsurance/android/commit/da0f344d83c6905741ee7d376d9e3519f8f8e1a6) and go back to having kotlin as the directory name.

The -4+4 line code changes happen on a doc at `sample-module/README.md` and referring to the new file name at `app/build.gradle.kts`.


oscarnylander (approved):
This seems like it could be the kind of thing to cause a lot of pain when merging.

Let's see 😅

PR #1207, opened 2022-09-01
Giraffe cleanup

StylianosGakis:
Some preparations for when this https://github.com/HedvigInsurance/giraffe/pull/1672 is merged.
I was simply checking if we're using the removed endpoints, and it turned out we still had the graphql files but we were simply not using the generated code. This is the first step, but after that PR is properly merged we can download the new schema and see if there are any more things that we are generating code for but not using.


hugokallstrom (approved):
Nice!

PR #1210, opened 2022-09-06
APP-1910: Charity -> Business model

StylianosGakis:


hugokallstrom (approved):
Nice! Could you please give a short description on what a business model is in this context? Totally out of the loop on this, just curious :)

StylianosGakis:
Absolutely!
There basically has been a change in how things works regarding the leftover money that *may* be left over from the pool of money that gets used to pay out claims.
With the deal that we had before, with the re-insurers or whatnot, there was a guarantee that some percentage (I think 1%) would always be donated to the charities that we had a deal with, therefore this could be considered charity since we were consistently giving money there.
With the new deal, the re-insurers can not guarantee that they'll cover that % in order to be paid to charities. That means that now they'll only be paid if there is money leftover in that pool of money. Which means that there's not a guarantee that any money will reach their pockets every year, which I assume means that it can no longer be considered that we're contributing to charity. So with the new deal, and with the new parter, milkywire, we do not provide this guarantee and we're renaming the entire feature to "business-model" instead of charity, and the information that is shown to the members need to be updated accordingly.
So I took the opportunity to also rename our internal naming of this to not be a source of confusion down the line. The one thing that's still named "charity" now is the event inside `hAnalytics` but we may/may not need to keep that the same to retain historical data consistent. In any way I didn't touch that part since I'm not super familiar with what the expectations are for that kind of data from our analytics team.

PR #1211, opened 2022-09-07
Replace inhouse language solution with official APIs

oscarnylander:


app/src/main/kotlin/com/hedvig/app/feature/adyen/payin/AdyenExtensions.kt
```diff
@@ -37,7 +37,7 @@ fun Activity.startAdyenPayment(market: Market?, paymentMethods: PaymentMethodsAp
     )
     .addCardConfiguration(cardConfig)
     .addGooglePayConfiguration(googlePayConfig)
-    .setShopperLocale(getLocale(this, market))
+    .setShopperLocale(getLocale())
```
StylianosGakis, line 40: An example of using the top-level one instead of the one provided by LanguageService.
I guess it's hard to use that one in extension functions like this, but this will mean that it'll always be easier to reach for the top-level one from everywhere. Even by accident like we do in "ClaimDetailActivity".
Any thoughts on this? Maybe not the end of the world, but if we keep both I'd rather the LanguageService to also call this top level getLocale as an implementation detail so that no matter where you're coming from you're ending up in the same one implementation, to keep it simple to understand and change without having to do it in multiple places (if we ever need to do that)

app/src/main/kotlin/com/hedvig/app/feature/chat/ui/ChatActivity.kt
```diff
@@ -76,8 +76,6 @@ class ChatActivity : BaseActivity(R.layout.activity_chat) {
 
   private var forceScrollToBottom = true
 
-  override val preventRecreation = true
```
StylianosGakis, line 79: Since this isn't really gonna be supported anymore I wonder, what was the use case in the first place?
From what I understand some screens were done in a way where we did not want to restart on config change triggered from locale change for whatever reason. If this reason is still something we care about do we need to do something special about it?

app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt
```diff
@@ -79,7 +84,7 @@ class MarketingActivity : BaseActivity() {
 
   private fun openOnboarding(market: Market) {
     val baseUrl = getString(R.string.WEB_BASE_URL).substringAfter("//")
-    val uri = market.createOnboardingUri(this, baseUrl)
+    val uri = market.createOnboardingUri(baseUrl, languageService.getLanguage() ?: Language.EN_SE)
```
StylianosGakis, line 87: A̶n̶ ̶e̶x̶a̶m̶p̶l̶e̶ ̶w̶h̶e̶r̶e̶ ̶t̶h̶e̶ ̶g̶e̶t̶L̶a̶n̶g̶u̶a̶g̶e̶ ̶b̶e̶i̶n̶g̶ ̶n̶u̶l̶l̶ ̶b̶e̶c̶o̶m̶e̶s̶ ̶a̶ ̶b̶i̶t̶ ̶t̶e̶d̶i̶o̶u̶s̶ ̶t̶o̶ ̶u̶s̶e̶.̶ ̶S̶i̶n̶c̶e̶ ̶w̶h̶e̶n̶ ̶w̶e̶ ̶c̶a̶l̶l̶ ̶i̶t̶ ̶w̶h̶a̶t̶ ̶w̶e̶'̶r̶e̶ ̶p̶r̶o̶b̶a̶b̶l̶y̶ ̶g̶o̶n̶n̶a̶ ̶d̶o̶ ̶(̶a̶s̶ ̶w̶e̶ ̶d̶o̶ ̶h̶e̶r̶e̶)̶ ̶i̶s̶ ̶`̶?̶:̶ ̶L̶a̶n̶g̶u̶a̶g̶e̶.̶E̶N̶_̶S̶E̶`̶ ̶d̶o̶e̶s̶ ̶i̶t̶ ̶m̶a̶k̶e̶ ̶s̶e̶n̶s̶e̶ ̶t̶o̶ ̶b̶a̶k̶e̶ ̶t̶h̶i̶s̶ ̶i̶n̶t̶o̶ ̶t̶h̶e̶ ̶`̶g̶e̶t̶L̶a̶n̶g̶u̶a̶g̶e̶`̶ ̶i̶t̶s̶e̶l̶f̶?̶ ̶O̶r̶ ̶d̶o̶ ̶w̶e̶ ̶e̶v̶e̶r̶ ̶e̶x̶p̶e̶c̶t̶ ̶o̶u̶r̶s̶e̶l̶v̶e̶s̶ ̶t̶o̶ ̶d̶e̶a̶l̶ ̶w̶i̶t̶h̶ ̶t̶h̶e̶ ̶n̶u̶l̶l̶ ̶c̶a̶s̶e̶ ̶i̶n̶ ̶a̶ ̶d̶i̶f̶f̶e̶r̶e̶n̶t̶ ̶w̶a̶y̶?̶
Addressed on another comment

app/src/main/kotlin/com/hedvig/app/feature/marketing/data/GetInitialMarketPickerValuesUseCase.kt
```diff
@@ -9,19 +8,20 @@ import com.hedvig.android.hanalytics.featureflags.flags.Feature
...
-  private val context: Context,
   private val apolloClient: ApolloClient,
   private val featureManager: FeatureManager,
+  private val languageService: LanguageService,
 ) {
   suspend operator fun invoke(): Pair<Market, Language> {
     val currentMarket = marketManager.market
     if (currentMarket != null) {
-      val currentLanguage = Language.fromSettings(context, currentMarket)
+      val currentLanguage = languageService.getLanguage() ?: currentMarket.defaultLanguage()
```
StylianosGakis, line 24: Perfect, now I see why we need this to be nullable.
Can we then maybe instead do the opposite, have LanguageService be the one which knows how to handle that case by having a dependency on MarketManager so that each time we're using them we don't need to think of such logic?
Have it go smth like
1. pick selected lang
2. if none is picked, ask market for current market and get default lang
3. if there is no market fallback to en_SE
and have all this logic happen in LanguageService to make using it dead simple and always return a non-null result.

app/src/main/kotlin/com/hedvig/app/feature/marketing/data/SubmitMarketAndLanguagePreferencesUseCase.kt
```diff
@@ -13,14 +10,13 @@ import com.hedvig.app.util.apollo.toEither
...
-  private val context: Context,
   private val localeManager: LocaleManager,
   private val marketManager: MarketManager,
 ) {
-  suspend fun invoke(language: Language, market: Market) = apolloClient
+  suspend fun invoke() = apolloClient
     .mutation(
       UpdateLanguageMutation(
-        makeLocaleString(language.apply(context), market),
+        makeLocaleString(),
```
StylianosGakis, line 19: Should this also be smth coming from LanguageService instead?

app/src/main/kotlin/com/hedvig/app/feature/marketing/data/UpdateApplicationLanguageUseCase.kt
```diff
@@ -1,19 +1,16 @@
...
 class UpdateApplicationLanguageUseCase(
   private val marketManager: MarketManager,
-  private val localeBroadcastManager: LocaleBroadcastManager,
-  private val context: Context,
 ) {
   operator fun invoke(market: Market, language: Language) {
     marketManager.market = market
-    Language.persist(context, language)
-    localeBroadcastManager.sendBroadcast()
+    AppCompatDelegate.setApplicationLocales(LocaleListCompat.forLanguageTags(language.toString()))
```
StylianosGakis, line 14: This happens both here and inside the LanguageService. Maybe make this use-case have a dep on it and call that instead?

app/src/main/kotlin/com/hedvig/app/feature/settings/SettingsViewModel.kt
```diff
@@ -2,29 +2,42 @@ package com.hedvig.app.feature.settings
...
+
+class ChangeDeviceLanguageAndUploadPreferredLanguageUseCase(
+  private val apolloClient: ApolloClient,
+  private val languageService: LanguageService,
+  private val cacheManager: NetworkCacheManager,
+) {
+  suspend fun invoke(language: Language) {
+    apolloClient
+      .mutation(
+        UpdateLanguageMutation(language.toString(), language.toLocale()),
```
StylianosGakis, line 38: Just a generic question, but what is the reason this mutation exists in the first place? Is it that the backend is somehow storing our language preference as well on top of doing it in app? Is this a result of us not using a `Accept-language` header everywhere in the backend therefore it needs to have this stored somewhere? I am thinking there's just some historical knowledge here that you may have.

StylianosGakis (approved):
Looking very good overall. Would be nice to streamline this even more by not giving us too many options to choose from when dealing with language. If possible even make the language selection retrievable from 1 single place to avoid doing it differently from different places resulting in confusion X amount of time in the future when this won't make that much sense again :D

app/src/engineering/kotlin/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
```diff
@@ -237,7 +237,7 @@ class EmbarkStoryTesterViewModel(
   fun onMarketClick(market: Market) {
     marketManager.market = market
     val language = Language.getAvailableLanguages(market).first()
-    Language.persist(context, language)
+    // TODO: Update application language
```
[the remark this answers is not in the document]
oscarnylander: Nope, this is a mistake, which I need to fix 🙂

app/src/main/kotlin/com/hedvig/app/feature/chat/ui/ChatActivity.kt
oscarnylander: To be honest, I really don't remember. 🤔

StylianosGakis: Haha that's also a valid response. In which case I'd say since I don't think we're actively using this let's just let it get deleted and observe if anything breaks 👀

app/src/main/kotlin/com/hedvig/app/LanguageService.kt
```diff
@@ -0,0 +1,60 @@
...
+    return Language.from(locale.toLanguageTag())
+  }
+
+  fun performOnLaunchLanguageCheck() {
+    val currentLanguage = AppCompatDelegate.getApplicationLocales()
+    // Always default to English, Sweden, if nothing else is set
+    if (currentLanguage.isEmpty) {
+      setLanguage(Language.EN_SE)
+    }
+  }
```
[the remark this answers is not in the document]
StylianosGakis: Yes my thought was the following.
Before this is ever called, we *must* have gone through the SplashScreen first, which means that the `performOneTimeLanguageMigration` has been called before, which means that the old language was saved in `AppCompatDelegate.getApplicationLocales()`. Which in turn would mean that there would be no need to check for this again in the MarketingScreen, since it'd be set for sure already.

The only case I imagine this not being the case is that the migration never happens, due to the sharedprefs not having anything saved in there in the first place. But which case could that storage be empty, is there any? And if there is any such chance, maybe this should be happening also in the splash screen and not in the marketing screen which is not guaranteed to show, so exactly as you said merging them into one function.

With that said, on a fresh app install, what happens there? Is the `AppCompatDelegate.getApplicationLocales()` populated with anything, or do we need to set it ourselves? The docs unfortunately don't seem to mention this as far as I can tell 👀

app/src/main/kotlin/com/hedvig/app/feature/settings/SettingsViewModel.kt
oscarnylander: It is for the backend to know in certain contexts which language to present information, but I'm not quite sure which still use it, if any.

`Accept-Language` as a standard would indeed probably be better.

StylianosGakis (approved):
I feel like this is in a great state to merge already, and if there's anything else that we may think of doing with this we simply do it later in some other PR.

PR #1219, opened 2022-09-09
Feature Notification service register device notification

StylianosGakis:
Adds the second mutation to run along with the previous one when registering a new push notification token.

Also added a constraint on the worker to only work when there's internet. Since there's no way we'd ever want this work to run without internet.


hugokallstrom (approved):
LGTM!

hugokallstrom:
[Here](https://hedvig.atlassian.net/jira/software/c/projects/APP/boards/6?modal=detail&selectedIssue=APP-1909&quickFilter=51) is the ticket for this issue btw!

StylianosGakis:
Ah thank you! I didn't find it when looking 👀

PR #1223, opened 2022-09-13
Rename some safe apollo util functions

StylianosGakis:
Just some renamings that make more sense now with apollo 3.x and how things work. Doing mutation.safeQuery() 

safeQuery -> safeExecute
In reality, safeQuery is an extension on ApolloCall<D> where D is an instance of Operation.Data. Operation.Data can be all of Mutation.Data, Query.Data or Subscription.Data, not just regarding a query. For this, apollo has the name of just "operation" and we can use that instead to indicate this.

safeSubscription -> toSafeFlow
The call-site already looks like `apolloClient.subscription (FooSubscription).safeSubscription` which feels redundant. The subscription is already indicated by the `subscription` call, and what the now `toSafeFlow` function does is simply `toFlow()` plus some things to turn the response into the result type including the errors. It's also already an extension on ApolloCall<D> where D is Subscription.Data so it's type-safe by itself.

QueryResult -> OperationResult
Same argument as the `safeQuery -> safeExecute` change. It's not just for a query result, but for all apollo operations.


hugokallstrom (approved):
LGTM!

PR #1224, opened 2022-09-13
Move apollo safe calls to apollo module

StylianosGakis:
Enables their use in other modules which are not only the :app one.


hugokallstrom (approved):
LGTM!

PR #1228, opened 2022-09-19
APP-1917: Show odyssey claims flow depending on feature toggle

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsFlowActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.content.Intent
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.app.authenticate.AuthenticationTokenService
+import com.hedvig.common.nav.OdysseyNavigationComposable
+import org.koin.android.ext.android.inject
+
+class ClaimsFlowActivity : ComponentActivity() {
```
StylianosGakis, line 13: Is this the way we're going to be working with different flows?
Will it not be more like OdysseyActivity, and then the inner router will take care of any and all navigation within it? Or maybe something else?

app/src/main/AndroidManifest.xml
```diff
@@ -270,6 +270,7 @@
             android:name=".feature.marketing.MarketingActivity"
             android:theme="@style/Hedvig.Theme.Marketing"
             android:windowSoftInputMode="adjustResize" />
+        <activity android:name=".feature.claims.ui.ClaimsFlowActivity" />
```
StylianosGakis, line 273: Do we want this to have a Theme too? Do we rely 100% on everything coming from Odyssey?
Will this mean that if there's anything that is no defined 1:1 equivalent with the local theme, things inside the odyssey world may look off?

StylianosGakis (approved):

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsFlowActivity.kt
hugokallstrom: The way the library API looks like now, we take a `navigationRootUrl` which specifies which router Odyssey will show as "root". For now we only have one flow, claims recording, so this URL can be hard coded in this activity. 

Later on we may want to pass a URL to this activity (and by then also rename the activity). Or we just start a general "/claim" router which builds different flows depending on user input. Whatever we decide, it will be easy to modify in the future.

app/src/main/AndroidManifest.xml
hugokallstrom: We have set a theme for `<application>`, that should be enough right?

StylianosGakis: Yep totally, my bad, read up on this a bit. This means that on the app manifest in 18 places where we define `android:theme="@style/Hedvig.Theme"` those could be removed too!

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsFlowActivity.kt
StylianosGakis: Yeap was just making sure I understand, this sounds reasonable I think.

StylianosGakis:
Just resolved the merge conflict too, and put the deps in alphabetical order as the rest are which made this conflict in the first place.

ci-jitpack-properties.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+set +uex
+
+echo "" >> ~/.gradle/gradle.properties
+echo "odysseyRepositoryUsername=${JITPACK_AUTH_TOKEN}" >> ~/.gradle/gradle.properties
+echo "odysseyRepositoryPassword=" >> ~/.gradle/gradle.properties
```
StylianosGakis, line 6: The empty password is what the `PasswordCredentials` is expecting I guess right? Or is this leftover from something we no longer need?

StylianosGakis (approved):

PR #1232, opened 2022-09-27
Splash screen

StylianosGakis:
Removes the artificial delay and launches the app faster, especially after coming back after you've backgrounded the app by pressing back button on Android 12+

Might have to look into changing the icon used not be the mipmap but an actual SVG
The mipmap made it so that
Pixel: It worked fine
Samsung: It looks very small, layout inspection makes this obvious
Xiaomi: The outer ring gets shrinked somehow (no clue why? maybe one svg will clear this issue)

Longer term ideas, make MainActivity be the actual LoggedInActivity, which itself can route you to other places if need be. To avoid showing the splash screen every time after going *back* out of the app and in again.

Leaving this here to revisit sometime later if we wish to do so.


StylianosGakis:
Some screenshots. Samsung is asa always special and renders the icon smaller than normal. But this works all the way to API 23 (21 really) as tested on an emulator.

| Pixel API 33 | Pixel API 33 + grid | Samsung API 32 | Samsung API 32 + grid | Pixel API 23 |
|:---:|:---:|:---:|:---:|:---:|
| image | image | image | image | image | |

StylianosGakis:
Very interestingly, while the docs say that we got 72dp space for our icon, when we do in fact use all of it Samsung cuts it off on the top/down/left/right edges of the circle 🙈
So keeping it 70dp as the screenshots above show should work.

Table with the icon being 72dp:
| Pixel API 33 | Samsung API 31 | Pixel API 23 |
|:---:|:---:|:---:|
| image | image | image |

oscarnylander (approved):
lgtm

hugokallstrom (approved):

PR #1233, opened 2022-09-27
Fix relying on `.first()` on list which can very well be empty

StylianosGakis:
https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/de950c400675366c7eaca57c86487b2c?time=1661682966000:1664274966000&sessionEventKey=632A11AD02A2000179993FC3B51FDF76_1723988857574285009

main diff is changing `private fun OfferModel.getBundleVariant(` return type to a nullable one, and handle it from there.


hugokallstrom (approved):
LGTM!

PR #1235, opened 2022-09-30
Remove Showkase

StylianosGakis:
We're really not using this atm plus it's introducing kapt into our build. If we are to use it again we should go with the KSP version. But since we're not using it, it's easier to simply delete it for the time being, we can always revert this commit when we reintroduce it.


hugokallstrom (approved):
LGTM

PR #1237, opened 2022-10-03
CLX-48: AdjustResize claims flow activity

hugokallstrom:
Add adjustResize in android manifest to allow for keyboard to not push view off screen


StylianosGakis (approved):
👍

PR #1241, opened 2022-10-04
CLX-59: Update proguard rules

hugokallstrom:


hugokallstrom:
> If there was some place where you fetched these proguard rules from, could you please paste a link to that source as a comment inline in this file? I've found that looking at this file makes no sense even like a week later after we've edited it, since it's not information we need too often. So having an easy way to reference back to where the rules come from is quite good imo.

This was needed because yoga is used by odyssey components, and we are never calling those components in the app code - hence R8 will remove yoga code for optimisation.

StylianosGakis:
Yeah I understand why we need to instruct R8 to do this, but I was mostly thinking that these rules then become stale and if we bump a library version and they update the rules we wouldn't know about it.
Also even if we do and we add the new configuration, we wouldn't the feel confident to remove the old lines in this file since there's no guardrails to protect us from removing something crucial, so we let all the code in here untouched.
So if the lines that exist in here are accompanied with a link to where these rules come from then we will be able to more confidently make these changes when there is such a need.

hugokallstrom:
The update in this commit is not from a libraries rules. Its because how use odyssey and how R8 works. I will add some comments explaining why we need that line.

StylianosGakis:
Aha so yoga themselves don't suggest this change, it was just something we specifically encountered in this case.
I wonder why it's the case then that only we have had this issue but not other people using yoga.
This dependency is expected to be called from Android, which is normal for all apps to be running R8 before releasing, so I'd imagine if there's something special one needs to do to retain some class name from being optimized away or renamed then it'd be part of the yoga rules, that's why I'm asking this basically.

hugokallstrom:
The thing is we are using yoga as part of our SDUI system. [Here](https://github.com/facebook/litho/issues/592#issuecomment-879237766) is a similar problem, but with facebook litho.

PR #1242, opened 2022-10-05
CLX-63: Remove insurely

hugokallstrom:
Remove insurely. Insurely was used for onboarding, but since we switched to using the web onboarding we no longer need insurely.


StylianosGakis (approved):
RIP 😅

PR #1243, opened 2022-10-06
CLX-57: Do not show honesty pledge for odyssey claims 

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/claims/ui/commonclaim/CommonClaimActivity.kt
```diff
@@ -5,27 +5,26 @@ import android.content.Intent
...
-import com.hedvig.app.feature.claims.ui.pledge.HonestyPledgeBottomSheet
+import com.hedvig.app.feature.claims.ui.startClaimsFlow
 import com.hedvig.app.util.extensions.compatSetDecorFitsSystemWindows
-import com.hedvig.app.util.extensions.view.applyNavigationBarInsets
-import com.hedvig.app.util.extensions.view.applyStatusBarInsets
-import com.hedvig.app.util.extensions.view.disable
-import com.hedvig.app.util.extensions.view.enable
-import com.hedvig.app.util.extensions.view.setHapticClickListener
-import com.hedvig.app.util.extensions.view.setupToolbarScrollListener
+import com.hedvig.app.util.extensions.view.*
```
StylianosGakis, line 16: ktlint will prob complain about this

It seems that now it failed here https://github.com/HedvigInsurance/android/actions/runs/3197099130/jobs/5219881747 with:
`The following Gradle properties are missing for 'odysseyRepository' credentials:`
so could it be that CI wasn't setup for it maybe, so it didn't have a chance to run

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsUtil.kt
```diff
@@ -0,0 +1,27 @@
+package com.hedvig.app.feature.claims.ui
+
+import android.content.Context
+import android.content.Intent
+import androidx.fragment.app.FragmentManager
+import com.hedvig.app.feature.claims.ui.pledge.HonestyPledgeBottomSheet
+import com.hedvig.hanalytics.HAnalytics
+
+suspend fun startClaimsFlow(
```
StylianosGakis, line 9: This seems to be called only from 1 place if I see correctly, maybe just move it as a private function there?

StylianosGakis (approved):
lgtm

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsUtil.kt
hugokallstrom: Its called from common claims and from home

PR #1244, opened 2022-10-06
Pure hanalytics module

StylianosGakis:
This was done in the process of removing all Android dependencies on
that module in order to re-use a lot of its logic in odyssey, so might
as well merge these changes here too.

In the process, a core-common was split into core-common and a core-commonAndroid to allow having an android agnostic common module as well.
core-datastore was also turned into a jvm module since it didn't need android for anything.


hanalytics/hanalytics-android/src/main/kotlin/com/hedvig/android/hanalytics/AndroidHAnalyticsService.kt
```diff
@@ -45,12 +40,12 @@ internal class HAnalyticsServiceImpl(
       "properties" to event.properties,
       "graphql" to event.graphql,
     )
+    d { "Stelios: requestJsonObject:$requestJsonObject" }
```
hugokallstrom, line 43: Used when debugging I suppose

hugokallstrom (approved):
LGTM

hanalytics/hanalytics-android/src/main/kotlin/com/hedvig/android/hanalytics/AndroidHAnalyticsService.kt
StylianosGakis: I do this way too often 🙈
Yes will remove!

PR #1246, opened 2022-10-07
Android libraries without staging build type

StylianosGakis:
Figured a way to remove build types from android library modules. They don't need to know about staging, pullrequest or whatever anyway.


hugokallstrom (approved):
Cool!

PR #1249, opened 2022-10-13
Update odyssey

hugokallstrom:


app/build.gradle.kts
```diff
@@ -271,4 +271,5 @@ dependencies {
 
   implementation(libs.datadog.sdk)
   implementation(libs.odyssey)
+  implementation(libs.kotlin.reflect)
```
StylianosGakis, line 274: Interesting that we need to add this on our end, I wonder why this didn't work when defined as `api` on the library itself, but oh well not big deal to add it here I think.

StylianosGakis (approved):

PR #1252, opened 2022-11-03
CLX-120: fix updating language before logging in

StylianosGakis:
Call the mutation only after there was an auth success on BankID success, and on simple sign success. Basically looked at the options that start from here https://github.com/HedvigInsurance/android/blob/ce396d51a817a5d620b5762e627e9fccb9a4ef64/app/src/main/kotlin/com/hedvig/app/util/navigation/NavUtil.kt#L20-L32 , https://github.com/HedvigInsurance/android/blob/dabb98db9d2042e9b61da630a6ab4a4e97d85e44/app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt#L100-L111 and for the OTP case.

We're still setting these values in the marketing screen, and in the settings screen and the single source of truth about them is `LanguageService` where this use case is fetching the info from.

hasSelectedMarket was only being set in two places but seemingly not used anywhere, just removed that.

### DIFF
--- a/app/src/main/kotlin/com/hedvig/app/feature/marketing/data/UploadMarketAndLanguagePreferencesUseCase.kt
+++ b/app/src/main/kotlin/com/hedvig/app/feature/marketing/data/UploadMarketAndLanguagePreferencesUseCase.kt
@@ -21,8 +21,8 @@ class UploadMarketAndLanguagePreferencesUseCase(
       .safeExecute()
       .toEither()
       .fold(
-        { i { "Succeeded uploading language preferences to language:$languageTag | locale:$locale" } },
-        { e { "Failed to to upload language preferences to language:$languageTag | locale:$locale" } },
+        ifLeft = { e { "Failed to to upload language preferences to language:$languageTag | locale:$locale" } },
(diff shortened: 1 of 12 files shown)


app/src/main/kotlin/com/hedvig/app/feature/marketing/data/UploadMarketAndLanguagePreferencesUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+) {
+  suspend fun invoke() {
+    val languageTag: String = languageService.getLocale().toLanguageTag()
+    val locale: Locale = languageService.getGraphQLLocale()
+    apolloClient
+      .mutation(UpdateLanguageMutation(languageTag, locale))
+      .safeExecute()
+      .toEither()
+      .fold(
+        { i { "Succeeded uploading language preferences to language:$languageTag | locale:$locale" } },
```
hugokallstrom, line 24: Good that we are at least logging something here, instead of doing nothing when this errors 👍

hugokallstrom (approved):

PR #1253, opened 2022-11-03
CLX-122: Update to odyssey which needs an image loader

StylianosGakis:
Needs Odyssey to make a new release and bump version before merge.

Also removes all last pieces where we were using functions from the `coil` artifact which can rely on the singleton ImageLoader instance.

Import changes since we took in the functions that are inside `coil` into "app/src/main/kotlin/com/hedvig/app/ui/coil/CoilUtil.kt". This is however now safe since there's no default value for `ImageLoader` which means we *need* to pass it in.

https://github.com/coil-kt/coil/issues/1525 maybe for updates regarding why these functions are not part of the `coil-base` artifact instead, because they could be since they don't need to use any of the singleton mechanisms.


app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsFlowActivity.kt
```diff
@@ -30,6 +32,7 @@ class ClaimsFlowActivity : ComponentActivity() {
         apiUrl = odysseyUrl,
         authorizationToken = token,
         locale = languageService.getLocale().toString(),
+        imageLoader = imageLoader,
```
StylianosGakis, line 35: This adapts to the new Odyssey API

hugokallstrom (approved):
Nice!

PR #1297, opened 2022-11-14
Bump odyssey

StylianosGakis:


core-navigation/src/main/kotlin/com/hedvig/android/navigation/Navigator.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.android.navigation
+
+import android.content.Context
+
+class Navigator(
```
hugokallstrom, line 5: Would it make sense to add other navigation related code here? Like opening web urls, keyboard etc? Could keep that in mind :)

feature-odyssey/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+  id("hedvig.android.library")
+  id("hedvig.android.library.compose")
+  id("hedvig.android.ktlint")
+}
+
+dependencies {
```
hugokallstrom, line 7: Nice to have this as a separate module 👍

gradle/libs.versions.toml
```diff
@@ -184,7 +184,7 @@ materialComponents = { module = "com.google.android.material:material", version.
 mockk-android = { module = "io.mockk:mockk-android", version.ref = "mockk" }
 mockk-jvm = { module = "io.mockk:mockk", version.ref = "mockk" }
 moneta = { module = "org.javamoney.moneta:moneta-core", version.ref = "moneta" }
-odyssey = { module = "com.github.HedvigInsurance.odyssey:common-android", version.ref = "odyssey" }
```
hugokallstrom, line 187: I really like moving to github packages. What is the difference compared to jitpack? Is the build time the same?

hugokallstrom (approved):

gradle/libs.versions.toml
StylianosGakis: My main goal was 2 things.
1. Have easy to fetch dev build for *every* merge on main, without having to make a GH release which is what Jitpack needed. This gives us more freedom to tinker with the name of those releases, plus standardize how one makes a new Odyssey client release just by following the RELEASING.md file inside Odyssey.
2. Be in full control of the build. Both in terms of seeing how fast it goes (with anecdotal data, it does feel faster with GH actions, running on a macos runner). And in terms of if something goes bad we can see in the GH Actions logs what is it that goes wrong.

Nice to have side effects of this change however are also that:

* We don't need to rely on availability of Jitpack. Lately I just had this one day where it was not showing the build logs etc. to me and I got quite annoyed tbh, because I had no idea what was going wrong and no idea that it was Jitpack being down the problem.
* We don't need to first go on say Android and *ask* for the build version for Jitpack to then start building it. If you wanted to fetch the latest version fast this was a big blocker, and had to use publishing to mavenLocal() instead. Now it kicks off the build immediately on merge to main. When working on the feature yourself of course it's still better to work with mavenLocal(), nothing will beat that, but this is mostly for the cases where you just want the latest main.
* No extra place to host the artifacts at Jitpack, instead do what every other library at Hedvig does, which is host itself in GitHub Packages, under the Hedvig organization
* We can simply rely on the `common` and not `common-android`, since we're publishing all targets now, and simply asking for `common` knows by itself which one to fetch, in our case android since we are on an Android target. This means that if we ever make any other library which wants to consume Odyssey, it can even be a KMP library and it will all work just fine, we're no longer *only* exposing the android .aar artifact.

Potential negatives:

* Not sure how much space this takes in our GH Packages and if it will be a problem.

core-navigation/src/main/kotlin/com/hedvig/android/navigation/Navigator.kt
StylianosGakis: Yeah in general there's definitely room to explore what we can do for navigation in our project.
In this case this was needed because I needed access to the `Context.startChat()` function but it lives inside the :app module, so :feature-odyssey couldn't see it.
Now, I could move that entire function to the newly created :core-navigation, but `startChat()` needs a reference to the ChatActivity because that's how it knows what to launch, and it doesn't make sense to extract that out of :app this way.
So this way, :app keeps the logic of how to navigate to the ChatActivity, by keeping the `startChat()` function in there, and sets up this `Navigator` by passing it to its lambda, while making this `Navigator` available to whichever other module needs to navigate to chat.
But yeah, since I am not 100% confident with what's the best solution to this, especially with Activity navigation as opposed to Compose-only navigation like we do inside of Odyssey, I think this is good enough for now.

gradle/libs.versions.toml
StylianosGakis: Yeah that was my main frustration too.
Luckily it turned out making GH Packages as the host for the artifacts isn't even that hard. Before trying it out it seemed impossible, but it all worked out perfectly fine in the end.
Now if we wanted the library to be OSS which wouldn't work in GH Packages, maybe that'd require some different kind of work. But I'll learn that when/if I need to in the future :D

core-navigation/src/main/kotlin/com/hedvig/android/navigation/Navigator.kt
hugokallstrom: Makes total sense

PR #1304, opened 2022-11-18
CLX-166: Pass itemtype from phone common claim card

hugokallstrom:
This is used to get questions related to the type of claim. For phone claims we want to ask specifically for location and product info, but not for "general" claims.


app/src/main/kotlin/com/hedvig/app/feature/claims/ui/commonclaim/CommonClaimActivity.kt
```diff
@@ -90,3 +91,8 @@ class CommonClaimActivity : AppCompatActivity(R.layout.activity_common_claim) {
       }
   }
 }
+
+private fun CommonClaimsData.itemTypeFromId() = when (id) {
+  "6" -> "PHONE"
```
StylianosGakis, line 96: So 6 has some significance on the backend I presume right?

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsUtil.kt
```diff
@@ -14,9 +14,15 @@ suspend fun startClaimsFlow(
   context: Context,
   fragmentManager: FragmentManager,
   registerForResult: ((Intent) -> Unit)? = null,
+  itemType: String? = null,
```
StylianosGakis, line 17: Could we potentially create a class for this so it's a bit more straightforward about what it is? So the type could be itemType: ItemType? = null
```kotlin
@JvmInline
value class ItemType(val name: String)
```
But it is true that we then pass it through a bundle, I wonder how this interaction works, didn't find anything by googling. Maybe it'd need to be
```kotlin
@kotlinx.parcelize.Parcelize
@JvmInline
value class ItemType(val name: String) : Parcelable
```

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/ClaimsFlowActivity.kt
```diff
@@ -26,8 +28,15 @@ class ClaimsFlowActivity : ComponentActivity() {
...
     requireNotNull(token)
     val odysseyUrl = intent.getStringExtra(ODYSSEY_URL_KEY) ?: error("ODYSSEY_URL_KEY needs to be passed in")
+    val itemType = intent.getStringExtra(EXTRA_ITEM_TYPE)
+
     val locale = languageService.getLocale().toString()
 
+    val scopeValues = ScopeValues()
+    if (itemType != null) {
+      scopeValues.setValue(InitialDataScopeValueKey, mapOf("itemType" to itemType))
+    }
```
StylianosGakis, line 38: So if itemType is null, we simply send an empty ScopeValue object right? Which I guess is perfectly fine if there's nothing to send anyway.

StylianosGakis (approved):
lgtm, don't *need* to do the value class thing, especially since I'm not sure if it plays well with bundles

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/commonclaim/CommonClaimActivity.kt
hugokallstrom: Yeah its the id for broken phone common claim card

app/src/main/kotlin/com/hedvig/app/feature/claims/ui/ClaimsUtil.kt
hugokallstrom: Fixing!

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/ClaimsFlowActivity.kt
hugokallstrom: Exactly!

PR #1307, opened 2022-11-22
CLX-157: Implement start login endpoint

hugokallstrom:


StylianosGakis (commented):
I feel like this is a signal for us to look into Retrofit.
Have you tried it before? I think we're starting to do quite a lot of non GQL requests at this point where introducing Retrofit to do the heavy lifting for us like mapping the objects, awaiting responses and so on is totally worth it.
What do you think?

auth/build.gradle.kts
```diff
@@ -1,9 +1,13 @@
...
   id("hedvig.android.library")
   id("hedvig.android.ktlint")
+  alias(libs.plugins.serialization)
 }
 
 dependencies {
   implementation(libs.koin.android)
   implementation(libs.okhttp.core)
+  implementation(libs.serialization.json)
+  implementation(projects.coreCommon)
```
StylianosGakis, line 12: I usually try to keep the `projects.` dependencies at the top and with a gap with the rest, so that at a glance you can see which other modules a module depends on. So smth like like 

```suggestion
  implementation(projects.coreCommon)

  implementation(libs.koin.android)
  implementation(libs.okhttp.core)
  implementation(libs.serialization.json)
```

StylianosGakis (approved):
Lgtm!

PR #1308, opened 2022-11-22
CLX-158: Poll login status

hugokallstrom:


auth/src/main/kotlin/com/hedvig/android/auth/NetworkAuthRepository.kt
```diff
@@ -33,19 +42,42 @@ class NetworkAuthRepository(
...
+            .newCall(request)
+            .await()
+            .toLoginStatusResult()
+
+          emit(loginStatusResult)
+
+          isPending = loginStatusResult.isPending()
+          if (isPending) {
+            delay(POLL_DELAY_MILLIS)
+          }
```
StylianosGakis, line 75: Hmm I *think* we can instead have this be a `while(true)` and when isPending is false we  `break` the look instead of relying on this mutable local state? I *think* there's no concurrency issues here, but it'd still be a bit more clear that way.

auth/src/main/kotlin/com/hedvig/android/auth/network/LoginStatusResponse.kt
```diff
@@ -0,0 +1,42 @@
...
+
+fun Response.toLoginStatusResult(): LoginStatusResult {
+  val responseBody = body?.string()
+  val result = if (isSuccessful && responseBody != null) {
+    Json.decodeFromString<LoginStatusResponse>(responseBody).toLoginStatusResult()
+  } else {
+    LoginStatusResult.Failed(message = message)
+  }
+  return result
+}
```
StylianosGakis, line 29: Hmm what do you think about putting these functions inside the file where it was used?
Otherwise I'd imagine we'd start getting quite a namespace pollution on `Response`, getting a lot of not always relevant extension functions on it in places of the module where it wouldn't make a lot of sense.
But not sure how exactly it'd look like and if we do want access to this function in other places, so giving the idea here to get feedback.

auth/src/main/kotlin/com/hedvig/android/auth/network/LoginStatusResponse.kt
```diff
@@ -0,0 +1,42 @@
...
+
+private fun LoginStatusResponse.toLoginStatusResult() = when (status) {
+  LoginStatusResponse.LoginStatus.PENDING -> LoginStatusResult.Pending(seBankidHintCode)
+  LoginStatusResponse.LoginStatus.FAILED -> LoginStatusResult.Failed("Login status failed")
+  LoginStatusResponse.LoginStatus.COMPLETED -> {
+    require(authorizationCode != null) {
+      "Login status completed but did not receive authorization code"
+    }
```
StylianosGakis, line 37: Maybe we can consider introducing an extra option for this case, where it's clearly backend error but we can still model it as another LoginStatusResult, or even just the existing Failed

StylianosGakis (approved):
💪

auth/src/main/kotlin/com/hedvig/android/auth/network/LoginStatusResponse.kt
hugokallstrom: Good point, fixing

PR #1309, opened 2022-11-22
CLX-159: Complete login attempt

hugokallstrom:


auth/src/main/kotlin/com/hedvig/android/auth/network/SubmitAuthorizationCodeResponse.kt
```diff
@@ -0,0 +1,43 @@
...
+  @SerialName("refresh_token")
+  val refreshToken: String,
+  @SerialName("refresh_token_expires_in")
+  val refreshTokenExpiresIn: Int,
+)
+
+fun Response.toAuthTokenResult(): AuthTokenResult {
+  val responseBody = body?.string()
+  val result = if (isSuccessful && responseBody != null) {
+    Json.decodeFromString<SubmitAuthorizationCodeResponse>(responseBody).toAuthAttemptResult()
```
StylianosGakis, line 27: One thing I thought of for this, we use the global `Json` in all these samples. Is the default configuration enough? In Odyssey we had to make one ourselves for a couple of reasons if I remember correctly. Will just have to double check if we want to do that here too ☺️

auth/src/main/kotlin/com/hedvig/android/auth/network/SubmitAuthorizationCodeResponse.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+private fun SubmitAuthorizationCodeResponse.toAuthAttemptResult() = AuthTokenResult.Success(
+  accessToken = AccessToken(
+    token = accessToken,
+    expiryInSeconds = accessTokenExpiresIn,
+  ),
+  refreshToken = RefreshToken(
+    token = refreshToken,
+    expiryInSeconds = refreshTokenExpiresIn,
```
StylianosGakis, line 41: A thought again. I'm not sure exactly how we'll be using these expiration times, but I'd imagine that having the time in a number won't be that useful by itself, since when checking it out later it won't mean much if we don't also know the starting time of when this will expire.
I wonder if it's a good idea to take the milliseconds it will expire in, and at this level immediately truth it into a kotlinx.datetime.Instant object (or kotlinx.datetime.WhateverSuitsThis) so that later we can always compare that with current time 🤔

StylianosGakis (approved):
Looks good ☺️

auth/src/main/kotlin/com/hedvig/android/auth/network/SubmitAuthorizationCodeResponse.kt
hugokallstrom: Good idea! Ill add that in another PR when we actually use these if thats ok :)

StylianosGakis: Yeah and *if* we use them too, we might not if we don't want to do anything special with them and simply try with whatever we have, and re-fetch if it's gonna expire.

PR #1311, opened 2022-11-22
CLX-194: Submit otp

hugokallstrom:


auth/src/main/kotlin/com/hedvig/android/auth/AuthRepository.kt
```diff
@@ -32,23 +33,26 @@ sealed interface AuthAttemptResult {
...
   data class OtpProperties(
     val id: String,
-    val statusUrl: String,
+    val statusUrl: StatusUrl,
     val validationUrl: String,
   ) : AuthAttemptResult
 }
 
+@JvmInline
+value class StatusUrl(val url: String)
```
StylianosGakis, line 54: Love this way of avoiding passing around error-prone primitives on APIs like this 🤩

auth/src/main/kotlin/com/hedvig/android/auth/AuthRepository.kt
```diff
@@ -67,6 +71,13 @@ sealed interface LoginStatusResult {
 
 fun LoginStatusResult.isPending() = this is LoginStatusResult.Pending
 
+
+sealed interface SubmitOtpResult {
+  data class Error(val message: String) : SubmitOtpResult
+  data class Success(val loginAuthorizationCode: LoginAuthorizationCode) : SubmitOtpResult
+}
```
StylianosGakis, line 78: Feel free to not do this if you don't want to, but something I sometimes like in these cases is instead to have
```kotlin
data class SubmitOtpError(val message: String)
data class SubmitOtpSuccess(val loginAuthorizationCode: LoginAuthorizationCode)
```
And have the function return `Either<SubmitOtpError, SubmitOtpSuccess>` instead.
This is so that then on the call site you got the full power of Either to do stuff instead of having to manually deal with the custom result type.
Along with being able to construct this return type by using the `either {}` api like we do in [here](https://github.com/HedvigInsurance/android/blob/1a49ac6baed33b03680f6ae889ad337ccdbff775/app/src/main/kotlin/com/hedvig/app/feature/offer/OfferRepository.kt#L39-L58) for example. So gives some goodies on both receiving that type, and creating it.

It does of course depend on how we want to use this, maybe sometimes it doesn't make sense, but just saying :D

StylianosGakis (approved):
lgtm!

auth/src/main/kotlin/com/hedvig/android/auth/AuthRepository.kt
StylianosGakis: Agree yes!

PR #1312, opened 2022-11-22
CLX-195: Logout

hugokallstrom:


StylianosGakis (approved):
Lgtm.

So this is for the network related parts I guess. On the client side, if the LogoutResult ends up being Error, we had discussed with Fredrik that then just tossing our tokens away and deleting them from storage should also be fine, since then they'll expire themselves.
This is so that one can still log out instantaneously without requiring  a network connection. But that's for another PR I guess, where we deal with storing the received tokens in some sort of persistent storage right? Basically when we also write the client to the AuthRepository

hugokallstrom:
> Lgtm.
> 
> So this is for the network related parts I guess. On the client side, if the LogoutResult ends up being Error, we had discussed with Fredrik that then just tossing our tokens away and deleting them from storage should also be fine, since then they'll expire themselves. This is so that one can still log out instantaneously without requiring a network connection. But that's for another PR I guess, where we deal with storing the received tokens in some sort of persistent storage right? Basically when we also write the client to the AuthRepository

Yes lets figure something out when we replace the current impl in the client code :)

PR #1318, opened 2022-11-28
CLX-209: Refresh auth tokens

hugokallstrom:


auth/src/main/kotlin/com/hedvig/android/auth/network/AccessTokenAuthenticator.kt
```diff
@@ -0,0 +1,43 @@
...
+  private val authRepository: AuthRepository,
+) : Authenticator {
+
+  override fun authenticate(route: Route?, response: Response): Request? {
+    val refreshToken = authenticationTokenService.refreshToken ?: return null
+    val result = synchronized(this) {
+      runBlocking {
+        authRepository.submitAuthorizationCode(refreshToken.token)
+      }
+    }
```
StylianosGakis, line 23: Hmm this syncs on not doing the network request in parallel, but what are we protecting from exactly?
If two of them come at the same time, get the `refreshToken` in a varible, one enters this, finishes with AuthTokenResult.Error and goes down and clears the tokens inside authService, then the other one resuming will try and reuse the old refreshToken which was just cleared.
What we want to do with the sync I guess would be not to use old data like that.
As a first step adding the `refreshToken` inside the sync block would be a good idea, but I think also the parts further down where the authTokenService gets updated information should also be inside the synchronized block, to make sure that the new request gets the latest data available.

This https://github.com/StylianosGakis/paw-finder/blob/daa332532860280d64e0580c7fe423ec4c041859/app/src/main/java/com/stylianosgakis/pawfinder/api/authenticator/TokenAuthenticator.kt#L27-L56 is how I've done this in the past, with a mutex on getting the token, and if you get the same twice (nobody tried to do this in-between) then you go ahead and try and refresh the token. If it's not the same, it means that in the meantime another call went through and refreshed the token while you were waiting for the mutex.

where `withAuthenticationToken` in that code base is just appending the token to the request:
```
fun Request.withAuthenticationToken(
    token: String,
): Request {
    return this.newBuilder()
        .header(AUTHORIZATION_HEADER_NAME, "Bearer $token")
        .build()
}
```

StylianosGakis (approved):
Got two comments, I think it'd be good to address the null token one in case I'm missing smth, and the comment about sync may not affect correctness, but I'm not sure, always tricky with doing stuff in parallel. I'd say better to address it too.

auth/src/main/kotlin/com/hedvig/android/auth/SharedPreferencesAuthenticationTokenService.kt
```diff
@@ -12,4 +14,20 @@ internal class SharedPreferencesAuthenticationTokenService(
...
+      sharedPreferences.edit()
+        .putString(SHARED_PREFERENCE_REFRESH_TOKEN, value?.token?.code)
+        .apply()
+
+      sharedPreferences.edit()
+        .putInt(SHARED_PREFERENCE_REFRESH_TOKEN_EXPIRY, value?.expiryInSeconds ?: 0)
+        .apply()
+    }
+    get() {
+      val code = sharedPreferences.getString(SHARED_PREFERENCE_REFRESH_TOKEN, "") ?: ""
```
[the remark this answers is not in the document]
hugokallstrom: Yes very good catch, fixing

auth/src/main/kotlin/com/hedvig/android/auth/network/AccessTokenAuthenticator.kt
hugokallstrom: Makes sense, fixing! That project looks great btw 😮

PR #1319, opened 2022-11-28
CLX-206: Replace zignsec

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/zignsec/SimpleSignAuthenticationActivity.kt
```diff
@@ -48,11 +49,16 @@ class SimpleSignAuthenticationActivity : AppCompatActivity(R.layout.simple_sign_
...
-        viewModel.subscribeToAuthSuccessEvent().collect()
+    viewModel.statusUrl.observe(this) { statusUrl ->
+      lifecycleScope.launch {
+        lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
+          viewModel.subscribeToAuthSuccessEvent(statusUrl)
+            .distinctUntilChanged()
+            .collect()
+        }
       }
     }
```
StylianosGakis, line 60: the outer observe on LiveData won't cancel the last thing that was run, meaning that if say statusUrl brings one value, then another, the `subscribeToAuthSuccessEvent` will be running for both of them at the same time. Will this be a problem in our case? I'd guess yes since then the old URL might trigger an error, and then the that would send an Error event and so on.
Since this part inside the activity only exists so that the flow only is read when the view is at least on STARTED. I'd suggest this logic is moved inside the body of `subscribeToAuthSuccessEvent` instead.
So like that function itself can listen on statusUrl, then if we need it to be LiveData, there's a `asFlow()` function, and can do `collectLatest()` on it. Smth like:

```
fun subscribeToAuthSuccessEvent(): Flow<*> {
    return statusUrl.asFlow().collectLatest { statusUrl ->
        authRepository.observeLoginStatus(statusUrl).collect { loginStatusResult ->
            when (loginStatusResult) {
                is LoginStatusResult.Completed -> {
                    onSimpleSignSuccess(loginStatusResult)
                    _events.postValue(Event.Success)
                }
                is LoginStatusResult.Failed -> _events.postValue(Event.Error)
                is LoginStatusResult.Pending -> {}
            }
        }
    }
}
```

This'd make it so that if there's a new `statusUrl` emission, due to the collectLatest, the previous observeLoginStatus (with the old statusUrl) would end itself, and there's no need for the activity to be able to see the statusUrl either that way, can be private I think.

auth/src/main/kotlin/com/hedvig/android/auth/NetworkAuthRepository.kt
```diff
@@ -37,6 +37,8 @@ class NetworkAuthRepository(
     val requestBody = buildJsonObject {
       put("method", loginMethod.name)
       put("country", market)
+      put("callbackSuccess", "https://google.se?q=success")
+      put("callbackFailure", "https://google.se?q=failure")
```
StylianosGakis, line 41: So from what I understand, we also need to edit our ZignSec webview to match these. As a first step it could be just changing the "failure" -> "fail"
https://github.com/HedvigInsurance/android/blob/4074740bd0cb72fdb37693c492ea55ceb377116c/app/src/main/kotlin/com/hedvig/app/feature/zignsec/ui/ZignSecWebViewFragment.kt#L47
But in reality we could even match the entire URL so we have full control over this, instead of matching just part of the URL. But regardless, the "failure" and "fail" need to match is my understanding there.

StylianosGakis (approved):
2 comments that need some addressing imo 😊

PR #1320, opened 2022-11-28
CLX-210: Replace otp apis

hugokallstrom:


auth/src/main/kotlin/com/hedvig/android/auth/NetworkAuthRepository.kt
```diff
@@ -26,6 +27,7 @@ class NetworkAuthRepository(
 
   private val jsonBuilder = Json {
     ignoreUnknownKeys = true
+    explicitNulls = false
```
StylianosGakis, line 30: Which part of the API responds without a specific key at all and we need to infer it as null? Was it the otp stuff? Is that potentially an error on the backend side or is it expected that it doesn't send that thing at all instead of sending it as null?

auth/src/main/kotlin/com/hedvig/android/auth/network/ResendOtpResponse.kt
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.android.auth.network
+
+import com.hedvig.android.auth.ResendOtpResult
+import okhttp3.Response
+
+fun toResendOtpResult(response: Response): ResendOtpResult {
+  val responseBody = response.body?.string()
+  return if (response.isSuccessful && responseBody != null) {
```
StylianosGakis, line 8: The spec for this reads:
```
POST <auth-base>/member-login/{id}/otp-resend
// no input, no output
```
Does `no output` also mean that responseBody is null, or how does this get shown exactly in reality I wonder? Just so we don't interpret a successful but null body as error while we don't need to.

StylianosGakis (approved):

auth/src/main/kotlin/com/hedvig/android/auth/NetworkAuthRepository.kt
```diff
@@ -107,6 +109,18 @@ class NetworkAuthRepository(
     }
   }
 
+  override suspend fun resendOtp(resendUrl: String): ResendOtpResult {
+    val request = createPostRequest("$url${resendUrl}", ByteArray(0).toRequestBody())
```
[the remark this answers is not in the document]
hugokallstrom: We need to pass a request body when doing this post, and here we are just creating an empty body.

auth/src/main/kotlin/com/hedvig/android/auth/NetworkAuthRepository.kt
hugokallstrom: This is because the response can omit some fields when we get error responses.

auth/src/main/kotlin/com/hedvig/android/auth/network/ResendOtpResponse.kt
hugokallstrom: Good catch, it should be enough to just check for successful response here

PR #1321, opened 2022-11-29
CLX-183: Handle cross sell web action and open browser with uri

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
```diff
@@ -73,6 +74,10 @@ class InsuranceFragment : Fragment(R.layout.fragment_insurance) {
             ?.navigate(requireContext())
             ?.also { insuranceViewModel.crossSellActionOpened() }
 
+          viewState.navigateWeb
+            ?.let { activity?.openWebBrowser(it) }
+            ?.also { insuranceViewModel.crossSellActionOpened() }
```
StylianosGakis, line 79: Inside crossSellActionOpened() we'd need to then also clear the `navigateWeb` variable so that it doesn't stay true right? Otherwise if going back here the navigateWeb would still be there, and this would retrigger. That's what I gather from the code at least.

app/src/main/kotlin/com/hedvig/app/feature/crossselling/ui/detail/CrossSellFaqActivity.kt
```diff
@@ -77,6 +79,10 @@ class CrossSellFaqActivity : AppCompatActivity() {
     navigateEmbark
       ?.navigate(this@CrossSellFaqActivity)
       ?.also { viewModel.actionOpened() }
+
+    navigateWeb
+      ?.let(::openWebBrowser)
+      ?.also { viewModel.actionOpened() }
```
StylianosGakis, line 85: This one is fine, it's cleared by creating a new ViewState()

app/src/main/kotlin/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt
```diff
@@ -33,6 +33,9 @@ data class CrossSellData(
 
     @Parcelize
     object Chat : Action()
+
+    @Parcelize
+    data class Web(val url: String): Action()
```
StylianosGakis, line 38: When do we get this from backend? The `CrossSellData::from` function seems to only ever create Embark and Chat here https://github.com/HedvigInsurance/android/blob/a9570c2c033f332eb78ca14d9dd87cfdea2a6567/app/src/main/kotlin/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt#L57-L59.
Is this a TODO for later, or maybe I'm missing smth

StylianosGakis (approved):
👍 I just had 2 small comments where in #1 I need some clarification and in #2 there might be a small loop with opening the web action

app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/tab/InsuranceFragment.kt
hugokallstrom: Good catch, fixing 🙏

PR #1324, opened 2022-12-09
CLX-219: Impersonation from hope

hugokallstrom:


app/src/engineering/kotlin/com/hedvig/app/feature/impersonation/ImpersonationReceiverActivity.kt
```diff
@@ -41,15 +46,16 @@ import org.koin.core.context.unloadKoinModules
 import org.koin.core.parameter.parametersOf
 import org.koin.dsl.module
 import kotlin.time.Duration.Companion.milliseconds
+import timber.log.Timber
```
StylianosGakis, line 49: I don't see timber being used here really, maybe this was leftover unused code? If I am simply missing smth, consider using `import slimber.log.e` instead so we use slimber instead of timber directly

StylianosGakis (approved):
Yeap that's looking super simple, really good.

PR #1325, opened 2022-12-12
Move token APIs all into AuthTokenService

StylianosGakis:
Mainly to remove the need for keeping a loggedIn boolean since in the past having a token wasn't enough of a reason to know you're logged in or not.

This then had to come with some little changes on the call sites of the old AuthenticationTokenService and LoggedInRepository.

Also introduced as an implementation detail of the :auth library that it stores the tokens in Datastore instead, in the `AuthTokenStorage`


app/src/engineering/kotlin/com/hedvig/app/DevelopmentScreenAdapter.kt
```diff
@@ -132,15 +134,19 @@ class DevelopmentScreenAdapter(
...
+          token.setText(authTokenService.getToken()?.token ?: "Token not found")
           saveToken.setHapticClickListener {
-            authenticationTokenService.authenticationToken = saveToken.text.toString()
-            saveToken.context.makeToast("Token saved")
+            authTokenService.updateTokens(
+              AccessToken(token.text.toString(), 600),
+              // todo maybe the dev screen shouldn't be doing this in the first place.
+              RefreshToken(token.text.toString(), 600),
+            )
+            token.context.makeToast("Token saved")
```
StylianosGakis, line 149: I haven't used this feature really myself, do you? I usually use the impersonation instead if I need to do this. Especially with the new login apis I don't know how often we'd want to edit the token here, especially considering we'd have to edit two of them.

*Maybe* we can consider keeping it here to be copied, but not really to be edited?

app/src/engineering/kotlin/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
```diff
@@ -253,33 +253,22 @@ class EmbarkStoryTesterViewModel(
   }
 
   fun setAuthToken(authToken: String) {
-    authenticationTokenService.authenticationToken = authToken
+    authTokenService.updateTokens(AccessToken(authToken, 600), RefreshToken(authToken, 600))
```
StylianosGakis, line 256: Not sure this'd work when testing since we'd not have a proper refresh token. I know you've built and used the EmbarkStoryTesterActivity before, maybe this will need some touch up before it's gonna be usable again.

app/src/engineering/kotlin/com/hedvig/app/feature/impersonation/ImpersonationReceiverActivity.kt
```diff
@@ -112,37 +109,21 @@ class ImpersonationReceiverViewModel(
...
+        is AuthTokenResult.Error -> _state.update { ViewState.Error(result.message) }
+        is AuthTokenResult.Success -> {
+          authTokenService.updateTokens(result.accessToken, result.refreshToken)
           featureManager.invalidateExperiments()
-          _state.value = ViewState.Success
+          _state.update { ViewState.Success }
           delay(500.milliseconds)
-          _events.send(Event)
+          _events.send(GoToLoggedInActivityEvent)
         }
```
StylianosGakis, line 127: If I understood correctly from iOS impl of impersonation, this should just be enough. Take the token from the URL as we did before, and pass it as an AuthorizationCodeGrant in the `exchange` and we should be good I think.

app/src/main/kotlin/com/hedvig/app/ApplicationModule.kt
```diff
@@ -296,26 +283,28 @@ val apolloClientModule = module {
...
+  append("; ")
+  append(Build.BRAND)
+  append(" ")
+  append(Build.MODEL)
+  append("; ")
+  append(Build.DEVICE)
+  append("; ")
+  append(locale.language)
+  append(")")
+}
```
StylianosGakis, line 303: I think even ktlint was changing this all the time, so moved it to a `buildString` call to never have formatting problems again 😅

app/src/main/kotlin/com/hedvig/app/authenticate/LogoutUseCase.kt
```diff
@@ -1,52 +1,39 @@
...
   private val chatEventStore: ChatEventStore,
+  private val featureManager: FeatureManager,
+  private val hAnalytics: HAnalytics,
+  private val applicationScope: ApplicationScope,
 ) {
 
-  sealed class LogoutResult {
-    object Success : LogoutResult()
-    data class Error(val message: String?) : LogoutResult()
-  }
```
StylianosGakis, line 25: Logging out should not need a result anymore.
Even in the case where there's no internet connection, we can still clear the tokens locally and then we're good. They will expire soon anyway, so invalidating them can stay as best effort and okay if we don't succeed. Then going back to the marketing activity should be a result of the activity observing the token status and if there are any present or not.

app/src/main/kotlin/com/hedvig/app/authenticate/LogoutUseCase.kt
```diff
@@ -1,52 +1,39 @@
...
-
-  private fun clearLoginStatus() {
-    loginStatusService.isLoggedIn = false
-    authenticationTokenService.authenticationToken = null
+    applicationScope.launch { marketManager.market = null }
+    applicationScope.launch { pushTokenManager.refreshToken() }
+    applicationScope.launch { featureManager.invalidateExperiments() }
+    applicationScope.launch { chatEventStore.resetChatClosedCounter() }
+    applicationScope.launch { apolloClient.reconnectSubscriptions() }
```
StylianosGakis, line 37: Launch them all individually like that since it's okay for them to fail individually if they ever do (not possible for most but still doesn't hurt to do so).

app/src/main/kotlin/com/hedvig/app/authenticate/UserViewModel.kt
```diff
@@ -44,15 +42,7 @@ class UserViewModel(
...
   val viewState: StateFlow<ViewState>
-    get() = mutableViewState
-
-  private val _events = Channel<Event>(Channel.UNLIMITED)
-  val events = _events.receiveAsFlow()
-
-  sealed class Event {
-    object Logout : Event()
-    data class Error(val message: String?) : Event()
-  }
```
StylianosGakis, line 55: This event was a trampoline from
view interaction -> message to VM -> event to view -> view navigates somewhere
So just moved it in the activity itself directly.

app/src/main/kotlin/com/hedvig/app/authenticate/UserViewModel.kt
```diff
@@ -89,39 +79,32 @@ class UserViewModel(
...
+      authTokenResult.accessToken,
+      authTokenResult.refreshToken,
+    )
     uploadMarketAndLanguagePreferencesUseCase.invoke()
     mutableViewState.update {
       it.copy(navigateToLoggedIn = true)
     }
+    runCatching {
+      pushTokenManager.refreshToken()
+    }
```
StylianosGakis, line 104: Hmm I added this since it was here already at some previous state of this file.
Maybe this should be at a different place, where login happens for all markets? Maybe should look into this if we haven't already and I've missed it.

app/src/main/kotlin/com/hedvig/app/feature/chat/ui/ChatActivity.kt
```diff
@@ -147,11 +144,7 @@ class ChatActivity : AppCompatActivity(R.layout.activity_chat) {
         handleSingleSelectLink(
           value = value,
           onLinkHandleFailure = {
-            authenticationTokenService.authenticationToken = null
-            lifecycleScope.launch {
-              featureManager.invalidateExperiments()
-            }
-            startActivity(MarketingActivity.newInstance(this, true))
```
StylianosGakis, line 154: Another case where this should be handled by the to come PR where activities observe the token to do this automatically.

app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
```diff
@@ -319,19 +312,6 @@ class ChatViewModel(
     }
   }
 
-  fun restartChat() {
```
StylianosGakis, line 322: This was unused from any part of the app, never called so just deleted it.

app/src/main/kotlin/com/hedvig/app/feature/checkout/CheckoutViewModel.kt
```diff
@@ -87,7 +81,7 @@ class CheckoutViewModel(
     val checkout = model.checkout
     when (checkout?.status) {
       Checkout.CheckoutStatus.PENDING -> _events.trySend(Event.Loading)
-      Checkout.CheckoutStatus.SIGNED -> createAccessToken()
+      Checkout.CheckoutStatus.SIGNED -> offerRepository.fetchNewOffer(quoteCartId)
```
StylianosGakis, line 84: Fast forwarded to the success case of createAccessToken() since we should not be creating this token no more from the checkout.
As discussed before if we're inside the checkout we must already be logged in regardless.

app/src/main/kotlin/com/hedvig/app/feature/checkout/CheckoutViewModel.kt
```diff
@@ -180,8 +168,6 @@ class CheckoutViewModel(
   }
 
   private suspend fun onSignSuccess(): Event.CheckoutSuccess {
-    featureManager.invalidateExperiments()
-    loginStatusService.isLoggedIn = true
```
StylianosGakis, line 184: sign success no longer means we're logged in. But I also thought it shouldn't have something to do with invalidating the feature flags, so deleted.

app/src/engineering/kotlin/com/hedvig/app/DevelopmentScreenAdapter.kt
hugokallstrom: Not using this feature either. We can keep it for now and maybe take an off-piste to remove it later if we want to :)

app/src/main/kotlin/com/hedvig/app/feature/offer/OfferViewModel.kt
```diff
@@ -145,7 +140,6 @@ abstract class OfferViewModel : ViewModel() {
 
   abstract fun onOpenCheckout()
   abstract fun reload()
-  abstract fun onDiscardOffer()
```
StylianosGakis, line 148: Another case where the round trip to VM wasn't needed. Discarding the offer simply means going back to the main activity, so view does that directly.

app/src/engineering/kotlin/com/hedvig/app/feature/embark/EmbarkStoryTesterActivity.kt
hugokallstrom: No worries - dont think we will use this activity much now that we are working mostly with odyssey.

app/src/engineering/kotlin/com/hedvig/app/feature/impersonation/ImpersonationReceiverActivity.kt
hugokallstrom: Yes!

app/src/main/kotlin/com/hedvig/app/feature/offer/ui/OfferActivity.kt
```diff
@@ -342,20 +357,6 @@ class OfferActivity : AppCompatActivity(R.layout.activity_offer) {
...
-    when (loginStatus) {
-      LoginStatus.Onboarding -> binding.offerToolbar.inflateMenu(R.menu.offer_menu)
-      LoginStatus.LoggedIn -> {
-        binding.offerToolbar.inflateMenu(R.menu.offer_menu_logged_in)
-        menu.getItem(0).actionView?.setOnClickListener {
-          handleMenuItem(menu[0])
-        }
-      }
-    }
-  }
```
StylianosGakis, line 357: We had two versions of the top app bar options. One with 

- chat
- app settings
- app info
- login
- discard offer

And when already logged-in, it was just one option
- chat

This should make it show 1 menu at all times, which is the first menu minus the login option, since we are already logged in when in the offer activity.

app/src/main/kotlin/com/hedvig/app/authenticate/LogoutUseCase.kt
hugokallstrom: Nice!

auth/src/main/kotlin/com/hedvig/android/auth/AuthTokenServiceImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.coroutines.launch
+
+class AuthTokenServiceImpl(
+  private val authTokenStorage: AuthTokenStorage,
+  private val authRepository: AuthRepository,
+  private val coroutineScope: CoroutineScope,
+) : AuthTokenService {
+
+  @Suppress("NAME_SHADOWING")
+  private val authStatus: StateFlow<AuthStatus?> = authTokenStorage.getTokens()
```
hugokallstrom, line 23: Can this be public, and then we can remove `fun authStatus()`?

app/src/test/kotlin/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
```diff
@@ -1,192 +1,159 @@
...
+    val viewModel = testViewModel(authRepository, authTokenStorage)
 
-    val events = mutableListOf<OtpInputViewModel.Event>()
-    val eventCollectingJob = launch {
-      viewModel.events.collect { events.add(it) }
+    authTokenStorage.getTokens().first().also { (resultAccessToken, resultRefreshToken) ->
+      assertThat(resultAccessToken?.token).isNull()
+      assertThat(resultRefreshToken?.token).isNull()
+    }
+    viewModel.events.test {
```
StylianosGakis, line 98: So the .test {} function coming from [Turbine](https://github.com/cashapp/turbine) is great because it takes care of what we had to do manually here before.
It ensures that by the end of it, no more events were made other than the ones we observed using `awaitItem()`, so we don't need to check for size == 1 in a list we're populating ourselves.
And it makes sure that at the end it cancels the collection, we don't need to do .cancel on the event collecting job.

app/src/test/kotlin/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
```diff
@@ -1,192 +1,159 @@
...
+      runCurrent()
+      assertThat(viewModel.viewState.value.loadingCode).isEqualTo(false)
+      assertThat(viewModel.viewState.value.loadingResend).isEqualTo(true)
+
+      authRepository.resendOtpResponse.add(ResendOtpResult.Success)
+      runCurrent()
+      assertThat(viewModel.viewState.value.loadingResend).isEqualTo(false)
+      val event = awaitItem()
+      assertThat(event).isEqualTo(OtpInputViewModel.Event.CodeResent)
     }
```
StylianosGakis, line 150: So the reason I changed these a tiny bit is as follows:

If we call `runCurrent()` directly after `resetCode()` is called, we'd get the wrong assertions for 
```
assertThat(viewModel.viewState.value.loadingCode).isEqualTo(false)
assertThat(viewModel.viewState.value.loadingResend).isEqualTo(true)
```
since they would not be true if the code contained in the `viewmodelScope.launch` inside `resendCode()` would run before them.
So in the test previously we depended on the way the test schedules the jobs for this assertion to work, basically making the test racy, meaning if we changed the impl details of how the coroutines were launched (for example if we moved the .update of the viewState inside the viewModelScope.launch{} block this test would fail) we might get different results.

This change makes it so that after we call `resendCode()` we call `runCurrent()` to let it do everything it's supposed to do, and then we assert that we're in the loading state since we know that authRepository has *not* responded anything yet, and we know that since we haven't fed anything to the `authRepository.resendOtpResponse` yet. *Then* we give it the response, `runCurrent()` to let OtpInputViewModel to process whatever that means for it when the repository has responded with something, and then we can safely make our assertions again. Nice and one by one, making reasoning about what's happening in which order a bit simpler rather than relying on how the test coroutine scheduling might work or not.

For example if in the previous test we changed the dispatcher it was running on to an `UnconfinedTestDispatcher()` it'd start failing since we wouldn't be able to see this in-between loading state, with this approach this would still pass since we're also controlling when the authRepository responds with something.

app/src/test/kotlin/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
```diff
@@ -201,18 +168,60 @@ class OtpInputViewModelTest {
...
+      authRepository = authRepository,
+      uploadMarketAndLanguagePreferencesUseCase = mockk(relaxed = true),
+    )
+  }
+
+  private fun TestScope.testAuthTokenStorage(): AuthTokenStorage {
+    return AuthTokenStorage(
+      dataStore = TestPreferencesDataStore(
+        datastoreTestFileDirectory = testFolder.newFolder("datastoreTempFolder"),
+        backgroundScope,
```
StylianosGakis, line 207: backgroundScope is a field inside `TestScope` which we get inside the `runTest{}` block. It serves as a CoroutineScope which can be used to run stuff which is okay to still be running at the end of the test without making the test hang.
This is needed since for this example running a DataStore needs a CoroutineScope but it'd stay open forever and make the test timeout otherwise.
This lets it run in the background using our test CoroutineScope

Discussion where I learned this here https://slack-chats.kotlinlang.org/t/8113862/i-got-a-coroutine-test-where-i-use-the-testscope-provided-in

app/src/test/kotlin/com/hedvig/app/feature/swedishbankid/sign/SwedishBankIdSignViewModelTest.kt
```diff
@@ -122,54 +98,19 @@ class SwedishBankIdSignViewModelTest {
...
-
-      fakeObserveQuoteCartCheckoutUseCase.results.add(Checkout(Checkout.CheckoutStatus.SIGNED, null, null).right())
-      assertThat(awaitItem()).isEqualTo(BankIdSignViewState.BankIdSuccess)
-
-      fakeCreateAccessTokenUseCase.results.add(ErrorMessage().left())
-      assertThat(awaitItem()).isInstanceOf(BankIdSignViewState.Error::class)
-
-      assertThat(loginStatusService.isLoggedIn).isEqualTo(false)
     }
   }
```
StylianosGakis, line 106: No longer doing a access token call, hence the test removal.

auth-test/src/main/kotlin/com/hedvig/android/auth/FakeAuthRepository.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.hedvig.authlib.RevokeResult
+import com.hedvig.authlib.StatusUrl
+import com.hedvig.authlib.SubmitOtpResult
+import kotlinx.coroutines.flow.Flow
+
+class FakeAuthRepository : AuthRepository {
+
+  val resendOtpResponse = Turbine<ResendOtpResult>()
+  val submitOtpResponse = Turbine<SubmitOtpResult>()
+  val exchangeResponse = Turbine<AuthTokenResult>()
```
StylianosGakis, line 20: Learned this approach from https://www.youtube.com/watch?v=uk8kc9jgzbM&t=1610s mostly.

So this is what enables what I wrote above in a previous comment regarding giving the AuthRepository the ability to respond when `we` want to.

Turbine<T> is basically Channel<T> with some convenience functions on top of it. It allows in our case to simply do `awaitItem()` on one of these which does this. When the test calls this function, it's suspending so it suspends until there's an item. Initially it's empty, so it just waits there for the response. When we do for example `submitOtpResponse.add(Foo)`, the channel gets an item, then the `awaitItem()` suspending function resolves with the result that we passed in, and it continues on with the test given that response. Gives us full control of when and what this fake auth repository responds and enables the tests in the way I wrote in that other bigger comment.

auth/src/main/kotlin/com/hedvig/android/auth/AuthTokenServiceImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+    .stateIn(
+      coroutineScope,
+      SharingStarted.Eagerly,
+      null,
+    )
+
+  override fun getToken(): AccessToken? {
+    return (authStatus.value as? AuthStatus.LoggedIn)?.accessToken
+  }
```
StylianosGakis, line 37: Technically this method is just convenience over the `authStatus()` and doing this safe cast, but it's such a common use case that I think it's worth the extra function.

auth/src/main/kotlin/com/hedvig/android/auth/interceptor/AccessTokenAuthenticator.kt
```diff
@@ -0,0 +1,32 @@
...
+        val newAccessToken = authTokenService.getToken()
+        if (newAccessToken != null && newAccessToken != accessToken) {
+          return@withLock response.request.withAuthorizationToken(newAccessToken.token)
+        }
+        // Still same old token, so fetch a new one.
+        val refreshedToken = authTokenService.refreshAndGetToken() ?: return@withLock null
+        response.request.withAuthorizationToken(refreshedToken.token)
+      }
+    }
+  }
```
StylianosGakis, line 31: Took the liberty to copy this from my old project as I had shown you before. Mainly so that it defers the actual logic of refreshing the token over to the AuthTokenService, while the Authenticator itself is only responsible for locking and calling the right things, nothing more.

auth/src/main/kotlin/com/hedvig/android/auth/storage/AuthTokenStorage.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
+import com.hedvig.authlib.AccessToken
+import com.hedvig.authlib.RefreshToken
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+// TODO Work with expiration dates.
```
StylianosGakis, line 12: Not needed yet since we'll just rely on 401 to know when to retry. If we want to do some preemptive fetches and so on later on we can expand this,

auth/src/test/kotlin/com/hedvig/android/auth/AccessTokenAuthenticatorTest.kt
```diff
@@ -125,28 +75,39 @@ class AccessTokenAuthenticatorTest {
     assertThat(firstRequest.headers["Authorization"]).isEqualTo(expiredToken)
     // Second request intercepted by authenticator, adding an updated token
     val secondRequest = mockWebServer.takeRequest()
-    assertThat(secondRequest.headers["Authorization"]).isEqualTo("newToken")
+    assertThat(secondRequest.headers["Authorization"]).isEqualTo(refreshedAccessToken.token)
   }
 
   @Test
-  fun shouldPersistNewTokenWhenResponseIs401AndCallIsSuccessFull() {
+  fun shouldPersistNewTokenWhenResponseIs401AndCallIsSuccessful() = runTest(UnconfinedTestDispatcher()) {
```
StylianosGakis, line 82: These tests rely on using the `Authenticator` we have. But that one uses `runBlocking` to go from non-coroutines into coroutines world. This means that we don't quite have the same control over it to pass the test dispatcher and be able to do `runCurrent()` on it, so we're using the UnconfinedTestDispatcher to simply have everything run eagerly and it works.

core-datastore-test/src/main/kotlin/com/hedvig/android/core/datastore/TestPreferencesDataStore.kt
```diff
@@ -0,0 +1,40 @@
...
+class TestPreferencesDataStore(
+  datastoreTestFileDirectory: File,
+  coroutineScope: CoroutineScope,
+) : DataStore<Preferences> by PreferenceDataStoreFactory.create(
+  scope = coroutineScope,
+  produceFile = {
+    val fileNameWithDatastoreExtension = "$testPreferencesDatastoreFileName$preferencesDatastoreFileExtension"
+    File(datastoreTestFileDirectory, "$preferencesDataStoreFileDirectory/$fileNameWithDatastoreExtension")
+  },
+)
```
StylianosGakis, line 40: With this we can simply use a real DataStore<Preferences> in tests, with a temporary file as described in the comment above. Super convenient and lets us avoid having to abstract some classes, but we can use the real implementation in tests with the test DataStore, super convenient that DataStore lets you do this.

core-common/src/main/kotlin/com/hedvig/android/core/common/ApplicationScope.kt
```diff
@@ -0,0 +1,10 @@
+package com.hedvig.android.core.common
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.MainScope
+
+/**
+ * A [CoroutineScope] which lives for the entire lifecycle of the application. Can be used to launch top-level actions
+ * which need to out-live the current CoroutineScope, like even after a ViewModel would be killed.
+ */
+class ApplicationScope : CoroutineScope by MainScope()
```
StylianosGakis, line 10: Serves as a top-level CoroutineScope to be used to run anything that is not tied to the specific ViewModel we're currently in or something like that, but things that should rather be run on a global level, even if we go back and kill the ViewModel for example. MainScope also has a SupervisorJob like viewModelScope does, which means that if any child coroutine fails that's okay and keeps this scope alive too.

This is what we had done in notification-service too https://github.com/HedvigInsurance/notification-service/blob/master/service/src/main/java/com/hedvig/notificationservice/utils/ApplicationScope.kt and it worked well for this use case.

hugokallstrom (approved):

app/src/test/kotlin/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
hugokallstrom: Cool!

app/src/test/kotlin/com/hedvig/app/feature/genericauth/OtpInputViewModelTest.kt
hugokallstrom: Makes sense 👍

auth/src/main/kotlin/com/hedvig/android/auth/AuthTokenServiceImpl.kt
StylianosGakis: Hmm yeah, why not actually, I made the change, I don't see a reason why this wouldn't be a good idea. Removes one useless function from this class!

PR #1326, opened 2022-12-13
Migrate token

hugokallstrom:


StylianosGakis (approved):
At first thought, I thought this would be smth we do in our MainActivity once at startup. This interceptor introduces a sharedprefs call on each reques.
If we're gonna keep this interceptor for like a year, I think I'd rather we do it at startup instead, just where we used to have the database migration as well. If we want to add this for now and maybe in 2-3 releases remove this and for the late adopters we simply let them logout, then I think it's okay to leave this interceptor in.

hugokallstrom:
> At first thought, I thought this would be smth we do in our MainActivity once at startup. This interceptor introduces a sharedprefs call on each reques. If we're gonna keep this interceptor for like a year, I think I'd rather we do it at startup instead, just where we used to have the database migration as well. If we want to add this for now and maybe in 2-3 releases remove this and for the late adopters we simply let them logout, then I think it's okay to leave this interceptor in.

Yes good point. I think in practice this should be fine, if we also make sure to remove this when we know most users have updated.

PR #1327, opened 2022-12-19
CLX-246: Use time based auth token refreshing

StylianosGakis:
Handles the entire process of checking the current tokens, refreshing
them if the access token is expired, and clearing them if the refresh
token is also expired. Does so by *only* looking at the expiration 
date and assuming the token is OK if it's not expired.

Make AuthTokenService work so that all the suspending functions fetch
directly from storage, ensuring no possible race conditions, while
keeping the `authStatus` StateFlow to be used for reactive purposes
(logging out) or for a non-blocking fast read on the interceptor in the
happy path where the access token simply exists and is not expired.


auth/src/main/kotlin/com/hedvig/android/auth/AuthTokenService.kt
```diff
@@ -6,9 +6,9 @@ import kotlinx.coroutines.flow.StateFlow
 
 interface AuthTokenService {
   val authStatus: StateFlow<AuthStatus?>
-  fun getToken(): AccessToken?
-  suspend fun refreshAndGetToken(): AccessToken?
-  fun updateTokens(accessToken: AccessToken, refreshToken: RefreshToken)
-  fun invalidateTokens()
+  suspend fun getTokens(): AuthTokens?
+  suspend fun refreshAndGetAccessToken(): AccessToken?
+  suspend fun updateTokens(accessToken: AccessToken, refreshToken: RefreshToken)
```
hugokallstrom, line 11: Let this take AuthTokens as paremeter as well?

hugokallstrom (approved):
Looks good! Nice tests as well 👍

auth/src/main/kotlin/com/hedvig/android/auth/AuthTokenService.kt
StylianosGakis: I liked the idea, but I went ahead and tried to do it and realized that `AuthTokens` takes in the `Local` versions of the tokens, which means that ones with the expiration as an `Instant` instead of an `int` with seconds to expire from now, so this wouldn't work. Since we don't want to do this conversion in all call sites, and keep that logic inside `AuthTokenStorage` so gonna keep it like this, and let this take in the types that authlib returns directly, to not overload this with another `AuthTokens` version which is for those versions of the tokens.

PR #1329, opened 2022-12-20
CLX-211: replace how logout is handled

StylianosGakis:
First PR with just the code around the logout functionality, another one is incoming which just picks the activities which need this functionality


hugokallstrom (approved):
GJ!

PR #1330, opened 2022-12-20
Adds the logout observer to the necessary activities

StylianosGakis:
Tried to do this as carefully as I can, but I can totally see me making the mistake of adding this in a place where it doesn't need to be, or not adding it to a place where it does need to be in.
Gonna double check the activities once again to make sure.

In reality, the majority of the times, if someone opens the app after the refresh token is expired and they need to be logged out, they will have been redirected since the MainActivity will have caught the token being null. But this is so that even if we change what MainActivity does and they happen to be in any of these activities, they will still be redirected to the MarketingActivity.

### DIFF
--- a/app/src/main/kotlin/com/hedvig/app/feature/adyen/payin/AdyenConnectPayinActivity.kt
+++ b/app/src/main/kotlin/com/hedvig/app/feature/adyen/payin/AdyenConnectPayinActivity.kt
@@ -7,6 +7,7 @@ import androidx.appcompat.app.AppCompatActivity
 import com.adyen.checkout.components.model.PaymentMethodsApiResponse
 import com.adyen.checkout.dropin.DropIn
 import com.adyen.checkout.dropin.DropInResult
+import com.hedvig.android.auth.android.AuthenticatedObserver
 import com.hedvig.android.language.LanguageService
 import com.hedvig.app.R
 import com.hedvig.app.feature.adyen.AdyenCurrency
@@ -31,6 +32,7 @@ class AdyenConnectPayinActivity : AppCompatActivity(R.layout.fragment_container_
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+    lifecycle.addObserver(AuthenticatedObserver())
 
     val c = intent.getSerializableExtra(CURRENCY) as? AdyenCurrency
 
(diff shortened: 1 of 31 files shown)


hugokallstrom (approved):
I could see a situation where one would forget to add this `AuthenticatedObserver` in a new activity or fragment. Could we solve that somehow? Would it be possible to only add this functionality to the application somehow?

hugokallstrom:
We use `ProcessLifecycleOwner.get().lifecycle.addObserver()` in HedvigApplication for tracking. Could we use that for this use case as well?

StylianosGakis:
> I could see a situation where one would forget to add this `AuthenticatedObserver` in a new activity or fragment. Could we solve that somehow? Would it be possible to only add this functionality to the application somehow?

Yes, I can see that happening as well, and he problem with that is that we don't want to do that in the activities where we are OK with being logged out. If we did it globally like that, if for example we open the debug activity, we will be automatically redirected to the marketing activity. In fact if we are already in the marketing activity, we will be in a loop of being redirected to the marketing activity again and again since we'll be considered logged out there. Same for any other activity where we may let the member go to at a logged out state.

In any way we do this, we'll have to either opt-in to specific activities as we're doing right now. Or opt-out from specific activities, by idk somehow adding this to the entire application but have it be a no-op by checking which activity we're currently in? Not sure how that'd work either.

So I haven't had any better idea of how to make it log one out in all the other activities.

And again, in reality, what will happen is that if someone comes back to the app after not having it open for more time than the refresh token exists for, they will open the app in a cold state, the MainActivity will run and the token will be checked again and they simply won't be redirected to inside the app in the first place. The observers in each activity is for potential cases where
1) The app was alive there and it was not killed in the time the refresh token expired (unlikely since the token lives long enough)
2) They've deep linked into it without going through the MainActivity, which atm is not even possible afaik, but depends on how our navigation structure evolves it may be possible in the future
3) They've literally kept the app open for the refresh token expiration time without doing any network request which would automatically refresh their token, and then after that time passed they try to do a request. Then if they are in an activity where this observer is there, the request would fail, the tokens would be cleared but they wouldn't be redirected to the marketing screen, until they navigate to an activity which does, or they force close the app

All of these are so unlikely that I think this is fine.

StylianosGakis:
Otherwise, if we can come up with a way to could apply this globally using `ProcessLifecycleOwner` yet still not get this problem with MarketingActivity and all other activities which are fine to be logged out in, then that'd work too, it's just that I couldn't come up with one.

hugokallstrom:
> Otherwise, if we can come up with a way to could apply this globally using `ProcessLifecycleOwner` yet still not get this problem with MarketingActivity and all other activities which are fine to be logged out in, then that'd work too, it's just that I couldn't come up with one.

Good points. Yeah I guess the tricky thing is to know which activity your currently in.. Lets revisit that later on if we feel its a problem :) Good job with this!! 💯

StylianosGakis:
Yeah exactly, if we figure it's a problem or if we come up with something better I'd also love to change this, since I also don't like that it's so prone to making a mistake. But let's merge for now then :D

PR #1331, opened 2022-12-21
CLX-252: Fix authentication token refreshing with odyssey

StylianosGakis:
Moved the token fetching and refreshing into the `AccessTokenProvider` interface, which the interceptor is also using under the hood. Tests still pass fine so it's all nice and dandy. The Ktor interceptor inside Odyssey does the exact same thing.

Bumps odyssey to get the new OdysseyRoot API, and bumps Koin too as it was crashing for some reason, same as Sam posted today in odyssey slack channel, but this combination seemed to work allright.

Tested with artificially small expiration times, and it worked both to take an expired AccessToken and a non-expired RefreshToken and to refresh them both, and for both of them to be expired and to be force logged out at that moment.

p.s. It notes `AuthTokens` as deleted, but I simply moved it to the same package as the LocalAccessToken is.


hugokallstrom (approved):
🚀

PR #1333, opened 2022-12-22
CLX-256: Do not open bankid twice

StylianosGakis:
Did a little overhaul of the handling there.
Turns out we had an abstract class which was only subclassed from that dialog, so I merged them together to simplify things.
Then made `loginStatusResult` which should handle the results of `authRepository.observeLoginStatus` only when the bankIdProperties are already loaded.
One case we're not covering really is the case where the initial `authRepository.startLoginAttempt` fails for whatever reason, right now and before it'd get stuck at a Loading state, since there's no retry functionality, nor something to inform the member, but this should only happen in a no-internet situation. I will see if we have some ready to use strings to simply add a retry button in there, but the meat of the PR is ready.

I can also add a test or two tomorrow, just to be extra sure.


hugokallstrom (approved):
LGTM!

PR #1334, opened 2022-12-23
BankIdLoginViewModel tests

StylianosGakis:
Also made the case where the initial attempt to fetch the `BankIdProperties` also represent an Error case, with the generic `We couldn\'t reach Hedvig right now, are you sure you have an internet connection?` error message showing on the dialog, since that is most likely to happen when the network request fails, which would probably mean the internet connection is down, or our backend is dead, so it will like 100% of the time be a network error.


hugokallstrom (approved):
Nice!

PR #1335, opened 2022-12-23
Set portrait orientation for claims flow activity 

hugokallstrom:
In order to not reset odyssey when rotated


StylianosGakis (approved):
lgtm! Any information about what exactly was breaking just so I know for future reference?

PR #1346, opened 2023-01-19
Bump Android Gradle Plugin (AGP 7.4.0) and Gradle (7.6.0)

StylianosGakis:


StylianosGakis:
Android Studio Electric Eel is now stable, so we can bump to AGP 7.4.
I think this may not work with old Android Studio versions, so make sure to update to the Electric Eel build.

hugokallstrom (approved):

PR #1349, opened 2023-01-20
Send down common claim ID instead of itemType+itemProblem

vonElfvin:
It breaks when entering the flow. Might be due to `1.2.10` not working properly.


StylianosGakis:
Yeah let's see if I can get Odyssey to not crash with this, merge that and then we merge this in after we make sure it's not crashing, gonna hold off merging this until then

PR #1360, opened 2023-02-02
CLX-254: Show switching message correctly for NO members

hugokallstrom:


StylianosGakis (approved):
So we're both changing that we do a `==` comparison on the id instead of `contains` which I assume was problematic for some cases? Do we know some specifically?
And also checking that the insurance provider is switchable in the first place, I guess from some providers automatic switching is just not possible right?
Just trying to get the context so I know if we ever revisit this in the future

hugokallstrom:
> So we're both changing that we do a `==` comparison on the id instead of `contains` which I assume was problematic for some cases? Do we know some specifically? And also checking that the insurance provider is switchable in the first place, I guess from some providers automatic switching is just not possible right? Just trying to get the context so I know if we ever revisit this in the future

Yeah exactly, im not entirely sure in what cases doing contains was problematic - but now we have the same logic as iOS has which we now works 👍

PR #1364, opened 2023-02-03
Remove showing images inline in AttachPicker

StylianosGakis:
Mainly to remove the need to run the getImagesPath() function which needs storage permission unnecessarily. Picking files from the picker with the appropriate activity result contracts should be enough instead.

This code was necessary to show some images inline in the bottom sheet, along with the camera icon which opens the camera, and the other icon which opens the other bottom sheet asking to pick either files or images.

The main function removes was `getImagesPath()` which then made one of the viewHolders in the `AttachFileAdapter` redundant so removed that too, and all related functions after that. Now the RecyclerView is only ever hosting 1 item, but didn't change the code there to keep the PR as small as possible while not keeping dead code in there.


hugokallstrom (approved):
Nice!!

PR #1365, opened 2023-02-07
CLX-339: Entry points triaging

hugokallstrom:
Two implementations for fetching `searchableClaim` - common claims from giraffe and entrypoints from claims-service.

Screenshot 2023-02-07 at 11 43 42


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchResults.kt
```diff
@@ -0,0 +1,154 @@
...
+          } else {
+            Text(text = "Start a claim")
+          }
+        },
+        text = { Text(text = claim.displayName) },
+        modifier = Modifier
+          .animateItemPlacement()
+          .clickable {
+            if (claim.isCovered) {
+              onClaimSelected(claim)
```
StylianosGakis, line 100: I wonder for these, if it would make it more complicated or less complicated if this composable just had something like `claimClick : (SearchableClaim) -> Unit` and the decision on what is happening is determined on a higher up point, maybe even the activity, so that the compose code can focus as much as possible on just the UI and not the business logic.
Not 100% sure though, this isn't something that I've myself settled on and every time find myself thinking about it in different ways.
Feel free to do nothing or tell me your thoughts on this if you have a stronger opinion on it.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchViewModel.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.hedvig.android.odyssey.model.ItemType
+import com.hedvig.android.odyssey.model.SearchableClaim
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+class SearchViewModel(
+  private val getClaimEntryPoints: GetClaimEntryPoints,
+) : ViewModel() {
+  private val searchableClaims = emptyList<SearchableClaim>()
```
StylianosGakis, line 14: This doesn't seem to be used anywhere, is this leftover or WIP in some way? Otherwise maybe _viewState cna be initiated with `results = emptyList()` inline without the need of this variable?

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/ClaimsFlowActivity.kt
```diff
@@ -28,7 +28,7 @@ class ClaimsFlowActivity : ComponentActivity() {
     super.onCreate(savedInstanceState)
     lifecycle.addObserver(AuthenticatedObserver())
     val odysseyUrl = intent.getStringExtra(ODYSSEY_URL_KEY) ?: error("ODYSSEY_URL_KEY needs to be passed in")
-    val commonClaimId = intent.getStringExtra(COMMON_CLAIM_ID)
+    val commonClaimId = "1001"
```
StylianosGakis, line 31: What's 1001 in this case? And what should happen with `COMMON_CLAIM_ID`, which isn't used somewhere atm?

feature-odyssey/src/main/res/drawable/ic_claim_covered.xml
```diff
@@ -0,0 +1,13 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="17dp"
+    android:height="22dp"
```
StylianosGakis, line 3: This isn't 24.dp like those usually are, is this from figma? Looks a bit odd for it to be this way no?

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchAppBar.kt
```diff
@@ -0,0 +1,135 @@
...
+private fun AddressInput(
+  viewState: SearchViewState,
+  setNewTextInput: (String) -> Unit,
+  focusRequester: FocusRequester,
+  closeKeyboard: () -> Unit,
+  onFocus: (Boolean) -> Unit,
+) {
+  Row(modifier = Modifier.padding(horizontal = 22.dp, vertical = 8.dp)) {
+    Image(
+      painter = painterResource(R.drawable.ic_search),
```
StylianosGakis, line 79: For these, I think we can simply use `Icons.Default.Search` which comes inside compose, to also avoid having to add the xml icons ourselves which may not be the correct ones anyway.

StylianosGakis (commented):
Hey had a lot of questions, some are due to me missing some context, some becuase I am not sure which parts of this is WIP and which ones are ready to be used, and then some more general ones.
Please fill me in so I understand a bit better 😊

PR #1408, opened 2023-02-22
Add customer io SDK and setup

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
```diff
@@ -68,6 +72,17 @@ class LoggedInViewModelImpl(
...
+    viewModelScope.launch {
+      when (val result = memberIdUseCase.memberId()) {
+        is GetMemberIdUseCase.MemberIdResult.Error -> {
+          slimber.log.e { "Could not identify member to customer.io - could not fetch member id" }
+        }
+        is GetMemberIdUseCase.MemberIdResult.Success -> {
+          CustomerIO.instance().identify(identifier = result.id, attributes = mapOf("os" to "android"))
+        }
+      }
+    }
```
StylianosGakis, line 85: For getting the memberID, there's a new way apprently, which we just started using. Basically take a look at this https://github.com/HedvigInsurance/android/blob/6c86b2a6f988600effeca786a0a3f63a3bdb910c/datadog/src/main/kotlin/com/hedvig/android/datadog/memberid/DatadogMemberIdUpdatingAuthEventListener.kt#L30-L48 and then take a look at this discussion https://hedviginsurance.slack.com/archives/C03DTURB621/p1675757933975849

The gist is that the new auth tokens we get do properly handle embedding the member ID in there, so you can fetch it locally from the stored token locally. 

Maybe this is an opportunity to also have this customer-io functionality also hook into the `AuthEventListener` events, and do this on login, and clear the token on logout? You can check `DatadogMemberIdUpdatingAuthEventListener` for how we do that for Datadog.

StylianosGakis: And would need to extract that `extractMemberIdFromAccessToken` to somewhere else. Didn't do it before since it didn't need to be used elsewhere, but now we can think of where it can fit. core-common-andoird is the first candidate, as it's a simple function of type `(String) -> String?`, but it does have an android dependency since it's using `android.util.Base64`.

app/src/main/kotlin/com/hedvig/app/service/push/PushNotificationService.kt
```diff
@@ -34,10 +35,12 @@ class PushNotificationService : FirebaseMessagingService() {
       )
       .build()
     WorkManager.getInstance(this).enqueue(work)
```
StylianosGakis, line 37: I wonder how `PushNotificationWorker` fits in this scenario now. And if that one can be removed completely, as I see it does two mutations: `RegisterPushTokenMutation`, `NotificationRegisterDeviceMutation`. Maybe those two aren't even needed anymore? It even has this bug there right now, where if there's no token it simply retries forever, [here](https://github.com/HedvigInsurance/android/blob/bf87684a61daab10d17174076f276bf4519d7e1a/app/src/main/kotlin/com/hedvig/app/service/push/PushNotificationWorker.kt#L31-L33) if we can get rid of it altogether it'd be great. If not it'd be nice to help me understand what it does and it needs to stay.

app/src/main/kotlin/com/hedvig/app/HedvigApplication.kt
```diff
@@ -22,11 +25,28 @@ open class HedvigApplication : Application() {
...
+      appContext = this,
+    ).apply {
+      addCustomerIOModule(
+        ModuleMessagingPushFCM(),
+      )
+      setRequestTimeout(8000L)
+      setRegion(Region.US)
+      build()
+    }
+  }
```
StylianosGakis, line 48: Might be a good idea to do what [DatadogInitializer](https://github.com/HedvigInsurance/android/blob/73e90644d4bd8e3442e1bca478c0ef565a57e836/datadog/src/main/kotlin/com/hedvig/android/datadog/DatadogInitializer.kt#L23) is doing to initialize itself, basically using androidx.startup. Can also be done that way so that the customer-io stuff can be in its own module. Can be done at a later PR though of course if we wanna do it.

StylianosGakis (commented):
Do you also know if the SDK sets up their own notification channel? And if that comes with a good name so that people don't disable it in their notification settings + has sound on by default and that kind of stuff. Would be good to check it out.

Other than that, I think the important bit is that we can use the existing token to get the member-id without doing a extra network request, which is what Bystam was suggesting in the linked slack discussion.

But the rest is all good, we can address them or not, not blocking this PR though imo.

StylianosGakis (approved):
Oops, wanted to approve before 🫣

app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInViewModel.kt
hugokallstrom: Thats great, I was looking for something like this :) Will add a new PR with these changes together with a new module 👍

app/src/main/kotlin/com/hedvig/app/service/push/PushNotificationService.kt
hugokallstrom: Good idea. I will investigate if its possible to remove this!

PR #1410, opened 2023-02-22
Use data store to observe changes to market

hugokallstrom:


hedvig-market/src/main/kotlin/com/hedvig/android/market/MarketManager.kt
```diff
@@ -1,51 +1,56 @@
...
-  private val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context)
-
   override val enabledMarkets = listOfNotNull(
     Market.SE,
     Market.NO,
     Market.DK,
     if (isDebug) Market.FR else null,
   )
+  override val market: Market?
+    get() = runBlocking { observeMarket().first() }
```
StylianosGakis, line 32: Yeah we were reading this in a blocking manner anyway before, I guess we're keeping it to avoid having to update *all* call sites. Just gonna have to be a bit careful if this is blocking for too long cause it's gonna happen on the UI thread. But we can let if be for now, since I am hoping it won't be too long anyway.

hedvig-market/src/main/kotlin/com/hedvig/android/market/MarketManager.kt
```diff
@@ -1,51 +1,56 @@
...
 
-  @SuppressLint("ApplySharedPref") // We need to do this right away
-  private fun setMarketLocally(market: Market) {
-    sharedPreferences.edit()
-      .putString(Market.MARKET_SHARED_PREF, market.name)
-      .commit()
+  override suspend fun removeMarket() {
+    dataStore.edit { preferences ->
+      preferences.remove(marketPreferenceKey)
+    }
```
StylianosGakis, line 50: I think we may have had some cases where removing the market completely makes the marketing screen act a bit weird. Could be wrong though. Just gonna have to make sure that when this happens, we are also logged out for sure (there's no guard-rails for this yet, but I see we're only calling this on logout() so let's make sure we keep it that way) otherwise a lot of weirdness will happen.
And that the marketing activity handles this case well, which I do not really know if it does ATM, would need to check.

StylianosGakis (approved):
Couple of comments but not blocking, just things to keep in our minds
LGTM :)

hedvig-market/src/main/kotlin/com/hedvig/android/market/MarketManager.kt
hugokallstrom: Good point. I think we could remove this functionality altogether tbh.

StylianosGakis: Yeah, defaulting to SE is the safe bet usually. But this does potentially come with the side-effect that in the cases where things are actually going wrong, for non-SE members it may break completely. But handling the no-market case in the marketing screen should be a separate PR regardless, if we ever want/need to do it.

PR #1411, opened 2023-02-22
Initialize customer io

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/service/push/CustomerIoInitializer.kt
```diff
@@ -0,0 +1,58 @@
...
+private const val REQUEST_TIMEOUT_MILLIS = 8000L
+
+class CustomerIoInitializer(
+  private val marketManager: MarketManager,
+  private val applicationScope: ApplicationScope,
+) {
+
+  private val region = Region.EU
+
+  fun setupCustomerIo(application: Application) {
```
StylianosGakis, line 24: Koin can also inject the Application directly, when on startKoin() we call [androidContext()](https://github.com/HedvigInsurance/android/blob/1c64759443394da16107100242999105ef3aa28b/app/src/main/kotlin/com/hedvig/app/feature/di/KoinInitializer.kt#L68) it is provided to the list of things that koin can provide [here](https://github.com/InsertKoinIO/koin/blob/cb3ed3a85f7a0f86982f59a0a7439c0935cace67/android/koin-android/src/main/java/org/koin/android/ext/koin/KoinExt.kt#L63).

So I think you can make that a parameter of the class itself, and not having to pass it through the setupCustomerIo function itself.
Which will also mean if we want to make this an androidx.startup.Initializer, we can do that directly, since we won't rely on `HedvigApplication.kt` to call this.

StylianosGakis (approved):
Great with the introduction of the initializer class, lgtm!

app/src/main/kotlin/com/hedvig/app/service/push/CustomerIoInitializer.kt
hugokallstrom: Ill add that in a PR on top of this!

PR #1414, opened 2023-02-23
CLX-377: Design showcase sample

StylianosGakis:
Add the design-showcase app, which shows m3 components

*note* A lot of the code in here is from adding a bunch of m3 components to test showing in the UI of the sample app. If you look through the file structure while reviewing this PR it will make it easier for you, since you know you can skip reviewing the com.hedvig.android.sample.design.showcase.ui.components and everything inside there. They are all called inside com.hedvig.android.sample.design.showcase.ui.MaterialComponents to build this showcase screen.

These are how everything should look like out of the box for material3
components called inside HedvigTheme {}.

image

And share some run-configurations through the VCS, in this case to have an easy way to run the sample design-showcase app.

This way, we can share some common run configs, like in this case:
* Running the app
* Running the app with a deep link for authentication
* Running the design-showcase app
No need to look into the xml files themselves, only that the IDE will
show them on the top right to choose from. We can even share more of
them in the future if we feel like there's any value in them


hugokallstrom (approved):
Super!!

PR #1415, opened 2023-02-23
Add a Hedvig adjusted card and date picker

StylianosGakis:
The card to conform to our DS levitation and color The DatePicker to the above, plus removing the unnecessary title part and the empty space for the buttons which we do not use when using this component inline, and not in a Dialog


hugokallstrom (approved):
👌

PR #1418, opened 2023-02-23
CLX-350: Cancel insurance feature

StylianosGakis:
Draft for now.

Needs the backend obviously to exist so that the real operation can happen. Other than that, need to make sure that the correct insurance ID is passed in from the insurance tab, and a success screen to be made too. For that screen the UI is dead simple, and if we make yet another activity it will be straightforward, but might be a chance to consider alternatives. Or not :D 

Some screenshots to give an idea of how this looks like with the m3 color scheme too.
|||||
|---|---|---|---|
| image |image | image | image |


feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/ui/CancelInsuranceScreen.kt
```diff
@@ -0,0 +1,159 @@
...
+      HedvigDatePicker(
+        datePickerState = datePickerState,
+        dateValidator = dateValidator,
+      )
+    }
+  }
+}
+
+@Composable
+fun ErrorSnackbar(hasError: Boolean, showedError: () -> Unit, modifier: Modifier = Modifier) {
```
hugokallstrom, line 136: Can we extract this so that we can reuse it? Or do we have something similar we can use already, like a dialog?

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/CancelInsuranceViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+import kotlinx.datetime.toLocalDateTime
+import kotlin.time.Duration.Companion.seconds
+
+@OptIn(ExperimentalMaterial3Api::class)
+internal class CancelInsuranceViewModel(
+  private val insuranceId: InsuranceId,
+  private val cancelInsuranceUseCase: CancelInsuranceUseCase,
+  clock: Clock = Clock.System,
+) : ViewModel() {
+  private val dateSubmissionError = MutableStateFlow(false)
```
hugokallstrom, line 33: Why not keep these in the viewstate?

hugokallstrom (approved):
Looks good!

app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/detail/ContractExt.kt
```diff
@@ -45,6 +44,7 @@ fun InsuranceQuery.Contract.toMemberDetailsViewState() =
       null
     },
     change = YourInfoModel.Change,
+    cancelInsurance = YourInfoModel.CancelInsuranceButton(id), // todo ensure this is the correct ID
```
hugokallstrom, line 47: Add this behind a feature flag (or just remove this line for now) so that you can merge this branch to dev. That way you avoid conflicts later (even though they should be minimal).

hugokallstrom (approved):

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/CancelInsuranceViewModel.kt
StylianosGakis: These are part of the UIstate through the `combine` done a few lines down from here.
(btw, uistate <-> viewstate, I use those interchangeably. I do think I like UIstate more lately though, as ViewState makes me think of the XML view, and UI-state for UI in compose, so I tend to use uiState instead, I hope that's not confusing)

Basically the reason they are separate is so that I can individually set the to true/false without having to take the existing UIState, copy it with this value changed, and set it back again.
Now to switch the error you just do `dateSubmissionError.update { true }` and `combine` triggers automatically and produces the new UiState.
With all of it being in a local _viewState it'd be 
```
_viewState.update { viewState ->
  viewState.copy(
    dateSumissionError = true
  )
}
```

This achieves the same result but in a pretty simple way imo. Basically like having all the puzzle pieces individually, and then putting them all together nicely in a UIState through the combine.

What do you think about it in general, does it make sense?

app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/detail/ContractExt.kt
StylianosGakis: Done in https://github.com/HedvigInsurance/android/pull/1433

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/ui/CancelInsuranceScreen.kt
StylianosGakis: Extracted it here https://github.com/HedvigInsurance/android/pull/1434
For this one I used a snackbar since we don't really have a meaningful message to show to the member about what's wrong. In general we don't have any guidelines on how to handle such cases, so I think either approach would be fine.
The Dialog is one that requires them to actively dismiss it and is kinda in your face, while just saying "Something went wrong", so I thought in this case the snackbar would suffice. Don't have strong feelings about this though, if anything I'd love to discuss with a designer how such bad scenarios should be handled in this screen but alas we can't do that atm

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/CancelInsuranceViewModel.kt
hugokallstrom: I like naming it UiState as well! 

Regarding separate flows for state - for me it makes it a bit harder to follow since I have to know that eg. `dateSubmissionError` will get combined and result in `CancelInsuranceUiState`. And that `isPerformingCancelInsuranceNetworkRequest` translates to `isLoading` in the UiState. 

Its a bit clearer when you only have one state class that gets manipulated directly.

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/cancelinsurance/CancelInsuranceViewModel.kt
```diff
@@ -0,0 +1,124 @@
...
+    viewModelScope.launch {
+      val result = cancelInsuranceUseCase.invoke(
+        insuranceId,
+        selectedDateMillis,
+      )
+      when (result) {
+        is Either.Left -> dateSubmissionError.update { true }
+        is Either.Right -> dateSubmissionSuccess.update { true }
+      }
+      isPerformingCancelInsuranceNetworkRequest.update { false }
```
hugokallstrom, line 80: In cases like this, wont you get two state emissions - first with the error/success update, and then with the network request update? Would that be a problem?

StylianosGakis: Yes this would trigger twice, and there would be an in-between where you'd get a uiState with (error = true or success = true) while loading is still true.

In case of the error, this'd mean that nothing changes in this UI, as it has no loading animation. If there was one, you'd get 1 frame where you'd get the snackbar while the loading is still showing.
In case of success it doesn't matter since success would mean you're navigating away directly, so you won't see the difference.

All that said, I did make this back into a single StateFlow in [this commit](https://github.com/HedvigInsurance/android/commit/e38151f6e0cb8d12b93a4faf49ee44e3a1790e93) 😊

The difference now is that we're not using `.stateIn()` which means that when we exit the app and go to the background, this StateFlow will keep alive forever until the app is killed vs how before with `.stateIn()` it was stopping this flow after 5 seconds of inactivity. Not a big deal in our case and in this screen, so let's go with it. If this turns out to be important in some other screen, where we have some hot flow ongoing then we'd need to figure out how to model this with `.stateIn()` while also not getting this double emission.

StylianosGakis:
Gonna merge this in now as we've discussed all the points we had for the existing code and we have the feature flag in place. Will do the rest in follow-up PRs so that this does not stay too far behind develop. Hooray for fflags 🎉

PR #1424, opened 2023-02-27
Rename the :sample module to :micro-apps

StylianosGakis:
To disambiguate from ":sample-module", it was a bit confusing I think to
 have these two things but they are two quite different things.

Also move components to m3 package, to prepare for an m2 version of them too.


hugokallstrom (approved):

StylianosGakis:
Gonna maybe wait 1 day for this, just to make sure the text spacings are correct. Will merge tomorrow and fix it later so as not to block ourselves since it's not a big issue

PR #1426, opened 2023-02-27
Offpiste: Material2 without theme adapter extras

StylianosGakis:
For the XML theme setup. The "Discount" text style was removed as it was only used in one spot and that one spot was adjusted accordingly.

Also remove some rogue direct calls on Hedvig styles. Instead use
material's ?textAppearanceFoo to properly fetch from the theme.

---

Change marketing screen to simply use Dark theme style

Instead of manually providing and overriding some color values, do this:
Since the background image right now is dark, it needs content on top
which acts more like what content would act on a dark background,
therefore how the dark theme handles this. Lean into that and "force"
dark theme style for this screen only, while allowing the bottom sheet
to follow the device's theme normally.

---

Provide custom theme for contained buttons

For contained buttons, we got the special case where for light mode we
want primary color, but for dark mode we want tertiary color. Material3
colorScheme however does not hold information about light/dark mode
available for us to use like we did with material2.
This makes us have to make a new composition local, which decides which
color to use at the time of when our original colorScheme is generated
by giving the same scheme and then providing that down through a
composition local.

Then by utilizing extension functions on ColorScheme, we can make these
extra options look like they come from the normal ColorScheme.

Relevant discussion at:
https://issuetracker.google.com/issues/208860800#comment7

### DIFF
--- a/app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt
+++ b/app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt
@@ -16,8 +16,6 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import coil.ImageLoader
 import com.hedvig.android.core.designsystem.theme.HedvigTheme
-import com.hedvig.android.core.designsystem.theme.hedvig_black
-import com.hedvig.android.core.designsystem.theme.hedvig_off_white
 import com.hedvig.android.language.LanguageService
 import com.hedvig.android.market.Language
 import com.hedvig.android.market.Market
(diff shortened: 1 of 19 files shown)


app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt
```diff
@@ -46,16 +44,7 @@ class MarketingActivity : AppCompatActivity() {
...
-        colorOverrides = { colors ->
-          colors.copy(
-            primary = hedvig_off_white,
-            onPrimary = hedvig_black,
-            secondary = hedvig_off_white,
-            onBackground = hedvig_off_white,
-          )
-        },
-      ) {
+      HedvigTheme(darkTheme = true) { // Force dark theme as the background is dark
```
hugokallstrom, line 47: Nice workaround :)

hugokallstrom (approved):

PR #1438, opened 2023-03-02
Add permission dialog for audio recorder screen

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/audiorecorder/AudioRecorder.kt
```diff
@@ -44,6 +50,9 @@ fun AudioRecorderScreen(
 ) {
   val audioRecorderViewState by audioRecorderViewModel.viewState.collectAsState()
 
+  val openPermissionDialog = remember { mutableStateOf(false) }
```
StylianosGakis, line 53: I think it's a bit more natural to write these as 
```suggestion
  var openPermissionDialog by remember { mutableStateOf(false) }
```
So that you can just edit the value itself instead of having to call `.value` on it.

Having it as a var immediately shows us that we have control over its value and can edit it locally. If it's a val, you have to look inside the remember lambda, see that it's a `mutableState` and then you know you can edit it. Because you may well have a `remember { false }` which is just an immutable `false` value which is not editable.

So I think these are some compelling reasons to follow that style 😊

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/audiorecorder/PermissionDialog.kt
```diff
@@ -0,0 +1,42 @@
...
+        onClick = {
+          openDialog.value = false
+          recordAudioPermissionState.launchPermissionRequest()
+        },
+      ) {
+        Text(stringResource(android.R.string.ok))
+      }
+    },
+  )
+}
```
StylianosGakis, line 42: I'd propose to change this to
```suggestion
fun PermissionDialog(
  recordAudioPermissionState: PermissionState,
  permissionTitle: String,
  permissionMessage: String,
  dismiss: () -> Unit,
) {
  AlertDialog(
    onDismissRequest = { dismiss() },
    title = { Text(permissionTitle) },
    text = { Text(permissionMessage) },
    dismissButton = {
      TextButton(
        onClick = { dismiss() },
      ) {
        Text(stringResource(android.R.string.cancel))
      }
    },
    confirmButton = {
      TextButton(
        onClick = {
          dismiss()
          recordAudioPermissionState.launchPermissionRequest()
        },
      ) {
        Text(stringResource(android.R.string.ok))
      }
    },
  )
}
```

This dialog doesn't need to know about the status of this variable, in fact it should only be shown when it's true, it never does anything when the value is false, cause on the call-site it's simply not called therefore not shown on screen.
It just needs to know that it can dismiss itself, and on the call-site you can call this like:
```kotlin
  if (openPermissionDialog.value) {
    PermissionDialog(
      recordAudioPermissionState,
      "Permission required",
      "We need permission to record audio",
      dismiss = { openPermissionDialog = false }
    )
  }
```

StylianosGakis (approved):
Lgtm!

Not passing MutableState<> around I think is something we should avoid doing generally, posted a comment on how I'd suggest to write that function.
More context about why to avoid passing MutableState as parameter exists in this chat from one of the compose devs here if you're interested https://slack-chats.kotlinlang.org/t/506690/i-have-a-composable-function-that-gets-3-state-lt-t-gt-param#b2194ed3-39aa-45cd-9dad-224f8d6f8347

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/audiorecorder/AudioRecorder.kt
hugokallstrom: Fixing!

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/audiorecorder/PermissionDialog.kt
hugokallstrom: Excellent, fixing 👍

PR #1440, opened 2023-03-02
Single item price input

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/MonetaryAmountInput.kt
```diff
@@ -0,0 +1,181 @@
...
+        )
+      },
+      cursorBrush = SolidColor(cursorColor),
+      decorationBox = { innerTextField ->
+        BoxWithTrailingIcon(
+          textField = innerTextField,
+          showTrailingIcon = text.isNotEmpty(),
+          onClick = {
+            text = ""
+            focusRequester.freeFocus()
```
hugokallstrom, line 153: This does not seem to work. I can only press the "Purchase price"-button (not the input field) once in order to get focus. If I close the keyboard and press the button again, nothing happens.

hugokallstrom:
Screenshot 2023-03-02 at 14 39 50 

Getting this behaviour when inputing some text, height goes back to normal when clearing though. Any ideas?

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/MonetaryAmountInput.kt
StylianosGakis: Hmmm I think I've seen smth like this before.
Can you try fetching the manager using
`val focusManager = LocalFocusManager.current`
And then on that do
`focusManager.clearFocus()`
Maybe that helps? Or something like that?

Otherwise not sure, we can figure that out later though.

StylianosGakis (approved):

StylianosGakis:
Hmm interesting with the height change.
Maybe inside SingleItem.kt in the row where we do `horizontalArrangement = Arrangement.SpaceBetween,` we can also make them verticalAlign to center. Now why the height itself changes, not sure, kinda hard to figure out like this, but we can dig into it at some later point.

hugokallstrom:
> Hmm interesting with the height change. Maybe inside SingleItem.kt in the row where we do `horizontalArrangement = Arrangement.SpaceBetween,` we can also make them verticalAlign to center. Now why the height itself changes, not sure, kinda hard to figure out like this, but we can dig into it at some later point.

Needed to set icon size! Fixed now :)

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/MonetaryAmountInput.kt
hugokallstrom: Worked! Thanks

PR #1442, opened 2023-03-03
CLX-354: Success screen

StylianosGakis:
The screen itself was super straight-forward obviously, what I took some extra time to figure out was navigation and to start going away from 1 activity per screen if possible.

Using the same library as we did with Odyssey to navigate to screens (the official androidx.navigation), I explored a bit how this could look like to use in our app as well, to give us a bit more power in how we navigate.

This can be a nice little sample of how this can be done, but I was thinking of writing down my thoughts on this and how we can improve our app experience by doing this + how to incrementally adopt such a navigation system to go away from Activities if possible. Key being "incrementally" so that we don't need to sit down and spend time on this to do a all-out migration if we don't want to.

Might also be useful to have this plan down for when we get the navigation refresh that we might do in the near future


build-logic/convention/src/main/kotlin/com/hedvig/android/KotlinAndroid.kt
```diff
@@ -57,10 +57,12 @@ private fun KotlinJvmOptions.configureKotlinOptions(
   allWarningsAsErrors = project.properties["warningsAsErrors"] as? Boolean ?: false
 
   freeCompilerArgs = freeCompilerArgs + listOf(
+    "-opt-in=androidx.compose.material3.ExperimentalMaterial3Api",
+    "-opt-in=androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi",
```
StylianosGakis, line 61: I had forgotten we can just just opt-in globally like this. It's imo much preferred over having to have @OptIn(ExperimentalMaterial3Api) in like tens of places all over the codebase.

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/CancelInsuranceActivity.kt
```diff
@@ -28,24 +20,19 @@ class CancelInsuranceActivity : AppCompatActivity() {
...
-          showedError = viewModel::showedError,
-          navigateBack = { onBackPressedDispatcher.onBackPressed() },
+          navController = rememberNavController(),
+          insuranceId = insuranceId,
+          navigateUp = {
+            if (!onSupportNavigateUp()) {
+              finish()
+            }
+            true
+          },
```
StylianosGakis, line 35: Context here https://developer.android.com/guide/navigation/navigation-navigate#back-stack
Doing navigateUp() is what's expected by the system when clicking on the topAppBar back icon, as that should *not* exit the app if thats at all possible, but if `onSupportNavigateUp` returns false it means that that was not possible, so then we fallback to finish(). Taken from the docs linked above + this discussion https://slack-chats.kotlinlang.org/t/9661532/is-there-a-way-when-using-navigation-compose-to-immitate-exa#a951c61b-a697-4cd3-9f03-6299d916ec6d

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/data/CancelInsuranceUseCase.kt
```diff
@@ -7,22 +7,26 @@ import com.apollographql.apollo3.ApolloClient
...
     selectedDateMillis: Long,
   ): Either<OperationResult.Error, InsuranceId> {
     coroutineContext.ensureActive()
-    return if (failTimes != 0) {
-      failTimes--
-      OperationResult.Error.NetworkError("manual failure").left()
+    val submittedDate = Instant.fromEpochMilliseconds(selectedDateMillis).toLocalDateTime(TimeZone.UTC)
+    return if (submittedDate < Clock.System.now().plus(3.days).toLocalDateTime(TimeZone.UTC)) {
+      // Artificial failure when selecting in the next 3 days
+      OperationResult.Error.GeneralError("Too early").left()
```
StylianosGakis, line 29: This is still a mock implementation, just made it fail if you click on the first 3 available days to make it easy to test both success and error cases without waiting for anything

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/navigation/CancelInsuranceDestinations.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.kiwi.navigationcompose.typed.navigation
+import com.kiwi.navigationcompose.typed.popUpTo
+import kotlinx.serialization.Serializable
+import org.koin.androidx.compose.koinViewModel
+import org.koin.core.parameter.parametersOf
+
+internal sealed interface Destinations : Destination {
+  @Serializable
+  object CancelInsurance : Destinations
+}
```
StylianosGakis, line 24: This is like a route to the entire feature. In such a future, there'd be many destinations globally available where you can navigate from any other point in the app.
So navigating to this is like saying "Go to the terminate insurance feature, and it will handle its own navigation internally"

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/navigation/CancelInsuranceDestinations.kt
```diff
@@ -0,0 +1,68 @@
...
+
+internal sealed interface CancelInsuranceDestinations : Destination {
+  @Serializable
+  object TerminationDate : CancelInsuranceDestinations
+
+  @Serializable
+  object TerminationSuccess : CancelInsuranceDestinations
+}
```
StylianosGakis, line 32: And these are the possible screens *inside* this feature. In our case this is now picking the date, and the success screen. Here would fit the error screen, the confirmation screen, and so on, from Figma.

p.s. I wanna rename everything from "cancelinsurance to insurance-termination, I've made a bit of a mess using both terms interchangeably

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/navigation/CancelInsuranceDestinations.kt
```diff
@@ -0,0 +1,68 @@
...
+
+internal fun NavGraphBuilder.cancelInsuranceGraph(
```
StylianosGakis, line 34: This "graph" is what would be exposed from this module to the :app module, which would add all these paths to the graph of possible routes to navigate to.
I envision this future where each module can expose one graph like this, and the :app module can just add all of them to one NavHost, for it to have access to navigate to all of them.
For now though, the activity itself can simply host their own NavHost, which *only* has this one graph, of the destinations defined above.

hugokallstrom: Thats great. I just ran into this as well when building the summary screens. Ill look at this when doing more work on that feature!

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/navigation/CancelInsuranceDestinations.kt
hugokallstrom: Super cool

hugokallstrom (approved):

feature-cancel-insurance/src/main/kotlin/com/hedvig/android/feature/cancelinsurance/navigation/CancelInsuranceDestinations.kt
StylianosGakis: Yeah I figured at this point as we do more stuff like this we might benefit from taking a few moments to think how to improve our experience with navigation. I think this time is as good as it gets, we're gonna touch navigation soon and the androidx navigation library is in a good spot to try this out with from what I've seen/read and I think I have a plan in my head that could work. Just need to try it out a bit and see how it feels.

PR #1445, opened 2023-03-06
CLX-354: Success screen *with* screen transitions

StylianosGakis:
Make use of Accompanis AnimatedNavHost instead of just NavHost, so that we can leverage proper animation support.
Make use of https://m3.material.io/styles/motion/transitions/transition-patterns#61e7e3f3-5035-4f00-a17f-b1f9b16689f9 the lateral movement animation for this, also known as "shared X axis" transition.


hugokallstrom (approved):
Nice!

PR #1446, opened 2023-03-06
Fix: Improve compose previews

StylianosGakis:
This is a PR which basically does these things:
* Create a _HedvigPreview_ which contains both the two most common preview annotations we used, which was the standard Preview + the Preview in Dark mode. Context: https://developer.android.com/jetpack/compose/tooling/studio#preview-multipreview
* Make all previews `private` visibility. This was something that was not possible earlier on which meant that you could accidentally call previews from other places in the app which is definitely not something we ever want to do.
* Make all previews start with the `Preview...` suffix in the name, so that again while inside that file, when you want to call the composable you don't call the preview one by accident. Again a fix to improve what you can access with autocomplete to improve developer experience a bit
* Wrap all composables with `HedvigTheme {}`. This represents how they really are going to be shown in the app, and now that we don't use the m2 material theme adapter we can use that in child modules too.
* Introduce a @HedvigMultiScreenPreview as an alternative to @HedvigPreview just as a way to see how a screen looks like also in portrait mode and in bigger screens. To be used in screen-level composables.


hugokallstrom (approved):
Very good cleanup

PR #1447, opened 2023-03-06
CLX-339: Entry points triaging

hugokallstrom:
Screenshot 2023-03-06 at 11 23 17

Getting these with REST from odyssey. Will change to GraphQL when its ready. 

We should also include colors in the schema so that we can avoid heard coding those on the clients. 

Other than that this works well!


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -0,0 +1,112 @@
...
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.platform.LocalSoftwareKeyboardController
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.auth.android.AuthenticatedObserver
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.core.ui.appbar.CenterAlignedTopAppBar
+import com.hedvig.android.odyssey.OdysseyClaimsFlowActivity
+import org.koin.androidx.viewmodel.ext.android.getViewModel
+
+class SearchActivity : ComponentActivity() {
```
hugokallstrom, line 29: This is not used

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchAppBar.kt
```diff
@@ -0,0 +1,135 @@
...
+import androidx.compose.ui.text.TextRange
+import androidx.compose.ui.text.input.TextFieldValue
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.hedvig.android.core.ui.appbar.CenterAlignedTopAppBar
+import com.hedvig.android.odyssey.R
+
+@Composable
+fun SearchAppBar(
```
hugokallstrom, line 41: Not used

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+    lifecycle.addObserver(AuthenticatedObserver())
+
+    setContent {
+      val viewModel: SearchViewModel = getViewModel()
+      val viewState by viewModel.viewState.collectAsState()
+
+      val commonClaimId = viewState.selectedClaim?.id
+      if (commonClaimId != null) {
+        startClaimsFlow(viewModel, commonClaimId)
+      }
```
StylianosGakis, line 42: This'd need to be:
```suggestion
      LaunchedEffect(commonClaimId) {
        if (commonClaimId != null) {
          startClaimsFlow(viewModel, commonClaimId)
        }
      }
```
Since otherwise `startClaimsFlow()` is called in composition, which may be called more than once without you wanting it to. 

Kinda covered in [this section](https://developer.android.com/jetpack/compose/mental-model#simple-example) in the docs which asks for composable functions to be idempotent, meaning if called again with the same inputs, it shouldn't change anything, in which case here it will, by calling startClaimsFlow() again without the commonClaimId changing.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  private fun startClaimsFlow(viewModel: SearchViewModel, commonClaimId: String?) {
+    val intent = OdysseyClaimsFlowActivity.newInstance(
+      context = this,
+      odysseyUrl = intent.getStringExtra(ODYSSEY_URL) ?: throw java.lang.IllegalArgumentException("No url found"),
```
StylianosGakis, line 82: Can this intent.getStringExtra() be top-level, kinda as we do in [here](https://github.com/HedvigInsurance/android/blob/2f6afcf235ecf2a0677ef225cf0d219baa9b828f/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/TerminateInsuranceActivity.kt#L21-L22) for example. Makes it easier to at a glance see what inputs an activity is asking for, along with the aforementioned "newInstance()" function which would help here.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.hedvig.android.odyssey.model.SearchableClaim
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+class SearchViewModel(
+  private val getClaimEntryPoints: GetClaimEntryPoints,
+) : ViewModel() {
+  private val searchableClaims = emptyList<SearchableClaim>()
+  private val _viewState = MutableStateFlow(SearchViewState(results = searchableClaims))
```
StylianosGakis, line 15: This isn't used anywhere else aside for the initial status of the _viewState. Do you think it'd be clearer if it was inlined there instead
```suggestion
  private val _viewState = MutableStateFlow(SearchViewState(results = emptyList<SearchableClaim>()))
```
or would you find that confusing? This is a bit of a personal preference, no super strong feelings for it, but I would've inlined it I think.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+        }
+        is CommonClaimsResult.Success -> _viewState.update {
+          it.copy(
+            commonClaims = result.searchableClaims,
+            isLoading = false,
+          )
+        }
+      }
+    }
+  }
```
StylianosGakis, line 35: This ATM means that at first load, if there's an error there's no way to reset this error state right? Always a bit tricky with VMs and doing stuff like this in `init{}`.
But I guess we also have no specs for a "retry" screen, so might have to keep it this way and have them go back and go into the screen again to handle such cases for now.

StylianosGakis (approved):
Great work with this!

Had some thoughts here and there, most important is the LaunchedEffect one and maybe catching the JSONException, the rest aren't as critical before this is merged.

app/src/main/AndroidManifest.xml
```diff
@@ -265,6 +265,14 @@
             android:screenOrientation="portrait"
             android:exported="true"
             android:windowSoftInputMode="adjustResize" />
+        <activity
+            android:name="com.hedvig.android.odyssey.OdysseyClaimsFlowActivity"
+            android:screenOrientation="portrait"
+            android:exported="true"
```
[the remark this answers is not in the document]
hugokallstrom: Forgot to remove, good catch!

app/src/main/kotlin/com/hedvig/app/ApplicationModule.kt
```diff
@@ -686,3 +690,7 @@ val claimsViewModelModule = module {
   }
   viewModel { (resolution: Resolution) -> ResolutionViewModel(get(), resolution) }
 }
+
+val claimsSearchViewModelModule = module {
+  viewModel { SearchViewModel(get<GetClaimEntryPoints>()) }
+}
```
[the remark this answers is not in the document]
hugokallstrom: I need the odyssey url string here: 
`GetClaimEntryPointsUseCase(get(), get<Context>().getString(R.string.ODYSSEY_URL))`

Would like to move that string to core-resources I suppose, but then we need different build flavors for that module right? Can you help me with that if thats something we want to do? :)

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchViewModel.kt
hugokallstrom: Yeah no retry screen for now, we will hopefully get more elaborate designs soon!

PR #1450, opened 2023-03-07
CLX-427: Odyssey schema in android repo pt.1

StylianosGakis:
Split apollo in three modules

giraffe: To contain giraffe GQL operations
octopus: To contain octopus GQL operations
core: To contain apollo related convenience functions, like the arrow
extension functions.

This will mean that new modules can only depend on octopus for example
and not giraffe, so that they don't get access to gql operations that
they don't want to call in the first place.

Also make sure CI is calling the right commands to download the schemas.
Make ApolloConventionPlugin to generate said gradle tasks, along with
applying the apollo gradle plugin and sharing any other common logic
needed.


hugokallstrom (approved):
Looks great

PR #1451, opened 2023-03-07
CLX-427: Odyssey schema in android repo pt.2

StylianosGakis:
Adds the qualifier to all `get<ApolloClient>()` calls, and creates the octopus client too to be used in the future.

The important bit here is the now we can't do `get<ApolloClient>()` anymore without a qualifier, we need to specify which one it has to be,  a̶n̶d̶ ̶u̶s̶i̶n̶g̶ ̶t̶h̶e̶ ̶w̶r̶o̶n̶g̶ ̶c̶l̶i̶e̶n̶t̶ ̶i̶s̶ ̶t̶o̶t̶a̶l̶l̶y̶ ̶u̶p̶ ̶t̶o̶ ̶u̶s̶,̶ ̶t̶h̶e̶r̶e̶ ̶i̶s̶n̶'̶t̶ ̶a̶ ̶w̶a̶y̶ ̶t̶o̶ ̶f̶i̶x̶ ̶a̶s̶k̶i̶n̶g̶ ̶f̶o̶r̶ ̶t̶h̶e̶ ̶w̶r̶o̶n̶g̶ ̶q̶u̶e̶r̶y̶ ̶u̶s̶i̶n̶g̶ ̶t̶h̶e̶ ̶w̶r̶o̶n̶g̶ ̶c̶l̶i̶e̶n̶t̶.̶ This is fixed in https://github.com/HedvigInsurance/android/pull/1452 with :app no longer depending on :apollo:octopus.

The only way to fix this is for a module to *only* depend on `:apollo:giraffe` or only to `:apollo:octopus` which will mean that they *can't* even see the other queries/mutations etc, so you are unable to make the mistake at all.

### DIFF
--- a/app/src/androidTest/kotlin/com/hedvig/app/TestApplication.kt
+++ b/app/src/androidTest/kotlin/com/hedvig/app/TestApplication.kt
@@ -1,9 +1,6 @@
 package com.hedvig.app
 
 class TestApplication : HedvigApplication() {
-  override val graphqlUrl = "http://localhost:$PORT/"
-  override val graphqlSubscriptionUrl = "http://localhost:$PORT/"
-
   companion object {
     const val PORT = 8080
   }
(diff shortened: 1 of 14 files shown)


hugokallstrom (approved):
💯

PR #1453, opened 2023-03-07
Have termination feature use only octopus apollo client

StylianosGakis:
By changing the dependency from `implementation(projects.apollo.giraffe)` to `implementation(projects.apollo.octopus)`, now this module *can not* reference giraffeClient, nor any of the generated classes it has, so no longer can make a mistake of asking for the wrong apolloClient in these cases where we can only depend on one of the two.


hugokallstrom (approved):
Super clean!

PR #1454, opened 2023-03-08
Check feature flag when starting claims flow from search activity

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -40,8 +48,10 @@ class SearchActivity : ComponentActivity() {
 
       val commonClaimId = viewState.selectedClaim?.id
       LaunchedEffect(commonClaimId) {
-        if (commonClaimId != null) {
-          startClaimsFlow(viewModel, commonClaimId)
+        lifecycleScope.launch {
+          if (commonClaimId != null) {
+            startClaimsFlow(viewModel, commonClaimId)
+          }
```
StylianosGakis, line 54: I think inside the LaunchedEffect you're in a suspending context already, so you can call startClaimsFlow directly without launching a new coroutine, can you try that and see if it works?

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -68,10 +78,12 @@ class SearchActivity : ComponentActivity() {
...
-                startClaimsFlow(
-                  viewModel = viewModel,
-                  commonClaimId = null,
-                )
+                lifecycleScope.launch {
+                  startClaimsFlow(
+                    viewModel = viewModel,
+                    commonClaimId = null,
+                  )
+                }
```
StylianosGakis, line 86: I think since we're inside compose right here, it's preferable to use the coroutine context provided through `rememberCoroutineScope`.
By that I mean basically in line 76, add `val coroutineScope = rememberCoroutineScope()`, and inside here, instead of doing `lifecycleScope.launch {` you can do `coroutineScope.launch {`.
This means that it will both follow the composable's lifecycle better, but also if we ever extract this function outside of the activity (where we're getting the lifecycleScope from), then it'd still work.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -80,16 +92,23 @@ class SearchActivity : ComponentActivity() {
     }
   }
 
-  private fun startClaimsFlow(viewModel: SearchViewModel, commonClaimId: String?) {
-    val intent = OdysseyClaimsFlowActivity.newInstance(
-      context = this,
-      odysseyUrl = odysseyUrl,
-      commonClaimId = commonClaimId,
-    )
+  private suspend fun startClaimsFlow(viewModel: SearchViewModel, commonClaimId: String?) {
```
StylianosGakis, line 95: Maybe not in the scope of this PR, maybe in the future we can consider removing passing the entire viewModel here, but instead pass a lambda to call the viewmodel would be a bit nicer.

StylianosGakis (approved):
lgtm 💪

PR #1455, opened 2023-03-08
Improve phone number input. Align payout summary according to design.

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/input/ui/PhoneNumber.kt
```diff
@@ -54,11 +58,22 @@ fun PhoneNumber(
...
+          autoCorrect = false,
+          keyboardType = KeyboardType.Phone,
+          imeAction = ImeAction.Done,
+        ),
+        keyboardActions = KeyboardActions(
+          onDone = {
+            updatePhoneNumber()
+            onNext()
+          },
+        ),
```
StylianosGakis, line 71: Nice! Should be a better experience this way

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/resolution/ui/PayoutSummary.kt
```diff
@@ -31,59 +34,89 @@ fun PayoutSummary(
...
   onPayout: (MonetaryAmount) -> Unit,
 ) {
-  Box(
-    Modifier
-      .fillMaxHeight()
-      .padding(all = 16.dp),
-  ) {
-    Column {
-      Spacer(modifier = Modifier.padding(top = 28.dp))
+  Surface(color = MaterialTheme.colors.background) {
```
StylianosGakis, line 37: Was this surface needed for this particular composable, or was it needed for the preview to look better?
In general, there's always a Surface top-level, on the screen-level basically. In this particular case, [this Scaffold](https://github.com/HedvigInsurance/android/blob/01dfdad98012ef19a7f247a4a80b23d1ad3538f9/feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/resolution/ui/ResolutionRoot.kt#L30) (which internally uses a Surface with color = MaterialTheme.colors.background) should cover this case.

My guess here is that you added this because the preview was looking odd, but I'll suggest you try moving this to the preview instead, we do this in most other previews you can look there for inspiration.

Correct me if I am wrong though of course 😅

StylianosGakis (approved):
lgtm 🎨👨‍🎨

Just a commment about what I *think* is a Surface which we can remove, but I could be wrong 🫣

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/resolution/ui/PayoutSummary.kt
hugokallstrom: Oh that makes total sense. Ill move it to the preview instead. Thanks

PR #1458, opened 2023-03-08
CLX-390: Termination network communication

hugokallstrom:
Add mutations for starting and setting date for termination. Observe step and show appropriate UI.


apollo/octopus/src/main/graphql/com/hedvig/android/apollo/octopus/graphql/flowTerminationDateNext.graphql
```diff
@@ -0,0 +1,21 @@
+mutation FlowTerminationDateNext($context: FlowContext!, $input: FlowTerminationDateInput!) {
+  flowTerminationDateNext(context: $context, input: $input) {
+    context
+    currentStep {
+      ... on FlowTerminationDateStep {
+        id
+        minDate
+        maxDate
+      }
```
StylianosGakis, line 9: So these 3 we can try turning into a fragment to see if this gives us an easier experience working with it right?

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/TerminateInsuranceViewModel.kt
```diff
@@ -18,25 +20,37 @@ import kotlinx.datetime.TimeZone
 import kotlinx.datetime.atStartOfDayIn
 import kotlinx.datetime.toInstant
 import kotlinx.datetime.toLocalDateTime
+import java.time.Instant
+import java.time.ZoneId
```
StylianosGakis, line 24: Can we try using the kotlinx.datetime.Instant and same for zone? In general they're easier to work with in my experience, and it's nice to be consistent as that is what the Date response from octopus is returning to us anyway.

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/TerminateInsuranceRepository.kt
```diff
@@ -0,0 +1,56 @@
...
+      is OperationResult.Success -> {
+        terminationContext = res.data.flowTerminationDateNext.context
+        res.data.flowTerminationDateNext.currentStep.toTerminationStep()
+      }
+    }
+  }
+}
+
+private fun FlowTerminationStartMutation.Data.FlowTerminationStart.CurrentStep.toTerminationStep(): TerminationStep = when (this) {
+  is FlowTerminationStartMutation.Data.FlowTerminationStart.FlowTerminationDateStepCurrentStep -> TerminationStep.Date(minDate, maxDate)
```
StylianosGakis, line 44: Would suggest importing the entire thing here so we only see the last type, not everything that comes before it. This might also be easier after we introduce the fragments

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/TerminationStep.kt
```diff
@@ -0,0 +1,19 @@
...
+  ) : TerminationStep
+
+  class Success(
+    val terminationDate: LocalDate,
+    val surveyUrl: String,
+  ) : TerminationStep
+
+  class Failed(
+    val message: String?,
+  ) : TerminationStep
```
StylianosGakis, line 18: I think it's a good idea to make those `data class` to get the free `equals` implementation out of them. Especially if these ever reach down to compose levels and they need to be compared it's nice to get it for free.

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
```diff
@@ -42,15 +41,7 @@ internal fun NavGraphBuilder.terminateInsuranceGraph(
...
-      )
-    }
-    animatedComposable<TerminateInsuranceDestinations.TerminationSuccess> {
-      TerminationSuccessDestination(
-        windowSizeClass = windowSizeClass,
-        // Can't use navigateUp() from non-starting destination while it's the only one in the backstack due to not
-        // having a fixed startDestination. This may be hard to achieve our this mixed compose-activity situation.
-        // https://issuetracker.google.com/issues/271549886
-        navigateBack = finishTerminationFlow,
+        finishTerminationFlow = finishTerminationFlow,
```
StylianosGakis, line 44: Yeah this way it means that we are no longer really using the navigation, but everything is in one destination with the inner content changing, and the AnimatedContent doing the work.
I am sure the way it is right now if you are X steps in and press back the entire thing gets closed down since this function is called, and since there's nothing else in the backstack it exits.

Will definitely try and work with it though, I can try and look into this into another PR

StylianosGakis (approved):
Looking good, gonna need to do something for the navigation yeah, but we can look into this in another PR.

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/TerminateInsuranceViewModel.kt
```diff
@@ -83,20 +88,17 @@ private class DatePickerConfiguration(clock: Clock) {
     .date
     .atStartOfDayIn(TimeZone.UTC)
     .toLocalDateTime(TimeZone.UTC)
-  private val todayAtStartOfDayEpochMillis = todayAtStartOfDay.toInstant(TimeZone.UTC).toEpochMilliseconds()
   private val yearsRange = todayAtStartOfDay.year..2100
 
   val datePickerState = DatePickerState(null, null, yearsRange)
-  val dateValidator: (Long) -> Boolean = { selectedDateEpochMillis ->
-    selectedDateEpochMillis >= todayAtStartOfDayEpochMillis
-  }
```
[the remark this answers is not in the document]
hugokallstrom: Yeah I create this in the composable with the min and max date instead

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
hugokallstrom: Thats true. We will have to figure that out when we introduce more steps for sure!

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/TerminateInsuranceRepository.kt
hugokallstrom: Fragments does not help here unfortunately... apollo still generates two different implementations, one in `FlowTerminationStartMutation` and one in `FlowTerminationDateNextMutation`:
```
public sealed interface CurrentStep : StepFragment {
        public override val __typename: String

        public companion object {
          public fun CurrentStep.stepFragment() = this as? StepFragment

          public fun CurrentStep.asFlowTerminationDateStep() = this as?
              FlowTerminationDateStepCurrentStep

          public fun CurrentStep.asFlowTerminationFailedStep() = this as?
              FlowTerminationFailedStepCurrentStep

          public fun CurrentStep.asFlowTerminationSuccessStep() = this as?
              FlowTerminationSuccessStepCurrentStep
        }
      }
```

I fixed the imports though, so its a bit easier on the eyes.

PR #1463, opened 2023-03-09
CLX-349: Update peril design

StylianosGakis:
Create a specialized request for insurance coverage only

Now we're reusing ContractDetailViewState in separate fragments making
it hard to change any of them as they are bound to each other. This
extracts the coverage fragment to handle its own stuff.
This is done by removing the dependency of this screen to
ContractDetailViewState and introducing GetContractCoverageUseCase and CoverageViewModel
along with a new GQL query


hugokallstrom (approved):
LGTM! Good clean up!

PR #1467, opened 2023-03-10
Feature: termination flow navigation

StylianosGakis:
Use GQL fragments for FlowSteps

Make termination destinations use androidx.navigation

Use a normal backstack approach of adding the steps on top of each
other on the backstack.
When reaching any of the unrecoverable steps, which is the unknown step,
the finished successfully step and the termination error step, make it
so that the entire flow is popped off the backstack instead

Use a ..Destination suffix naming convention for the composable holding for each of the destinations


feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
```diff
@@ -1,47 +1,101 @@
...
+      BackHandler { finishTerminationFlow() }
+      TerminationSuccessDestination(
+        terminationDate = this.terminationDate,
+        surveyUrl = this.surveyUrl,
+        windowSizeClass = windowSizeClass,
+        navigateBack = finishTerminationFlow,
+      )
+    }
+    animatedComposable<TerminateInsuranceDestination.TerminationFailure> {
+      BackHandler { finishTerminationFlow() }
```
hugokallstrom, line 87: Should we make this go back to the previous step? In case you would want to retry

hugokallstrom (approved):
Really nice!

feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
StylianosGakis: Took this to Slack as I don't know myself either. If backend supports this then we will make it so.

PR #1468, opened 2023-03-13
Fix app icon handling by adding easylauncher plugin

StylianosGakis:
Add the plugin and configure it to give a "debug" banner for the "debug"
 build, and a "staging" bottom-attached banner for the staging build.
This replaces having different icons in XML for the build types. Also enables monochromatic icons for the OS versions supporting it
Also enables monochromatic icons for the OS versions supporting it and for 
the users who have opted in.

In this process, also clear some unused XML resources here and there and bump gradle version

| normal | themed dark mode | themed light mode |
| --- | --- | --- |
| image | image | image |


StylianosGakis:
Note, I don't even know how the old icon inside /release was not picked up in favor of the one inside /main. If you put a string in /main and in /release, it seems like it prefers the one inside /release by default. But anyway, this removes them completely and does not leave room for a mistake there :D

app/src/engineering/res/xml/network_security_config.xml
```diff
@@ -1,11 +0,0 @@
...
-<network-security-config>
-    <base-config cleartextTrafficPermitted="true">
-        <trust-anchors>
-            <!-- Trust preinstalled CAs -->
-            <certificates src="system" />
-            <!-- HERE: Additionaly trus user added CAs -->
-            <certificates src="user"/>
-        </trust-anchors>
-    </base-config>
-</network-security-config>
```
StylianosGakis, line 11: This was used by Martyna to install some network interceptor to log the network. We no longer are using that approach so removing it completely makes sense.

app/build.gradle.kts
```diff
@@ -291,3 +294,25 @@ dependencies {
   debugImplementation(libs.androidx.compose.uiTestManifest)
   androidTestImplementation(libs.androidx.compose.uiTestJunit)
 }
+
+easylauncher {
+  buildTypes.register("staging") {
+    setFilters(
+      customRibbon(
```
hugokallstrom, line 301: So cool :D

hugokallstrom (approved):
Super!!

PR #1475, opened 2023-03-17
CLX-339: Claims triaging design

hugokallstrom:
Screenshot_20230316_160805

Not getting the correct claim flows for some of these entrypoints, talking to Linus about that. We also want to get the "Other" option last in the list.

### DIFF
--- a/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigTypography.kt
+++ b/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigTypography.kt
@@ -17,21 +17,21 @@ internal val HedvigTypography: Typography
   @Composable
   get() = Typography(
     displayLarge = MaterialTheme.typography.displayLarge.copy(
-      fontFamily = SerifBookSmall,
+      fontFamily = SansStandard,
       fontWeight = FontWeight.Normal,
       fontSize = 57.sp,
       lineHeight = 64.sp,
       letterSpacing = (-1f).percentage.em,
     ),
     displayMedium = MaterialTheme.typography.displayMedium.copy(
-      fontFamily = SerifBookSmall,
+      fontFamily = SansStandard,
       fontWeight = FontWeight.Normal,
       fontSize = 45.sp,
       lineHeight = 52.sp,
       letterSpacing = (-0.5f).percentage.em,
     ),
     displaySmall = MaterialTheme.typography.displaySmall.copy(
-      fontFamily = SerifBookSmall,
+      fontFamily = SansStandard,
       fontWeight = FontWeight.Normal,
       fontSize = 36.sp,
       lineHeight = 44.sp,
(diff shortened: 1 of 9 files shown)


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/ui/CommonClaims.kt
```diff
@@ -51,77 +51,26 @@ internal fun CommonClaims(
...
-
-              Spacer(modifier = Modifier.padding(12.dp))
-
               Text(claim.displayName, Modifier.align(Alignment.CenterVertically))
+              Icon(
+                painter = painterResource(id = com.hedvig.android.odyssey.R.drawable.ic_list_item_indicator),
+                contentDescription = "Arrow",
+              )
+            }
+            if (index < commonClaims.size - 1) {
```
StylianosGakis, line 64: A nice convenience Kotlin offers is being able to change this with `lastIndex`. Exactly like it's used here https://github.com/HedvigInsurance/android/blob/fbe54efb7b913ae465b3edda3b8f4cd8ab23a67d/feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/step/location/LocationDestination.kt#L127-L129
```suggestion
            if (index != commonClaims.lastIndex) {
```

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/ui/CommonClaims.kt
```diff
@@ -130,37 +79,36 @@ fun PreviewCommonClaims() {
...
       commonClaims = listOf(
         SearchableClaim(
-          id = "1",
+          entryPointId = "1",
           displayName = "Broken phone",
           itemType = ItemType("PHONE"),
         ),
         SearchableClaim(
-          id = "4",
+          entryPointId = "4",
```
StylianosGakis, line 87: Since the entryPointId is used as a `key` for LazyColumn, I think if it's not unique it will crash, making this preview also not work? Not sure, but might be worth checking and if it does not work adding a unique entryPointId on each of these preview items.

StylianosGakis (approved):
lgtm! 🤩

core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigTypography.kt
```diff
@@ -11,26 +11,27 @@ import androidx.compose.ui.tooling.preview.Preview
...
-      fontFamily = SansStandard,
+      fontFamily = SerifBookSmall,
       fontWeight = FontWeight.Normal,
       fontSize = 45.sp,
       lineHeight = 52.sp,
       letterSpacing = (-0.5f).percentage.em,
     ),
     displaySmall = MaterialTheme.typography.displaySmall.copy(
-      fontFamily = SansStandard,
+      fontFamily = SerifBookSmall,
```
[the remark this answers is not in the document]
hugokallstrom: Good info! Ill just add the font on that particular text then

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/ui/CommonClaims.kt
hugokallstrom: NIce :D

PR #1477, opened 2023-03-17
Feature: claim flow file upload

StylianosGakis:
File uploading now works 🎉
And after it's done uploading, just fixed being able to navigate to the single item screen afterwards, by making its destination class properly serializable.

Use the correct flow ID to be passed in, previously was using the ID from the current step which was wrong.

Also add some rough UI for the case where they've permanently denied microphone access to the app. Otherwise we show the dialog, they press "OK" and nothing happens, which would be super odd for most of our members if I were to guess. I'd like us to have a discussion for these cases, and if we can gracefully let them continue the flow without submitting an audio recording at all. Since we have other data to base our decisions off of now. Or if not, if we can get some more thoughtful copy to explain why we need the audio permission to get their recording.

image


hugokallstrom (approved):
LGTM!

PR #1480, opened 2023-03-17
Termination flow touches

StylianosGakis:
* Fetch termination data from network

Would otherwise fetch stale data regarding the termination flow right
after coming back from the termination flow.

* Don't show termination flow when already terminated

Simply hide the button for already terminated contracts


hugokallstrom (commented):
LGTM! 

Should this be merged to develop?

hugokallstrom (approved):

StylianosGakis:
> LGTM!
> 
> Should this be merged to develop?

Yes, this should be fine there, is there something in particular you are concerned about before we merge this to develop?
Or do you perhaps mean that the target of this PR should be develop directly?

PR #1482, opened 2023-03-20
Fix: do not get stale contract detail data

StylianosGakis:
Avoids showing stale data when just finishing the termination flow which
 would also mean the button is shown again.
By using WhileSubscribed, it means that when coming back from the flow, the timeout has occurred, starting the flow again, and triggering the query again to refresh the data automatically, while still showing the old data initially while the new query is being executed.

In practice this basically just means that when you're done terminating your insurance, as you go back to the insurance detail screen, it automatically fetches the latest information, showing that your insurance is terminated, instead of having to go back in the insurance tab and pull to refresh for it to show.


hugokallstrom (approved):
Very nice

PR #1484, opened 2023-03-21
Also add a "Done" button to more easily exit the screen

StylianosGakis:
Before, it felt a bit unintuitive to leave the success screen. You had to either go back, or press the back arrow on the top left, but there wasn't any "ok I am done now" action, and the survey might feel like a dead end, since clicking it doesn't also leave this screen, only opens a website, and going back may feel like they're going back into the flow, so I thought this makes sense.

image


hugokallstrom (approved):
Agree, also think this makes more sense. GJ!

PR #1486, opened 2023-03-21
CLX-462: Triaging design tweaks

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/ui/CommonClaims.kt
```diff
@@ -58,7 +59,7 @@ internal fun CommonClaims(
             ) {
               Text(claim.displayName, Modifier.align(Alignment.CenterVertically))
               Icon(
-                painter = painterResource(id = com.hedvig.android.odyssey.R.drawable.ic_list_item_indicator),
+                painter = painterResource(id = R.drawable.ic_arrow_forward),
```
StylianosGakis, line 62: Is this any different from the one which comes from `Icons.Default.ArrowForward`?
I think we should just use that directly, as it's for sure the one which is what's expected for android, as seen in [here](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:arrow_forward:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=arrow+for) which is `Icons.Default.ArrowForward` in compose.
This way we also can remove the drawable XML completely.

StylianosGakis (changes requested):
I am pretty sure we can switch this with just `Icons.Default.ArrowForward`.
If they are the same, I'd suggest we just use it directly. If I am wrong and they are different, feel free to merge this.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/ui/CommonClaims.kt
hugokallstrom: Excellent, will change to using that instead 👍

PR #1499, opened 2023-03-28
CLX-468: Termination unknown step and claim unknown step open play store

StylianosGakis:
Moves `tryOpenPlayStore()` to core-common-android to make it accessible in the feature module.
I didn't move the `makeToast()` function but just copied the code as I hope we won't be using this so much anyway in new features.

image


hugokallstrom (commented):
Neat!

PR #1506, opened 2023-03-30
Fix crash when no peril items exist

StylianosGakis:
Crash:
https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/e98ea7e8fd8195f23e69ca44c5bd1590?time=last-thirty-days&versions=10.3.0%20(1679681923)&sessionEventKey=642349D10118000179DE3299BA90214A_1794138022370128837

Fixes the crash by making sure we never send no items inside the grid, and instead do not render the grid nor the title above it at all when the list is empty


hugokallstrom (approved):
LGTM!!

PR #1508, opened 2023-04-05
CLX-502: Claim single item payout screen

StylianosGakis:
All states:

| Loading | Error | Success |
| --- | --- | --- |
| image | image | image |

The error screen with the three buttons is smth to be decided, just added something there so it works for now, but it does not look all that great, until we make a better decision. Currently there's no design about it.

And some video of it working

https://user-images.githubusercontent.com/44558292/230097742-dce6c2a2-6862-4bc2-badb-4b1bb6eef0fc.mp4


hugokallstrom (approved):
Looks great! Can we format the currency string to not include any decimals if its .0? And otherwise show two decimals, like 3990.50?

StylianosGakis:
Yes you are totally right. I had it in my mind to try and find the right formatting configuration (Maybe using java.text.NumberFormat or something like that), for now I've just been doing a simple toString on all the currency related stuff in this flow.
Will fix on another PR!

PR #1510, opened 2023-04-05
CLX-496: Query type of contract and hide co-insured button depending on result

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/detail/ContractExt.kt
```diff
@@ -74,3 +75,48 @@ fun InsuranceQuery.Contract.toDocumentsViewState() = ContractDetailViewState.Doc
     ),
   ),
 )
+
+fun InsuranceQuery.Contract.canChangeCoInsured() = when (typeOfContract) {
```
StylianosGakis, line 79: This list is decided to be a client-side check I suppose? Might be easy to both make mistakes on android-ios, and/or have different setup in each client, especially if this list is ever to be updated.
But just smth to keep in mind, no need to move to backend if this wasn't in the plans anyway.

StylianosGakis (approved):
lgtm!

app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/detail/ContractExt.kt
hugokallstrom: Yes this lived in giraffe before, and we still have for example `canChangeAddress`. I think logic like this is fine to have in the clients though, its super easy to find and change - and the frequency of which we need to change this should be pretty low. Its also not critical business logic, this will just hide or show a button - the business logic (can I change co-insured) still lives in the backend.

PR #1511, opened 2023-04-05
CLX-456: Claims flow small UI touches

StylianosGakis:
Had an issue with some items not looking disabled while loading

| Before | After |
| --- | --- |
| image | image |

Now allows the user to select multiple items in the damage section, instead of instantly dismissing the dialog as soon as the first selection is done

image

And replace the title in the summary step with the one coming from the backend as it now exists in the GQL schema

| Before | After |
| --- | --- |
| image | image |


hugokallstrom (approved):
LGTM

PR #1516, opened 2023-04-12
CLX-520: Show correct charge amount

hugokallstrom:
`insuranceCost` is not working correctly and will be deprecated soon. `chargeEstimation` is the new way to go!


StylianosGakis (approved):
Nice, lgtm

PR #1517, opened 2023-04-13
CLX-457: Register push token after login

hugokallstrom:
Fix FCM token on login/logout

This fixes previous issues with how the FCM token was handled and how
registering it to the backend was not consistent.
This also solves the issue of people receiving notifications after
they've logged out, potentially receiving notifications for a use they
no longer should have access to.

It does so by doing the following:
* No longer generate a FirebaseMessaging token on app startup. This is
now disabled in the AndroidManifest.xml
* In the event of logging in, a new token is requested and is tehrefore
generated. This is then stored and a WorkManager persistent job is
kicked off to upload this to the backend. It's done in that way so that
there is no chance that we get some error and fail to do so.
* If there is a logout event, the stored token is cleared, and any
previous workmanager jobs that may have been ongoing are cancelled as we
no longer should upload that token to the backend
* In the logout event, we also explicitly delete the FirebaseMessaging
token as that makes sure that the new notifications do not reach our
device, and we rely on having the token auto-generation turned off to
ensure that we only get a new token on the login event again.
* When we receive the `onNewToken` callback from our
FirebaseMessagingService, we also make sure to update the existing
stored token, while also informing the backend about this new token that
we now should be using for that member.

Also:
Use koin for WorkManager initialization
Cleanup deprecated registerPushToken mutation


app/src/main/kotlin/com/hedvig/app/service/push/RegisterPushTokenAuthEventListener.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.hedvig.android.apollo.OperationResult
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.auth.event.AuthEventListener
+import giraffe.NotificationRegisterDeviceMutation
+import slimber.log.e
+import slimber.log.i
+
+class RegisterPushTokenAuthEventListener(
+  private val apolloClient: ApolloClient,
```
hugokallstrom, line 16: Getting a circular dependency here because OkHttp depends on `AuthTokenRefreshingInterceptor`. Not sure how we can solve that without making a larger refactoring, do you have any ideas @StylianosGakis?

StylianosGakis: Gonna pick this up now and see what's up

PR #1518, opened 2023-04-13
CLX-467: Search claim error handling

hugokallstrom:
Screenshot_20230413_114146

### DIFF
--- a/feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
+++ b/feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
@@ -9,18 +9,22 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.CircularProgressIndicator
 import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
+import androidx.compose.material.TextButton
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.ArrowBack
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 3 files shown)


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -99,6 +108,22 @@ class SearchActivity : ComponentActivity() {
...
+      horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+      Text(text = message)
+      Spacer(modifier = Modifier.padding(top = 8.dp))
+      TextButton(onClick = onRetry) {
+        Text(text = stringResource(id = R.string.GENERAL_RETRY))
+      }
+    }
+  }
+
```
StylianosGakis, line 126: Hmm maybe we can try and use what we have in figma here https://www.figma.com/file/tpp00CvD8ALUKdjDRzyygv/Android%E2%80%A8-UI-Kit?node-id=1863-4046 which is this https://github.com/HedvigInsurance/android/blob/76ae326b892b07d1aa42c9d0c9b6b9718231eb8c/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/information/AppStateInformation.kt#L22-L38 and then you can put a LargeContainedButton under it with a 40dp spacing as that figma file describes.

Maybe we should have a component for this use case btw, right now we do have `GenericInfoScreen` which is exactly that, but it forces the button to be at the bottom of the screen, as that figma shows it should be if it's a full screen error item.

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/SearchActivity.kt
```diff
@@ -84,10 +88,15 @@ class SearchActivity : ComponentActivity() {
 
               Spacer(modifier = Modifier.padding(8.dp))
 
-              CommonClaims(
-                selectClaim = viewModel::onSelectClaim,
-                commonClaims = viewState.commonClaims,
-              )
+              val errorMessage = viewState.errorMessage
+              if (errorMessage != null) {
+                Error(errorMessage) { viewModel.loadSearchableClaims() }
```
StylianosGakis, line 93: The error message as far as I understand is more often than not something that the user would like to see. Not sure if this case in particular the error is localized properly and is relevant for a normal member to see by the backend, but if not, maybe we can do with a generic network error message as we often do instead.

StylianosGakis (approved):
Might wanna not show the error message just as we receive it.
And might want to use the AppStateInformation for the error UI to make it look more closely to how other similar cases look like.

But lgtm in terms of how it works.

PR #1519, opened 2023-04-14
CLX-537: Use graphql instead of http for getting claim entrypoints

hugokallstrom:


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/GetClaimEntryPointsUseCase.kt
```diff
@@ -1,12 +1,11 @@
...
 internal interface GetClaimEntryPointsUseCase {
-  suspend operator fun invoke(): CommonClaimsResult
+  suspend operator fun invoke(): Either<ErrorMessage, CommonClaimsResult>
 }
 
-internal sealed interface CommonClaimsResult {
-  data class Error(val message: String) : CommonClaimsResult
-  data class Success(val searchableClaims: List<SearchableClaim>) : CommonClaimsResult
-}
+data class CommonClaimsResult(val searchableClaims: List<SearchableClaim>)
```
StylianosGakis, line 11: Super nice 🔥

apollo/octopus/src/main/graphql/com/hedvig/android/apollo/octopus/graphql/claimflow/QueryEntryPointSearch.graphql
```diff
@@ -0,0 +1,6 @@
+query EntrypointSearch($input: EntrypointSearchInput) {
+  entrypointSearch(input: $input) {
+    displayName
+    id
+  }
+}
```
StylianosGakis, line 6: Since I see that the input we give this in the code is always:
```
EntrypointSearchInput(
  type = EntrypointType.CLAIM,
  limit = Optional.present(NR_OF_ENTRYPOINTS),
)
```
So we can change this so that in the graphql file itself we just be
```suggestion
query EntrypointSearch {
  entrypointSearch(input: { type: CLAIM, limit: 20 }) {
    id
    displayName
  }
}
```
And we don't need to give it any more input.

If we ever need to change the limit, or add the search string we can adjust accordingly. Wdyt?

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/GetClaimEntryPointsUseCase.kt
```diff
@@ -1,12 +1,11 @@
 package com.hedvig.android.odyssey.search
 
+import arrow.core.Either
+import com.hedvig.android.core.common.ErrorMessage
 import com.hedvig.android.odyssey.model.SearchableClaim
 
 internal interface GetClaimEntryPointsUseCase {
-  suspend operator fun invoke(): CommonClaimsResult
+  suspend operator fun invoke(): Either<ErrorMessage, CommonClaimsResult>
```
StylianosGakis, line 8: Hmm I know it was this way before too, but when using the `operator fun invoke` instead of just `fun invoke` we are bumping into this bug https://youtrack.jetbrains.com/issue/KTIJ-1050/Go-to-implementation-fails-on-an-override-invoke-function-call-in-both-operator-and-function-forms where trying to do Cmd + B (Or Cmd + click) on the function it does not find its usages when calling this function.
It's an annoying IDE bug which has been there since forever, but imo just getting rid of the `operator` is our best bet, since we can still call it by doing `useCase.invoke()` instead of `useCase()`, and it's consistent with the rest of the code base.

PR #1520, opened 2023-04-17
Bump to AGP 8.0.0

StylianosGakis:
buildconfig is now false by default.
Same for renderScript and aidl.

This allows us to use the new capabilities of Android Studio Flamingo which is the current latest stable and remove some workarounds we've had for a couple of little things.

Also, this requires for you to use JDK 17. There's this section which helps with that https://developer.android.com/build/releases/gradle-plugin#jdk-17-agp


hugokallstrom (approved):
Nice!

PR #1521, opened 2023-04-17
Bump androidx core

StylianosGakis:
Use the new bundle/intent functions which API 33 and above expects, while still using the old APIs on older versions

The new functions are inside IntentCompatExt and BundleCompatExt

context here:
https://issuetracker.google.com/issues/242048899


hugokallstrom (approved):
Cool!

PR #1523, opened 2023-04-18
Claim flow audio permission

StylianosGakis:
Now uses the proper flow of asking for the permission on first action
click, then asking again if we are allowed to give them the rationale,
and allow them to go directly to settings if we are not allowed to show
them a rationale and we've been denied at least once.


hugokallstrom (approved):
LGTM!

PR #1524, opened 2023-04-20
New cross sell API

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt
```diff
@@ -50,36 +47,37 @@ data class CrossSellData(
...
+      },
+      highlights = data.productVariants.first().highlights.map {
+        Highlight.from(it)
+      },
+      faq = data.productVariants.first().faq.map {
+        FAQItem.from(it)
+      },
+      insurableLimits = data.productVariants.first().insurableLimits.map {
+        InsurableLimitItem.InsurableLimit.from(it)
       },
```
StylianosGakis, line 81: If `productVariants` is an empty list (it can be from the GraphQL perspective, nothing preventing it), all the `first()` calls will crash the app.
Should we maybe just do a `firstOrNull()` call and let it be null if the list is empty? 
Also in general, why is `productVariants` a list, but before we were getting a single item? What was the shift that happened here, and do we maybe want to at any point not ignore all the other items in this `productVariants` list?

StylianosGakis (approved):
Just a couple of questions I may be missing some context on. And once concern about the empty list case.

Otherwise looking good 😊

app/src/main/kotlin/com/hedvig/app/ApplicationModule.kt
```diff
@@ -223,6 +224,7 @@ val applicationModule = module {
             .newBuilder()
             .header("User-Agent", makeUserAgent(languageService.getLocale()))
             .header("Accept-Language", languageService.getLocale().toLanguageTag())
+            .header("hedvig-language", languageService.getLocale().toLanguageTag())
```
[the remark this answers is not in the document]
hugokallstrom: This was needed for storefront I think, not sure why they dont use Accept-Language...

app/src/main/kotlin/com/hedvig/app/feature/crossselling/ui/CrossSellData.kt
hugokallstrom: Good catch, fixing

PR #1525, opened 2023-04-20
Show existing audio file in claim flow + claims flow fixes

StylianosGakis:
Add handling in the claim flow to handle when an audio file already exists when entering the audio recording step.

Move HedvigAudioComponent in a module, it's used in both :app for claim details and :odyssey for the audio replaying in the claim flow

This also makes a bunch of touches in the claim flow, like changing the color of a cursor which was wrong, some fixes with how ClaimFlowActivity was handling dark mode.


hugokallstrom (approved):
Really good work! 👍

PR #1528, opened 2023-04-24
Log notification senders + Permission check + customerio data

StylianosGakis:
Extract the data that comes from customerIO as explained here https://www.customer.io/docs/send-push/#standard-payload
Specifically, it takes the body and the title coming from there.

Consolidate all notification senders to one place so that if there is no notification permission we do not try to send


app/src/main/kotlin/com/hedvig/app/feature/chat/service/ChatNotificationSender.kt
```diff
@@ -170,14 +182,14 @@ class ChatNotificationSender(
     )
   }
 
-  private fun hedvigPerson() = Person.Builder()
-    .setName(context.getString(hedvig.resources.R.string.NOTIFICATION_CHAT_TITLE))
+  private val hedvigPerson: Person = Person.Builder()
+    .setName("Stelios")
```
hugokallstrom, line 186: Do we need to set a name?

PR #1530, opened 2023-04-26
Replace chat TextField with compose and fix single-line problem

StylianosGakis:
I actually gave up trying to fix the existing code when I saw the way that the click was handled with trying to use `onTouchEvent` and trying to find the touch bounds to construct a Rect of where the icon should be and if we are in bounds or not. We can do much better than that so I just made a tiny composable View replacing the previous `ChatTextInput` custom view, but now it's super straightforward to read and understand.

This solves two issues:
The text field can now expand appropriately vertically, to allow the user to see the entire text that they are writing. The clickable icon on the right side is now properly clickable with the correct touch size and a ripple to indicate when it's clicked. This also helps with the TextField text not to go too close to that icon.

As a side effect of this, make those items be 48x48 instead of 40x40, to
make them look the same size by default, and follow proper touch size
standards.
 
| small text | long text + clicked send icon |
| --- | --- |
| image | image |


hugokallstrom (approved):
GJ

PR #1531, opened 2023-04-27
Chat dialog on send error

StylianosGakis:
By the way, as I see from the code here, we did not clear the text itself even before my changes here (I didn't touch that part), if the network request failed. So that means that for the folks that were having issues, the network request seemed to work fine, and then the text was cleared. The problem was that even though the network request was OK, the message never arrived, which means it was for sure on the backend, and it also means that this change won't necessarily catch this issue again. Since from our perspective, sending the value worked fine.
Just some context on what I see here, but regardless, this is a good change.

* Add email opening function with hedvig support email prefilled

This does not open a bottom sheet to pick a mail, just opens it directly
 
* Remove unused FileUploadOutcome

Was only used to get a report that it's done, without using the result
in any way other than informing the UI that the operation is done. So
changed the model to reflect that fact.

Before, if it failed,
`attachPickerDialog?.uploadingTakenPicture(false)` would never be
called, and it'd be stuck on a loading state forever.
 
* Use safeExecute instead of runCatching

RunCatching is problematic as it catches even CancellationExceptions and
then reports them as slimber.e errors which is definitely not what we
want to do.
 
* Make the error dialog prompt to open an email app

The dialog now says that something went wrong and to check their
internet connection, and gives an option to contact us through opening
an email with the hedvig support email prefilled.

| When the error happens | After you click "Contact us" and you get the email prefilled |
| --- | --- |
| image | image |

### DIFF
--- a/app/src/main/kotlin/com/hedvig/app/feature/chat/FileUploadOutcome.kt
+++ b/app/src/main/kotlin/com/hedvig/app/feature/chat/FileUploadOutcome.kt
@@ -1,8 +1,0 @@
-package com.hedvig.app.feature.chat
-
-import android.net.Uri
-
-data class FileUploadOutcome(
-  val uri: Uri,
-  val wasSuccessful: Boolean,
-)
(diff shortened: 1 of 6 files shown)


hugokallstrom (approved):
GG!

PR #1532, opened 2023-04-28
Cleanup: Remove ODYSSEY SDUI

StylianosGakis:
Most of the code is about removing the dependencies we were using in our code, importing them with
```
import com.hedvig.odyssey.Foo
```
And moving them to our code. Most notable the AudioPlayer inside odyssey was coming from that library, now we simply use the AudioPlayer which we had in the claim details already, which takes in a URL and just plays the audio, even when coming from a local file like here.
Other cases like the KeepScreenOn composable which was copied inside Odyssey which was originally in our codebase, now just moved to core-ui so that both places that need it can use it directly.

Also removes all usages of the Odyssey feature flag too, now we're simply using the local/embark claim flow, nice and simple.

And this completely removes the odyssey library from the repo, we can also remove the CI building it in the Odyssey repo whenever we want.


hugokallstrom (approved):
🔥

PR #1550, opened 2023-05-02
Cleanup remove entire engineering directory plus all lint rules I could remove easily

StylianosGakis:
This was me doing a wiiiide sweep of things, in the nature of the spring cleaning week we had before.

This PR mainly removes all the dead code from inside the `engineering` directory, which I haven't used in well over a year now. And I assume you don't either.
While doing so, this cleaned up some other stuff, like some extra modules we had for HAnalytics to work with the engineering mode and so on.
This moves all the necessary parts we had in engineering, like the impersonation functionality into just the `debug` directory.

Also in this process I did a couple of version bumps.

And replaced the https://github.com/HedvigInsurance/gradle-unused-resources-action we were using in CI. Problem with that was that it as using this https://github.com/konifar/gradle-unused-resources-remover-plugin which is 5 years old, and as I was checking some lint errors we were getting, it wasn't picking up everything that it should be picking up. Replaced it with this https://github.com/irgaly/android-remove-unused-resources-plugin/tree/main which is actively maintained right now, and as I was checking to see what it did, it found more cases of unused resources, and no false positives, aka removing things that are in fact used in some odd way. It does so by using the output of the official lint check from AGP, which should always be the most trustworthy source of knowing what is used and what isn't.


hugokallstrom (approved):
Really good!!👌

PR #1552, opened 2023-05-02
Remove all unused permissions, most importantly, the camera one

StylianosGakis:
This was part of what we wanted to do when we used the contract for fetching a file from the phone without using the permission. This was a tiny bit more involved, but still the same concept applies. No longer have to ask for a permission to see their camera, since we are not accessing anything, just letting their own camera app do the work, and we only get to see the resulting picture if they want to send it to us, we are not allowing ourselves to see anything more than that.

Use the camera activity result contracts instead to fetch a picture.
Rewrite a bit of the logic of where it's saved as a consequence of this change.
All this is explained in a comment inside provider_paths.xml

This change also fixes compared to before that in case we start the picture taking, while we're there our application is killed by the system, and then the user picks a picture and they come back, this flow still works, as we are saving the URI string through process death.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ run `./gradlew ktlintFormat` to make ktlint to format all files according to the
 
 ![Module graph](misc/images/modularization-graph.png "Image showing the module dependencies graph")
 
-Generated from `./gradlew :generateProjectDependencyGraph`*\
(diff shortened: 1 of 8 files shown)


hugokallstrom (approved):
💯

PR #1553, opened 2023-05-02
Add module and repository used for changing address

hugokallstrom:


app/src/main/kotlin/com/hedvig/app/feature/di/KoinInitializer.kt
```diff
@@ -129,6 +129,7 @@ class KoinInitializer : Initializer<KoinApplication> {
         viewModelModule,
         whatsNewModule,
         workManagerModule,
+        changeAddressModule,
```
StylianosGakis, line 132: These are alphabetically sorted, if you don't mind doing that too.
I typically highlight everything, do (shift + shift) which brings up a window and type "sort lines" in there. Then only pressing enter does it for you on the highlighted text.

feature-changeaddress/src/main/kotlin/com/example/feature/changeaddress/data/ChangeAddressRepository.kt
```diff
@@ -0,0 +1,153 @@
...
+        .moveIntentCreate
+
+      val moveIntent = result.moveIntent
+      val userError = result.userError
+
+      when {
+        userError != null -> shift(ErrorMessage(userError.message))
+        moveIntent != null -> moveIntent.toMoveIntent()
+        else -> shift(ErrorMessage("No data found in MoveIntent"))
+      }
```
StylianosGakis, line 43: Yes perfect!

StylianosGakis: A question, since the backend returns these two fields separately, there isn't any guarantee in the type at least that only one of the two is going to be non-null, or that neither of them are going to be non-null. Despite that, have they communicated such a guarantee?

For example in this implementation, if the response contains both an error and a moveIntent response, we are going to be using only the error. Or should that be an impossible return value from them anyway?

settings.gradle.kts
```diff
@@ -71,3 +71,4 @@ include(":notification-badge-data")
 include(":notification:notification-core")
 include(":notification:firebase")
 include(":testdata")
+include(":feature-changeaddress")
```
StylianosGakis, line 74: Consider alphabetically sorting this too

feature-changeaddress/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+  id("hedvig.android.library.compose")
+  @Suppress("DSL_SCOPE_VIOLATION")
+  alias(libs.plugins.serialization)
+}
+
+android {
+  namespace = "com.hedvig.android.feature.changeaddress"
+}
+
+dependencies {
```
StylianosGakis, line 13: I see that you added a lot of the things that this module may need but doesn't yet. This is fine, and maybe we can keep it like this while we're working on it, and strip away the ones we're not interested in after this is more or less done.

This can be done by running
`./gradlew :feature-changeaddress:projectHealth` and there's a gradle plugin which will help figure out which dependencies this module doesn't actually need.

StylianosGakis (approved):
Super nicely done, I like all of this!

feature-changeaddress/src/main/kotlin/com/example/feature/changeaddress/data/ChangeAddressRepository.kt
hugokallstrom: We just have to assume that both will be non-null. The purchase journey team talked about modelling this with return types so that the schema got more expressive. It got too complex in the end though and they ended up with this design instead. Ill check if the moveIntent is null first though, that makes more sense.

feature-changeaddress/build.gradle.kts
hugokallstrom: Oh that is very nice! We can do that later though, I think we will use more of these dependencies during development

feature-changeaddress/src/main/kotlin/com/example/feature/changeaddress/data/ChangeAddressRepository.kt
StylianosGakis: Alright, let's go with that then yeah

PR #1559, opened 2023-05-03
Bump to latest material3 dependency

StylianosGakis:
This now removes the workaround we were doing on DatePicker which didn't
 fit out needs.
Also removes calling some internal code which is now properly public, to
 set the date manually
And now we get a proper DatePickerDialog from the library so we don't have to make our own Dialog to wrap it.


hugokallstrom (approved):
👌

PR #1561, opened 2023-05-03
Add screen and state for selecting apartment owner type

hugokallstrom:


feature-changeaddress/src/main/kotlin/com/example/feature/changeaddress/navigation/ChangeAddressGraph.kt
```diff
@@ -37,6 +37,18 @@ internal fun NavGraphBuilder.changeAddressGraph(
...
           navController.navigate(ChangeAddressDestination.MoveQuotes(quotes))
         },
+        onClickHousingType = {
+          navController.navigate(ChangeAddressDestination.SelectHousingType)
+        },
+      )
+    }
+
+    animatedComposable<ChangeAddressDestination.SelectHousingType> {
+      val viewModel: ChangeAddressViewModel = koinViewModel()
```
StylianosGakis: Hmm okay about this one we gotta see how we do this instead. I now also understand why that UIState object had a lot of fields and all of them were null to begin with. Because we're planning to have all state in one shared ViewModel.

So in the other flows, each step had its own ViewModel, which only needed to know about that one step, and not about everything, relying on the backend to hold the more wide-level state instead.

Right now, this `= koinViewModel<ChangeAddressViewModel>()` in multiple destinations, won't get you the same ViewModel, but you'd get separate instances of the ViewModel which isn't what we'd want in our case. To fix that there's a way to scope the ViewModel to the navigation graph instead of the specific destination (Scope to the animatedNavigation<Destinations.ChangeAddress> basically in this case) so we may want to take that approach

We need to see if we want to also not go down that path and keep all screens more separate from each other, and hold this more wide state somewhere else instead, even in some persistent storage, like DataStore, and we can manually clear it whenever we finish this moving flow, or start over and so on.

app/src/main/kotlin/com/hedvig/app/feature/di/KoinInitializer.kt
```diff
@@ -15,7 +14,6 @@ class KoinInitializer : Initializer<KoinApplication> {
   override fun create(context: Context): KoinApplication = startKoin {
     androidLogger(Level.ERROR)
     androidContext(context.applicationContext)
-    workManagerFactory()
```
StylianosGakis: Was leftover from the previous PR I suppose, since you addressed this there.

feature-changeaddress/src/main/kotlin/com/example/feature/changeaddress/navigation/ChangeAddressGraph.kt
StylianosGakis: Reading the docs, this section shows how to use this API https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-apis#vm-api-navgraph and since we're using Koin, our equivalent would be something like this
```kotlin
val parentEntry = remember(navController, navBackStackEntry) {
  navController.getBackStackEntry(createRoutePattern<Destinations.ChangeAddress>())
}
val viewModel: ChangeAddressViewModel = koinViewModel(viewModelStoreOwner = parentEntry)
```

StylianosGakis (approved):
I think I see a bunch of code here that has already been merged in the previously approved PRs, not sure if I am somehow misunderstanding something though 🤔

But it's looking good, just have one comment about sharing this viewmodel across all the screens.

app/src/main/kotlin/com/hedvig/app/feature/di/KoinInitializer.kt
StylianosGakis: Ok yeah after your rebase fix these are now gone 👍

PR #1567, opened 2023-05-05
Show error dialog for network errors

hugokallstrom:


feature-changeaddress/src/main/kotlin/com/feature/changeaddress/ChangeAddressViewModel.kt
```diff
@@ -127,6 +127,10 @@ internal class ChangeAddressViewModel(
     }
   }
 
+  fun onErrorDialogDismissed() {
+    _uiState.update { it.copy(errorMessage = null) }
+  }
```
StylianosGakis, line 132: So this is going to be for a wide error for the entire screen, which will be there along with the errors per-field which can also have an error right?

StylianosGakis (approved):

feature-changeaddress/src/main/kotlin/com/feature/changeaddress/ChangeAddressViewModel.kt
hugokallstrom: Correct!

PR #1568, opened 2023-05-05
Add overview and open chat button to offer destination

hugokallstrom:


StylianosGakis (approved):
Is the pillow image as a placeholder there? Will those live in the app and not be fetched from a url instead? What's the plan there?

PR #1569, opened 2023-05-05
Fix proguard rules to new okhttp release

StylianosGakis:
OkHttp 4.11.0 now has all these rules built-in
https://github.com/square/okhttp/compare/parent-4.10.0...parent-4.11.0#diff-2c8b9c303079dd83e7937b68b4074e19bbdcb65a72ed8fb43b4057ab6d8629acR10-R14 We now only need the ones for Retrofit which aren't merged into the main library yet.


hugokallstrom (approved):
Good find!

PR #1570, opened 2023-05-05
Add the shape and the colors for now

StylianosGakis:
Let's merge this in just to have them available for everyone for future PRs

Note that a big chunk of the diff is me running a ktlintformat and applying quick fixes by the IDE


core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/newtheme/Shape.kt
```diff
@@ -0,0 +1,112 @@
...
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import java.lang.Float.min
+
+// 27 magic number, matches the figma file close enough
+private val squircleRadius = 27.dp
+
+val SquircleShape = object : Shape {
```
hugokallstrom, line 30: Great work :D

PR #1574, opened 2023-05-08
Format premium in offer screen

hugokallstrom:


feature-changeaddress/src/main/kotlin/com/feature/changeaddress/data/MoveQuote.kt
```diff
@@ -1,14 +1,37 @@
...
+  val premium: Premium,
   val startDate: LocalDate,
   val termsVersion: String,
 )
+
+@Serializable
+data class Premium(
+  val amount: Double,
+  val currencyCode: CurrencyCode,
+)
```
StylianosGakis, line 22: Maybe this https://github.com/HedvigInsurance/android/blob/487cd2f72b39befde541a378a8115da75c883dc9/feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/navigation/ClaimFlowDestination.kt#L207-L219 can be used instead here? And have it be moved to core-ui instead of inside the odyssey module.

And the formatting logic can be added to the `toString()` over there, so we use the same way of displaying text across the app.

In there, that class can look like this
```kotlin
@Immutable
@Serializable
internal data class UiMoney(val amount: Double, val currencyCode: CurrencyCode) {
  @kotlinx.serialization.Transient
  val decimalFormatter: DecimalFormat = DecimalFormat("0.#")
  
  override fun toString(): String {
    return buildString {
      append(decimalFormatter.format(amount))
      append(" ")
      if (currencyCode == CurrencyCode.UNKNOWN__) {
        error("Unknown currency code")
      }
      append(currencyCode)
    }
  }

  companion object {
    fun fromMoneyFragment(fragment: MoneyFragment): UiMoney {
      return UiMoney(fragment.amount, fragment.currencyCode)
    }
  }
}
```

Two questions here, I see in your impl here, you don't print SEK for swedish crowns, but "kr" instead, but you don't change this for the other currencies, is this just some standard when handling swedish crowns? If not maybe we can just use the automatic toString() which would print "SEK" instead.

For the case of the unknown type, if we want to crash maybe that can be handled too like
```
return buildString {
  append(decimalFormatter.format(amount))
  append(" ")
  if (currencyCode == CurrencyCode.UNKNOWN__) {
    error("Unknown currency code")
  }
  append(currencyCode)
}
```

And the second question is in the impl velow you add a " " and then another one before the currency text. Was this an accidental duplication or do we want two spaces between the number and the text?

StylianosGakis (changes requested):
I think it'd be nice to do this in one place, so maybe we can merge the nice formatting of this Premium class, and the fact that we have the UiMoney in the odyssey module and make it 1 thing, inside core-ui which can be used in both places.

StylianosGakis (approved):
Yes, perfect!

PR #1578, opened 2023-05-09
New UI kit textfield + preliminary work on the new colors etc.

StylianosGakis:
Add HedvigTextField

It's 90% what the m3 TextField does, but has been copied over to let us
customize it. Particularly so that we are able to edit a bunch of the
internal code that m3 had.

In our case, so far we are customizing:
* The shape of it
* The underline it had below it which we've removed, and instead have it
 look like a border around it when the text field has focus or it has an
 error.
* The supporting text paddings that were too big for our design
* The supporting text appearance now animating in vertically.

with signal 

https://github.com/HedvigInsurance/android/assets/44558292/ace370c8-edb7-4a9f-b670-1517c0b18800 

before signal + disabled

https://github.com/HedvigInsurance/android/assets/44558292/2e2e3120-ce93-4ef9-bf89-6252957b1477


hugokallstrom (approved):
Absolutely brilliant

PR #1581, opened 2023-05-10
Refactor search classes to instead fetch claim groups

hugokallstrom:
Screenshot_20230510_153423


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/groups/GetNetworkClaimEntryPointGroupsUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.core.common.ErrorMessage
+import octopus.EntrypointGroupsQuery
+
+internal class GetNetworkClaimEntryPointGroupsUseCase(
+    private val apolloClient: ApolloClient,
+) {
+
+  suspend fun invoke(): Either<ErrorMessage, GroupedClaimsResult> {
```
StylianosGakis, line 15: Return type could even be `Either<ErrorMessage, List<ClaimGroup>` directly and could remove GroupedClaimsResult if we don't use GroupedClaimsResult in any other significant way, but no hard opinions here.

StylianosGakis (approved):
Would just be nice to remove that one null ImageLoader along with the `!!`, gave a suggestion there too.
Lgtm otherwise 😊

PR #1582, opened 2023-05-10
Query first name and show in claim group screen

hugokallstrom:


StylianosGakis (approved):
I see how the GroupedClaimsResult now has a use since it has more than 1 field, great work!

PR #1585, opened 2023-05-11
Show claim group after selecting group id

hugokallstrom:
Screenshot_20230511_112108


feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/group/ClaimGroupViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+
+internal class ClaimGroupViewModel(
+  private val getClaimEntryPoints: GetClaimEntryGroupUseCase,
+  private val groupId: String,
+) : ViewModel() {
+  private val _uiState = MutableStateFlow(ClaimGroupUiState())
+  val uiState = _uiState
+
+  init {
+    loadClaimGroup()
```
StylianosGakis, line 18: Doesn't have any way to show error and allow retrying right now right, something we wanna add later?

StylianosGakis (approved):

feature-odyssey/src/main/kotlin/com/hedvig/android/odyssey/search/group/ClaimGroupViewModel.kt
hugokallstrom: Good point, ill add it

PR #1586, opened 2023-05-12
Show multiple quotes for move offer

hugokallstrom:


feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/ChangeAddressOfferDestination.kt
```diff
@@ -89,194 +77,57 @@ internal fun ChangeAddressOfferDestination(
...
+              vertical = 8.dp,
+            )
+            .clip(SquircleShape)
+            .clickable { openChat() }
+            .background(shape = SquircleShape, color = MaterialTheme.colorScheme.primary)
+            .padding(
+              horizontal = 16.dp,
+              vertical = 8.dp,
+            ),
         )
```
StylianosGakis, line 125: Might wanna use `androidx.compose.material3.TextButton` if we want a text which is also clickable? That way you get the slot for the shape, automatically get the ripple to follow that shape, and other guarantees like the touch size being big enough and so on. Would it fit in this case?

feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/QuoteCard.kt
```diff
@@ -0,0 +1,210 @@
...
+  isExpanded: Boolean,
+) {
+  Column(
+    modifier = Modifier
+      .padding(horizontal = 16.dp)
+      .clip(SquircleShape)
+      .clickable { onExpandClicked() }
+      .background(shape = SquircleShape, color = Color(0xFFF0F0F0))
+      .padding(16.dp),
+  ) {
```
StylianosGakis, line 55: Since this is a card, it *may* be a better fit to have the top level composable be a HedvigCard, with the right clickable lambda and so on, which will also follow our theme well enough to not have to do any custom colors, and then have the column inside of it, without having to clip it or anything like that

feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/QuoteCard.kt
```diff
@@ -0,0 +1,210 @@
...
+            tint = Color(0xFFB4B4B4),
+          )
+        } else {
+          Icon(
+            painter = painterResource(id = com.hedvig.android.core.designsystem.R.drawable.ic_drop_down_indicator),
+            contentDescription = "",
+            tint = Color(0xFFB4B4B4),
+          )
+        }
+      }
```
StylianosGakis, line 124: If you're feeling like adding more animation to this, you can do something like this
```suggestion
var angle = animateFloatAsState(
    targetValue = if (isExpanded) { 0f } else { 180f },
    // animationSpec = spring(), // Might wanna play with the animation spec if the default spring feels odd
  )
  Icon(
    painter = painterResource(id = com.hedvig.android.core.designsystem.R.drawable.ic_drop_down_indicator),
    contentDescription = "",
    tint = Color(0xFFB4B4B4),
    modifier = Modifier.graphicsLayer {
      rotationZ = angle
    }
  )
```
So that you don't need to add the new icon resource at all, and you get the original icon to rotate clockwise to the new position.
If it rotates on the wrong side, play around with the initial and target values, you may wanna start at 180 degrees and go to 360 or something like that, not sure, I didn't run this but it should work.

feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/QuoteCard.kt
```diff
@@ -0,0 +1,210 @@
...
+        style = MaterialTheme.typography.titleMedium,
+        fontSize = 18.sp,
+      )
+    }
+
+    AnimatedVisibility(
+      visible = isExpanded,
+      enter = expandVertically(),
+      exit = shrinkVertically(),
+    ) {
```
StylianosGakis, line 140: You *may* wanna do something like this https://github.com/HedvigInsurance/android/blob/eafa3492a8c9c2ce1bfb1971ef95d72a4000226c/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitem/SingleItemDestination.kt#L129-L130 here, to also have the thing fade in as it's coming in, so it comes into the view more naturally, instead of immeditelly look like it's being clipped above.
The `clip = false` is maybe something you shouldn't bring over here, as it will most likely look odd as they come from above, but mostly taking about the `fadein() + expandVertically()` and the same for exit.

StylianosGakis (approved):
Great work! 😊
Just some suggestions I had to make this even crispier with some animations and so on.

feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/QuoteCard.kt
hugokallstrom: Super nice!

hugokallstrom:
Great comments, thanks!!

PR #1587, opened 2023-05-12
New UI kit in ChangeAddressSelectHousingTypeDestination

StylianosGakis:
Add the "Info" colors, the blue ones basically for the info cards.
Add the HedvigTheme overload which allows flipping the surface and background color so that it follow the new UI kit reccomendation.
Add a HedvigScaffold to assist with making typical flow content without having to repeat the inset handling along with the top app bar hooking up with the scrollable column.
Fix AddressInfoCard to use the colors from the theme, also fixing the information icon to have the right blue tint.

And finally apply all this to the ChangeAddressSelectHousingTypeDestination screen.

image

image


hugokallstrom (approved):
💯

PR #1588, opened 2023-05-12
Update change address new destination UI

StylianosGakis:
Same treatment as the ChangeAddressSelectHousingTypeDestination. Fix the scrolling, and add our theme everywhere where applicable

As seen in the demo video too.
image


hugokallstrom (approved):
Beautiful

PR #1590, opened 2023-05-16
Show "Generate travel certificate" in home tab

hugokallstrom:
Screenshot_20230516_132201


app/src/main/kotlin/com/hedvig/app/feature/home/ui/HomeViewModel.kt
```diff
@@ -82,14 +87,26 @@ class HomeViewModelImpl(
...
   private suspend fun createViewState(forceReload: Boolean) {
     _viewState.value = ViewState.Loading
-    _viewState.value = when (val result = getHomeUseCase.invoke(forceReload)) {
-      is Either.Left -> ViewState.Error(result.value.message)
-      is Either.Right -> ViewState.Success(
-        homeData = result.value,
+    either {
+      val homeData = getHomeUseCase.invoke(forceReload).bind()
+      val travelCertificateData = if (featureManager.isFeatureEnabled(Feature.TRAVEL_CERTIFICATE)) {
+        getTravelCertificateUseCase.invoke().bind()
```
StylianosGakis, line 93: Good job with the use of `either` here, I like it 😊

One thing we may or may not want, right now if the travel certificate query would fail, the entire home ui state would result in a failure showing nothing.
We could also consider if the travel certificate query fails, that we just make this null, and render the rest of the home screen just without the travel certificate.
Both approaches make sense to me, it really depends on how much we want that certificate to be there. What are your thoughts on it?

If we do want to go that way, the code would be something like changing the `bind()` with `.getOrNull()` and it automatically turns a Left into null.

StylianosGakis (approved):
Great job! Had a question, could go either way, depends what we prefer really, more in that comment.

app/src/main/kotlin/com/hedvig/app/feature/home/ui/HomeViewModel.kt
hugokallstrom: Good point, I think we can do getOrNull and log the error instead to avoid not showing anything at all 👍

PR #1591, opened 2023-05-16
Rest of moving flow

StylianosGakis:
Room for improvement:
* Since there's a lot of content which just isn't there yet, and isn't coming from backend, it was added as sample data just so we can get the UI working at least.
* Also a bunch of localization stuff.

From the preview (It's a bit blurry as it's quite long) 
phone


core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3ColorScheme.kt
```diff
@@ -115,17 +134,39 @@ val ColorScheme.onTypeElement: Color
...
 
-val ColorScheme.onTypeSignalContainer: Color
+val ColorScheme.onTypeContainer: Color
   @ReadOnlyComposable
   @Composable
   get() = LocalHedvigMaterial3ColorScheme.current.onTypeContainer
 // endregion
 
+// region info colors
+val ColorScheme.infoElement: Color
```
hugokallstrom, line 149: Nice that you can define own colors :)

core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/theme/HedvigTheme.kt
```diff
@@ -37,3 +37,28 @@ fun HedvigTheme(
     }
   }
 }
+
+@Composable
+fun HedvigTheme(
+  darkTheme: Boolean = isSystemInDarkTheme(),
+  flipBackgroundAndSurface: Boolean,
```
hugokallstrom, line 44: Could we rename this to something like `useNewColorScheme`?

feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/offer/QuoteCard.kt
```diff
@@ -7,209 +7,196 @@ import androidx.compose.animation.fadeIn
...
-            Spacer(modifier = Modifier.padding(start = 5.dp))
-            Icon(
-              painter = painterResource(id = com.hedvig.android.core.designsystem.R.drawable.ic_info),
-              tint = Color(0xFF727272),
-              contentDescription = "",
-            )
+    modifier = modifier,
+  ) {
+    Column(Modifier.padding(start = 16.dp, top = 16.dp, end = 16.dp)) {
+      PillAndBasicInfo(movingDate)
```
hugokallstrom, line 53: Much better 🙌

hugokallstrom (approved):
Looks great!

core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/theme/HedvigTheme.kt
StylianosGakis: Yes absolutely!

core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3ColorScheme.kt
StylianosGakis: Yep, and can use them like `MaterialTheme.colorScheme.infoElement` which makes it feel like it was part of the material theme all along.

PR #1593, opened 2023-05-17
Cleanup, move packages inside a `app` directory. Also move core and feature modules inside their respective directory too

StylianosGakis:
Basically what the title says. 99.99% of the code changes are just moving files around and fixing the gradle strings to hook things together according to their new structure. And the rest 0.01% a ktlintFormat run after everything was done.

image

image


hugokallstrom (approved):
💯

PR #1594, opened 2023-05-17
Use gradle-build-action everywhere

StylianosGakis:
Should be better for caching the right amount of stuff for gradle to run
 faster.
Make sure to only write to the cache from the develop branch since that's our main branch where things end up in.

More details in:
https://github.com/gradle/gradle-build-action


hugokallstrom (approved):
💯

PR #1595, opened 2023-05-19
Move profile screen to compose

StylianosGakis:
Also had to move about app screen as they were sharing some UI


hugokallstrom (approved):
💯

PR #1598, opened 2023-05-22
Remove ReferralsAdapter and use compose for the fragment

StylianosGakis:
Still wraps some xml views to avoid having to recreate some more complex logic in compose. Can always revisit if we want to.


hugokallstrom (approved):
💯

PR #1599, opened 2023-05-22
Compose in insurances

StylianosGakis:
Same situation as for compose in forever screen. Still wraps some xml views to avoid having to recreate some more complex logic in compose.
In this process, remove some unused code from InsuranceAdapter. Can't
remove it yet since it's used elsewhere.
Remove the shared element transition in preparation for more compose
which doesn't really support that yet anyway.


hugokallstrom (approved):
Byebye xml! 👍

PR #1600, opened 2023-05-22
Add composables for travel certificate input and add co-insured views

hugokallstrom:
Screenshot_20230522_125655

### DIFF
--- a/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
+++ b/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
@@ -31,7 +32,8 @@ fun AddCoInsured(
   uiState: TravelCertificateUiState,
   onSsnChanged: (String) -> Unit,
   onNameChanged: (String) -> Unit,
-  onSave: () -> Unit,
+  onRemoveClicked: (String) -> Unit,
+  onSaveClicked: () -> Unit,
 ) {
 
   val selectedCoInsured = uiState.coInsured.input.firstOrNull { it.id == selectedCoInsuredId }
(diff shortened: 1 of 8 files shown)


app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
```diff
@@ -0,0 +1,130 @@
...
+  onSsnChanged: (String) -> Unit,
+  onNameChanged: (String) -> Unit,
+  onRemoveClicked: (String) -> Unit,
+  onSaveClicked: () -> Unit,
+) {
+
+  val selectedCoInsured = uiState.coInsured.input.firstOrNull { it.id == selectedCoInsuredId }
+
+  HedvigScaffold(
+    navigateUp = {},
```
StylianosGakis, line 42: Might wanna add smth here. Now pressing the arrow top left will do nothing.

app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
```diff
@@ -0,0 +1,130 @@
...
+      text = "Add co-insured",
+      style = MaterialTheme.typography.headlineSmall,
+      textAlign = TextAlign.Center,
+      modifier = Modifier.fillMaxWidth(),
+    )
+    Spacer(modifier = Modifier.height(64.dp))
+    HedvigTextField(
+      value = selectedCoInsured?.name ?: "",
+      onValueChange = { onNameChanged(it) },
+      errorText = "",
```
StylianosGakis, line 56: If error text is null, that entire section won't render, but if it's an empty string, the text that would show below the text field will be rendered, and it will add some weird spacing in this case, but simply won't show anything. This is handled here https://github.com/HedvigInsurance/android/blob/93966a0194e063771c25701f721ea26c42ab71b3/app/core/design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/textfield/HedvigTextField.kt#L106-L121

If we have no error, just pass null instead so that this section isn't rendered in the first place. Same for the other HedvigTextFields in this screen.

StylianosGakis (approved):
👌

app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
hugokallstrom: Fixed in a PR on top of this

app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
hugokallstrom: Fixed in a PR on top of this

PR #1604, opened 2023-05-22
Travel certificate viewmodel

hugokallstrom:


app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/AddCoInsured.kt
```diff
@@ -18,26 +22,23 @@ import com.hedvig.android.core.designsystem.component.textfield.HedvigTextField
...
-  onRemoveClicked: (String) -> Unit,
-  onSaveClicked: () -> Unit,
+  onRemoveCoInsured: (String) -> Unit,
+  onEditCoInsured: (CoInsured) -> Unit,
+  onAddCoInsured: (CoInsured) -> Unit,
 ) {
 
-  val selectedCoInsured = uiState.coInsured.input.firstOrNull { it.id == selectedCoInsuredId }
+  var name by remember { mutableStateOf(coInsured?.name ?: "") }
+  var ssn by remember { mutableStateOf(coInsured?.ssn ?: "") }
```
StylianosGakis, line 41: Since we're keeping this state here inside the composable itself and not the VM, it's a good idea to do a `rememberSaveable` instead of `remember` so it still survives configuration changes like it would if it were in the VM.

StylianosGakis (approved):

PR #1607, opened 2023-05-23
Compose in home screen

StylianosGakis:
This is a bigger one naturally as it's the screen with the more things in it.

Moved over each item, as none is too complex, like the pie in forever tab

The one thing I had to change in how the items are built was to bunch all the common claims together, so that they can be rendered together in a grid layout.

Then just removed the xml things which we do not need. Some of them were used in other screens too so they had to stay.

Next steps on this is to also cleanup the shared code inside LoggedInActivity which all these fragments are sharing, specifically stuff like the bottom nav, the toolbar and so on.
Also to break the code up a little bit more, since now the HomeFragment has a big chunk of code for all the composables in a big when statement, I'll just make them individual composable functions instead.
Will not merge to develop until that's done, since now the toolbar feels a bit awkward before I fix that too.

There's also a small change in how we construct Coil, so that we share the OkHttpBuilder with that which is passed to ApolloClient too. Should help with logging when that goes bad too. But removed the interceptors that add authentication, that's the only change. I did this since I had some issues with some icons not loading and Coil not surfacing those errors. This way we get the normal OkHttp logs just as we do for normal requests too.

Nevermind the background breaking in these pictures, it's because of the rolling screenshot taking separate screensots and snipping them together. Doesn't look like that in the app 😅

| dark | light |
| --- | --- |
| image | image |


hugokallstrom (approved):
Nice!

PR #1608, opened 2023-05-24
Create travel certificate

hugokallstrom:


app/apollo/octopus/build.gradle.kts
```diff
@@ -35,5 +35,6 @@ apollo {
     generateOptionalOperationVariables.set(false)
 
     mapScalar("Date", "kotlinx.datetime.LocalDate", "com.apollographql.apollo3.adapter.KotlinxLocalDateAdapter")
+    mapScalarToKotlinString("Url")
```
StylianosGakis, line 38: Are Url and URL considered as separate scalars then? https://github.com/HedvigInsurance/android/blob/8d5b1a9cda95a25db219c231d28442a030c71a7b/app/apollo/giraffe/build.gradle.kts#L43
One is provided by Giraffe and the other from the supergraph I suppose

StylianosGakis (approved):

app/apollo/octopus/build.gradle.kts
hugokallstrom: Correct!

PR #1613, opened 2023-05-26
Open overview screen after creating travel certificate

hugokallstrom:
Download certificate when pressing button in overview.
Extract info cards.

### DIFF
--- a/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
+++ b/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
@@ -1,0 +1,60 @@
+package com.hedvig.android.feature.travelcertificate.data
+
+import android.os.Environment
+import arrow.core.Either
+import arrow.core.raise.either
+import com.apollographql.apollo3.ApolloClient
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.core.common.ErrorMessage
+import com.hedvig.android.core.common.await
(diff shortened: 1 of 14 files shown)


app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
```diff
@@ -0,0 +1,60 @@
...
+          .url(travelCertificateUri.uri)
+          .build()
+
+        val downloadedFile = File(
+          Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS),
+          CERTIFICATE_NAME,
+        )
+
+        try {
+          val response = OkHttpClient().newCall(request).await()
```
StylianosGakis, line 44: Could we get the existing OkHttpClient instance which we have in our koin graph? So that we don't need to create a new client on each invocation of this function.

StylianosGakis (approved):
🔥

app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
hugokallstrom: I got some errors related to headers when I used that. This URL probably expects only "default" headers.

StylianosGakis: Right, it probably didn't like that it was getting some extra Authorization header with a token that it did not recognize. Let's leave it for now then

PR #1614, opened 2023-05-26
Share downloaded certificate

hugokallstrom:


app/feature/travel-certificate/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,14 @@
...
+            android:name="com.hedvig.android.feature.travelcertificate.data.MyFileProvider"
+            android:authorities="com.hedvig.fileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/file_paths" />
+        </provider>
+    </application>
+</manifest>
```
StylianosGakis, line 14: We did have this https://github.com/HedvigInsurance/android/blob/67d1a4d8ce67926029372071d51e673470c65f3b/app/app/src/main/AndroidManifest.xml#L285-L293 for working with the camera pictures we get and store remporarily before uploading for the chat feature.
Could we maybe reuse that same one, or should we reuse that one so that we don't have two of them? Not sure about it.

app/feature/travel-certificate/src/main/res/xml/file_paths.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths>
```
StylianosGakis, line 2: Same comment as the one above, we currently have this https://github.com/HedvigInsurance/android/blob/41758c154b9483826864a5f67f0f301caa053054/app/app/src/main/res/xml/provider_paths.xml#L25-L27. Again, not sure if we can reuse that (even if it means adding this new path as an entry there) so we don't have multiple ways to do this and confuse ourselves.

StylianosGakis (approved):

app/feature/travel-certificate/src/main/res/xml/file_paths.xml
hugokallstrom: Cool, Ill try to see if we can use this instead

hugokallstrom: Didnt quite get this to work since we needed to then add a dependecy on the app module from the travel cert module... I think its fine to have these separated by feature though. Even though it makes for a bit more and duplicated code.

StylianosGakis: Oh yeah absolutely. Optimally it'd be that the chat feature was also in a separate module, and then this could be shared somehow or something like that. But yeah let's go ahead like this now, totally forgot about that here.

PR #1615, opened 2023-05-26
Remove toolbar from activity

StylianosGakis:
Remove the toolbar from the activity which was shared across all screens.
Each fragment can use their own toolbar with their own implementation.
With this, cleanup a bunch of now unused menu xml files
Make Referrals fragment fetch its own data on what to show when clicking
 on the toolbar item
Make pull to refresh not show behind the system bars

Also remove the [viewtooltip](https://github.com/florent37/ViewTooltip) dependency which was  archived early last year, and implement the chat tooltip in compose.

Pic shows both the loading and the tooltip just to show them for this PR
image


app/app/src/main/kotlin/com/hedvig/app/feature/home/ui/HomeFragment.kt
```diff
@@ -231,6 +245,28 @@ class HomeFragment : Fragment() {
...
                 Spacer(Modifier.windowInsetsBottomHeight(WindowInsets.safeDrawing))
               }
+              Column {
+                TopAppBarWithActions {
+                  ToolbarChatIcon(
+                    onClick = { requireContext().startChat() },
+                  )
+                }
+                val shouldShowTooltip by produceState(false) {
+                  val currentEpochDay = LocalDate.now().toEpochDay()
```
hugokallstrom, line 255: Could extract this logic to its own function

hugokallstrom (approved):
GJ! Looking at the screenshot the tooltip text is repeated, is the text coming from lokalise like that?

StylianosGakis:
> GJ! Looking at the screenshot the tooltip text is repeated, is the text coming from lokalise like that?

Yeah sorry about that, the screenshot was just taken what I was trying to add more text to the tooltip to show how it'd react to more text showing there. It only shows the text once in reality.

app/app/src/main/kotlin/com/hedvig/app/feature/home/ui/HomeFragment.kt
StylianosGakis: Yeap, will do in another PR

PR #1618, opened 2023-05-26
Cleanup, remove trigger free form chat mutation

StylianosGakis:
Just open the chat directly without doing a round trip over the network for a mutation which does nothing anymore.


hugokallstrom (approved):
Great! 🧹

PR #1621, opened 2023-05-29
Only show first name on added coinsured

hugokallstrom:


app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/TravelCertificateInputState.kt
```diff
@@ -63,4 +63,6 @@ data class CoInsured(
   val id: String,
   val name: String,
   val ssn: String,
-)
+) {
+  fun firstNameOrFullName(): String = name.split(" ").firstOrNull() ?: name
```
StylianosGakis, line 67: What is the expected behavior for this when we say "orFullName"? When do we want to show the full name?
The current impl will always find something to print, since doing `split(" ")` will simply return a list with one item. It being the string itself if it does not find anything matching this predicate, which is an space in this case. So the `?: name` part will simply never be called.

Check this out https://pl.kotl.in/c7hDMOKrl to see what I mean. Is this what we're going for here?

StylianosGakis (approved):
lgtm, just didn't understand what we want from that one function, if you don't mind explaining to me.

app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/TravelCertificateInputState.kt
hugokallstrom: Good catch. Fixed

PR #1624, opened 2023-05-30
Compose logged in activity, and removal of top level fragments

StylianosGakis:
This removes all previous fragments which were already migrated to compose. This PR just makes it so that we don't go through the Fragment and that navigation to reach that point.
Also handles the top-level navigation from the bottom bar by making the NavHost be the single source of truth of what tab is selected. Before as we were setting it ourselves, there were situations where going back from the profile tab would show the profile bottom item selected but we were in the home screen.
Since we use the NavHost in the home screen too, this gives us the ability to now not have to have multiple activities for new features, but they can be destinations on this NavHost. Features like claim flow which have their own NavHost can be merged with this one eventually.

[nav_host_in_home.webm](https://github.com/HedvigInsurance/android/assets/44558292/917d9f46-21a8-49b0-be6d-4cd061463d86)

From the commit messages:

* Fix initial tab handling

* Fix Profile and Home destinations to fill the entire screen

All destinations need to ask themselves to be full-screen since it can't
be enforced on the content of a NavHost globally. Each destination must
be responsible for itself regarding this unfortunately.

* Make gradient animate and fix inset consuming

When there is a known background gradient, first show it immediately
and from then on animate between them when it changes.
Also consume the correct insets only from the content depending on if
the nav rail or the bottom nav is showing.
And flip the order of the column/row of the navrail/bottombar so that
the bottom bar coming in with its own background feels more natural.

* Add fadeThrough animation by default for top level destinations

Context:
https://m3.material.io/styles/motion/transitions/transition-patterns#8327d206-2a7d-423a-abf0-2be86130535b
https://m3.material.io/styles/motion/transitions/applying-transitions#b8ef3749-69b8-4294-a035-661462ec4262

* Make LoggedInActivity landscape mode look nicer

Add the nav rail instead of the bottom bar for wide screens.
Remove backgrounds from referrals an insurance destinations. Instead
should show the background gradient.

* Allow LoggedInActivity to be in landscape mode

* Fix bottom nav bar animating in properly

Now it "slides" in properly without squeezing the content

* Extract daysSinceLastTooltipShown

* Fix clearing the tooltipShown step earlier

Before you'd get that dialog more than once if you leave the screen to
early. This should fix it so that it clears itself as soon as it has
started showing once.

* Make notification dot for cross-sells sticky

When we receive a `true`, store it locally, and only clear it after
we've left the screen at least once

* Fix app_deep_link run config to default on deep link

* Remove a forgotten log

Also remove the "Stelios" from one place. I usually search for "Stelios"
so I make sure I haven't left some forgotten log in the code by
accident 😅
So might as well remove that not to confuse myself.

* Fix backgroundColors crash when in a single item list

Brush.linearGradient expects a list of 2 items or more, otherwise
crashes. SolidColor() is the solution to just using one color for those
cases.

* Fix notificationDot position in bottom nav bar

* Remove the Fragments from the four top level destinations.

Following up the work from the previous PR which made all these
destinations compose-first, this commit also removes the fragments
directly, and instead contribute these four destinations to the
HedvigNavHost through their extensions on NavGraphBuilder.

Functionality stays the same for all of them

* Add HedvigBottomBar

Uses HedvigAppState to trigger top-level navigation events

* Add HedvigNavHost, the centralized spot for app navigation

For now will contain the top level destinations that are driven from
bottom bar navigation

* Remove XML from LoggedInActivity and introduce HedvigAppState

Move some of the handling of the initial screen into HedvigAppState.
The NavHost is in there, enabling us to observe the current navigation
stack.
This allows us to make decisions for stuff like showing the
bottom nav bar.
what background gradient is used behind the content.
Triggering some logging side effects from navigation events.
Handling top-level navigation between the top level destinations.
What notification icons should show in each navigation item.
Whether we show the forever tab or not.
And more in the future.

* Make InsuranceViewModel use stateIn()

Fixes not having the UI to manually trigger a `load` function, but
instead depends on a flow consumer to exist for the flow to become
hot.

* Make openWebBrowser an extension on Context

* Add more dependencies that the :app module needs

Plus a couple of small fixes

* Replace LoggedInTabs with TopLevelDestination

* Rename LoggedInViewModel to ReviewDialogViewModel

It should only handle the review related stuff anyway. The rest is moved
to HedvigAppState

* Remove old LoggedInActivity gradient and menus from xml

Fix icons for bottom navigation


app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+  val backgroundColors: GradientColors?
+    @Composable
+    get() {
+      val isLightMode = !isSystemInDarkTheme()
+      return when (currentTopLevelDestination) {
+        TopLevelDestination.HOME -> {
+          if (isLightMode) {
+            GradientColors(Color(0xFFC0CAD8), Color(0xFFEDCDAB), Color(0xFFF6F6F6))
```
hugokallstrom, line 107: Could these be extracted to the UI module?

app/app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/tab/InsuranceViewModel.kt
```diff
@@ -24,25 +31,41 @@ class InsuranceViewModel(
...
+  private val loading = MutableStateFlow(false)
+  private val dataOrNullFlow: Flow<List<InsuranceModel>?> = retryChannel.transformLatest {
+    loading.update { true }
+    either {
+      val contracts = getContractsUseCase.invoke().bind()
+      val crossSells = getCrossSellsUseCase.invoke().bind()
+      createInsuranceItems(contracts, crossSells)
+    }.fold(
+      ifLeft = {
+        emit(null)
```
hugokallstrom, line 45: Are we logging this error somewhere?

hugokallstrom (approved):
Super nice!!

app/app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/tab/InsuranceViewModel.kt
StylianosGakis: Hmm no we don't, and we didn't use to either https://github.com/HedvigInsurance/android/blob/77eacb4f93942ec861ae691b4e92d4bf8f252572/app/app/src/main/kotlin/com/hedvig/app/feature/insurance/ui/tab/InsuranceViewModel.kt#L39
I can add something though for sure.

Right now when this happens, they get the error screen (since `hasError` evaluates to true further down where we do `hasError = insuranceModels == null && loading == false,`), and they can always retry loading by clicking the button. So it's an easily recoverable error if you were worried about that part.

PR #1633, opened 2023-06-01
Try to show review dialog more than just once

StylianosGakis:
Before we would try to show the review dialog just once and absolutely never again.
This probably means that most of our members who've had this show at least once will never see it again, even if they never responded. This change should make it so that as the users use the app more, this dialog will trigger again eventually.
We should not be concerned regarding showing this too much, as the review APIs by themselves rate limit when this is shown. In reality they will see it less often than every month for example, as explained here https://developer.android.com/guide/playcore/in-app-review#quotas


hugokallstrom (approved):
🧠

PR #1652, opened 2023-06-26
Show error message for failed login

hugokallstrom:
Update authlib and use the error message from backend.

Screenshot_20230626_150922


app/app/src/main/kotlin/com/hedvig/app/authenticate/BankIdLoginViewModel.kt
```diff
@@ -68,12 +69,20 @@ class BankIdLoginViewModel(
...
+        }
+
+        is AuthAttemptResult.ZignSecProperties -> {
+          e { "Got ZignSec properties when signing in with BankId" }
+          startLoginAttemptFailed.update { "Got ZignSec properties when signing in with BankId" }
+        }
+
+        is AuthAttemptResult.OtpProperties -> {
+          e { "Got Otp properties when signing in with BankId" }
+          startLoginAttemptFailed.update { "Got Otp properties when signing in with BankId" }
```
hugokallstrom, line 84: Should we use a generic error message for the view state here?

app/app/src/main/kotlin/com/hedvig/app/authenticate/BankIdLoginViewModel.kt
```diff
@@ -133,9 +142,11 @@ sealed interface BankIdLoginViewState {
   object Loading : BankIdLoginViewState {
     override fun toString(): String = "Loading"
   }
-  object Error : BankIdLoginViewState {
+
+  data class Error(val message: String) : BankIdLoginViewState {
     override fun toString(): String = "Error"
```
StylianosGakis, line 147: Maybe edit the toString here to have the message in there too. This is used for the logs only, so more info would always be good. Something like `...: String = "Error: message:$message"`

StylianosGakis (approved):
Some test complaints only left I think. Probably need to catch the new cases in the sealed interfaces.

app/app/src/main/kotlin/com/hedvig/app/authenticate/BankIdLoginDialog.kt
```diff
@@ -80,6 +80,10 @@ class BankIdLoginDialog : DialogFragment(com.hedvig.app.R.layout.dialog_authenti
               startLoggedInActivity()
             }
           }
+          is LoginStatusResult.Exception -> {
```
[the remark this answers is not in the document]
hugokallstrom: Ill log them differently, good idea

app/app/src/main/kotlin/com/hedvig/app/ApplicationModule.kt
```diff
@@ -675,6 +676,7 @@ private val authRepositoryModule = module {
         AuthEnvironment.STAGING
       },
       additionalHttpHeaders = mapOf(),
+      callbacks = Callbacks("", ""), // Not used
```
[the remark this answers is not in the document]
hugokallstrom: Ill change to those just to be on the safe side.

app/app/src/main/kotlin/com/hedvig/app/authenticate/BankIdLoginViewModel.kt
hugokallstrom: I see. The error is not a network error though, so it will be confusing for us if members report that they get network errors but in reality there is something wrong with our systems. Its not pretty having hard coded error messages like this, but it would make identifying member reports so easy - we would immediately know what is wrong without having to dig into logs, check members in hope, etc. 

What do you think?

StylianosGakis: Hmm yeah I see what you mean.
The only slow thing with this would be that we need to localize the error message whatever that may be.
We could in this case instead have a message that says something like "There's something wrong on our end", and in the logs make sure to log which one of the alternatives has happened, so we can look into it.
That's at least as you say going to help with not mistakenly making them think it's a network error

hugokallstrom: A localized message would be nicer for the member, on the other hand this error "should never happen". I think in edge cases like this the benefit of knowing exactly what went wrong, directly from the member reporting the error, outweighs having a good UX with member readable error messages.

hugokallstrom: We dont have to do this for all features, but we have had quite a lot of trouble with specifically auth.

PR #1654, opened 2023-06-28
Download travel certificate to internal app storage directory

hugokallstrom:
Do this to avoid "Failed to find configured root"-crash

### DIFF
--- a/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
+++ b/app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
@@ -33,7 +33,7 @@ internal class DownloadTravelCertificateUseCase(
           .build()
 
         val downloadedFile = File(
-          context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS),
+          context.filesDir,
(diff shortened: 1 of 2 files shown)


app/feature/travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/data/DownloadTravelCertificateUseCase.kt
```diff
@@ -33,7 +33,7 @@ internal class DownloadTravelCertificateUseCase(
           .build()
 
         val downloadedFile = File(
-          context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS),
+          context.filesDir,
```
StylianosGakis, line 36: In the other place we handle files we're doing this and it was working okay, what do we think is different in this case which breaks it?

Also the file name is of the date today,  Iwas thinking if they try to download two files in one day, one will overwrite the other, should we maybe add also the hour-minute-second to the file name too?

hugokallstrom: Good point.. Im not sure if this fixes the problem to be honest. It seems to work on dev builds so it makes it quite hard to reproduce.

I added the current time to the file path.

hugokallstrom:
Ill merge this in order to get a staging build out

PR #1661, opened 2023-07-05
Last claims flow fixes

StylianosGakis:
More or less all changes included here are paddings and spacings.
Color changes.
Using the correct strings in various places.
Adapting some screens to the new design from the ground up, for those which looked a lot different before and after.

I'd encourage you to if you want to take a look at this to mostly run the app itself, go through the flow and optimally compare it with iOS too to get a feeling for what's different, and if there are any more bugs in either of the implementations.

Things still missing:

Loading from inside the button itself. Now it's the spinning H.
No progress bar at the top.

But those can be done later.


app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
```diff
@@ -77,7 +79,7 @@ class ChatViewModel(
                   message.fragments.chatMessageFragment,
                 )
             } else {
-              e { "Chat: subscription was not allowed to write" }
```
hugokallstrom, line 80: Should this even be a warning? We are setting the `isSubscriptionAllowedToWrite` to false in a lot of places

hugokallstrom (approved):
Super! 🔥

app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
StylianosGakis: Not sure if a warning is appropriate either. I was hoping wouldn't show for on-call which is the reason why I changed this, because it's a non-actionable trigger there which basically makes it noise that everyone ignores.
If a warning would also trigger it there, I'd be happy to move this even further down to just `i` for `info` so that we still have the data point, but it silences it for people on-call.

PR #1664, opened 2023-07-07
Improve chat upload failure logging

StylianosGakis:
Differentiates failure on chat upload, file/image picking and picture (taken at that moment) upload


hugokallstrom (approved):
👍

PR #1668, opened 2023-07-10
Log compose destinations to datadog

hugokallstrom:
With this we can create metrics for errors in claims flow, as well as number of successful claims

Example:  https://app.datadoghq.eu/rum/sessions?query=%40application.id%3A4d7b8355-396d-406e-b543-30a073050e8f%20env%3Adev%20%40type%3Aview&cols=&event=AgAAAYlAK9tYrpuf6QAAAAAAAAAYAAAAAEFZbEFLOXRZQUFBelRkeFFBWHpjNE9lbwAAACQAAAAAMDE4OTQwMmUtYzYwYi00NzJkLTljZTQtZDMwNTg3Y2E0ZmUz&viz=stream&from_ts=1688998380133&to_ts=1688999280133&live=true

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
@@ -17,6 +17,7 @@ import androidx.navigation.NavGraph.Companion.findStartDestination
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.navOptions
+import com.datadog.android.rum.GlobalRum
 import com.google.accompanist.navigation.animation.rememberAnimatedNavController
 import com.hedvig.android.hanalytics.featureflags.FeatureManager
 import com.hedvig.android.hanalytics.featureflags.flags.Feature
@@ -208,6 +209,10 @@ private fun NavigationTrackingSideEffect(navController: NavController) {
           }
         }
       }
+      GlobalRum.get().startView(
+        key = destination,
+        name = destination.route ?: destination.displayName,
+      )
     }
     navController.addOnDestinationChangedListener(listener)
     onDispose {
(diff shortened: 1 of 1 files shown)


app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -208,6 +209,10 @@ private fun NavigationTrackingSideEffect(navController: NavController) {
           }
         }
       }
+      GlobalRum.get().startView(
+        key = destination,
+        name = destination.route ?: destination.displayName,
```
StylianosGakis, line 214: `destination.displayName` seems to have `@get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` on it, which means that they optimally want to restrict the callers only to the library itself.

Basically they'd want it to have the `internal` modifier but they can't for some reason.
So the tl;dr is that we probably should avoid calling it, as it does not have any guarantees that it will still be there forever, or that the implementation won't change (so our metrics might start reporting changes even though nothing we did changed) and so on.

I think we can do something like `destination.route ?: "Unknown route",` instead, since I don't thin there's realistically ways that we'll be getting a null route anyway. And if we do we can see if it is a real problem or not. For our destinations, none of it should be null anyway.

StylianosGakis (approved):
Very cool, let's see how the metrics that come out of this look like!

A question, if we never do a paired `stopView` after these `startView` will datadog have an issue with it? Do their docs mention anything about this?

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
hugokallstrom: Good catch, fixing

hugokallstrom:
> Very cool, let's see how the metrics that come out of this look like!
> 
> A question, if we never do a paired `stopView` after these `startView` will datadog have an issue with it? Do their docs mention anything about this?

Did not mention that in the docs for the compose use case. I checked the logs and it seems to be logged correctly.

PR #1670, opened 2023-07-10
GEN-729: new design icons

StylianosGakis:
image

image

Used https://github.com/DevSrSouza/svg-to-compose to get these files.

Made an object named Hedvig so that as we are used to doing
`Icons.Default.Foo`
We can now do
`Icons.Hedvig.Foo`
And it should still work.
If the namespace being polluted by this `Hedvig` ends up being annoying we can think of renaming it


StylianosGakis:
Should've named the branch using the gen-729 too, can do next time

hugokallstrom (approved):

PR #1672, opened 2023-07-11
GEN-732: Vector flags

StylianosGakis:
With these, important *not* to use `androidx.compose.material3.Icon` but to use `androidx.compose.foundation.Image` instead. Since Icon puts the LocalContentColor as tint, which the flags do not want. Image just puts the image as it is, without applying any color change to it.

image

image


hugokallstrom (approved):
👍

PR #1677, opened 2023-07-11
GEN-717: Add composables for documents

hugokallstrom:
image

Not sure about the colors from the theme here. Should we change those later?

### DIFF
--- a/app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
+++ b/app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
@@ -1,23 +1,25 @@
 package com.hedvig.android.core.ui.insurance
 
-import androidx.compose.foundation.background
-import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.material.icons.Icons
 import androidx.compose.material3.Icon
+import androidx.compose.material3.LocalContentColor
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.buildAnnotatedString
+import androidx.compose.ui.text.style.BaselineShift
+import androidx.compose.ui.text.withStyle
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
-import com.hedvig.android.core.designsystem.newtheme.SquircleShape
+import com.hedvig.android.core.designsystem.component.card.HedvigCard
 import com.hedvig.android.core.designsystem.preview.HedvigPreview
 import com.hedvig.android.core.designsystem.theme.HedvigTheme
 import com.hedvig.android.core.icons.Hedvig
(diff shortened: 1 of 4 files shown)


app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.hedvig.android.core.icons.hedvig.small.hedvig.ArrowNorthEast
+
+@Composable
+fun DocumentRow(
+  modifier: Modifier = Modifier,
+  title: String,
+  subTitle: String,
+  onClick: () -> Unit,
+) {
+  Row(
```
StylianosGakis, line 33: So in Figma these icons seem to have a background color of f0f0f0.
In our design system, the color which maps to this in light mode is `surfaceVariant`. If you follow the references from here https://github.com/HedvigInsurance/android/blob/93966a0194e063771c25701f721ea26c42ab71b3/app/core/design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt#L82 you'll see.

I think you can wrap this entire thing with a `HedvigCard` and do something like
```
HedvigCard(
  onClick = onClick,
  colors = CardDefaults.outlinedCardColors(containerColor = MaterialTheme.colorScheme.surfaceVariant),
) {
 Box() ... this code here
 // And remove the background() modifier from it since HedvigCard now sets the color, same for onClick. Keep the inner Padding that you have though.
}
```

Oh and in this process, maybe consider changing the default shape of the HedvigCard to `shape: Shape = MaterialTheme.shapes.squircle,` like HedvigBigCard does. That Big card is an alternative version of the card which needs a minimum size which is bigger, hence the name :D That one is newer so that has squirce by default. Now that we're doing all this redesign it's okay to propagate such changes to random places, we're gonna take a look at them anyway.

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
```diff
@@ -0,0 +1,71 @@
...
+      )
+      .fillMaxWidth()
+      .padding(horizontal = 16.dp, vertical = 12.dp)
+      .clickable { onClick() },
+  ) {
+    Column {
+      Row {
+        Text(title)
+        Spacer(modifier = Modifier.padding(1.dp))
+        Text("PDF", fontSize = 11.sp)
```
StylianosGakis, line 49: For this you can reuse what I had done here https://github.com/HedvigInsurance/android/blob/42a2b7379f8237e42e5c5edf7d67541b401d46c5/app/feature/changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressOfferDestination.kt#L385-L398
Basically you can make parts of a text feel like a "superscript" by using an annotated string.

Here's a SOE post about this https://stackoverflow.com/a/66801935/9440211. I just didn't use superscript I remember since it was not looking as we wanted it to look. You can see for yourself if it is what we want in this scenario though.

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
```diff
@@ -0,0 +1,71 @@
...
+      }
+      Text(subTitle, color = MaterialTheme.colorScheme.secondary)
+    }
+
+    Icon(
+      imageVector = Hedvig.ArrowNorthEast,
```
StylianosGakis, line 55: My intention with those was to be able to do `Icons.Hedvig.ArrowNorthEast` which looks a lot like `Icons.Default.Foo` as we're doing for the built-in ones.
Otherwise just doing `Hedvig.anything` feels quite a big leap, like what is `Hedvig` by itself?
If it was the IDE which removed the initial `Icons` by itself, we need to re-think this naming since I think it will be super confusing to have things like `Hedvig.Plus` for example.

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
```diff
@@ -0,0 +1,71 @@
...
+      contentDescription = "link",
+    )
+  }
+}
+
+@Composable
+@HedvigPreview
+fun PreviewDocumentRow() {
+  HedvigTheme {
```
StylianosGakis, line 64: With previews it's easier to work with when you already have the surface (and as a result the content colors too) by using a surface composable top level.

Recently I've made a Live Template which looks like this

image

The text there is:
```
@com.hedvig.android.core.designsystem.preview.HedvigPreview
@androidx.compose.runtime.Composable
private fun Preview$NAME$() {
  com.hedvig.android.core.designsystem.theme.HedvigTheme {
    androidx.compose.material3.Surface(color = androidx.compose.material3.MaterialTheme.colorScheme.background) {
        $NAME$($END$)
    }
  }
}
```

And this makes it so that if you type "prev" and then "enter" it auto-fills the preview for you with some sane defaults.

And for this transition period, you may also want to do `HedvigTheme(useNewColorScheme = true) {` instead, so that you get the new color changes in your previews too.

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
```diff
@@ -0,0 +1,71 @@
...
+        color = MaterialTheme.colorScheme.background
+      )
+      .fillMaxWidth()
+      .padding(horizontal = 16.dp, vertical = 12.dp)
+      .clickable { onClick() },
+  ) {
+    Column {
+      Row {
+        Text(title)
+        Spacer(modifier = Modifier.padding(1.dp))
```
StylianosGakis, line 48: I always find it confusing having to do *2 in my head if it's a padding vs if this were a width instead.
Makes it easy to miss spots if you quickly glance over the numbers and see a different number from what you see in figma which is 2 instead of 1. So could we do that instead?
```suggestion
        Spacer(Modifier.width(2.dp))
```

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/DocumentRow.kt
hugokallstrom: Wow that is so good. I love live templates.

PR #1678, opened 2023-07-12
Fix: Remove decimal handling in monetary input

StylianosGakis:
There was a crash with an edge case on the monetary amount, but we are not interested in handling decimal amounts anyway. So remove the entire handling code for it instead, the tests still pass for the non-decimal situation 😊

### DIFF
--- a/app/data/claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowRepository.kt
+++ b/app/data/claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowRepository.kt
@@ -195,7 +195,7 @@ internal class ClaimFlowRepositoryImpl(
               itemModelInput = Optional.presentIfNotNull(itemModelInput),
               itemProblemIds = Optional.presentIfNotNull(itemProblemIds),
               purchaseDate = Optional.presentIfNotNull(purchaseDate),
-              purchasePrice = Optional.presentIfNotNull(purchasePrice),
+              purchasePrice = Optional.presentIfNotNull(purchasePrice.takeIf { it: Double? -> it != 0.0 }),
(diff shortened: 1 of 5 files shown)


app/data/claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowRepository.kt
```diff
@@ -195,7 +195,7 @@ internal class ClaimFlowRepositoryImpl(
               itemModelInput = Optional.presentIfNotNull(itemModelInput),
               itemProblemIds = Optional.presentIfNotNull(itemProblemIds),
               purchaseDate = Optional.presentIfNotNull(purchaseDate),
-              purchasePrice = Optional.presentIfNotNull(purchasePrice),
+              purchasePrice = Optional.presentIfNotNull(purchasePrice.takeIf { it: Double? -> it != 0.0 }),
```
hugokallstrom, line 198: Should we log if we get decimals here?

hugokallstrom (approved):

app/data/claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowRepository.kt
StylianosGakis: So my idea was that the input field does not allow decimals at all in the monetary amount, so we do not need to check for that.
This change here is so that if someone writes "0000" which is a valid input in the text field, then we don't send an amount of value "0.0" over the network. That fails the next step basically, probably causes an exception on the backend somehow. So if someone has written nothing (null) or has written 0, we also turn that to null and don't send the purchase price at all to the backend.

PR #1687, opened 2023-07-17
Turn ContractDetailActivity into a compose screen

StylianosGakis:
Hook it up with the rest of the application's graph


hugokallstrom (approved):
🚀

PR #1689, opened 2023-07-18
RND-183: Move terminate button to documents tab. Change color.

hugokallstrom:


StylianosGakis (approved):
Hmm can you also after this is merged, merge new develop and the 688 branch which contains the redesign, basically by dropping all the changes that are done in this PR? So that 688 is also up to date with develop ad we don't get a huge merge conflict later if we do more changes.

If you can do this merge in a way that idk, basically completely ignores these changes over here would be best, since all those screens have changed enough that they don't need all this work with adapters and whatever to make this change happen. It's much simpler with the pure compose implementation

hugokallstrom:
> Hmm can you also after this is merged, merge new develop and the 688 branch which contains the redesign, basically by dropping all the changes that are done in this PR? So that 688 is also up to date with develop ad we don't get a huge merge conflict later if we do more changes.
> 
> If you can do this merge in a way that idk, basically completely ignores these changes over here would be best, since all those screens have changed enough that they don't need all this work with adapters and whatever to make this change happen. It's much simpler with the pure compose implementation

Yes good idea, i will do that

PR #1690, opened 2023-07-18
GEN-631: Add select contract screen to claims flow. Map GraphQL UUID to String.

hugokallstrom:


app/feature/odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/di/OdysseyModule.kt
```diff
@@ -57,4 +59,7 @@ val odysseyModule = module {
   viewModel<SingleItemPayoutViewModel> { (singleItemPayout: ClaimFlowDestination.SingleItemPayout) ->
     SingleItemPayoutViewModel(singleItemPayout, get<ClaimFlowRepository>())
   }
+  viewModel<SelectContractViewModel> { (selectedContract: String?, contractOptions: List<LocalContractContractOption>) ->
```
StylianosGakis, line 62: This could get the entire `(ClaimFlowDestination.SelectContract)` instead of `(String, List<LocalContractContractOption>)`. Should help for if the options change in the future, and also helps with issues like if you want to send over 2 things of type `String` since koin has no way to know which one is which otherwise.

We've done the same above for SingleItemViewModel for example.

app/feature/odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/selectcontract/SelectContractViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.hedvig.android.data.claimflow.ClaimFlowStep
+import com.hedvig.android.data.claimflow.LocalContractContractOption
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+internal class SelectContractViewModel(
+  initialSelectedContract: String?,
```
StylianosGakis, line 15: Aha I see what this is here then. For the initial selection.
I don't think we're ever gonna decide on the initial selection from this parameter, in fact I see it's just `null` all the time where we instantiate this ViewModel.

Can we instead not have this parameter at all and do the "selecting the first contract by default" logic just in here instead?

app/feature/odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/selectcontract/SelectContractViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+      locationOptions: List<LocalContractContractOption>,
+    ): SelectContractUiState {
+      val selectedLocation = locationOptions
+        .firstOrNull { it.id == initialSelectedLocation }
+      return SelectContractUiState(
+        locationOptions,
+        selectedLocation,
+      )
+    }
+  }
```
StylianosGakis, line 89: From what we briefly discussed when we were talking about this we thought having a contract pre-selected always would be a sane default, since we don't have a design for what happens when there's no selected contract in the first place.

Can we make selectedContract be a non-null field, and have in the intialization logic something which auto-selects the first contract instead? 

Something like this is what I had in mind.

And the list of contractOptions should be guarnateed to be of at least size 2, since otherwise we should not be getting this step in the first place, so doing `.first()` on the list should be okay, if we ever crash we have a bigger issue than that.

```suggestion
internal data class SelectContractUiState(
  val contractOptions: List<LocalContractContractOption>,
  val selectedContract: LocalContractContractOption,
  val isLoading: Boolean = false,
  val error: Boolean = false,
  val nextStep: ClaimFlowStep? = null,
) {
  val canSubmit: Boolean = !isLoading && !error && nextStep == null

  companion object {
    fun fromInitialSelection(
      locationOptions: List<LocalContractContractOption>,
    ): SelectContractUiState {
      return SelectContractUiState(
        contractOptions = locationOptions,
        selectedContract = locationOptions.first(),
      )
    }
  }
```

StylianosGakis (approved):
Some thoughts on the pre-selection of the selected contract. We can bring the iOS folks into this discussion too

app/feature/odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/selectcontract/SelectContractViewModel.kt
hugokallstrom: Great, thanks

PR #1697, opened 2023-07-19
GEN-796: Refactor app info screen to compose, implement new design

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/aboutapp/AboutAppDestination.kt
```diff
@@ -0,0 +1,170 @@
...
+    memberId = memberId,
+    onBackPressed = onBackPressed,
+    showOpenSourceLicenses = { showWebView = true },
+    isProduction = isProduction,
+  )
+
+  if (showWebView) {
+    val state = rememberWebViewState(licensesUrl)
+    WebView(state = state)
+  }
```
StylianosGakis, line 78: Does this show on top of the rest as a full-screen thing?
How does one exit it, does it have an X on the top or something like that?
I haven't checked out the accompanist compose WebView myself yet 👀

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/aboutapp/AboutAppDestination.kt
```diff
@@ -0,0 +1,170 @@
...
+) {
+  val memberIdState by viewModel.state.collectAsStateWithLifecycle()
+  val memberId: String? by remember {
+    derivedStateOf {
+      when (val latestMemberIdState = memberIdState) {
+        is MemberIdViewModel.State.Success -> latestMemberIdState.id
+        else -> null
+      }
+    }
+  }
```
StylianosGakis, line 64: I think since here the result (memberId) is a simple mapping of what the status of `memberIdState` is, we can skip the remember and just do

```suggestion
  val memberId: String? = when (val latestMemberIdState = memberIdState) {
    is MemberIdViewModel.State.Success -> latestMemberIdState.id
    else -> null
  }
```
derivedStateOf usually is helpful when the state onbject changes often, but the result of whatever is inside the calculation does not change as often, which I don't think is the case here.

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/aboutapp/AboutAppDestination.kt
```diff
@@ -0,0 +1,170 @@
...
+          }
+        }
+        Row(
+          horizontalArrangement = Arrangement.SpaceBetween,
+          verticalAlignment = Alignment.CenterVertically,
+          modifier = Modifier
+            .fillMaxWidth()
+            .padding(16.dp),
+        ) {
+          Text("Version")
```
StylianosGakis, line 125: We don't need this to be localized I suppose right?

StylianosGakis (approved):

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/aboutapp/AboutAppDestination.kt
hugokallstrom: Yeah the navigation here was a bit off, moved this to its own destination.

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/aboutapp/AboutAppDestination.kt
hugokallstrom: Added string resource just in case 👍

PR #1698, opened 2023-07-19
GEN-716: insurance tab use octopus

StylianosGakis:
Octopus returns the correct colors
And the UI has no room for the exceptions or generic info per peril, so we don't query those, at least not for now.


hugokallstrom (approved):
Nice!

PR #1699, opened 2023-07-20
GEN-798: Settings screen

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/settings/SettingsDestination.kt
```diff
@@ -0,0 +1,309 @@
...
+    val observer = LifecycleEventObserver { _, event ->
+      if (event == Lifecycle.Event.ON_RESUME) {
+        notificationsEnabled = notificationManager.areNotificationsEnabled()
+      }
+    }
+    lifecycleOwner.lifecycle.addObserver(observer)
+    onDispose {
+      lifecycleOwner.lifecycle.removeObserver(observer)
+    }
+  }
```
StylianosGakis, line 83: I'd guess that doing `NotificationManagerCompat` is an at least somewhat expensive operation that we'd be happy to avoid doing on each potential recomposition, so I'd suggest we `remember` that one and keep the same instance across recompositions.
And with that, then also a good idea to use notificationManager as a key to the DisposableEffect, since if that one changes somehow, the effect itself would still be referencing the old one instead.

```suggestion
  val notificationManager = remember(context) { NotificationManagerCompat.from(context) }
  var notificationsEnabled by remember { mutableStateOf(notificationManager.areNotificationsEnabled()) }
  val uiState by viewModel.uiState.collectAsStateWithLifecycle()
  val lifecycleOwner = LocalLifecycleOwner.current

  DisposableEffect(lifecycleOwner, notificationManager) {
    val observer = LifecycleEventObserver { _, event ->
      if (event == Lifecycle.Event.ON_RESUME) {
        notificationsEnabled = notificationManager.areNotificationsEnabled()
      }
    }
    lifecycleOwner.lifecycle.addObserver(observer)
    onDispose {
      lifecycleOwner.lifecycle.removeObserver(observer)
    }
  }
```

app/app/src/main/kotlin/com/hedvig/app/feature/settings/SettingsDestination.kt
```diff
@@ -0,0 +1,309 @@
...
+  HedvigBigCard(
+    onClick = { showThemePickerDialog = true },
+    hintText = stringResource(R.string.SETTINGS_THEME_TITLE),
+    inputText = stringResource(id = selectedTheme.getLabel()),
+    enabled = enabled,
+    modifier = Modifier.fillMaxWidth(),
+  )
+}
+
+private fun startSettingsActivity(context: Context) {
```
StylianosGakis, line 293: Hmm think it'd be a tiny bit nicer to name this something like `startAndroidSettingsActivity`?
At first glance I was wondering why our settings are an activity but I realize now this is for the android settings themselves, not an activity we control.
If you don't like the name feel free to not change it though, or if you have a better one just pick that one instead 😊

app/core/datastore/src/main/kotlin/com/hedvig/android/core/datastore/SettingsDataStore.kt
```diff
@@ -0,0 +1,44 @@
...
+  private val languageKey = stringPreferencesKey("settings-language")
+  private val themeKey = stringPreferencesKey("settings-theme")
+
+  override suspend fun setTheme(theme: String) {
+    dataStore.edit {
+      it[themeKey] = theme
+    }
+  }
+
+  override fun observeTheme(): Flow<String> {
```
StylianosGakis, line 30: Do you think this could return an instance of our `enum class Theme` directly? So that we can use the default of Theme.Light for example instead of the String "LIGHT", and also not have to put the burden of doing this translation on the consumer of this class.

This might require that Theme enum to move to another module and that's perfectly fine. It could move into a app:theme, exactly like we have app:market and app:language.
Then this datastore module could depend on that so it knows about Theme, and :app:app would also be able to know about it and use it.

But could be another PR and if you feel like doing it.

app/core/datastore/src/main/kotlin/com/hedvig/android/core/datastore/SettingsDataStore.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  override suspend fun setLanguage(language: String) {
+    dataStore.edit {
+      it[languageKey] = language
+    }
+  }
+
+  override fun observeLanguage(): Flow<String> {
+    return dataStore.data.map { it[languageKey] ?: "EN_SE" }
+  }
```
StylianosGakis, line 42: I think for language we should use [LanguageService](https://github.com/HedvigInsurance/android/blob/0d5f2f0a3ca4ca42b18eb98dcabb721648fa3d20/app/language/language-core/src/main/kotlin/com/hedvig/android/language/LanguageService.kt#L15) instead, since that one uses `AppCompatDelegate.getApplicationLocales()` and `AppCompatDelegate.setApplicationLocales` to get and set the language.

The benefit of those APIs is that you can then in compose just use [getLocale()](https://stackoverflow.com/a/75172481/9440211) and you get the latest value automatically. It works with the [per-app language settings](https://developer.android.com/guide/topics/resources/app-languages) that Android 13 and on allow you to edit.

That way we don't need 2 sources of truth for what language the app is in, our settings and Android settings. We just use those APIs and we get 1 place where the truth is in.

From those docs: "These APIs automatically sync with system settings; therefore, apps that use these APIs to create custom in-app language pickers will ensure their users have a consistent user experience regardless of where they select their language preferences".

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/dialog/SingleSelectDialog.kt
```diff
@@ -1,4 +1,4 @@
-package com.hedvig.android.feature.odyssey.ui
+package com.hedvig.android.core.ui.dialog
```
StylianosGakis, line 1: 👌👌

StylianosGakis (approved):
Awesome work!

Mostly have a comment about storing the language decision, I know we may have not been doing it this way before, but maybe it's a good opportunity to rely on LanguageService fully for this from now on.

app/app/src/main/kotlin/com/hedvig/app/feature/settings/SettingsDestination.kt
hugokallstrom: Good idea!

app/core/datastore/src/main/kotlin/com/hedvig/android/core/datastore/SettingsDataStore.kt
hugokallstrom: Should definitely do that, ill add another commit

PR #1700, opened 2023-07-21
Payment screen composable

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+  Spacer(Modifier.height(12.dp))
+  Row(
+    modifier = Modifier
+      .fillMaxWidth()
+      .padding(vertical = 12.dp),
+    horizontalArrangement = Arrangement.SpaceBetween,
+  ) {
+    Text(stringResource(R.string.PAYMENTS_NEXT_PAYMENT_SECTION_TITLE))
+    Text(
+      text = uiState.nextChargeDate.toString(),
```
hugokallstrom, line 118: Should I use kotlin LocalDate here? Or java?

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+        )
+        Spacer(Modifier.width(8.dp))
+        Text(insuranceCost.displayName)
+      }
+      Text(
+        text = insuranceCost.cost.format(locale),
+        color = MaterialTheme.colorScheme.onSurfaceVariant,
+      )
+    }
+    Divider()
```
StylianosGakis, line 153: Is the divider at the end of the list here on purpose?
Maybe the `InsuranceCosts` component by itself can only render the dividers between its items, and if the entire layout does require a divider there for some reason it can be part of the column of items inside `PaymentScreen` instead?
Basically what I mean is so that `InsuranceCosts` itself does not make any assumptions about where it's placed, it just works no matter where it's placed. Right now I *think* that it makes the assumption that the item below it also wants a divider between that and this InsuranceCosts component. So if that items leaves from the `PaymentScreen`, then this composable would need to be edited, which makes this a bit non-obvious.

Another way to think about it, is consider if you wanted to reuse this composable from another screen completely. You'd again want it to just render itself, and not make any assumptions about what's around it.

Does my thinking make sense? Maybe the design literally here asks for the divider at the end, even if it's the last item, but in that case the design is most likely wrong.

StylianosGakis: NextPayment right above for example does this well I think. It does not have the divider at all, but the screen composable itself knows that it wants a divider between it and the InsuranceCosts, so the screen puts it instead.

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+@Composable
+private fun NextPayment(uiState: PaymentUiState, locale: Locale) {
+  HedvigCard(modifier = Modifier.fillMaxWidth()) {
+    Text(
+      text = uiState.nextChargeAmount.format(locale),
+      textAlign = TextAlign.Center,
+      style = MaterialTheme.typography.displayMedium,
+      modifier = Modifier.padding(vertical = 6.dp),
+    )
+  }
```
StylianosGakis, line 108: Maybe this can be not part of the `NextPayment`, or maybe the Row below this? Or are these two components considered one in the design?

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+        containerColor = MaterialTheme.colorScheme.typeContainer,
+        contentColor = MaterialTheme.colorScheme.primary,
+      ),
+      modifier = Modifier.widthIn(min = 127.dp),
+    )
+  }
+}
+
+@Composable
+fun PaymentDetails(uiState: PaymentUiState) {
```
StylianosGakis, line 227:
```suggestion
private fun PaymentDetails(uiState: PaymentUiState) {
```

StylianosGakis, line 227: This function takes in the UiState I think only to use one field from it, the display name. Maybe it can just take in that one parameter only?
```suggestion
fun PaymentDetails(paymentMethodDisplayName: String) {
```

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+      Image(
+        imageVector = Icons.Hedvig.Payments,
+        contentDescription = null,
+        modifier = Modifier.size(24.dp),
+      )
+      Spacer(Modifier.width(16.dp))
+      Text(uiState.paymentMethod.displayName)
+    }
+    Text(
+      text = uiState.paymentMethod.displayValue,
```
StylianosGakis, line 251: This *will* be the "31489*****" text in the future right?

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+        InsuranceCosts(uiState, locale)
+        Spacer(Modifier.height(4.dp))
+        TotalDiscount(uiState)
+        Spacer(Modifier.height(4.dp))
+        Divider()
+        Spacer(Modifier.height(16.dp))
+        AddDiscount(uiState)
+        Spacer(Modifier.height(32.dp))
+        PaymentDetails(uiState)
+        Spacer(Modifier.height(4.dp))
```
StylianosGakis, line 86: This 4dp here is added because of the line 231 `Spacer(Modifier.height(4.dp))`, so it centers the item appropriately.
So now this screen is adding 4 dps under it, and the item itself is adding the 4dp over the item.
This would make it quite tricky to change if we ever re-order how these items come, or if the spacings change etc.

I'd suggest either:
* The item itself takes the space + 12dp vertical padding that it needs, and the screen up here does
```
Spacer(Modifier.height(4.dp))
PaymentDetails(uiState)
Spacer(Modifier.height(4.dp))
```
* The PaymentDetails item makes the vertical padding be 16dp directly, and the screen here does not add any special spacing around it
```
PaymentDetails(uiState)
```

image

Considering the design here maybe in this case it's easier to just add the 16.dp paddings and then no need to add those 4.dp extra here nor in the component itself

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+    }
+    Text(
+      text = uiState.paymentMethod.displayValue,
+      color = MaterialTheme.colorScheme.onSurfaceVariant,
+    )
+  }
+}
+
+@Composable
+fun PaymentHistory(onClick: () -> Unit) {
```
StylianosGakis, line 258:
```suggestion
private fun PaymentHistory(onClick: () -> Unit) {
```

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+      imageVector = Icons.Hedvig.ChevronRight,
+      contentDescription = null,
+      modifier = Modifier.size(16.dp),
+    )
+  }
+}
+
+@Composable
+@HedvigPreview
+fun PreviewPaymentScreen() {
```
StylianosGakis, line 288:
```suggestion
private fun PreviewPaymentScreen() {
```

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+        modifier = Modifier.size(24.dp),
+      )
+      Spacer(Modifier.width(16.dp))
+      Text(stringResource(id = R.string.PAYMENTS_PAYMENT_HISTORY_BUTTON_LABEL))
+    }
+    Icon(
+      imageVector = Icons.Hedvig.ChevronRight,
+      contentDescription = null,
+      modifier = Modifier.size(16.dp),
+    )
```
StylianosGakis, line 282: I vote for removing the chevron :D

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -0,0 +1,321 @@
...
+  }
+}
+
+@Composable
+fun PaymentHistory(onClick: () -> Unit) {
+  Row(
+    modifier = Modifier
+      .fillMaxWidth()
+      .padding(vertical = 12.dp)
```
StylianosGakis, line 262: I think this is also 16.dp from looking at the design maybe?

StylianosGakis (approved):
Great work! Had some comments regarding some approaches to doing these screens, feel free to discuss with me everything that I've mentioned 😊

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
hugokallstrom: I just though it made sense to group this since it contains both the next charge amount and the next charge date

PR #1701, opened 2023-07-21
Payment screen viewmodel

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -196,30 +239,36 @@ private fun AddDiscount(uiState: PaymentUiState) {
...
+        text = stringResource(id = R.string.PAYMENTS_ADD_CODE_BUTTON_LABEL),
+        onClick = onAddDiscountCode,
+        contentPadding = PaddingValues(horizontal = 16.dp, vertical = 16.dp),
+        colors = ButtonDefaults.buttonColors(
+          containerColor = MaterialTheme.colorScheme.typeContainer,
+          contentColor = MaterialTheme.colorScheme.primary,
+        ),
+        modifier = Modifier.widthIn(min = 127.dp),
+      )
+    }
```
StylianosGakis, line 271: I think what we want in this case for the small button is for it to be exactly as big as the text field on the left. And then for it to take as much space as it needs, and the rest can be give to the TextField to the left
```suggestion
    Row(
      modifier = Modifier.fillMaxWidth().height(IntrinsicSize.Max),
      horizontalArrangement = Arrangement.SpaceBetween,
    ) {
      HedvigTextField(
        value = uiState.discountCode?.code ?: "",
        onValueChange = { onDiscountCodeChanged(CampaignCode(it)) },
        errorText = uiState.discountError,
        label = {
          Text(stringResource(id = R.string.REFERRAL_ADDCOUPON_INPUTPLACEHOLDER))
        },
        modifier = Modifier.weight(1f),
      )
      Spacer(Modifier.width(8.dp))
      HedvigContainedSmallButton(
        text = stringResource(id = R.string.PAYMENTS_ADD_CODE_BUTTON_LABEL),
        onClick = onAddDiscountCode,
        contentPadding = PaddingValues(horizontal = 32.dp, vertical = 16.dp),
        colors = ButtonDefaults.buttonColors(
          containerColor = MaterialTheme.colorScheme.typeContainer,
          contentColor = MaterialTheme.colorScheme.primary,
        ),
        modifier = Modifier.fillMaxHeight(),
      )
    }
```

Would this work instead?
This way we make the row use IntrinsicSize.Max, so that it finds which one of the children is bigger in height (the text field for us) and then it also measures that high. This allows for the button to then be able to do `fillMaxHeight()` and that will make it be as big as that row, which in turn is as big as the TextField.

And then by adjusting the contentPadding to have 32.dp horizontal padding we get the result that design seems to want from this specific button.

Now if this makes it so that the button takes way too much space on the right side, leaving nothing to the input, maybe we can just put the horizontal padding back to 16, but still not have the hardcoded 127 width as that'll probably make it look weird depending on the localization and how big the inner text is.

StylianosGakis (commented):
Still a draft, just had one thought :D

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -70,23 +73,29 @@ fun PaymentDestination(
...
-    onChangeBankAccount = {
-      context.startActivity(
-        connectPayinIntent(
-          context,
-          PaymentType.TRUSTLY,
-          Market.SE,
-          false,
-        ),
+
+  AnimatedContent(targetState = uiState.isLoading) { loading ->
```
StylianosGakis, line 77: This by default does a slight scale transition (from 92% to 100% size) and a fade.
There's also just `CrossFade` which only does the fade without the scale effect.
I don't have a personal opinion on which one we'd prefer, just letting you know what easy options we got so you can pick the one you feel like is best for simple transitions like these.

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel2.kt
```diff
@@ -70,39 +75,27 @@ class PaymentViewModel2(
...
-  totalDiscount = totalDiscount.format(locale),
+  totalDiscount = totalDiscount?.negate()?.format(locale),
   activeDiscounts = redeemedCampagins.map {
-     PaymentViewModel2.PaymentUiState.Discount(
-       code = it.incentive.toString(),
-       displayName = it.displayValue ?: "Unknown"
-     )
+    PaymentViewModel2.PaymentUiState.Discount(
+      code = it.code,
+      displayName = it.displayValue ?: "-",
```
StylianosGakis, line 97: To print and format potentially nullable monetary amount values we got this https://github.com/HedvigInsurance/android/blob/90856a690cb8049e9715901c7af7fa9629cb6ee8/app/core/ui-data/src/main/kotlin/com/hedvig/android/core/uidata/UiNullableMoney.kt#L12-L22 is it something that would be useful in this context here?
It also contains this logic of adding "-" if there's no money in there for example.

If not disregard this comment, I am not 100% sure it'd fit here, just thought looks kinda familiar.

StylianosGakis (approved):
Looking nice! 🔥

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
hugokallstrom: Good to know, we can check with William later! I dont have any strong opinions, think both look good :)

StylianosGakis: Yeah, then just keep it and we can discuss in the future

app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentViewModel2.kt
hugokallstrom: We can switch to this after we migrate the payment types to octopus. `CurrencyCode` is not present in Giraffe currently which makes this class hard to use without refactoring

PR #1702, opened 2023-07-24
Insurance feature module

StylianosGakis:
Moves the entire insurances feature into its own module.
This means the Insurance tab (the 2nd screen in the bottom nav), the insurance detail screen (with the three horizontal pager tabs which we just redesigned) and the terminated contracts screen.

This process made it so that these screens no longer rely on shared code of some adapters, model mappers and so on which were also used in the offer screen etc. but those should be deleted completely soon anyway, so this split had to be done.

This also removes the gradient background from the home screens, since that's gonna go soon anyway.


app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceViewModel.kt
```diff
@@ -0,0 +1,201 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.first
+import octopus.CrossSalesQuery
+
+internal class InsuranceViewModel(
+  private val getInsuranceContractsUseCase: GetInsuranceContractsUseCase,
+  private val getCrossSellsUseCase: GetCrossSellsUseCase,
+  private val crossSellCardNotificationBadgeService: CrossSellCardNotificationBadgeService,
+) : MoleculeViewModel<InsuranceScreenEvent, InsuranceUiState>() {
```
StylianosGakis, line 34: This is this https://github.com/cashapp/molecule with some video about it here https://www.youtube.com/watch?v=q9p4ewk-9E4.
I wanted to bring it up in the code camp we got soon, so I thought what perfect way to give it a shot in the app too (I had tried it in sample projects before) so that we got something to reference in the code camp.

app/core/ui/src/main/kotlin/com/hedvig/android/core/ui/MoleculeViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.WhileSubscribed
+import kotlinx.coroutines.flow.stateIn
+import kotlin.time.Duration.Companion.seconds
+
+abstract class MoleculeViewModel<Event, UiState> : ViewModel() {
```
StylianosGakis, line 18: (Wrote this in another place too, but just to not confuse you if you see this first)

This is this https://github.com/cashapp/molecule with some video about it here https://www.youtube.com/watch?v=q9p4ewk-9E4.
I wanted to bring it up in the code camp we got soon, so I thought what perfect way to give it a shot in the app too (I had tried it in sample projects before) so that we got something to reference in the code camp.

hugokallstrom (approved):

PR #1706, opened 2023-07-25
Show connect payout for DE and NO markets

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/profile/ui/payment/PaymentDestination.kt
```diff
@@ -347,14 +382,49 @@ fun PaymentDetails(uiState: PaymentUiState) {
...
+  Row(
+    modifier = Modifier
+      .fillMaxWidth()
+      .padding(vertical = 12.dp, horizontal = 16.dp),
+    horizontalArrangement = Arrangement.SpaceBetween,
+    verticalAlignment = Alignment.CenterVertically,
+  ) {
+    Row(
+      verticalAlignment = Alignment.CenterVertically,
+    ) {
```
StylianosGakis, line 403: There seems to maybe be an extra Row in here that is not necessary? Maybe it was here before for the chevron or something like that, but if it's so I think you can just delete it and it should work.

StylianosGakis (approved):
lgtm 😊

PR #1711, opened 2023-07-26
Insurance viewmodel with molecule

StylianosGakis:


app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
```diff
@@ -90,7 +95,12 @@ internal fun InsuranceDestination(
   openChat: () -> Unit,
   imageLoader: ImageLoader,
 ) {
-  val uiState by viewModel.uiState.collectAsStateWithLifecycle()
+  val uiState: InsuranceUiState by viewModel.models.collectAsStateWithLifecycle()
+  LaunchedEffect(Unit) {
+    snapshotFlow { uiState }.collect {
+      d { "Stelios:$it" }
+    }
+  }
```
StylianosGakis, line 103: mb, will remove this

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -145,23 +68,96 @@ internal data class InsuranceUiState(
...
+        ),
+      )
+    }
+    var isLoading by remember { mutableStateOf(true) }
+    var didFailToLoad by remember { mutableStateOf(false) }
+    val retryChannel = remember { RetryChannel() }
+
+    val showNotificationBadge by crossSellCardNotificationBadgeService
+      .showNotification()
+      .collectAsState(seed.showNotificationBadge)
```
StylianosGakis, line 95: I realized before we were reacting to the status of the notification only on each time we refresh, which is not what we were really looking for here.
Since the state comes from inside DataStore as a `Flow<Boolean>`, we can directly listen to it and turn it into state with a normal `collectAsState(initialValue = ...)`.
No we get the reactivity for free, if DataStore ever updates, this auto-updates and the presenter will emit a new version of the UI state

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -145,23 +68,96 @@ internal data class InsuranceUiState(
...
+
+    CollectEvents { event ->
```
StylianosGakis, line 97: This is a custom Composable effect, same as LaunchedEffect, but we get the event directly in here, much more convenient this way

app/feature/insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenterTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import com.hedvig.android.feature.insurances.data.InsuranceContract
+import com.hedvig.android.molecule.test.test
+import com.hedvig.android.notification.badge.data.crosssell.card.FakeCrossSellCardNotificationBadgeService
+import giraffe.type.TypeOfContract
+import kotlinx.coroutines.test.runTest
+import octopus.CrossSalesQuery
+import org.junit.Assert
+import org.junit.Test
+
+internal class InsurancePresenterTest {
```
StylianosGakis, line 24: Please do give this class a nice look, I feel like here is where we get the biggest benefit of using Molecule.
The tests read so nicely like:
Start the test -> see initial state
Do something -> See that the UI is in the appropriate state
Make the backend return the response -> Automatically the UI has already reacted to it.

It just feels so easy to reason about, especially without having to know about the inner workings of how the ViewModel is doing things and when you need to `advanceUntilIdle()` or not.

app/feature/insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenterTest.kt
```diff
@@ -0,0 +1,263 @@
...
+  fun `when cross sells fail to load, we get an error state`() = runTest {
+    val getInsuranceContractsUseCase = FakeGetInsuranceContractsUseCase()
+    val getCrossSellsUseCase = FakeGetCrossSellsUseCase()
+    val presenter = InsurancePresenter(
+      getInsuranceContractsUseCase,
+      getCrossSellsUseCase,
+      FakeCrossSellCardNotificationBadgeService(),
+    )
+    presenter.test(InsuranceUiState.InitialState) {
+      skipItems(1)
```
StylianosGakis, line 112: We've got a test which tests the initial state already, so for the rest of the tests since we know we're testing UI state, we know the first emission is always the initial state, so we can just skip the first one.
This is the same as doing `awaitItem()` and then not doing anything with the response.

app/feature/insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenterTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+      getInsuranceContractsUseCase.errorMessages.add(ErrorMessage())
+      expectNoEvents() // No events while we're still waiting for all backend calls to finish
```
StylianosGakis, line 137: We can also test that after something has happened (in this case, only 1 of the backend responses has arrived) that we've not gotten any UiState updates in this moment in time.

app/molecule/molecule-public/src/main/kotlin/com/hedvig/android/molecule/public/MoleculePresenter.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.compose.runtime.LaunchedEffect
+import kotlinx.coroutines.flow.Flow
+
+fun interface MoleculePresenter<Event, Model> {
+  @Composable
+  fun MoleculePresenterScope<Event>.present(seed: Model): Model
+}
+
+class MoleculePresenterScope<Event>(
+  val events: Flow<Event>,
```
hugokallstrom, line 13: This can be private

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
```diff
@@ -1,43 +1,52 @@
...
 import kotlinx.coroutines.flow.WhileSubscribed
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.stateIn
 import kotlin.time.Duration.Companion.seconds
 
-abstract class MoleculeViewModel<Event, UiState> : ViewModel() {
+abstract class MoleculeViewModel<Event, Model>(
+  initialState: Model,
+  presenter: MoleculePresenter<Event, Model>,
+  val sharingStarted: SharingStarted = SharingStarted.WhileSubscribed(5.seconds),
```
hugokallstrom, line 22: This can be private

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
```diff
@@ -1,43 +1,52 @@
...
+) : ViewModel() {
   private val scope = CoroutineScope(viewModelScope.coroutineContext + AndroidUiDispatcher.Main)
 
   // Events have a capacity large enough to handle simultaneous UI events, but
   // small enough to surface issues if they get backed up for some reason.
   private val events = MutableSharedFlow<Event>(extraBufferCapacity = 20)
 
+  private var seed: Model = initialState
```
hugokallstrom, line 30: Can we rename this to something like `cachedState`, `storedState`, `lastState`? I think that better reflects what we are using it for

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsuranceViewModel.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.hedvig.android.feature.insurances.data.GetCrossSellsUseCase
+import com.hedvig.android.feature.insurances.data.GetInsuranceContractsUseCase
+import com.hedvig.android.feature.insurances.insurance.presentation.InsurancePresenter
+import com.hedvig.android.feature.insurances.insurance.presentation.InsuranceScreenEvent
+import com.hedvig.android.feature.insurances.insurance.presentation.InsuranceUiState
+import com.hedvig.android.molecule.android.MoleculeViewModel
+import com.hedvig.android.notification.badge.data.crosssell.card.CrossSellCardNotificationBadgeService
+
+internal class InsuranceViewModel(
+  private val getInsuranceContractsUseCase: GetInsuranceContractsUseCase,
```
hugokallstrom, line 12: You can remove `private val` from all these

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
```diff
@@ -1,43 +1,52 @@
...
+
   fun take(event: Event) {
```
hugokallstrom, line 32: Not sure about this name either, what do you think about `emit` or `consume`?

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
```diff
@@ -1,43 +1,52 @@
...
-  abstract val initialValue: UiState
-
-  val uiState: StateFlow<UiState> = moleculeFlow<UiState>(RecompositionMode.ContextClock) {
-    models(events)
-  }.stateIn(
-    scope = scope,
-    started = SharingStarted.WhileSubscribed(5.seconds),
-    initialValue = initialValue,
-  )
+  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
```
hugokallstrom, line 38: `uiState` and `State`, or `UiState` woul make more sense for our use case I think. I think about viewmodels when I read`models` and `Model` in this code base. What do you think about the naming here?

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -128,13 +50,14 @@ internal data class InsuranceUiState(
...
+    val id: String,
     val title: String,
     val subtitle: String,
-    val uri: Uri,
+    val storeUrl: String,
   )
 
   companion object {
-    val Loading = InsuranceUiState(
+    val InitialState = InsuranceUiState(
```
hugokallstrom, line 60: lower case variable name

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -145,23 +68,96 @@ internal data class InsuranceUiState(
...
+      crossSells = insuranceData.crossSells,
+      showNotificationBadge = showNotificationBadge,
+      quantityOfCancelledInsurances = insuranceData.quantityOfCancelledInsurances,
+      hasError = didFailToLoad == true && isLoading == false,
+      loading = isLoading,
+    )
+  }
+}
+
 private suspend fun loadInsuranceData(
```
hugokallstrom, line 147: This is beautiful

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -145,23 +68,96 @@ internal data class InsuranceUiState(
...
+  private val getCrossSellsUseCase: GetCrossSellsUseCase,
+  private val crossSellCardNotificationBadgeService: CrossSellCardNotificationBadgeService,
+) : MoleculePresenter<InsuranceScreenEvent, InsuranceUiState> {
+  @Composable
+  override fun MoleculePresenterScope<InsuranceScreenEvent>.present(
+    seed: InsuranceUiState,
+  ): InsuranceUiState {
+    var insuranceData by remember {
+      mutableStateOf<InsuranceData>(
+        InsuranceData(
```
hugokallstrom, line 82: I would create a function here like `InsuranceData.fromUiState(seed)` in order to keep this present function as clean as possible.

hugokallstrom (approved):
Very cool to see molecule in action, great job.

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
StylianosGakis: Nice catch! Turns out this didn't even need to be a `val` since we use it on initialization, like we do with the other two, so removed val completely, which does make it `private` by default too

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
StylianosGakis: Let's go with `lastState` since that's basically exactly what it is. The last state that this has ever seen. On initialization, the last state *is* technically the initial state, and from then on it's the last state that was emitted. Will fix!

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsuranceViewModel.kt
StylianosGakis: Thanks! 🤩

app/feature/insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenterTest.kt
hugokallstrom: Looks super good

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
StylianosGakis: Solved here https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1690447797256339

app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt
StylianosGakis: Discussed here https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1690447797256339

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
StylianosGakis: 🤩

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
StylianosGakis: Done

app/feature/insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
StylianosGakis: done

PR #1712, opened 2023-07-26
Hedvig theme always use new color scheme

StylianosGakis:
The new theme colors option basically was simply flipping background with surface, so this is now done in the color definition itself, and the extra HedvigTheme overload is no longer necessary.


hugokallstrom (approved):
LGTM!

PR #1714, opened 2023-07-27
Sort dependencies gradle plugin

StylianosGakis:
Running `./gradlew sortDependencies` just sorts it all in one go, no need ever think about this again 😅
Should help with not accidentally applying the same dependency twice and so on.


hugokallstrom (approved):
Great!

PR #1717, opened 2023-07-27
Forever view model and composable

hugokallstrom:
Not included: 
* List of referrals and status
* Information bottom sheet
* Piechart in compose

Will be included in future PRs


app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
```diff
@@ -217,6 +218,72 @@ private fun ForeverScreen(
...
+    coroutineScope.launch {
+      if (uiState.showEditCode) {
+        sheetState.show()
+      } else {
+        sheetState.hide()
+      }
+    }
+  }
+
+  if (uiState.showEditCode) {
```
hugokallstrom, line 235: I couldnt get this to animate properly, do you know what Im doing wrong?

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
```diff
@@ -217,6 +218,72 @@ private fun ForeverScreen(
...
+    ) {
+      Text(
+        text = stringResource(id = hedvig.resources.R.string.referrals_change_change_code),
+        textAlign = TextAlign.Center,
+        modifier = Modifier
+          .fillMaxWidth()
+          .padding(horizontal = 16.dp),
+      )
+      Spacer(Modifier.height(32.dp))
+      HedvigTextField(
```
hugokallstrom, line 252: This text field gets a white container color. Like this: 
Screenshot 2023-07-27 at 14 40 09

Any idea on how to fix that??

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
StylianosGakis: Yeah I think it's the same as I had to do here https://github.com/HedvigInsurance/android/blob/7e85f0db5781027e32700587c32c517454c95dac/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt#L57-L86

When you want it to go away, if you just remove it from composition by passing false to the `if` it will just disappear the moment the if turns false.
In the place I linked you at, when you try to dismiss it what you need to do is ask for the SheetState to dismiss, and after it's finished (hide() is suspending) then flip the flag, so it will animate out, and the moment it's already animated all the way then the if statement is flipped and it leaves composition.

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
StylianosGakis: Hmm do you meant that the text field has the exact same color as the sheet background?

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsViewModel.kt
```diff
@@ -2,72 +2,217 @@ package com.hedvig.app.feature.referrals.ui.tab
...
 ) : ViewModel() {
-  private val retryChannel = RetryChannel()
 
-  private val isLoading = MutableStateFlow(false)
+  private val _uiState = MutableStateFlow(ReferralsUiState())
+  val uiState: StateFlow<ReferralsUiState> = _uiState.stateIn(
+    viewModelScope,
+    SharingStarted.WhileSubscribed(5.seconds),
+    ReferralsUiState(),
+  )
```
StylianosGakis, line 34: In this case since `_uiState` already is a `StateFlow`, `stateIn` won't do anything, since that one usually takes a `Flow<>` and turns it into a `StateFlow<>`.
So if we are going with the normal always hot `StateFlow` in this screen we can omit the `stateIn()` and instead replace it with just `_uiState.asStateFlow()`

StylianosGakis (approved):
👌

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
StylianosGakis: In this case where the state does come from the ViewModel however, where you can't defer flipping it like it's done in that sample.
I think that makes this a bit more involved, might be a bit harder to fix, maybe we should look into it later?

hugokallstrom: I think that works, thanks!

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
hugokallstrom: Exactly

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsViewModel.kt
hugokallstrom: I see! Will probably rewrite this to using molecule after new design is in place :)

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
StylianosGakis: I think the issue is actually that the bottom sheet uses the same color as the text field.
While the text field expects the colors behind it to be `background` but the material sheet is probably using `surface`, or maybe `surfaceVariant`, which the text field may also be using.

Maybe see if you can use `surface` for the MaterialBottomSheet itself and see how that looks and go from there?

That's my assumption at least, if I'm wrong we can see what else we can do.

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsViewModel.kt
StylianosGakis: If you wanna try and do something like that together some time in the future I am 100% up for it. Gonna be a very fun exercise so we both learn how to use it better 🤩

hugokallstrom: Yeah absolutely! 👍

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/tab/ReferralsGraph.kt
StylianosGakis: Right, I meant to write "Maybe see if you can use **background**" 🤦‍♂️
Glad it worked!

PR #1720, opened 2023-07-28
Information cards data

StylianosGakis:
Fetching the information of when these cards will be shown.
For now only has data for the reminder and the connect payment, not sure about the "Will be renewed" situation. Asked here https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1690537259832729 for clarification.

I am having a really hard time coming up with a good names for the module btw, if you got any input on any of the names here please lmk.

Hooking it up to the UI will come in a separate PR.

This PR also introduces the idea of using the multi-module support of apollo-kotlin https://www.apollographql.com/docs/kotlin/advanced/multi-modules/ to have the specific query only present in the module that needs it.
Can still re-use fragments and whatever from the apollo-giraffe module and such with how this works, but should make it much more manageable if we start moving more .graphql files only in the place where they are needed instead of in 1 global space while it may only be used in 1 spot.


app/app-information/app-information-public/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
+plugins {
+  id("hedvig.android.ktlint")
+  id("hedvig.android.library")
+  alias(libs.plugins.squareSortDependencies)
+  alias(libs.plugins.apollo)
+}
+
+dependencies {
+  apolloMetadata(projects.apolloGiraffePublic)
```
StylianosGakis, line 9: This is the part that hooks up this module with giraffe, and we get access to the generated apollo classes.

hugokallstrom (approved):
LGTM! 

Regarding module name, Im thinking in terms of "member", "reminder", and "notification". `member-reminders`, `member-notifications`, or something along those lines?

StylianosGakis:
> LGTM!
> 
> Regarding module name, Im thinking in terms of "member", "reminder", and "notification". `member-reminders`, `member-notifications`, or something along those lines?

Yes, `member` that's it!
I really like `member`. And I also wanna avoid using `notification` as that can be confused with actual Android notifications, but `reminders` should be fine, so `member-reminders` seems to hit the spot, thank you!

PR #1722, opened 2023-07-28
Refactor forever module. Change naming from Referral to Forever.

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -110,7 +110,7 @@ internal class HedvigAppState(
       listOfNotNull(
         TopLevelGraph.HOME,
         TopLevelGraph.INSURANCE,
-        TopLevelGraph.REFERRALS.takeIf { isReferralsEnabled },
+        TopLevelGraph.FOREVER.takeIf { isReferralsEnabled },
```
StylianosGakis, line 113: Might be able to rename the feature flag locally too, so it's `isForeverEnabled` and `featureManager.isFeatureEnabled(Feature.FOREVER)` but I know it's always hard to do renames, we'll definitely miss a thing or two here and there.

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ForeverViewModel.kt
```diff
@@ -16,14 +17,14 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 
-class ReferralsViewModel(
-  private val referralsRepository: ReferralsRepository,
+class ForeverViewModel(
```
StylianosGakis, line 20: Most likely can be `internal` now too.

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ForeverViewModel.kt
```diff
@@ -93,7 +94,7 @@ class ReferralsViewModel(
   }
 }
 
-data class ReferralsUiState(
+data class ForeverUiState(
```
StylianosGakis, line 97: Also probably can be `internal`.
Likely that almost everything in the entire module can be internal now, aside from the koin module and the foreverGraph().
Maybe a thing or two more too, but those for sure.
Doesn't have to be part of this PR necessarily, we can alway do it as we see those things in the future.

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/di/ForeverModule.kt
```diff
@@ -1,22 +1,20 @@
...
-import com.hedvig.android.feature.forever.GetReferralsInformationUseCase
-import com.hedvig.android.feature.forever.ReferralsViewModel
-import com.hedvig.android.feature.forever.data.ReferralsRepository
-import com.hedvig.android.language.LanguageService
+import com.hedvig.android.feature.forever.ForeverViewModel
+import com.hedvig.android.feature.forever.data.GetReferralsInformationUseCase
 import org.koin.androidx.viewmodel.dsl.viewModel
 import org.koin.dsl.module
 
 val foreverModule = module {
```
StylianosGakis, line 9: Could we check if the entire `foreverModule` can move inside the forever feature?
This is the thing that would allow everything else like the ViewModels and so on to be internal from that point on, and :app doesn't need to know about any of them.

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/EditCodeBottomSheet.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.button.HedvigContainedButton
+import com.hedvig.android.core.designsystem.component.button.HedvigTextButton
+import com.hedvig.android.core.designsystem.component.textfield.HedvigTextField
+import hedvig.resources.R
+
+@Composable
+fun EditCodeBottomSheet(
```
StylianosGakis, line 22:
```suggestion
internal fun EditCodeBottomSheet(
```

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ForeverContent.kt
```diff
@@ -0,0 +1,165 @@
...
+  LaunchedEffect(uiState.showEditCode) {
+    coroutineScope.launch {
+      if (uiState.showEditCode) {
+        sheetState.expand()
+      } else {
+        sheetState.hide()
+      }
+      showEditBottomSheet = uiState.showEditCode
+    }
+  }
```
StylianosGakis, line 84: Doesn't this make it so that when the sheet comes in it just pops in instead of animating in? Since if `showEditCode` is true, it will first try to animate the state (at that point `showEditBottomSheet` is still false, so no UI is showing) and then it sets it to true, so the bottom sheet shows, already in the expanded state?
That's what I see from reading the code at least.

That's why I thought this is a bit tricky to do with the UI state determining the sheet state while also having to have a local representation of it too. If I am wrong ignore this. If I am right we can look into this later I think.

StylianosGakis (approved):
lgtm!

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ForeverContent.kt
```diff
@@ -67,6 +71,7 @@ internal fun ForeverContent(
     onRefresh = reload,
     refreshingOffset = PullRefreshDefaults.RefreshingOffset + systemBarInsetTopDp,
   )
+  val locale = uiState.locale ?: Locale.ENGLISH
```
StylianosGakis, line 74: We can do `getLocale()` inside a composable https://github.com/HedvigInsurance/android/blob/7e85f0db5781027e32700587c32c517454c95dac/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/GetLocale.kt#L12 which is always non-null, should be easier to use here.

StylianosGakis: Which makes me think, why do we have a nullable Locale here in the first place? maybe from the old code? `LanguageService.getLocale()` also does not return a nullable locale, so we shouldn't even have this issue.
But yeah, since we can get locale in the composable this way, I think you can probably omit the Locale from inside the UiState completely

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/di/ForeverModule.kt
hugokallstrom: `ForeverRepository` is still used by payments and offer, is that okay or should we create separate use cases for those features?

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ForeverContent.kt
hugokallstrom: Yes that is true. Lets figure this out together!

PR #1724, opened 2023-07-31
Add list of referrals with status and discount

hugokallstrom:


app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ReferralList.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+    Row(
+      modifier = Modifier
+        .padding(vertical = 16.dp)
+        .fillMaxWidth(),
+      horizontalArrangement = Arrangement.SpaceBetween,
+    ) {
+      Text(text = stringResource(id = R.string.FOREVER_TAB_TOTAL_DISCOUNT_LABEL))
+      Row {
+        CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.onSurfaceVariant) {
```
hugokallstrom, line 75: Not sure if this is the correct way of setting the text color? It should be gray, but im not getting that with this currently

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ReferralList.kt
```diff
@@ -0,0 +1,183 @@
...
+import com.hedvig.android.core.ui.getLocale
+import com.hedvig.android.feature.forever.ForeverUiState
+import hedvig.resources.R
+import javax.money.CurrencyContext
+import javax.money.CurrencyUnit
+import org.javamoney.moneta.Money
+
+@Composable
+internal fun ReferralList(uiState: ForeverUiState) {
+  if (uiState.referrals.isEmpty()) {
```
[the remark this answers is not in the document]
StylianosGakis: Yeah this works. It also works to have on the call site instead something like
```
if (uiState.referrals.isNotEmpty()) {
  ReferralList(uiState)
}
```
So that it's not called in the first place at all.
I don't know if there's any *good* reason to chose one over the other in this particular situation, it may just be a stylistic choice, in which case do whatever feels right to you.

The one scenario where doing the if check on the call site would be for sure better is if we need to know there that if it shows or not, something else should change. For example
```
@Composable
fun Foo() {
  SomeText()
  if (someList.isNotEmpty()) {
    ReferralsList()
  } else {
    Spacer()
    Divider()
  }
}
```
Or something like that.

In that case it's important for it to be on the call site, since the alternative would be
```
@Composable
fun Foo() {
  SomeText()
  ReferralsList()
  if (someList.isEmpty()) {
    Spacer()
    Divider()
  }
}
```
And it would not be immediately obvious that if the list is not empty ReferralsList() shows, but if it's empty only then the spacer + divider shows.

I hope I am getting my point across here :D

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ReferralList.kt
StylianosGakis: Ah yeah this is interesting.

So Text() has somewhat of an order in which it looks at to know what color it should have.

Text composable has a `color: Color = Color.Unspecified,` parameter, which you can pass a color directly in it.
Then, TextStyle also has a `color: Color = Color.Unspecified,` inside of it.
Finally, we got LocalContentColor which also is a color.

If you don't provide it anything `Text("someText")`, it will try and use the LocalTextStyle and see if it has a color inside there. If it does not, then it will look inside `LocalContentColor.current`.
Our styles [here](https://github.com/HedvigInsurance/android/blob/7e85f0db5781027e32700587c32c517454c95dac/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigTypography.kt#L23) don't provide any style-specific colors, so by default TextStyle will never have color in our app at least.

Then you can call your text like this
```
Text(
 "foo",
 style = LocalTextStyle.current.copy(color = someColor)
)
```
then since there is a color inside textStyle, it won't fall down to using LocalContentColor, and use the color inside the style directly.

Finally, if you call your text like this:
```
Text(
  "foo",
  color = someColor,
)
```
Then that takes precedence over everything,

To summarize this:
```
CompositionLocalProvider(LocalContentColor provides Color.Yellow) {
  Text(
    text = "foo",
    color = Color.Red,
    style = LocalTextStyle.current.copy(color = Color.Blue)
  )
}
```
Will *not* use the local content color, since a style color exists.
Will *not* use the style color, since a specific color for this particular text composable exists
*Will* use Color.Red since that's the specific color passed to this composable.

It's done this way so that if you don't have any styles, it at least falls back to LocalContentColor so that it's light on dark surfaces, and dark on light surfaces.
And so that you can define a style-wide color for the entire app if you wish using TextStyle.
And you can also pass specific ones for specific composables you wish.

tl;dr: For these situations you most likely just wanna do:
```
Text(
  text = "",
  color = MaterialTheme.colorScheme.onSurfaceVariant,
)
```

And in situations where you've got a whole section which should have these colors, just do:
```
CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.onSurfaceVariant) {
  Column {
    Text("I am gray")
    Text("I am also gray")
    Text("I am also gray")
    Text("I am also gray")
  }
}
```

StylianosGakis (approved):
Wrote a little story on each comment, sorry 😅
Lgtm though, fell free to do whatever you wish for question number 1

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ReferralList.kt
hugokallstrom: Awesome, thanks 💯

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/ReferralList.kt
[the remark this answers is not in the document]
hugokallstrom: Thats a good point. I also think its good to know on the call site if the function will actually run or not in this case.

PR #1725, opened 2023-07-31
payment and profile feature module

StylianosGakis:
This PR is only moving things around, so that now Profile feature is all in a separate module.

In this process, a payment module was also introduced to host `PaymentRepository`.
Extracted `NetworkCacheManager.clearCache` so feature modules can also use it.

Added a module at dir data/dataForever was also introduced to hold the information regarding the forever feature, which is needed in more than just the Forever feature module which contains the main UI that interacts with that information.


hugokallstrom (approved):
LGTM!

PR #1727, opened 2023-08-01
Import contract image resources and map to contract type

hugokallstrom:
Also use smaller titles on root destinations

Screenshot_20230801_121431


app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -126,15 +123,20 @@ private fun ProfileScreen(
         .matchParentSize()
         .pullRefresh(pullRefreshState)
         .verticalScroll(rememberScrollState())
-        .windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal)),
+        .windowInsetsPadding(WindowInsets.safeDrawing),
```
StylianosGakis, line 126: This change here also means that it's applying the bottom padding in the entire content here. But we have `      Spacer(Modifier.windowInsetsBottomHeight(WindowInsets.safeDrawing))` also in here which means that it will be applied 2 times afaik.
Maybe if we need the top padding here too, we can instead use 
```suggestion
        .windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal + WindowInsetsSize.Top),
```
Or remove the bottom padding from line 162.

StylianosGakis (approved):

PR #1729, opened 2023-08-01
Set correct payment button label if no payment is connected

hugokallstrom:


app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/payment/PaymentViewModel.kt
```diff
@@ -114,23 +114,25 @@ private fun PaymentData.toUiState(locale: Locale): PaymentViewModel.PaymentUiSta
...
+      null
+    } else {
+      bankAccount?.let {
+        PaymentViewModel.PaymentUiState.PaymentMethod(
+          displayName = it.name,
+          displayValue = it.accountNumber,
+        )
+      } ?: PaymentViewModel.PaymentUiState.PaymentMethod(
+        displayName = when (paymentMethod) {
+          is PaymentData.PaymentMethod.CardPaymentMethod -> paymentMethod.brand ?: "Unknown"
```
StylianosGakis, line 127: Do we need "Unknown" localized?

StylianosGakis (approved):

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/payment/PaymentViewModel.kt
hugokallstrom: This ~should~ never happen so I think its ok to not have it localized. We can look at handling this better when we move the API to octopus

PR #1730, opened 2023-08-01
Do not show title for logout dialog. Change dismiss button label.

hugokallstrom:


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/dialog/HedvigAlertDialog.kt
```diff
@@ -12,19 +12,23 @@ import hedvig.resources.R
...
 @Composable
 fun HedvigAlertDialog(
-  title: String,
+  modifier: Modifier = Modifier,
+  title: String?,
   text: String,
+  confirmButtonLabel: String = stringResource(R.string.GENERAL_YES),
+  dismissButtonLabel: String = stringResource(R.string.GENERAL_NO),
   onDismissRequest: () -> Unit,
   onConfirmClick: () -> Unit,
```
StylianosGakis, line 21:
```suggestion
  title: String?,
  text: String,
  onDismissRequest: () -> Unit,
  onConfirmClick: () -> Unit,
  modifier: Modifier = Modifier,
  confirmButtonLabel: String = stringResource(R.string.GENERAL_YES),
  dismissButtonLabel: String = stringResource(R.string.GENERAL_NO),
```
So that Modifier is the first parameter which does not have a default parameter, as the compose API guidelines (and the built-in IDE lint check) require.

StylianosGakis (approved):

PR #1731, opened 2023-08-01
urls.xml in resources module 

hugokallstrom:


app/core/core-resources/src/main/res/values/urls.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="GRAPHQL_URL" translatable="false">https://graphql.dev.hedvigit.com/graphql</string>
+    <string name="WS_GRAPHQL_URL" translatable="false">wss://graphql.dev.hedvigit.com/subscriptions</string>
+    <string name="OCTOPUS_GRAPHQL_URL" translatable="false">https://apollo-router.dev.hedvigit.com</string>
+    <string name="BASE_URL" translatable="false">https://graphql.dev.hedvigit.com</string>
+    <string name="WEB_BASE_URL" translatable="false">https://www.dev.hedvigit.com</string>
+    <string name="HANALYTICS_URL" translatable="false">https://hanalytics.dev.hedvigit.com</string>
+    <string name="ODYSSEY_URL" translatable="false">https://odyssey.dev.hedvigit.com</string>
+</resources>
```
StylianosGakis, line 10: Unfortunately this won't work, as we're relying on the string values to exist inside the `debug` directory so that we get different one for prod/staging/dev.
I can make a PR on top of this with a fix for this actually, since I wanted to fix this anyway recently, so this is a good opportunity.

My idea will be that we can have one class named something like `HedvigAppInfo` or something like that (name suggestions welcome), and inside there we can hold these strings, along with the app version code/ app version name which are used in the "About app" part of the app.
Would be nice to get it all in one place basically in one nice object.
And then :app module can create this object by looking at the resources from debug/staging/develop when appropriate, and all other modules can just fetch it from Koin.

But yeah, we must do that, otherwise this will be resolving to wrong strings in production

StylianosGakis (changes requested):
Basically for what I describe in the comment, I will work on a PR on top of this to fix it though

app/core/core-resources/src/main/res/values/urls.xml
hugokallstrom: I see, good catch. What about `HedvigBuildInfo` or `HedvigBuildConstants`?

PR #1732, opened 2023-08-01
Add pie chart in forever screen

hugokallstrom:


app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/DiscountPieChart.kt
```diff
@@ -0,0 +1,100 @@
...
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.material3.typeElement
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+
+@Composable
+fun ColumnScope.DiscountPieChart(
+  containerColor: Color = MaterialTheme.colorScheme.typeElement,
+  discountColor: Color = MaterialTheme.colorScheme.secondaryContainer,
+  incentiveColor: Color = MaterialTheme.colorScheme.surface,
```
hugokallstrom, line 30: Not sure about these colors, I think we want a darker green for the discount color here

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/DiscountPieChart.kt
```diff
@@ -0,0 +1,100 @@
...
+    animationSpec = InfiniteRepeatableSpec(
+      animation = KeyframesSpec(
+        KeyframesSpec.KeyframesSpecConfig<Float>().apply {
+          0f at 800 with FastOutSlowInEasing
+          30f at 1300
+          30f at 2800 with FastOutSlowInEasing
+          durationMillis = 2800
+        },
+      ),
+      repeatMode = RepeatMode.Reverse,
```
StylianosGakis, line 48: I feel like maybe we don't need to have the animation reverse here but just have it be `RepeatMode.Restart`.
that way we can specify the animation completely in one shot using keyframes again perhaps, and then let it replay from the start again.

something like
```
animation = keyframes {
  durationMillis = 5400
  0f at 0
  0f at 1800
  30f at 2300 with SomeEasingHere
  30f at 3600
  0f at 4100 with SomeEasingHere
  0f at 5400
},
```
So the animations that play from 1800 -> 2300 get the easing, same with 3600 -> 4100

StylianosGakis (approved):
Just a thought I had for using the repeating animation + the Easings which I *think* they don't apply as you expect them to do here. If I got smth wrong please lmk!

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ui/DiscountPieChart.kt
hugokallstrom: The easings actually have to be on `0f at 1800` and `30f at 3600` to have an effect here. But this looks much better, thanks!

PR #1735, opened 2023-08-04
Fix logging for both android and jvm only modules. Delegating to slimber as we used to do already.

StylianosGakis:
All logs now simply go through
```
inline fun logcat(
  priority: LogPriority = LogPriority.DEBUG,
  throwable: Throwable? = null,
  noinline message: () -> String,
)
```
This https://github.com/HedvigInsurance/android/commit/24605197b601c036470494de29875169028668d5 is what adds this functionality, the rest of the commits are to migrate the rest of the app to use this.


hugokallstrom (approved):
Really clean! LGTM!

PR #1736, opened 2023-08-04
Redesign/smooth corners

StylianosGakis:
Now we can actually specify corner radius as it comes from Figma, using the corner math from https://github.com/c5inco/smoother

This also now allows us to easily specify top-only corner for the bottom sheets as the design specified.


hugokallstrom (approved):
LGTM!

PR #1737, opened 2023-08-07
Redesign/text color palette

StylianosGakis:
WIP


StylianosGakis:
Didn't need to go that route afterall

PR #1738, opened 2023-08-09
Redesign member reminders UI in Profile + Profile>Settings

StylianosGakis:
image
image


app/core/core-datastore-public/src/main/kotlin/com/hedvig/android/core/datastore/SettingsDataStore.kt
```diff
@@ -8,24 +8,22 @@ import com.hedvig.android.theme.Theme
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
-interface SettingsDataStore {
-  suspend fun setTheme(theme: Theme)
-  fun observeTheme(): Flow<Theme>
-}
-
-internal class SettingsDataStoreImpl(
+class SettingsDataStore(
```
StylianosGakis, line 11: Testing this by faking the datastore storage https://github.com/HedvigInsurance/android/blob/feaf403b4550d25d005ff5bacc33e52a684faaf8/app/feature/feature-profile/src/test/kotlin/com/hedvig/android/feature/profile/settings/SettingsPresenterTest.kt#L25-L30 instead of faking the UseCase itself was more convenient here.

Not always sure which way to go with these, but testing the real implementation with a fake in-memory DataStore is really convenient, and you're testing with the real implementation. In the tests I used it for here it was convenient enough that not having an interface to create a test double here made sense.

app/core/core-datastore-public/src/main/kotlin/com/hedvig/android/core/datastore/SettingsDataStore.kt
```diff
@@ -8,24 +8,22 @@ import com.hedvig.android.theme.Theme
...
     dataStore.edit {
       it[themeKey] = theme.name
     }
   }
 
-  override fun observeTheme(): Flow<Theme> {
-    return dataStore.data.map { it[themeKey]?.let { themeString -> Theme.valueOf(themeString) } ?: Theme.LIGHT }
+  fun observeTheme(): Flow<Theme> {
+    return dataStore.data.map {
+      it[themeKey]?.let { themeString -> Theme.valueOf(themeString) } ?: Theme.SYSTEM_DEFAULT
```
StylianosGakis, line 22: Thought `Theme.SYSTEM_DEFAULT` would make more sense as a default, until we decide what happens on the first release of this. In any case, this is now not really hooked up with actually changing the app theme, selecting a new theme doesn't actually change the UI. So if we do in fact release without this being an option the better option would be to not even have this button in the first place. So lets keep the implementation here realistic, aka supporting dark mode, and if we do not have dark mode on for the release, let's just hide the button instead.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/settings/NotifyBackendAboutLanguageChangeUseCase.kt
```diff
@@ -3,21 +3,19 @@ package com.hedvig.android.feature.profile.settings
...
   private val cacheManager: NetworkCacheManager,
-) {
-  suspend fun invoke(language: Language) {
-    apolloClient
-      .mutation(
-        UpdateLanguageMutation(language.toString(), language.toLocale()),
-      ).safeExecute()
+) : NotifyBackendAboutLanguageChangeUseCase {
+  override suspend fun invoke(language: Language) {
+    apolloClient.mutation(UpdateLanguageMutation(language.toString(), language.toLocale())).safeExecute()
```
StylianosGakis, line 18: We were doing `languageService.setLanguage(language)` here, but this was bringing some confusion as we were as some point calling the languageService function both from this UseCase and calling the real one too.
I figured it's better to just split those up, and call each one separately. This is also important as we do want to make sure we change the language locally first and have that apply instantly, and then do a best effort call to the backend too. Because to be honest I am not even sure why we're doing this mutation in the first place, probably legacy reasons as we should be appending a language header on every invocation anyway, informing the backend for this doesn't make sense to keep for the future.

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
```diff
@@ -0,0 +1,228 @@
...
+  openUrl: (String) -> Unit,
+  notificationPermissionState: NotificationPermissionState,
+  snoozePermission: () -> Unit,
+  modifier: Modifier = Modifier,
+) {
+  Column(
+    modifier = modifier,
+    verticalArrangement = Arrangement.spacedBy(8.dp),
+  ) {
+    AnimatedVisibility(
```
hugokallstrom, line 60: A MemberReminder could then contain the data needed to construct its own ReminderCard. Then we would only need to add new MemberReminders, we have a generic way of creating the cards.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/settings/SettingsPresenter.kt
```diff
@@ -0,0 +1,90 @@
...
+import androidx.compose.runtime.setValue
+import com.hedvig.android.core.datastore.SettingsDataStore
+import com.hedvig.android.language.LanguageService
+import com.hedvig.android.market.Language
+import com.hedvig.android.memberreminders.EnableNotificationsReminderManager
+import com.hedvig.android.molecule.public.MoleculePresenter
+import com.hedvig.android.molecule.public.MoleculePresenterScope
+import com.hedvig.android.theme.Theme
+
+internal class SettingsPresenter(
```
hugokallstrom, line 17: Awesome!

hugokallstrom (approved):

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
```diff
@@ -0,0 +1,228 @@
...
+import kotlinx.datetime.Clock
+import kotlinx.datetime.LocalDate
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.atStartOfDayIn
+import kotlinx.datetime.daysUntil
+import kotlinx.datetime.toLocalDateTime
+
+@Composable
+fun MemberReminderCards(
+  memberReminders: ApplicableMemberReminders,
```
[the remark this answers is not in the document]
StylianosGakis: Yes, and I did actually have it as a list at first.
Then what happened is that the reminder for UpcomingRenewals itself also could have a list of reminders, so it was a list in a list.
Then as I was trying to test this and use this, I found myself having to take care of things like the list containing 2 instances of the `EnableNotifications` reminder for example, which logically makes no sense.

And then in the UI, I didn't find a good way to make the notification items come into the UI in a smooth way. If they were not there on the initial frame, once the UI state updated and had them in the list they would just pop up instantly.
When I made it animate in, it was *always* animating in, even if you went back to the previous screen which should already have them in the first frame (going back to profile from settings for example).

Then doing this approach, it allowed me to just use AnimatedVisibility around each one, so that if the first frame does not have it, it doesn't show, and when it comes in, it animates in using the AnimatedVisibility animationSpec. If however you come back to it, if it already was there, it just shows fully from the start, without doing an unnecessary animation to come in again.

It's definitely possible that there was another way to approach this which would not have this issues, but as I was playing with it this was the one that ticked all the boxes of what I was going for.

If you wanna play with this and see if you can improve it absolutely feel free to do so of course

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
StylianosGakis: I was thinking that we're most likely not going to be adding many (if any) new reminders here, plus what I wrote in the comment above is what made me write it this way.
There still is a generic way to render a card, with the `MemberReminderCards` composable, which in order for it to be able to render any card, it needs to know how to handle the data needed + callbacks for them for all of them.

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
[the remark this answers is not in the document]
hugokallstrom: Thats what I thought :) in that case this solution makes sense!

PR #1742, opened 2023-08-16
Full screen error

hugokallstrom:


StylianosGakis (approved):
Super nice, thanks for clearing this up.
It may have tiny differences compared to the previous implementation, perhaps with some slight vertical padding etc.
But as long as nothing looks too odd (like things touching the edges etc. which is not the case here since HedvigErrorSection takes care of it by itself), and the retry functionality works then we should be good.

PR #1743, opened 2023-08-16
Add debounced loading indicator. Animate between states.

hugokallstrom:


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
```diff
@@ -132,60 +132,58 @@ private fun InsuranceScreen(
...
-            .fillMaxWidth()
-            .padding(horizontal = 16.dp),
-        ) {
-          Text(
-            text = stringResource(id = R.string.DASHBOARD_SCREEN_TITLE),
-            style = MaterialTheme.typography.titleLarge,
-          )
-        }
+      AnimatedContent(targetState = uiState.isLoading, label = "") { isLoading ->
         Spacer(Modifier.height(16.dp))
```
StylianosGakis, line 141: Since this Spacer now is just the child of the AnimatedContent (and not of the column as it was before) I think this does nothing now, so you can remove it.
If we still need this space however, it'd need to be added to the column below here in the isLoading -> false situation or something like that, can't really tell easily without looking at the emitted UI.

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/progress/HedvigFullScreenCenterAlignedProgress.kt
```diff
@@ -29,6 +37,25 @@ fun HedvigFullScreenCenterAlignedProgress(
   }
 }
 
+@Composable
+fun HedvigFullScreenCenterAlignedProgressDebounced(
+  modifier: Modifier = Modifier,
+  show: Boolean = true,
+  debounceMillis: Long = DEBOUNCE_MILLIS,
+) {
+  var debounce by remember { mutableStateOf(false) }
```
StylianosGakis, line 46: ok now I understood what this was for, initially I thought the debounce was for the opposite effect, that it should still show for X milliseconds even if show turns false.
So I understood the flipped thing of what this actually does. So this looks good yeah, now that I understood it. Sorry 😅

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileViewModel.kt
```diff
@@ -66,5 +66,5 @@ internal data class ProfileUiState(
   val euroBonus: EuroBonus? = null,
   val showPaymentScreen: Boolean = false,
   val memberReminders: MemberReminders = MemberReminders(),
-  val isLoading: Boolean = true,
+  val isLoading: Boolean = false,
```
StylianosGakis, line 69: This feels like a quite significant change, the initial state now has loading as false when you first open the screen. Is this expected? I don't think we set this to true ever either from looking inside this class.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/start/TerminationStartStepViewModel.kt
```diff
@@ -52,7 +52,7 @@ internal class TerminationStartStepViewModel(
 }
 
 internal sealed interface TerminationFlowUiState {
-  object Loading : TerminationFlowUiState
-  object Error : TerminationFlowUiState
+  data object Loading : TerminationFlowUiState
+  data object Error : TerminationFlowUiState
```
StylianosGakis, line 56: Nice, thanks for cleaning these up too 😊

StylianosGakis (approved):
lgtm! Just a comment about the default false loading state for Profile ViewModel, I am not sure if that change is on purpose or not 👀

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileViewModel.kt
hugokallstrom: Ill change it back!

PR #1745, opened 2023-08-17
Claim detail redesign

hugokallstrom:
Screenshot_20230817_120248


StylianosGakis (approved):
Looking great!
I will fix my PR on top of this one since you name the composable `ClaimInfoCard` now, and I also realize I did end up changing the parameter it takes a little bit in my PR. So this can be merged freely and I'll fix my changes to conform with this.

PR #1751, opened 2023-08-22
Feature/claim card fixes

StylianosGakis:
Remove the arrow, remove elevation and move the cards closer, make the progress bars circular, make the divider match the design color more closely

image


hugokallstrom (approved):
👌

PR #1752, opened 2023-08-22
Fix bottom and navigation rail to new designs

StylianosGakis:
Make them both use the surface color, with a hairline width divider-like line where the split is between them and the content. Also fix a bit how the insets were consumed in the navigation rail scenario, where the display cutout inset was not consumed. https://issuetracker.google.com/issues/296990873

Also fix cutout insets for the top app bar icons

image
image


hugokallstrom (approved):
LGTM!

PR #1753, opened 2023-08-28
Create Hedvig lint rules

StylianosGakis:
This sets us up to be able to write our own lint checks.

What we get is basically this from applying the Material2Detector check.
image
Free lint checks that are in the IDE itself, you don't need to run anything. It's the same way that lint checks happen for things like using non translated strings in your code and so on. They run in Android Studio itself automatically.

All this basically comes from here https://github.com/slackhq/compose-lints and we will in the future be able to bring in even more of those, but for now just getting it all to work first.

The long term goal is to bring in more such lints, as a first step to tackle for example compose problems like here https://twitter.com/mrmans0n/status/1507390768796909571 and going forward doing our own checks too if we wish


StylianosGakis:
The lint error here is OK since this https://github.com/HedvigInsurance/android/pull/1754 lint check passes, as over there we've setup the baseline of what errors we're ok with passing through

hugokallstrom (approved):
Love it, great initiative.

PR #1754, opened 2023-08-28
Remove all m2 references

StylianosGakis:
Removes all references in our app to any material2 dependencies (except for inside PickMarketScreen.kt since that one will be redesigned soon anyway so it was not worth the effort)

Does so by moving to m3 equivalents everywhere.

We removed the entire "Business model" feature, not present in the new redesign anyway.

For pull to refresh functionality which only exists in m2 for now, we simply brought in the core code to our project in the :pullrefresh module. It had almost no references to m2 anyway, it's just stuck in there for reasons. https://issuetracker.google.com/issues/261760718

Removes the entire HedvigMaterial2Theme

Makes use of https://github.com/HedvigInsurance/android/pull/1753 so that we're sure we have no references to m2, and we don't after this PR.


hugokallstrom (approved):
Amazing cleanup 🚀

PR #1755, opened 2023-08-29
Use androidx.graphics to generate the Figma shape

StylianosGakis:
After the discussion from here
https://github.com/c5inco/smoother/issues/1#issuecomment-1692445500 I learned about the new androidx.graphics:graphics-shapes library which seems to do exactly what we're looking for here with a very convenient and easy to use API.

FigmaShape is replaced entirely with a simple call to `RoundedPolygon` with the four corner locations and a CornerRounding configuration with our required radius and 100% smoothing as Figma specifies.

Example shape with 12.dp and 100% smoothing
image


hugokallstrom (approved):
Very nice!

PR #1765, opened 2023-09-13
Remove notification reminder from the home screen

StylianosGakis:
We simply do not want to show this specific reminder in the home screen


hugokallstrom (approved):
LGTM!

PR #1768, opened 2023-09-18
Set Kotlin JVM target

martinbonnin:
Makes it possible to build with Java > 17


StylianosGakis:
Thanks a lot!
Addressed here https://github.com/HedvigInsurance/android/pull/1769

PR #1770, opened 2023-09-19
Remove dark mode

StylianosGakis:
LoggedInActivity loads the feature flag and sets light mode before we ever allow the splash screen to leave, or go to the marketing screen. While it's still being loaded from network or whatever, the splash screen still shows through the use of `installSplashScreen().setKeepOnScreenCondition`.

In the settings screen, if the feature flag is on, we simply do not show the option to change the theme at all, and we also do **not** report back to the system through `AppCompatDelegate.setDefaultNightMode`, so if the flag is off we do not change the theme ever from the profile screen.


hugokallstrom (approved):
Looks great

PR #1771, opened 2023-09-19
Feature/chat feature flag

StylianosGakis:
We show the compose view on top of everything to hide the chat behind it completely. The `Surface` with the background color ensures that this is the case.
As soon as we have evaluated the flag, if it disables the chat, we show the error section. If it does not disable the chat, we completely hide the composable by making it `gone`.

The first commit hides the chat icon from the Home Screen too. This is basically removing the main entry point to the chat. People can still enter the chat through among other places:
* The claim status detail screen
* A notification coming from chat
* Some error screens which redirect to the chat

I figured this set of changes are the best balance of effort to what we get out of it with this. Disabling the chat button in all possible places in the app would be a basically impossible task. Disabling the chat itself is the most important and mandatory part, hiding the chat icon too is just a little extra touch.

image

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
+++ b/app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
@@ -5,6 +5,8 @@ import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.hedvig.android.core.common.RetryChannel
+import com.hedvig.android.hanalytics.featureflags.FeatureManager
+import com.hedvig.android.hanalytics.featureflags.flags.Feature
 import com.hedvig.android.logger.LogPriority
 import com.hedvig.android.logger.logcat
 import com.hedvig.app.feature.chat.data.ChatEventStore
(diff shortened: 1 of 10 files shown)


app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
```diff
@@ -17,24 +19,42 @@ import giraffe.GifQuery
...
 
   private val retryChannel = RetryChannel()
 
   private val _messages = MutableStateFlow<ChatMessagesQuery.Data?>(null)
   val messages = _messages.asStateFlow()
 
+  /**
+   * [null] implies that it's still not evaluated. In those situations, not showing the chat at all is preferred.
+   */
+  val isChatDisabled: StateFlow<Boolean?> = flow {
```
hugokallstrom, line 50: I think we still want to enable the chat if we cant evaluate the feature flag

hugokallstrom (approved):

app/app/src/main/kotlin/com/hedvig/app/feature/chat/viewmodel/ChatViewModel.kt
StylianosGakis: Yes, this is very important!
That's why the default value we've set in Unleash is "false".

What this will do is:
* Go here https://github.com/HedvigInsurance/android/blob/dbcb4ab8e582945584f730612c357e3204303117/app/hanalytics/hanalytics-core/src/main/kotlin/com/hedvig/android/hanalytics/HAnalyticsExperimentManager.kt#L24-L32
* If the experiments are not populated, ask the backend (at this point in the chat, this is basically never the case, we always have loaded experiments, but it'd work even if we did not)
* If it's in there, it will see what it was, otherwise, the generated code in HAnalytics looks like this:
```kotlin
suspend fun disableChat(): Boolean {
  try {
    val experiment = getExperiment("disable_chat")
    experimentEvaluated(experiment)

    return experiment.variant == "enabled"
  } catch (e: Exception) {
    experimentEvaluated(
      HAnalyticsExperiment(
        "disable_chat",
        "disabled",
      )
    )

    return false
  }
}
```
It will catch the exception if it was not in there, and default to `false`, aka the chat is *not* disabled.

So this should be fine because of this.

StylianosGakis: But despite all this, at this point in the app, there's also no way that we've still not cached the experiments, so this should be a super fast evaluation in all cases.

hugokallstrom: I see, sounds good!

PR #1775, opened 2023-09-19
Finalize moving flow

hugokallstrom:


app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressOfferDestination.kt
```diff
@@ -51,64 +48,72 @@ import androidx.compose.ui.tooling.preview.Preview
...
+    imageLoader = imageLoader,
   )
 }
 
 @Composable
 private fun ChangeAddressOfferScreen(
   uiState: ChangeAddressUiState,
   openChat: () -> Unit,
-  navigateBack: () -> Unit,
+  close: () -> Unit,
```
StylianosGakis, line 111: Regarding the `close` name change in some places.
Usually `navigateUp` is used explicitly in the top-left icon in screens, so that we can call `navController.navigateUp()`, which basically pops the backstack, but ensures we do not leave the app by doing so.
The name `close` kinda loses this meaning, is it that these screens are not possible to go back from? Is going back supposed to be a non allowed step in this process of moving? Could we in some way preserve being able to go back in those screens?

If there's a good reason why back doesn't work (and in this case I assume pressing the normal back android button/gesture also needs to do a custom thing) then we could rename this to something a bit more descriptive than close?

But since I don't see any `BackHandler` calls here in the first place, maybe we don't need to break back handling from the top left button? Please fill me in here.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressOfferDestination.kt
```diff
@@ -433,12 +361,15 @@ private fun PreviewChangeAddressOfferScreen() {
...
           movingDate = ValidatedInput(Clock.System.now().toLocalDateTime(TimeZone.UTC).date),
+          extraBuildingTypes = emptyList(),
         ),
         {},
         {},
         {},
         {},
         {},
+        {},
+        imageLoader = PreviewImageLoader(LocalContext.current),
```
StylianosGakis, line 372: Can use just `rememberPreviewImageLoader()` to not have to call the LocalContext
```suggestion
        imageLoader = rememberPreviewImageLoader(),
```

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressResultDestination.kt
```diff
@@ -1,31 +1,48 @@
...
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.button.HedvigTextButton
+import com.hedvig.android.core.designsystem.component.success.HedvigSuccessSection
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
 
 @Composable
 internal fun ChangeAddressResultDestination(
-  navigateUp: () -> Unit,
+  movingDate: String?,
```
StylianosGakis, line 19: Is it possible to have this be kotlinx.datetime.LocalDate?
Or does this come 100% straight from backend, so we don't want to deal with any conversions? If that is the case, then since I see we only use it to put it in a string, then we can keep it this way.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressSelectHousingTypeDestination.kt
```diff
@@ -1,55 +1,57 @@
...
   viewModel: ChangeAddressViewModel,
   navigateUp: () -> Unit,
-  onHousingTypeSubmitted: () -> Unit,
+  navigateToEnterNewAddressDestination: () -> Unit,
 ) {
   val uiState: ChangeAddressUiState by viewModel.uiState.collectAsStateWithLifecycle()
+
+  LaunchedEffect(uiState.moveIntentId) {
+    if (uiState.moveIntentId != null) {
+      navigateToEnterNewAddressDestination()
```
StylianosGakis, line 44: This doesn't also make sure to clear the `moveIntentId` in the ViewModel after it's consumed this state? Is this perhaps why we're also disabling back handling? Or am I wrong here?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/ChangeAddressSelectHousingTypeDestination.kt
```diff
@@ -70,43 +73,50 @@ private fun ChangeAddressSelectHousingTypeScreen(
...
       text = stringResource(hedvig.resources.R.string.CHANGE_ADDRESS_SELECT_HOUSING_TYPE_TITLE),
-      style = MaterialTheme.typography.headlineSmall,
+      style = MaterialTheme.typography.headlineMedium,
       textAlign = TextAlign.Center,
       modifier = Modifier
         .fillMaxWidth()
         .padding(horizontal = 16.dp),
     )
-    Spacer(modifier = Modifier.height(64.dp))
+    Spacer(modifier = Modifier.weight(1f))
```
StylianosGakis, line 97: Maybe also need some amount of space here for the "tiny screen" scenario?
```suggestion
    Spacer(modifier = Modifier.weight(1f))
    Spacer(modifier = Modifier.height(16.dp))
```

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/navigation/ChangeAddressGraph.kt
```diff
@@ -49,9 +52,27 @@ fun NavGraphBuilder.changeAddressGraph(
         navController = navController,
         backStackEntry = navBackStackEntry,
       )
+      BackHandler {
+        navController.popBackStack<ChangeAddressDestination.SelectHousingType>(inclusive = true)
```
StylianosGakis, line 56: If the goal with this is to exit the entire moving flow completely, I *think* it's a good idea to do it on `AppDestination.ChangeAddress` instead, so the actual navigation graph route, and not the first destination. Because if the start destination does change for whatever reason this would stop working.

This is only if I understood correctly that this should exit the entire flow of course.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/MovingDateButton.kt
```diff
@@ -0,0 +1,166 @@
...
+import hedvig.resources.R
+import kotlinx.datetime.Instant
+import kotlinx.datetime.LocalDate
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+
+@Composable
+internal fun MovingDateButton(
+  onDateSelected: (LocalDate) -> Unit,
+  uiState: ChangeAddressUiState,
```
StylianosGakis, line 44: Could this function perhaps take in `DatePickerUiState` and `movingDate` and all the things it needs from inside the `uiState` individually, so we don't need to do the `requireNotNull` here? I think it uses a small subset of the entire `ChangeAddressUiState` object anyway, it would probably be good to only take those anyway, besides doing it so that the nullability of datePickerUiState to be enforced when calling this function.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingContainer.kt
```diff
@@ -0,0 +1,126 @@
...
+                )
+                if (extraBuilding.hasWaterConnected) {
+                  Text(
+                    text = stringResource(id = R.string.CHANGE_ADDRESS_EXTRA_BUILDINGS_WATER_LABEL),
+                    color = MaterialTheme.colorScheme.secondary,
+                    style = MaterialTheme.typography.bodyMedium,
+                  )
+                }
+              }
+              Spacer(modifier = Modifier.weight(1f))
```
StylianosGakis, line 75: Maybe some space too here, in case the text is long enough that it touches the icon on the right?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingTypeContainer.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.card.HedvigCard
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.feature.changeaddress.data.ExtraBuildingType
+import com.hedvig.android.feature.changeaddress.data.stringRes
+
+@Composable
+fun ExtraBuildingTypeContainer(
```
StylianosGakis, line 28: internal maybe?

StylianosGakis (approved):
I feel like it's looking good. Some concerns around the back handling  in general, I probably lack the details regarding how the communication with backend works, but I feel like maybe we don't need to break normal back handling while in there? As I mention in some of the comments, please do fill me in so that I understand what we're doing there.
It will also be nice for me to try this in the app itself too once we merge it in, then it will be easier to review the UI as well, but any fixes around that can of course happen as follow-up PRs.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ChangeAddressViewModel.kt
```diff
@@ -170,21 +154,96 @@ internal class ChangeAddressViewModel(
...
+      )
+    }
+  }
+
+  fun onRemoveExtraBuildingClicked(clickedExtraBuilding: ExtraBuilding) {
+    _uiState.update {
+      val extraBuildings = it.extraBuildings.toMutableList()
+      extraBuildings.remove(clickedExtraBuilding)
+      it.copy(extraBuildings = extraBuildings)
+    }
```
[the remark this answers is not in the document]
hugokallstrom: Oh that is great

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/data/MoveQuote.kt
```diff
@@ -35,8 +38,37 @@ data class MoveQuote(
         numberCoInsured = index,
         premium = UiMoney(99.0 * index, CurrencyCode.SEK),
         startDate = LocalDate(2023, 5, 13),
-        termsVersion = "$index",
         isExpanded = index == 1,
+        productVariant = object : ProductVariantFragment {
```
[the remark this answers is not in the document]
hugokallstrom: I started creating a UI type for this, but realised that it looked exactly the same as the GraphQL type. I wonder if this is a case where we feel its ok to just use this fragment straight from the generated GraphQL code? What do you feel? You can decide :)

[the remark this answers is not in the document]
StylianosGakis: https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1695381566783129

PR #1780, opened 2023-09-25
Error day

StylianosGakis:
I think this way we can log all errors as we should.
If this results in a spike of error logs, we should try and assess if those logs are noise of if they are showing real problems.
`NetworkError` being an `INFO` level log is very deliberate, as those errors are the ones that come from network connection problems or when we ourselves cancel a request (from leaving a screen for example). So for those we do want them to be INFO just for tracing capabilities, while the rest are probably real errors that require our attention. That's at least my assumption now, we will change this as we see fit.

The TODO is there because currently we do get errors for being unauthenticated that have no way to differentiate from a real backend issue. What could be done in the near future is that backend returns inside the `extensions` field in the GQL response information about this. But until they do, we will simply continue logging it as we would any other error.


StylianosGakis:
I was thinking it'd be easier to make this work as we want after apollo-kotlin 4.x is released. They've done some very nice changed around queries not throwing exceptions around, but instead returning typed errors instead.
I think it's gonna be easier to build this feature on top of that instead.
Now they are on 4.0-beta0X so we can wait until next year to do this.

PR #1784, opened 2023-09-28
Market manager refactoring

StylianosGakis:
This is adapting the interface to mainly use a default non-null market, and as an opportunity also to move to DataStore.

This is part of the work to build the new login screen and all of its logic, as it's gonna be easier to work with this going forward by defaulting to SE as the chosen market.

This is *not* to be merged as it is right now, as it needs the login flow to properly adapt to these changes. And this will be done in a stacked PR which works on the Login flow logic + UI together.


hugokallstrom (approved):
LGTM!

PR #1786, opened 2023-09-29
Invalidate hanalytics flags on login/logout events

StylianosGakis:
Instead of having to manually do it on each place in the app which might perform a login/logout action.
FeatureFlagAuthEventListener catches all these events in a centralized spot.


StylianosGakis:
Tried to extract this from some more code, the CI might be failing as I didn't include everything, I will look into bringing everything together later anyway, but won't merge until CI is green of course

hugokallstrom (approved):
Nice!

PR #1788, opened 2023-10-03
Demo mode together with new login design

StylianosGakis:
* Remove AuthTokenServiceProvider

The token provider can remain being the real one at all times. This way,
if we do end up accidentally making real network requests to the backend
when we are in demo mode we know about it as they fail completely.
Then for knowing when we are to be force logged out, we also need to
take care of checking if demo mode is on or not, something that we'd
need to check for anyway.

For all the places that depend on the AuthTokenService, they can stay
exactly as they are. Places like the FCMTokenAuthEventListener etc would
be confusing to get a fake "logged in state" from the service.

* Add way to enter Demo mode

Long pressing the QR code triggers a dialog which sets the demo mode.
MarketingActivity always listens to demo mode to know when to move to
login screen directly

* Hookup logging in with email to GenericAuthActivity

* Add a maestro flow to show that this flow is possible to be done by pressing on some specific buttons with specific IDs.
This is added to ensure that we can give the same instructions (to press on specific items with specific resource IDs) to the people testing this flow. See inside maestro/demo-mode.yaml for the IDs pressed

Here's the test running.

https://github.com/HedvigInsurance/android/assets/44558292/fefb2113-6784-4839-99da-069ae1233e76

TODO:
The tests are not using the provider, will do ASAP


app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/redeemcode/RedeemCodeViewModel.kt
```diff
@@ -15,7 +15,7 @@ import kotlinx.coroutines.launch
 
 class RedeemCodeViewModel(
   private val quoteCartId: QuoteCartId?,
-  private val referralsRepository: ForeverRepository,
+  private val referralsRepository: ForeverRepositoryImpl,
```
hugokallstrom, line 18: This could still be ForeverRepository right?

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ForeverViewModel.kt
```diff
@@ -20,19 +22,19 @@ import giraffe.ReferralsQuery
 import giraffe.fragment.ReferralFragment
 
 internal class ForeverViewModel(
-  private val foreverRepository: ForeverRepository,
-  private val getReferralTermsUseCase: GetReferralsInformationUseCase,
+  foreverRepositoryProvider: Provider<ForeverRepository>,
```
hugokallstrom, line 25: Great that you can inject the Provider interface instead!

hugokallstrom (approved):
🚀

app/app/src/main/kotlin/com/hedvig/app/feature/referrals/ui/redeemcode/RedeemCodeViewModel.kt
StylianosGakis: Thank you, yes!

app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ForeverViewModel.kt
StylianosGakis: Yeah, this way the tests are able to just do `Provider { fakeImpl() }` as we discussed here https://hedviginsurance.slack.com/archives/D02DE2ZGLS1/p1695369859355719?thread_ts=1695367706.676759&cid=D02DE2ZGLS1
So the tests need not know anything about demo mode whatsoever.

PR #1789, opened 2023-10-03
Moving flow validation

hugokallstrom:


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/TypeOfContractExt.kt
```diff
@@ -63,3 +63,65 @@ fun TypeOfContract.toContractType(): ContractType = when (this) {
 
   is TypeOfContract.UNKNOWN__ -> ContractType.UNKNOWN
 }
+
+fun String.toContractType(): ContractType = when (this) {
```
StylianosGakis, line 67: Will we need to do this locally with the new API?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/data/MoveQuote.kt
```diff
@@ -34,22 +35,25 @@ data class MoveQuote(
...
-          product = Product(
-            displayNameFull = "Test",
-            pillowImageUrl = "",
-          ),
           perils = emptyList(),
-          insurableLimits = emptyList(),
+          insurableLimits = listOf(
+            InsurableLimit(
+              label = "test",
+              description = "longlonglonglonglonglonglonglonglonglonglonglonglonglonglonglong",
```
StylianosGakis, line 50: A thing I like to do in these scenarios is
```suggestion
              description = "long".repeat(10),
```
It's very convenient

StylianosGakis (approved):

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/insurance/TypeOfContractExt.kt
hugokallstrom: Good question, there is an enum in the schema from promise-cms actually. I can investigate if its possible to use that 👍

PR #1790, opened 2023-10-03
Demo mode tracking

StylianosGakis:
This way we will know in the logs if they come from a user in demo mode or not.
If we miss something which errors out a lot for example but only for demo users, we'll know that production works properly still.

DatadogDemoModeTracking.kt is the core of this change


hugokallstrom (approved):
💯

PR #1792, opened 2023-10-03
Moving flow design fixes

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -82,13 +81,7 @@ internal fun HedvigNavHost(
...
   fun startMovingFlow() {
     coroutineScope.launch {
-      if (featureManager.isFeatureEnabled(Feature.NEW_MOVING_FLOW)) {
-        hedvigAppState.navController.navigate(AppDestination.ChangeAddress)
-      } else {
-        context.startActivity(
-          LegacyChangeAddressActivity.newInstance(context),
-        )
-      }
+      hedvigAppState.navController.navigate(AppDestination.ChangeAddress)
```
StylianosGakis, line 84: Do we want to merge this like this?
If we make a new release already this won't be ready to ship already right?
Could we keep the feature flag check here, and if it makes it hard to test locally, make sure this https://github.com/HedvigInsurance/android/blob/b601360a2c8750ba038c01e563319792e5f8f36f/app/hanalytics/hanalytics-feature-flags-public/src/main/kotlin/com/hedvig/android/hanalytics/featureflags/flags/DevFeatureFlagProvider.kt#L28 is true.

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/textfield/HedvigTextField.kt
```diff
@@ -72,6 +72,7 @@ fun HedvigTextField(
   value: String,
   onValueChange: (String) -> Unit,
   errorText: String?,
+  isError: Boolean = errorText != null,
```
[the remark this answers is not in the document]
hugokallstrom: Then the textfield will be in a error state (warning backround) but with no text message. Request from William, turned out great for moving flow!

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
hugokallstrom: Good catch, fixing

PR #1795, opened 2023-10-04
Query contracts on member in new schema. 

hugokallstrom:
Refactor contracts and contract details.


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -131,16 +138,14 @@ private fun ContractDetailScreen(
               key = 1,
               contentType = "InsuranceCard",
             ) {
-              val contractCardData = state.contractDetails.contractCardData
+              val contract = state.insuranceContract
               InsuranceCard(
-                chips = contractCardData.chips,
-                topText = contractCardData.title,
-                bottomText = contractCardData.subtitle,
+                chips = contract.createChips(context),
```
StylianosGakis, line 143: Hmm I see the convenience that this brings. It is a bit confusing to have to bring the context all the way to the data layer for this.
In other places, we typically create some sort of sealed interface for the scenarios, and do the mapping to the string itself in the ui code, like we do here for example
https://github.com/HedvigInsurance/android/blob/03175abc23b2fa65a96c57f37229676490b55489/app/app/src/main/kotlin/com/hedvig/app/feature/chat/ui/ChatActivity.kt#L172-L173

But here it may be that we need access to this in more than one place.
Is there some other place, other than in the data module for this to live perhaps? And maybe it can also be a `@Composable` function, so that we can use `getString()` in there? Not sure though

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -158,28 +163,35 @@ private fun ContractDetailScreen(
...
                   0 -> {
                     YourInfoTab(
-                      coverageItems = state.contractDetails.overviewItems,
-                      allowChangeAddress = state.contractDetails.allowChangeAddress,
-                      allowEditCoInsured = state.contractDetails.allowEditCoInsured,
+                      coverageItems = state.insuranceContract.currentAgreement.displayItems
+                        .map { it.title to it.value }
+                        .toImmutableList(),
+                      allowEditCoInsured = true, // TODO
+                      allowChangeAddress = true, // TODO
```
StylianosGakis, line 170: Are these two just not present in the new Api then somehow?
We had this https://github.com/HedvigInsurance/android/blob/23cd4e0bb5e7c6678c5fa4c54631a01b5a230616/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/data/GetContractDetailsUseCase.kt#L162-L207 for the co-insured part before, can we have it again here? Do we not want to have it anymore perhaps? Is that the case?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailViewModel.kt
```diff
@@ -5,23 +5,27 @@ import androidx.lifecycle.viewModelScope
...
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.WhileSubscribed
 import kotlinx.coroutines.flow.stateIn
-import kotlin.time.Duration.Companion.seconds
 
 internal class ContractDetailViewModel(
   contractId: String,
-  private val getContractDetailsUseCase: Provider<GetContractDetailsUseCase>,
+  private val getInsuranceContractsUseCase: Provider<GetInsuranceContractsUseCase>,
```
StylianosGakis, line 18:
```suggestion
  private val getInsuranceContractsUseCaseProvider: Provider<GetInsuranceContractsUseCase>,
```

Maybe? We do follow that convention usually with this, so that doing `provide()` on it does not look odd further down. What do you think?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailViewModel.kt
```diff
@@ -5,23 +5,27 @@ import androidx.lifecycle.viewModelScope
...
   private val retryChannel = RetryChannel()
   val uiState: StateFlow<ContractDetailsUiState> = retryChannel.transformLatest {
     emit(ContractDetailsUiState.Loading)
     val uiState = either {
-      getContractDetailsUseCase.provide().invoke(contractId).bind()
+      getInsuranceContractsUseCase
+        .provide()
+        .invoke()
+        .bind()
+        .first { it.id == contractId }
```
StylianosGakis, line 28: Now I think this will throw an exception if the contract ID does not get found at all, since we do `first {}` on the List<> itself.
Could we perhaps do a `firstOrNull {}` and then do a [ensureNotNull](https://github.com/HedvigInsurance/android/blob/d2bff2658dad04316166214d4fe9e34a354af3cb/app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/GetTravelCertificateSpecificationsUseCase.kt#L44) and return a typed error instead, so it doesn't crash completely?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/coverage/CoverageTab.kt
```diff
@@ -150,6 +151,9 @@ private fun ColumnScope.PerilSection(
   }
 }
 
+val String.color
```
StylianosGakis, line 154: Maybe `private val` so that this does not get exposed to a lot of the project as an extension on any string?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/data/GetContractDetailsUseCase.kt
```diff
@@ -1,207 +0,0 @@
...
-                contractPerils = persistentListOf(),
-                insurableLimits = persistentListOf(),
-                documents = persistentListOf(),
-              )
-            },
-            ifRight = { it },
-          )
-        },
-        {
-          featureManager.isFeatureEnabled(Feature.TERMINATION_FLOW)
```
StylianosGakis, line 59: Do we not need to do this check anymore, or am I missing it from this PR?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
```diff
@@ -164,8 +171,8 @@ internal fun YourInfoTab(
       Spacer(Modifier.height(8.dp))
     }
     CoverageRows(coverageItems, Modifier.padding(horizontal = 16.dp))
-    if (allowChangeAddress || allowEditCoInsured) {
-      Spacer(Modifier.height(16.dp))
+    Spacer(Modifier.height(16.dp))
+    if (!isTerminated) {
```
StylianosGakis, line 175: We still do want to hide this button if both of these features are off.
We do use this check for Demo mode I know for sure.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
```diff
@@ -120,8 +125,10 @@ internal fun YourInfoTab(
       windowInsets = BottomSheetDefaults.windowInsets.only(WindowInsetsSides.Top),
     ) {
       UpcomingChangesBottomSheetContent(
-        infoText = upcomingChanges.title,
-        sections = upcomingChanges.sections,
+        infoText = "", // TODO
```
StylianosGakis, line 128: Missing from API or?

StylianosGakis (approved):
Some todos here and there, maybe we need to fix before the merge?

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
hugokallstrom: Thats great

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
hugokallstrom: Good catch. Im changing so that we look at ContractType for co insured. For address change we actually get from the backend now.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
hugokallstrom: Found in lokalise, good catch

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
hugokallstrom: Fixed 👍

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailViewModel.kt
```diff
@@ -15,17 +18,21 @@ import kotlinx.coroutines.flow.stateIn
...
+      val contract = getInsuranceContractsUseCaseProvider
         .provide()
         .invoke(forceNetworkFetch = false)
         .bind()
-        .first { it.id == contractId }
+        .firstOrNull { it.id == contractId }
+      ensureNotNull(contract) {
+        logcat(LogPriority.ERROR) { "No contract found with id: $contractId" }
+        ContractDetailsUiState.Error
+      }
```
StylianosGakis, line 35: Perfect! 🎉

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/ui/InsuranceContractExt.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.hedvig.android.feature.insurances.data.InsuranceContract
+import hedvig.resources.R
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.toPersistentList
+import kotlinx.datetime.Clock
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.todayIn
+
+@Composable
+fun InsuranceContract.createChips(): ImmutableList<String> {
```
StylianosGakis, line 19: Could it also be `internal` if we only use it in this module?

But yeah I *feel* like this is a bit nicer, since we do not need to add this code so close to the data layer. And no reference to context so close to the data layer either. Let's see if it proves to be harder to maintain or problematic if we need to use this in multiple places and so on.

StylianosGakis (approved):
Perfect! So if no more TODOs for strings and so on we can merge it.

StylianosGakis:
Ah yeah tests seem to need updating too.
Try and run `./gradlew testDebugUnitTest` to run them all (it's quite fast actually) so you can fix them all before pushing again.
And run `./gradlew ktlintFormat` too to fix PR / ktlint failing too.

PR #1797, opened 2023-10-05
This just moves all the code to the module

StylianosGakis:
I tried to do minimal if any changes. Most of the actual code changes was deleting some unused functions, or inlining some functions that were only used in 1 place.

The only real change was that I removed HAnalytics from the ViewModel since we do not use that anymore anyway.


app/app/src/androidTest/kotlin/com/hedvig/app/feature/loggedin/LoggedInScreen.kt
```diff
@@ -1,17 +1,17 @@
...
 import io.github.kakaocup.kakao.screen.Screen
 
 class LoggedInScreen : Screen<LoggedInScreen>() {
-  val root = KView { withId(R.id.loggedInRoot) }
-  val bottomTabs = KBottomNavigationView { withId(R.id.bottomNavigation) }
+  val root = KView { withId(R.id.chat) }
+  val bottomTabs = KBottomNavigationView { withId(R.id.chat) }
 
-  val openChat = KView { withId(R.id.openChat) }
+  val openChat = KView { withId(R.id.chat) }
```
StylianosGakis, line 14: This was broken so I just put whatever IDs to make it not complain. We wanna delete this anyway, but didn't do it since it may cascade more than it's worth right now.

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/legacy/LiveEvent.kt
```diff
@@ -0,0 +1,58 @@
+package com.hedvig.android.feature.chat.legacy
+
+import androidx.annotation.MainThread
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.MediatorLiveData
+import androidx.lifecycle.Observer
+
+// to be deleted asap when Chat gets reworked. Copied over from :app so we don't expose this everywhere in a common
+// module
+internal class LiveEvent<T> : MediatorLiveData<T>() {
```
StylianosGakis, line 10: This, like some other files here are just duplicated from the :app module. They are inside the `legacy` package indicating that we do want to just delete them asap!

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/navigation/ChatGraph.kt
```diff
@@ -0,0 +1,25 @@
...
+  hedvigBuildConstants: HedvigBuildConstants,
+) {
+  composable<AppDestination.Chat>(
+    deepLinks = listOf(
+      navDeepLink { uriPattern = hedvigDeepLinkContainer.chat },
+    ),
+    enterTransition = { slideIntoContainer(towards = AnimatedContentTransitionScope.SlideDirection.Up) },
+    exitTransition = { slideOutOfContainer(towards = AnimatedContentTransitionScope.SlideDirection.Down) },
+  ) {
+  }
```
StylianosGakis, line 24: Chat will be added here in follow-up PR which also removes the activity itself.

hugokallstrom (approved):
Great refactoring exercise :D Gj!

PR #1798, opened 2023-10-06
Add ChatRepositoryNew.kt with queries and mutations to octopus.

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/chat/data/ChatRepositoryNew.kt
```diff
@@ -0,0 +1,120 @@
...
+  )
+
+  is ChatMessagesQuery.Data.Chat.ChatMessageTextMessage -> ChatMessage.ChatMessageText(
+    id = id,
+    sender = when (sender) {
+      ChatMessageSender.MEMBER -> ChatMessage.Sender.MEMBER
+      ChatMessageSender.HEDVIG -> ChatMessage.Sender.HEDVIG
+      ChatMessageSender.UNKNOWN__ -> ChatMessage.Sender.HEDVIG
+    },
+    sentAt = null,
```
StylianosGakis, line 107: Is this something we want the API to include in the future perhaps?

app/app/src/main/kotlin/com/hedvig/app/feature/chat/data/ChatRepositoryNew.kt
```diff
@@ -0,0 +1,120 @@
...
+    if (error != null) {
+      raise(ErrorMessage(error.message))
+    } else if (message == null) {
+      raise(ErrorMessage("No data"))
+    } else {
+      ChatMessageResult(
+        message = message.toMessage(),
+        status = status?.message ?: "",
+      )
+    }
```
StylianosGakis, line 83: Ensure is pretty convenient in the context of an `either` as it just lets you take care of some scenarios one by one, without having to nest anything in if checks.
It'd look something like this here
```suggestion
    ensure(error == null) {
      ErrorMessage(error.message)
    }
    ensureNotNull(message) {
      ErrorMessage("No data")
    }
    ChatMessageResult(
      message = message.toMessage(),
      status = status?.message ?: "",
    )
```
Where you ensure that error is null, otherwise it raises the error and returns early.
Then it ensures that message is not null, early-returning again with the correct ErrorMessage otherwise.
And in the end, you're just left with the good path, and `error` and `message` are smart cast at this point to be null and non-null accordingly.

This way you avoid having to call `raise` completely, and the code reads a tiny bit easier without the chained if statement I think.

But this is a stylistic change, just sharing my opinion.

app/app/src/main/kotlin/com/hedvig/app/feature/chat/data/ChatRepositoryNew.kt
```diff
@@ -0,0 +1,120 @@
...
+      .bind()
+
+    ChatMessagesResult(
+      messages = result.chat.messages.mapNotNull { it.toMessage() },
+      nextUntil = result.chat.nextUntil,
+      hasNext = result.chat.hasNext,
+    )
+  }
+
+  suspend fun sendFile(uploadUrl: String) = either {
```
StylianosGakis, line 42: What will `uploadUrl` be here?
I suppose at some point we will need to convert the URI from the location in the device into a `File` and then into an apollo `Upload`, kinda how we do here https://github.com/HedvigInsurance/android/blob/69ef86b42ccc12b14c1ce6145d8a63815cd68c44/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatRepository.kt#L140

I wonder what the easiest work with type would be for the repository function to accept. Maybe it's String, while also being the least type-safe of course.

app/apollo/apollo-octopus-public/build.gradle.kts
```diff
@@ -35,8 +35,10 @@ apollo {
     generateOptionalOperationVariables.set(false)
 
     mapScalar("Date", "kotlinx.datetime.LocalDate", "com.apollographql.apollo3.adapter.KotlinxLocalDateAdapter")
+    mapScalar("Instant", "kotlinx.datetime.Instant", "com.apollographql.apollo3.adapter.KotlinxInstantAdapter")
     mapScalarToKotlinString("UUID")
     mapScalarToKotlinString("Url")
     mapScalarToKotlinString("FlowContext")
+    mapScalarToKotlinString("Upload")
```
StylianosGakis, line 42: There's `mapScalarToUpload("Upload")` which we use for Giraffe too.
This lets us use Upload https://github.com/apollographql/apollo-kotlin/blob/b4fb9fb7dea17b443fdce0fddce389bfd2331796/libraries/apollo-api/src/commonMain/kotlin/com/apollographql/apollo3/api/Upload.kt#L19 from apollo library directly, and it just works regarding setting the content type, length and everything that needs to be done for it to be sent over the network.
Also for us, we get the convenience extension function from `java.io.File.toUpload()` here https://github.com/apollographql/apollo-kotlin/blob/b4fb9fb7dea17b443fdce0fddce389bfd2331796/libraries/apollo-api/src/jvmMain/kotlin/com/apollographql/apollo3/api/FileUpload.kt#L9-L18

StylianosGakis (approved):
I suppose this would also now need to be done on top of current develop which moved the chat feature into its own module too.

And I think we should use `com.apollographql.apollo3.api.Upload` as we do for Giraffe too, it's very convenient.

app/apollo/apollo-octopus-public/build.gradle.kts
hugokallstrom: Excellent

PR #1799, opened 2023-10-06
New chat viewmodel+presenter

hugokallstrom:


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+) {
+  init {
+    emit(ChatEventNew.FetchMessages())
+    viewModelScope.launch {
+      while (true) {
+        delay(5.seconds)
+        emit(ChatEventNew.PollMessages)
+      }
+    }
+  }
```
StylianosGakis, line 46: This can be replaced with a launched effect inside the presenter itself like
```
LaunchedEffect(Unit) {
  while(isActive) {
    delay(5.seconds)
    fetchId++
  }
}
```

Doing it that way also ensures that it's part of the presenter's logic and does not come outside of it, so that when we run tests on the presenter itself we also make sure that this behavior exists in the tests as well.

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+      when (event) {
+        is ChatEventNew.FetchMessages -> fetchUntil = event.until
+        ChatEventNew.PollMessages -> fetchId++
+        is ChatEventNew.SendFileMessage -> fileState = event.file
+        is ChatEventNew.SendTextMessage -> messageInputState = event.message
+        is ChatEventNew.DismissError -> errorMessage = null
+      }
+    }
+
+    LaunchedEffect(fetchId, fetchUntil) {
```
StylianosGakis, line 80: Adding `fetchUntil` will re-trigger fetching new messages. It seems like to me when this call succeeds, fetchUntil always changes, so we'll basically always be doing 2 fetches here back to back, is this what we want? I suppose the 3rd time won't happen then, since the 2nd fetch will fetch the exact same fetchUntil value (old fetchUntil will be == the new fetchUntil) so this will not re-trigger.

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+
+internal class ChatPresenter(
+  private val chatRepository: Provider<ChatRepositoryNew>,
+  private val chatClosedTracker: ChatEventStore,
+  private val featureManager: FeatureManager,
+) : MoleculePresenter<ChatEventNew, ChatUiState> {
+
+  @Composable
+  override fun MoleculePresenterScope<ChatEventNew>.present(lastState: ChatUiState): ChatUiState {
+    var isLoadingChat by remember { mutableStateOf(lastState.isLoadingChat) }
```
StylianosGakis, line 57: This is just a loading state for the initial fetch right?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+    var isLoadingNewMessages by remember { mutableStateOf(lastState.isLoadingNewMessages) }
+    var messages by remember { mutableStateOf(lastState.messages) }
+    var isSendingMessage by remember { mutableStateOf(lastState.isSendingMessage) }
```
StylianosGakis, line 60: Do we want to do something with this? Like prevent new messages from being sent while the old one is sent or something like that?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+    var canFetchMoreMessages by remember { mutableStateOf(lastState.canFetchMoreMessages) }
```
StylianosGakis, line 61: Is this to be used by the UI itself, so that it know when *not* to send new requests for new messages? Or what is the idea?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+    var canFetchUntil by remember { mutableStateOf(lastState.canFetchUntil) }
```
StylianosGakis, line 62: What's the difference between fetchUntil and canFetchUntil?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    var fetchId by remember { mutableIntStateOf(0) }
+    var fetchUntil by remember { mutableStateOf<Instant?>(null) }
+    var fileState by remember { mutableStateOf<File?>(null) }
+    var messageInputState by remember { mutableStateOf<String?>(null) }
+
+    CollectEvents { event ->
+      when (event) {
+        is ChatEventNew.FetchMessages -> fetchUntil = event.until
+        ChatEventNew.PollMessages -> fetchId++
```
StylianosGakis, line 73: When does the UI need to ask the presenter for this event? If we're going to be polling every 5 seconds anyway, maybe this event does not need to exist, and we can just do it as the LaunchedEffect that I suggest further up.
Or does the UI itself need to send this event to the presenter at any point?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+    var errorMessage by remember { mutableStateOf(lastState.errorMessage) }
+
+    var fetchId by remember { mutableIntStateOf(0) }
+    var fetchUntil by remember { mutableStateOf<Instant?>(null) }
+    var fileState by remember { mutableStateOf<File?>(null) }
+    var messageInputState by remember { mutableStateOf<String?>(null) }
+
+    CollectEvents { event ->
+      when (event) {
+        is ChatEventNew.FetchMessages -> fetchUntil = event.until
```
StylianosGakis, line 72: How will the UI know what `until` is to send this event? Is the idea that the latest `fetchUntil` is passed down to the UI, and they send it through this event when we scroll all the way to the top?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModelNew.kt
```diff
@@ -0,0 +1,161 @@
...
+        .fold(
+          ifLeft = {
+            errorMessage = it.message
+            isSendingMessage = false
+          },
+          ifRight = {
+            it.message?.let(messages::add)
+            isSendingMessage = false
+          },
+        )
```
StylianosGakis, line 114: Should we be clearing the `fileState` after this effect is run? So we know it's now handled and is null again.
Same for `messageInputState`

PR #1800, opened 2023-10-09
Moving flow QA

hugokallstrom:


app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/offer/QuoteCard.kt
```diff
@@ -171,10 +178,45 @@ private fun ExpandedInformation(
       endSlot = { Text(quote.address.postalCode, textAlign = TextAlign.End) },
       spaceBetween = 4.dp,
     )
+    quote.squareMeters?.let {
```
hugokallstrom, line 181: This will soon change to a "display items" type, where we get localised key value pairs from backend.

PR #1802, opened 2023-10-09
Cleanup: Remove chat activity

StylianosGakis:
Made `hedvig-dev` firebase project, which holds the firebase configuration for only the dev app. This was required since the dev app has a different package name "com.hedvig.dev.app" and it had to be separate.
Hence the deep link url changes here in the string resources and in the HedvigBuildConstants. The google-services.json is also generated from that firebase project too, hence the change.

This change other than that, moves the ChatActivity simply over to being a fragment. A few code changes had to be made along with it. Namely some lifecycle related functions, and making sure that some listeners are not present forever, like `keyboardHeightListener` was done here, since doing `binding.chatRoot.viewTreeObserver.addOnGlobalLayoutListener` and never unregistering it meant that it survived globally, crashing the application when the fragment was removed and the binding was now null.

Besides that, all the links we've had from other places in the app were migrated like this:

For places which are part of the mono-activity already: Just navigate to the new AppDestination
For other activities which can't just do that: Navigate to the deep link, which does the right thing of putting the LoggedInActivity on top of that activity on the right destination. Pressing back puts them back to where they came from.


hugokallstrom (approved):
Looks good!

PR #1881, opened 2023-11-17
Add reminder cards for missing co insured information

hugokallstrom:
Screenshot_20231117_124012

Last reminder in above screenshot. Visible when any co insured has `needsMissingInfo` set.


app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
```diff
@@ -73,6 +74,24 @@ fun MemberReminderCards(
...
+      visible = memberReminders.coInsuredInfo != null,
+      enter = cardReminderEnterTransition,
+      exit = cardReminderExitTransition,
+      label = "coInsured animated visibility",
+    ) {
+      ReminderCoInsuredInfo(
+        navigateToContractDetail = {
+          memberReminders.coInsuredInfo
+            ?.coInsuredInfoList
+            ?.firstOrNull()
```
StylianosGakis, line 87: Since `coInsuredInfoList` is a NonEmptyList, I think you can just do `first()` without the `orNull` or even just call `coInsuredInfoList.head` and it returns back the non-null version of the first item.

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/AppDestination.kt
```diff
@@ -54,6 +54,14 @@ sealed interface AppDestination : Destination {
     val insuranceDisplayName: String,
   ) : AppDestination
 
+  @Serializable
+  data class InsuranceContractDetail(
+    val contractId: String,
+  ) : AppDestination
+
+  @Serializable
+  data object TerminatedInsurances : AppDestination
```
StylianosGakis, line 63: Same comment as above, I think we can avoid moving these destinations to the global AppDestination interface, and keep them only to the relevant module destinations.

StylianosGakis (changes requested):
If there is any reason why moving the destinations back wouldn't be possible please let me know and we can figure it out

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
```diff
@@ -99,6 +118,7 @@ fun MemberReminderCards(
 fun MemberReminderCards(
   memberReminders: ApplicableMemberReminders,
   navigateToConnectPayment: () -> Unit,
+  navigateToContractDetail: (contractId: String) -> Unit,
```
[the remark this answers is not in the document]
hugokallstrom: We want to show the co insured reminder everywhere

app/member-reminders/member-reminders-ui/src/main/kotlin/com/hedvig/android/memberreminders/ui/MemberReminderCards.kt
hugokallstrom: Awesome!

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/navigation/InsurancesNavigation.kt
```diff
@@ -1,20 +0,0 @@
...
-  ) : InsurancesDestination
-
-  @Serializable
-  data object TerminatedInsurances : InsurancesDestination
-}
-
-val insurancesBottomNavPermittedDestinations: List<String> = listOf(
-  createRoutePattern<InsurancesDestination.InsuranceContractDetail>(),
-  createRoutePattern<InsurancesDestination.TerminatedInsurances>(),
-)
```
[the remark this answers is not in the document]
hugokallstrom: That makes a lot of sense. Good catch.

StylianosGakis (approved):
Does it auto-fix the problems if you run `./gradlew ktlintFormat`?
if it does not, does running `./gradlew ktlintCheck` tell you in your code where the problematic code which makes CI fail lives?

PR #1883, opened 2023-11-20
Edit co-insured module and navigation

hugokallstrom:


StylianosGakis (approved):
Can run `./gradlew lint` when creating a new module.
Then it runs lint, and if creates the baseline file, which should be an empty file for now since there shouldn't be any issues with it already, and then just commit that auto-generated file too and push it. Will make the lint CI task succeed.

PR #1885, opened 2023-11-21
Show current insured people in list

hugokallstrom:
Screenshot_20231121_094042


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/CoInsured.kt
```diff
@@ -1,11 +1,22 @@
...
   val birthDate: LocalDate?,
   val ssn: String?,
   val hasMissingInfo: Boolean,
-)
+) {
+  val displayName: String? = if (firstName != null && lastName != null) {
+    "$firstName $lastName"
+  } else {
+    null
+  }
```
StylianosGakis, line 18: Potentially same comment as two PRs before, with the name and/or last name showing individually if either of them are non-null

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/CoInsured.kt
```diff
@@ -1,11 +1,22 @@
...
+
+  fun details(dateTimeFormatter: DateTimeFormatter): String? =
```
StylianosGakis, line 20: Not sure how `details` is used here, maybe rename this function to a more descriptive name? Otherwise, if it's only used in 1 very specific place where it makes sense only there, I wouldn't mind it to just be inlined there instead.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredDestination.kt
```diff
@@ -2,37 +2,160 @@ package com.hedvig.android.feature.editcoinsured.ui
...
+        displayName = it.displayName,
+        details = it.ssn,
+        hasMissingInfo = false,
+        allowEdit = false,
+        isMember = true,
+        onRemove = { },
+        onEdit = { },
+      )
+    }
+    Divider()
```
StylianosGakis, line 69: This divider goes all the way to the edges, we probably want
```suggestion
    Divider(Modifier.padding(horizontal = 16.dp))
```
here

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/InsuredRow.kt
```diff
@@ -0,0 +1,185 @@
...
+              modifier = Modifier.size(16.dp),
+            )
+          }
+        }
+      }
+    },
+    spaceBetween = 8.dp,
+    modifier = Modifier
+      .clickable(enabled = !isMember) {
+        if (hasMissingInfo) {
```
StylianosGakis, line 121: Should this also then check for if `allowEdit` is true?

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/navigation/EditCoInsuredDestinations.kt
```diff
@@ -4,4 +4,7 @@ import com.kiwi.navigationcompose.typed.Destination
 import kotlinx.serialization.Serializable
 
 @Serializable
-data class EditCoInsuredDestination(val contractId: String) : Destination
+data class EditCoInsuredDestination(
+  val contractId: String,
+  val allowEdit: Boolean,
```
StylianosGakis, line 9: Hmm I'd like to challenge whether we want to have a `allowEdit` on the same destination, or if we can instead have two distinct destinations, where both of them can potentially share some of the composables that they use.

My concern here is that the non editable and the editable versions of this screen may steer away from each other enough that having both in the same place might become quite hard to reason about? Reminds me of some older screens we've had which were clearly something like "either X or Y state" but those states ended up being different enough that completely separate screens would make more sense anyway.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/InsuredRow.kt
```diff
@@ -0,0 +1,185 @@
...
+      Row(
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = Modifier.padding(vertical = 16.dp),
+      ) {
+        Column {
+          if (displayName != null) {
+            Text(
+              text = displayName,
+              color = if (isMember) {
+                MaterialTheme.colorScheme.onSurfaceVariant.copy(alpha = 0.5f)
```
StylianosGakis, line 51: If we want a "disabled for interaction" sort of look here, onSurfaceVariant already achieves this.
If we want an even more toned down version you can probably use onSurface with alpha of this  https://github.com/HedvigInsurance/android/blob/03cecb9b5c957d0f15084f67cbe0e49d7e102d7d/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3ColorScheme.kt#L194 for the constant instead.

I'd like to say that just using `onSurfaceVariant` by itself is good enough for the "disabled text" situations, all other options become way too low-contrast with the background making the text downright unreadable for users with any sort of visual difficulties.
It's the exact same reason for this comment here 
image
and why sometimes I may look like I am using the wrong colors from design, but it's not by accident that I do so.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
```diff
@@ -64,4 +69,5 @@ internal data class EditCoInsuredState(
   val isLoading: Boolean = false,
   val errorMessage: String? = null,
   val coInsured: ImmutableList<CoInsured> = persistentListOf(),
+  val member: Member? = null,
```
StylianosGakis, line 72: This class atm can be of the possible 16 combinations of what is null and what is not null, (or false/true, or empty list/non-empty list).
Should a state like:
`loading = true, errorMessage = "error", coInsured = [blah], member = Member(...)` be possible?
How about the state:
`loading = false, errorMessage = null, coInsured = [], member = null`?

Both of those are possible here, but I am not sure if they should be possible in the first place?

Can use this https://github.com/HedvigInsurance/android/blob/5f1f762af6f5af12cc93946f7113bc4c9cee9a0a/app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/marketing/MarketingPresenter.kt#L60-L67 UiState as a good example. This could also be written as
```
internal data class MarketingUiState(
  val isLoading: Boolean,
  val market: Market?,
  val language: Language?
)
```
But then we make it possible for the state to be `isLoading = false, market = null, language = null`.
Or `isLoading = false, market = Market.SE, language = null`
Both of which are not useful for our UI to render, and are not possible to get to at all with the sealed interface solution instead.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredDestination.kt
```diff
@@ -2,37 +2,160 @@ package com.hedvig.android.feature.editcoinsured.ui
...
+        details = coInsured.details(dateTimeFormatter),
+        hasMissingInfo = coInsured.hasMissingInfo,
+        isMember = false,
+        allowEdit = allowEdit,
+        onRemove = { },
+        onEdit = { },
+      )
+      if (index < uiState.coInsured.size - 1) {
+        Divider()
+      }
```
StylianosGakis, line 82: I've realized lately that just doing it the other way around, putting it as the first thing in the loop, with a check on the first index is definitely an easier approach not to make any mistakes with looking for the wrong index or something like that. Now doing these `size - 1` etc checks are always prone to a programmer mistake I think.

Here https://github.com/HedvigInsurance/android/blob/03cecb9b5c957d0f15084f67cbe0e49d7e102d7d/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/history/PaymentHistoryDestination.kt#L129-L131 is an example of this, it's much harder to make a mistake on checking if you are simply not on the first item instead.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredDestination.kt
StylianosGakis, line 69: This divider would now show even if `member` is null, dividing nothing above it.
Or even when coInsured is an empty list (or both of those at the same time).
In either of these cases, the divider would look odd.
This would probably be solved completely by modeling the state in a way where such scenarios like the model being null are just impossible in this branch, and then the divider can just be added at the top of the list if there are any items in the list in the first place.

StylianosGakis (changes requested):
Would like to see if the `allowEdit` is the right thing to do here.
Also see if we can edit the UiState the presenter provides to make it impossible for it to represent some of the states which should be impossible to reach.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/navigation/EditCoInsuredDestinations.kt
hugokallstrom: I agree. Should design this as two different destinations. Ill fix

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/InsuredRow.kt
hugokallstrom: Good info, cheers

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
hugokallstrom: Makes sense, will change

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/InsuredRow.kt
hugokallstrom: Will change

PR #1886, opened 2023-11-22
Show bottom sheet and fetch co-insured from ssn

hugokallstrom:


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
```diff
@@ -52,22 +57,53 @@ internal class EditCoInsuredPresenter(
...
+          fetchCoInsuredPersonalInformationUseCaseProvider.provide()
+            .invoke(event.ssn)
+            .fold(
+              ifLeft = { coInsuredFromSsnError = it.message },
+              ifRight = {
+                coInsuredFromSsn = it
+                coInsuredFromSsnError = null
+              },
+            )
+        }
```
StylianosGakis, line 73: Hmm this we can talk about a bit too.

Right now I have concerns.
1. If the member clicks this button many times, it will launch many of these coroutines here, running the query potentially many times, all in parallel. And the one which will be set in the end is just the last one which finishes, not even the last click necessarily.
2. If the member clicks the button and sends this event, there is nothing right now indicating to them that there is something loading at this moment, that we're trying to fetch something.
3. If this fails to fetch the name + last name, one approach is to show an error and try again, or we could just decide that it's fine to fail here and just let them fill those fields themselves manually, so that we do not make this step impossible to go through if for example the place where we fetch this info from is down and the query keeps failing forever. (We've had that issue with danish autocomplete once, where the service was down and having a fallback was crucial for the flow to not be completely blocked)

Very conveniently, this https://youtu.be/q9p4ewk-9E4?t=1346 video at that specific time + ~1 minute forward very specifically explains this scenario and how you can use compose state to trigger a LaunchedEffect which will actually do the work for you instead.

That solves the first 2 issues:
1. If the button is pressed again with a new input for example, then the LaunchedEffect will *cancel* the old thing it was running, and start the new one, so we don't have parallel things going on at the same time
2. We can set a loading state (not the loading of the entire screen, a loading state specifically for this query, which can power when the button in the UI shows as loading) before the query is done, and set it back to false when it's done loading.

The 3rd, to fallback in case of an error is a decision we may or may not make, we can see what makes the most sense.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -0,0 +1,154 @@
...
+        stringResource(id = R.string.CONTRACT_SSN_FETCH_INFO)
+      },
+      enabled = ssn.isNotBlank(),
+      onClick = {
+        if (coInsured != null) {
+          onSave(coInsured)
+        } else {
+          onFetchInfo(ssn)
+        }
+      },
```
StylianosGakis, line 102: This may or may not be a good idea, but I wonder if we can have the button always just send the "I clicked this button" event to the presenter, and have that deduce what that means for the current state and if it should submit the thing, or if it should fetch the data for the member.
Maybe thinking out loud a bit here, to reduce the amount of logic in the UI itself. If it doesn't feel like it makes sense here we can skip.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -0,0 +1,154 @@
...
+      isLoading = isLoading,
```
StylianosGakis, line 103: Here then this button could be set as loading when we are fetching the member's info, or if we're submitting it. Not in the initial loading state for example. Not when the `getCoInsuredUseCaseProvider.provide().invoke(contractId)` is still loading initially on the presenter.
Perhaps 🤷

StylianosGakis (changes requested):
Let's try and fix the presenter with the concerns I had for the launch on the button click event.
Typically doing a `launch` from inside `CollectEvents` should be the last resort, if nothing else works for whatever reason that may be. Like we do here https://github.com/HedvigInsurance/android/blob/db3511f89a99c29cfd57dd6220c581c8c62b25bf/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt#L93 where marking the state as seen more than once is fine since it should be an idempotent action.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
```diff
@@ -0,0 +1,39 @@
...
+      .bind()
+
+    ensureNotNull(result.personalInformation) {
+      ErrorMessage("No personal information found")
+    }
+
+    CoInsured(
+      firstName = result.personalInformation.firstName,
+      lastName = result.personalInformation.lastName,
+      birthDate = null,
```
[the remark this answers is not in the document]
hugokallstrom: My thinking here was that these use cases will only be used in this module, and that we always work with this "CoInsured" data type in this module. Is this coupling too tight do you think? Should we return another datatype from this use case and it to CoInsured in the presenter instead?

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
hugokallstrom: This is added in the PR above this!

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
StylianosGakis: Ah found the place where we do this exact thing ourselves too, over here https://github.com/HedvigInsurance/android/blob/03cecb9b5c957d0f15084f67cbe0e49d7e102d7d/app/feature/feature-forever/src/main/kotlin/com/hedvig/android/feature/forever/ForeverViewModel.kt#L47

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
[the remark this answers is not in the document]
StylianosGakis: Yeah it's good that it's contained in the module, but what I just saw here is that in reality this doesn't want to return CoInsured, this just wants to return first name + last name, because that's what it fetches. Then the fact that birthDate or hasMissingInfo are what they are, is just a consequence of returning a CoInsured, but in reality it doesn't know anything about those. The call site might know that this is needed only for cases where we don't know those other two anyway, but it is a coincidence.

To put it in another scenario, if you have a CoInsured already, where you have their ID and you also know that their birthDate is 2000-01-01.
Then you'll have something like this to properly merge the two CoInsured objects
```
val existingCoInsured = CoInsured(
  id = "id",
  firstName = null,
  lastName = null,
  birthDate = "2000-01-01",
  hasMissingInfo = true,
)
val fetchedCoInsured: CoInsured = fetchCoInsuredPersonalInformationUseCase().invoke(existingCoInsured.id).getOrNull()!!
val resultingCoInsured = existingCoInsured.copy(
  id = "id",
  firstName = fetchedCoInsured.firstName,
  lastName = fetchedCoInsured.lastName,
  birthDate = "2000-01-01",
  hasMissingInfo = true,
)
```
This puts the burden of knowing that the use case returns a CoInsured, but actually only the first name and last name are the two important things, and you *should not* use the other fields, as they are just some sane defaults, they don't mean more than that.

The UseCase can return `data class CoInsuredPersonalInformation(val firstName: String, val lastName: String)` instead, and then you got an easy way of know that well you just got the first name + last name, do whatever you wish with it, there's no way to use the wrong information coming from it.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
hugokallstrom: Thanks, fixing

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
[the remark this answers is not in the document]
hugokallstrom: Makes sense, changing 👍

PR #1892, opened 2023-11-27
Commit co-insured changes and show success screen

hugokallstrom:


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredAddOrRemoveDestination.kt
```diff
@@ -117,6 +124,11 @@ private fun EditCoInsuredScreen(
 
       is EditCoInsuredState.Loaded -> {
         val coroutineScope = rememberCoroutineScope()
+
+        if (uiState.contractUpdateDate != null) {
+          onCompleted(uiState.contractUpdateDate)
+        }
```
StylianosGakis, line 130: This may trigger more than once since we're doing it in composition. Consider doing it inside a `LaunchedEffect` for example

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
```diff
@@ -184,6 +192,31 @@ internal class EditCoInsuredPresenter(
       }
     }
 
+    if (commit) {
+      LaunchedEffect(commit) {
+        intentId?.let {
```
StylianosGakis, line 197: `intentId` is used inside the LaunchedEffect here, but it's not part of the keys of `LaunchedEffect` this will mean that if `intentId` changes, this effect will not be run again, is this something we want here?

StylianosGakis (approved):

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
hugokallstrom: Yeah this should be run based on if you want to commit the intentid, so we dont want the key to be the intentid itself

PR #1896, opened 2023-11-30
Clear cache when success to show updated contract 

hugokallstrom:


app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
```diff
@@ -104,7 +104,9 @@ private fun darkColorScheme(hedvigTonalPalette: HedvigTonalPalette) = darkColorS
 
 internal fun darkHedvigColorScheme(hedvigTonalPalette: HedvigTonalPalette) = HedvigMaterial3ColorScheme(
   containedButtonContainer = hedvigTonalPalette.greyscale25,
-  onContainedButtonContainer = hedvigTonalPalette.greyscale1000,
+  onContainedButtonContainer = hedvigTonalPalette.greyscale100,
+  secondaryContainedButtonContainer = hedvigTonalPalette.greyscale700,
```
StylianosGakis, line 108: Was this because 700 looked more close to design compared to 800?
If yes, can we double check with iOS? Because I got 800 from them when I asked for this.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/CoInsuredList.kt
```diff
@@ -32,11 +30,9 @@ internal fun CoInsuredList(
...
       )
     }
-    Divider(Modifier.padding(horizontal = 16.dp))
-    uiState.coInsured.forEachIndexed { index, coInsured ->
-      if (index != 0) {
-        Divider()
-      }
+
+    uiState.coInsured.forEach { coInsured ->
+      Divider()
```
StylianosGakis, line 35: Do we want a divider at the top even if there are no more coinsured rows above it?

StylianosGakis (approved):

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/CoInsuredList.kt
hugokallstrom: We can assume the member is not null here. The state could be modeled better by using a sealed class with less non-nullable fields though.

PR #1898, opened 2023-12-01
Refactor member reminder cards to be horizontal scrollable

hugokallstrom:


app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetMemberRemindersUseCase.kt
```diff
@@ -20,79 +24,89 @@ internal class GetMemberRemindersUseCaseImpl(
...
 
-data class ApplicableMemberReminders(
-  val connectPayment: MemberReminder.ConnectPayment? = null,
-  val upcomingRenewals: MemberReminder.UpcomingRenewals? = null,
-  val enableNotifications: MemberReminder.EnableNotifications? = null,
-  val coInsuredInfo: MemberReminder.CoInsuredInfo? = null,
-) {
-  val hasAnyReminders: Boolean
-    get() = connectPayment != null || upcomingRenewals != null || enableNotifications != null || coInsuredInfo != null
-}
```
StylianosGakis, line 88: This class existed to model the fact that it's not a possible state to have a list of `[MemberReminder.ConnectPayment, MemberReminder.UpcomingRenewals, MemberReminder.ConnectPayment]` for example.
With the new Api introduced into this PR, we're making it possible for us to pass a list with 2 ConnectPayment objects, which should never be something that should be possible.
Do we feel the need for this API to change like this to stop modeling this limitation in the type itself?

StylianosGakis (changes requested):
Before we make this API change that I mention in the comment, would be nice for me to understand if we have a strong reason for doing so that I am missing

app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetMemberRemindersUseCase.kt
hugokallstrom: I see, thats a valid point. `GetMemberRemindersUseCase` will still return 
```
data class MemberReminders(
  val connectPayment: MemberReminder.ConnectPayment? = null,
  val upcomingRenewals: List<MemberReminder.UpcomingRenewal>? = null,
  val enableNotifications: MemberReminder.EnableNotifications? = null,
  val coInsuredInfo: List<MemberReminder.CoInsuredInfo>? = null,
)
```
so I think that model is still clear. The view can then take a list of `MemberReminder`, and dont care about the ordering or det underlying construction of that list. It made it quite a bit easier to work with in the view.

PR #1899, opened 2023-12-05
Add feature flag for edit coinsured

hugokallstrom:


app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetNeedsCoInsuredInfoRemindersUseCase.kt
```diff
@@ -17,9 +19,15 @@ internal interface GetNeedsCoInsuredInfoRemindersUseCase {
 
 internal class GetNeedsCoInsuredInfoRemindersUseCaseImpl(
   private val apolloClient: ApolloClient,
+  private val featureManager: FeatureManager,
 ) : GetNeedsCoInsuredInfoRemindersUseCase {
   override suspend fun invoke(): Either<CoInsuredInfoReminderError, NonEmptyList<MemberReminder.CoInsuredInfo>> {
     return either {
+
+      if (!featureManager.isFeatureEnabled(Feature.EDIT_COINSURED)) {
+        raise(CoInsuredInfoReminderError.NoCoInsuredReminders)
```
StylianosGakis, line 28: Can this raise a new type of error which just mentions that the feature flag is off?
Can be a good thing to differentiate so that we know that nothing is wrong, only the feature flag being flipped.

PR #1903, opened 2023-12-07
Payment details UI

hugokallstrom:
Screenshot_20231207_113809


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
```diff
@@ -0,0 +1,260 @@
...
+          ),
+          underTextContent = null,
+        )
+
+        MemberCharge.MemberChargeStatus.UNKNOWN -> {}
+      }
+
+      Spacer(Modifier.height(32.dp))
+
+      paymentConnection?.connectionInfo?.let {
```
StylianosGakis, line 195: If paymentConncetion is null, so the last item here is just the spacer above, it will mean that we'll have the 32.dp space anyway + the 16.dp space from the padding applied to the outer column, making this have 48.dp of space at the bottom right?
Could we only apply this space inside the let, so that it only gets added if it's needed?

Always tricky with such conditional UI + spacings of course, this one just caught my eye because it might feel like way too much spacing at the bottom

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
```diff
@@ -0,0 +1,260 @@
...
+        )
+        Divider()
+      }
+    }
+  }
+}
+
+@Composable
+@HedvigPreview
+internal fun PaymentDetailsScreenPreview() {
```
StylianosGakis, line 248:
```suggestion
private fun PaymentDetailsScreenPreview() {
```
If you did end up adding the live template for this, try changing it to `private` since all previews don't need to be referenced by anything else anywhere.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentStatusCard.kt
```diff
@@ -0,0 +1,88 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.card.HedvigInfoCard
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.core.icons.Hedvig
+import com.hedvig.android.core.icons.hedvig.normal.WarningFilled
+
+@Composable
+fun PaymentStatusCard(
```
StylianosGakis, line 33:
```suggestion
internal fun PaymentStatusCard(
```

StylianosGakis (approved):

app/feature/feature-connect-payment-trustly/src/main/kotlin/com/hedvig/android/feature/connect/payment/trustly/data/TrustlyCallback.kt
```diff
@@ -15,6 +15,6 @@ internal interface TrustlyCallback {
 internal class TrustlyCallbackImpl(
   hedvigBuildConstants: HedvigBuildConstants,
 ) : TrustlyCallback {
-  override val successUrl: String = "${hedvigBuildConstants.deepLinkHost}/payment-success"
-  override val failureUrl: String = "${hedvigBuildConstants.deepLinkHost}/payment-failure"
+  override val successUrl: String = "https://${hedvigBuildConstants.deepLinkHost}/payment-success"
+  override val failureUrl: String = "https://${hedvigBuildConstants.deepLinkHost}/payment-failure"
```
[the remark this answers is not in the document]
hugokallstrom: No, got an error from backend actually

PR #1905, opened 2023-12-07
Payment history

hugokallstrom:
Screenshot_20231207_150507

### DIFF
--- a/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
+++ b/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
@@ -77,7 +77,7 @@ internal fun PaymentDetailsDestination(
 }
 
 @Composable
-internal fun MemberChargeDetailsScreen(
+private fun MemberChargeDetailsScreen(
   memberCharge: MemberCharge,
   paymentOverview: PaymentOverview,
   selectedCharge: MemberCharge.ChargeBreakdown?,
@@ -217,9 +217,8 @@ internal fun MemberChargeDetailsScreen(
         MemberCharge.MemberChargeStatus.UNKNOWN -> {}
       }
 
-      Spacer(Modifier.height(32.dp))
-
       paymentOverview.paymentConnection?.connectionInfo?.let {
+        Spacer(Modifier.height(32.dp))
         HorizontalItemsWithMaximumSpaceTaken(
           startSlot = {
             Text(stringResource(id = R.string.PAYMENTS_PAYMENT_METHOD))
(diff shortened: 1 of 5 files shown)


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/HedvigDateTimeFormatter.kt
```diff
@@ -12,6 +12,14 @@ fun hedvigDateTimeFormatter(locale: Locale): DateTimeFormatter {
   return DateTimeFormatter.ofPattern("d MMM yyyy", locale)
 }
 
+/**
+ * Example output: "16 Jan"
+ */
+fun hedvigMonthDateTimeFormatter(locale: Locale): DateTimeFormatter {
```
StylianosGakis, line 18: This is not a blocker for this PR, but just writing it here so we take care of it eventually.

In the chat feature, I also needed like a bunch of different formatters, so I did this https://github.com/HedvigInsurance/android/blob/7c5bb59b2ff7128ab7c8a50ace63c254d2679dd4/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/HedvigDateTimeFormatter.kt#L44-L45 to make it easier to namespace them into `HedvigDateTimeFormatterDefaults`.
Taking inspiration from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Button.kt;l=293?q=ButtonDefaults as they do it in the material library.

We can probably after everything is merged and we have the time try to consolidate all of our date formatting usages inside there, to make it easy to reason about what formatting options we're using overall.

No action needs to be taken in this PR, that code was only merged with the chat feature. Just to keep in mind for the future.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
```diff
@@ -243,17 +270,46 @@ internal fun MemberChargeDetailsScreen(
...
+internal fun getNextCharge(memberCharge: MemberCharge, paymentOverview: PaymentOverview): MemberCharge? {
+  val index = (paymentOverview.pastCharges?.indexOf(memberCharge) ?: 0) - 1
+  return if (index < 0 && paymentOverview.memberCharge != null) {
+    paymentOverview.memberCharge
+  } else if (index >= 0) {
+    paymentOverview.pastCharges?.get(index)
+  } else {
+    null
+  }
+}
```
StylianosGakis, line 282: Do you think this piece of logic could live in the ViewModel instead? And the UI can just show the result on the screen directly without having to do these calculations itself?

StylianosGakis (approved):

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/details/PaymentDetailsDestination.kt
hugokallstrom: Good idea. I will fix it in the PRs above to avoid too many conflicts

PR #1907, opened 2023-12-08
Payment discounts overview

hugokallstrom:
Screenshot_20231208_063551

### DIFF
--- a/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/discounts/DiscountsDestination.kt
+++ b/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/discounts/DiscountsDestination.kt
@@ -1,0 +1,65 @@
+package com.hedvig.android.feature.payments2.discounts
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Divider
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import com.hedvig.android.core.designsystem.component.button.HedvigSecondaryContainedButton
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.core.ui.scaffold.HedvigScaffold
+import com.hedvig.android.feature.payments2.DiscountRow
+import com.hedvig.android.feature.payments2.data.Discount
+import com.hedvig.android.feature.payments2.discountsPreviewData
+import hedvig.resources.R
+
+@Composable
+fun DiscountsDestination(
+  discounts: List<Discount>,
+  navigateUp: () -> Unit,
+) {
+  HedvigScaffold(
+    topAppBarText = stringResource(id = R.string.PAYMENTS_DISCOUNTS_SECTION_TITLE),
+    navigateUp = navigateUp,
+  ) {
+    Column(modifier = Modifier.padding(horizontal = 16.dp)) {
+      if (discounts.isNotEmpty()) {
+        Spacer(modifier = Modifier.height(16.dp))
+        Text(stringResource(id = R.string.PAYMENTS_CAMPAIGNS_INFO_TITLE))
+        discounts.forEachIndexed { index, discount ->
+          DiscountRow(discount)
+          if (index < discounts.size - 1) {
+            Divider()
+          }
+        }
+      }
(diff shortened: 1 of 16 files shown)


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/discounts/DiscountsDestination.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.hedvig.android.core.designsystem.preview.HedvigPreview
+import com.hedvig.android.core.designsystem.theme.HedvigTheme
+import com.hedvig.android.core.ui.scaffold.HedvigScaffold
+import com.hedvig.android.feature.payments2.DiscountRow
+import com.hedvig.android.feature.payments2.data.Discount
+import com.hedvig.android.feature.payments2.discountsPreviewData
+import hedvig.resources.R
+
+@Composable
+fun DiscountsDestination(
```
StylianosGakis, line 25:
```suggestion
internal fun DiscountsDestination(
```

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/discounts/DiscountsDestination.kt
```diff
@@ -0,0 +1,65 @@
...
+        text = stringResource(id = R.string.PAYMENTS_ADD_CAMPAIGN_CODE),
+        onClick = { }, // TODO
+      )
+    }
+  }
+}
+
+@Composable
+@HedvigPreview
+internal fun PaymentDetailsScreenPreview() {
```
StylianosGakis, line 54:
```suggestion
private fun PaymentDetailsScreenPreview() {
```

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/navigation/PaymentsGraph.kt
```diff
@@ -65,8 +68,8 @@ fun NavGraphBuilder.paymentsGraph(
     composable<PaymentsDestinations2.History> {
       PaymentHistoryDestination(
         paymentOverview = paymentOverview.copy(
-          paymentConnection = null,
-        ), // Payment connection is not valid for historic payments
+          paymentConnection = null, // Payment connection is not valid for historic payments
```
StylianosGakis, line 71: Could this be a hint that we want a different type to be sent over at `PaymentDestination2.History`, if we need to manually in the UI be careful to make something null, but still have access to it? Even though we may not really need to have access to it in here?

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/navigation/PaymentsGraph.kt
```diff
@@ -36,7 +37,9 @@ fun NavGraphBuilder.paymentsGraph(
           navigator.navigateUnsafe(PaymentsDestinations2.History(paymentOverview))
         },
         onChangeBankAccount = navigateToConnectPayment,
-        onDiscountClicked = { }, // TODO
+        onDiscountClicked = { discounts ->
+          navigator.navigateUnsafe(PaymentsDestinations2.Discounts(discounts))
```
StylianosGakis, line 41: The `unsafe` variant exists for scenarios where we do not want to drop the event in case there is an already ongoing navigation going on.
To give examples:

We usually use the `safe` alternative, so that if there's an ongoing navigation, we simply do not do anything. This is important particularly when we know that this can be triggered more than once very quickly for some reason.

If we want to navigate as a response to an event coming from the ViewModel, where we know this will only happen once, we are okay with potentially doing `navigateUnsafe` since we want this navigation to happen even if they for example have a bottom sheet showing, which would put the lifecycle state in Paused, which would mean the `navigate` function would not perform the navigation. Because that's the difference between the normal one and the `unsafe` one. The normal one checks if the lifecycle state is just resumed (aka nothing is going on, no navigation, no bottom sheet is showing, the app is not on the background, etc.), while unsafe just navigated regardless.

However if this even happens on a click of a button, as we do here, and then we do `navigateUnsafe`, it means that if someone clicks this button twice really fast, we will navigate there two times actually. And pressing back would bring you back to the same screen and then you'd have to press back again.

StylianosGakis (approved):

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/navigation/PaymentsGraph.kt
hugokallstrom: Yes you are right. We should probably observe the state from the view model more instead of passing around the objects. Ill make some changes on a PR on top of everything to avoid too many conflicts.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments2/navigation/PaymentsGraph.kt
hugokallstrom: Awesome, thanks

PR #1918, opened 2023-12-11
String builders for name and ssn

hugokallstrom:


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
```diff
@@ -316,6 +319,12 @@ internal class EditCoInsuredPresenter(
     } else {
       (listState.coInsured + coInsured).toImmutableList()
     }
+
+  private fun getPaddedSsn(ssn: String) = if (ssn.length == 10) {
+    "19$ssn" // Need to add century if not included in SSN
```
StylianosGakis, line 324: We have people who are over 18 y.o born at 20xx nowadays. We either must make this a bit more sophisticated, like adding 20 for ssn starting in like [00..10] and 19 for [11..99] , which is still not 100% correct, or explain in the UI that we need the entire input, and force them to give us 12 characters instead.

StylianosGakis (approved):

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/EditCoInsuredPresenter.kt
hugokallstrom: Good point, got code from iOS 👍

PR #1920, opened 2023-12-12
QA payments

hugokallstrom:


app/core/core-ui-data/src/main/kotlin/com/hedvig/android/core/uidata/UiMoney.kt
```diff
@@ -12,7 +12,7 @@ import octopus.type.CurrencyCode
 @Serializable
 data class UiMoney(val amount: Double, val currencyCode: CurrencyCode) {
   @Transient
-  val decimalFormatter: DecimalFormat = DecimalFormat("#")
+  val decimalFormatter: DecimalFormat = DecimalFormat("")
```
StylianosGakis, line 15: What does this change do here?
Before we basically made it so that it only shows the decimal part of the number, does this now just raw print the number as it is in the UI?

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/bottomsheet/HedvigBottomSheet.kt
```diff
@@ -0,0 +1,87 @@
...
+        .padding(horizontal = 24.dp),
+    )
+    Spacer(Modifier.height(16.dp))
+    HedvigTextButton(
+      text = stringResource(id = R.string.general_close_button),
+      onClick = {
+        scope.launch {
+          sheetState.hide()
+        }.invokeOnCompletion {
+          onDismissRequest()
```
StylianosGakis, line 81: In this position, I'd say this is more of a "onDismissed" rather than a request. Since it's already done. Sometimes some of these APIs give a request callback so that one can decide what happens right as soon as the dismissal has started, but this is not the case here. We can change the name to avoid such confusions.

StylianosGakis (approved):
A question for the UiMoney formatter because it can be tricky if we get a decimal like 13.333333334 at some point and we just print that out to the UI as-is

app/core/core-ui-data/src/main/kotlin/com/hedvig/android/core/uidata/UiMoney.kt
hugokallstrom: We previously rounded all amounts to nearest integer. I think we should just display the amount exactly as we get it from back-end (which should be in the format of #.##) - but if we for some reason get something else its better to show it than to round it.

PR #1921, opened 2023-12-12
Hide co-insured info if contract does not support editing co-insured

hugokallstrom:


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
```diff
@@ -224,18 +224,21 @@ internal fun YourInfoTab(
...
-    CoInsuredSection(
-      coInsuredList = coInsured,
-      contractHolderDisplayName = contractHolderDisplayName,
-      contractHolderSSN = contractHolderSSN,
-      allowEditCoInsured = allowEditCoInsured,
-      onMissingInfoClick = onMissingInfoClick,
-      modifier = Modifier.padding(horizontal = 16.dp),
-    )
+
+    if (allowEditCoInsured) {
```
StylianosGakis, line 231: Yup also makes sense to hide all of it too. I did the minimum change here before just to keep people from going in to the coinsured flow when fflag was off.

StylianosGakis (approved):

PR #1923, opened 2023-12-13
Show uploaded files for a claim in claim details

hugokallstrom:
Screenshot_20231213_095812


app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimAsyncImage.kt
```diff
@@ -0,0 +1,69 @@
...
+import coil.request.ImageRequest
+import com.hedvig.android.core.designsystem.material3.rememberShapedColorPainter
+import com.hedvig.android.core.designsystem.material3.squircleMedium
+import com.hedvig.android.core.ui.layout.adjustSizeToImageRatio
+import com.hedvig.android.placeholder.PlaceholderHighlight
+import com.hedvig.android.placeholder.fade
+import com.hedvig.android.placeholder.placeholder
+
+@Composable
+fun ClaimAsyncImage(model: Any, imageLoader: ImageLoader, modifier: Modifier = Modifier, cacheKey: String? = null) {
```
StylianosGakis, line 24:
```suggestion
internal fun ClaimAsyncImage(model: Any, imageLoader: ImageLoader, modifier: Modifier = Modifier, cacheKey: String? = null) {
```

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
```diff
@@ -104,46 +126,100 @@ private fun ClaimDetailScreen(
...
+            shape = MaterialTheme.shapes.squircleMedium,
+            color = MaterialTheme.colorScheme.surface,
           )
+          .clickable {
+            openUrl(it.url)
+          },
+        contentAlignment = Alignment.Center,
+      ) {
+        if (it.mimeType.contains("image")) {
+          ClaimAsyncImage(
```
StylianosGakis, line 189: You'll probably want to pass in a cacheKey too, if the file contains some sort of unique ID that will be perfect.
Because otherwise each time from the backend we get a new URL to the image itself, and we end up having coil save into cache all those images (of the same image) in the cache, and we also never end up loading it quickly without doing another network request.

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimAsyncImage.kt
```diff
@@ -0,0 +1,69 @@
...
+          loadedImageIntrinsicSize.value = IntSize(
+            state.result.drawable.intrinsicWidth,
+            state.result.drawable.intrinsicHeight,
+          )
+          state
+        }
+      }
+    },
+    modifier = modifier
+      .adjustSizeToImageRatio(getImageSize = { loadedImageIntrinsicSize.value })
```
StylianosGakis, line 59: Yeah interesting idea to reuse this from chat.
I think that worked great there with the constraints that we had which was that it may take various widths too. In this scenario as far as I can tell we'll have a fixed width, since each item will be taking 1/3rd of the screen.

May I propose an alternative instead here. Could you look into using https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/lazy/staggeredgrid/LazyStaggeredGrid.kt;l=39?q=LazyStaggeredGrid&sq= instead?
And try not to give them any constraints vertically, as they will always have a fixed width, so they should be able to adjust their height as they wish. And using the Staggered grid instead will make the things fall into place naturally.

With all this said, I recognize that `adjustSizeToImageRatio` is not a very descriptive name, and maybe I should've picked a better name, but my idea was that it's only gonna be used in chat so I didn't need to make the name super big :D Hence it has some constants in there, especially around
```
private const val MaximumImageHeight = 250
private const val PreferredImageHeight = 180
private const val MinimumImageHeight = 100
private const val MinimumWidthTakenPercentage = 0.5f
```
if you see in there. Which describe that the images should be taking from 50% of the width to 100%. And with extra constraints of the heights being forced into 100dp..250dp.

StylianosGakis (approved):
Super nice to be able to add this functionality this easily there.
I have just one thing to consider trying which is to use LazyVerticalStaggeredGrid in there instead, potentially.

And to consider not reusing the ratio adjusting modifier there, since the width constraint is fixed, so there's not much that can be done on the height. Either we let the image adjust the height as it wishes, or we set the height to a fixed value, and then we crop the image so that it always fits the width but you can't see the entirety of it.

With this done, it I think shows that we really have the need here to be able to click on the image and have it come full-screen to be able to properly see what it shows. Something I need to do for chat too anyway.

hugokallstrom:
> Super nice to be able to add this functionality this easily there. I have just one thing to consider trying which is to use LazyVerticalStaggeredGrid in there instead, potentially.
> 
> And to consider not reusing the ratio adjusting modifier there, since the width constraint is fixed, so there's not much that can be done on the height. Either we let the image adjust the height as it wishes, or we set the height to a fixed value, and then we crop the image so that it always fits the width but you can't see the entirety of it.
> 
> With this done, it I think shows that we really have the need here to be able to click on the image and have it come full-screen to be able to properly see what it shows. Something I need to do for chat too anyway.

Great, thanks for the feedback. Regarding possibility to view image full screen - I am just opening the file URL, so if its an image it will show in the browser. If its another type of file it will download 👍

PR #1924, opened 2023-12-13
Emergency new UI and common claims new UI

StylianosGakis:
Common "emergency" screen for both claim's flow deflect and when clicking on the bottom sheet in home for sick abroad.

image

The common claims changes are the same as here https://github.com/HedvigInsurance/ugglan/pull/1148

This takes the opportunity to also remove another Activity which was previously used for common-claims (We got 4 more to go, eventually when the time allows). Which deletes some more XML code.


app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
```diff
@@ -130,6 +132,8 @@ internal fun lightHedvigColorScheme(hedvigTonalPalette: HedvigTonalPalette) = He
   onContainedButtonContainer = hedvigTonalPalette.greyscale1000,
   secondaryContainedButtonContainer = hedvigTonalPalette.greyscale100,
   onSecondaryContainedButtonContainer = hedvigTonalPalette.greyscale1000,
+  alwaysBlackContainer = hedvigTonalPalette.greyscale900,
```
hugokallstrom, line 135: Should be grayscale1000 according to design, for both light and dark

hugokallstrom (approved):
Looks great. We need to be careful though so that we dont miss any common claims coming from PCMS which have a layout other than "title and bullet points". We should revisit this API when we have time.

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
StylianosGakis: Yeap it is 1000 instead of 900, fixing that.
I posted a comment here https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1702551037977529 regarding dark mode because I was not so sure that 1000 is at visible with the dark background. And iOS does this differently too

PR #1926, opened 2023-12-14
Cleanup/unused dependencies and modules

StylianosGakis:
Bunch of now unused dependencies.
One really nice one is that material2 is now completely gone, making it easier to not accidentally reference something from there instead of m3 for example, among other things


hugokallstrom (approved):
🧹 👍

PR #1929, opened 2023-12-15
Add the "delivered" mark at the end of the last member sent message

StylianosGakis:
Keep the ChatMessage model as-is, and for the UI needs introduce a new one named UiChatMessage, which contains the domain model plus some UI-only information regarding which one of the messages is the one that should be marked as delivered.

The requirements for this is that when the member sends a message, when it is successfully sent to the backend, it should be marked as delivered. This can only be shown in the last delivered message, as all the ones above it are then expected to be also delivered. If there are any failed messages in the process, they do not interfere with this marking. If there are any messages coming from Hope, those are never expected to be marked as delivered, since they are incoming messages.

image

image

image

image

image

image


sladan-hedvig (approved):
Looks really awesome!

PR #1934, opened 2023-12-18
Remove custom jsonObjectOf()

StylianosGakis:
We can just use the built-in solution from kotlinx.serialization.json which we can be much more confident in


hugokallstrom (approved):
👍

PR #1936, opened 2023-12-18
Upload files for claim

hugokallstrom:
https://github.com/HedvigInsurance/android/assets/5929732/10f92fc9-6b53-4063-80e7-5c42532bf3e8


app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/data/UploadFileUseCase.kt
```diff
@@ -0,0 +1,87 @@
...
+  private val contentResolver: ContentResolver,
+) : UploadFileUseCase {
+  override suspend fun invoke(url: String, uri: Uri): Either<ErrorMessage, UploadSuccess> = either {
+    val claimId = Uri.parse(url).getQueryParameter("claimId") ?: raise(ErrorMessage("No claim id found in url"))
+    val result = uploadFileService.uploadFile(
+      claimId = claimId,
+      file = MultipartBody.Part.createFormData(
+        name = "files",
+        filename = fileService.getFileName(uri) ?: "media",
+        body = object : RequestBody() {
```
StylianosGakis, line 42: I see that you use the same function here for sending documents, but also for sending photos that one may snap right at the moment.
In the chat code, I had this snippet https://github.com/HedvigInsurance/android/blob/ba69a697a6a38147da88510f53d08d2b107ce7c4/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/ChatRepositoryImpl.kt#L124-L184 where `uploadFile` did things a tiny bit different for the photo scenario since we could just do a `uri.toFile()` to it and then upload it using the service + Retrofit.

But if this just works without it too, then this is a great opportunity to remove some extra code from inside ChatRepository too. Potentially share this part of the code between these two places, while still letting them upload to different services, as they need to. And even if those can not be shared, maybe the ChatRepository itself can reuse the file upload function for documents with the one which uploads the photo which was just taken.

But all this would then need to be extracted to a separate module which both of these can see, and would require some refactoring. Maybe not scoped for this PR, but it will long-term be a good idea not to have more or less the same code in two places.

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
```diff
@@ -127,19 +155,123 @@ private fun ClaimDetailScreen(
...
 ) {
-  LazyVerticalStaggeredGrid(
-    columns = StaggeredGridCells.Fixed(3),
+  var showFileTypeSelectBottomSheet by remember { mutableStateOf(false) }
+  val sheetState = rememberModalBottomSheetState()
+
+  val photoCaptureState = rememberPhotoCaptureState(appPackageId = appPackageId) { uri ->
+    logcat { "ChatFileState sending uri:$uri" }
+    onSendFile(uri)
+  }
```
StylianosGakis, line 171: Nice that this could just be reused too, I hope it was easy to plug-in

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
```diff
@@ -127,19 +155,123 @@ private fun ClaimDetailScreen(
...
+          TextButton(
+            onClick = {
+              filePicker.launch("*/*")
+              showFileTypeSelectBottomSheet = false
+            },
+            modifier = modifier
+              .fillMaxWidth()
+              .padding(horizontal = 16.dp),
+            shape = MaterialTheme.shapes.squircleMedium,
+          ) {
```
StylianosGakis, line 248: How do we want the ripple to play out here?
This view kinda looks like to me that clicking it would get the ripple to go all the way to the edges like in a typical list.
However if we want it to look/feel like a button which does not go all the way to the edges, maybe we can reuse the design that the bottom sheet in the home screen has? Where the buttons there look more like cards, with a slightly different background? What do you think?

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
```diff
@@ -127,19 +155,123 @@ private fun ClaimDetailScreen(
...
+  }
+
+  LazyVerticalGrid(
+    columns = GridCells.Fixed(3),
     horizontalArrangement = Arrangement.spacedBy(8.dp),
-    verticalItemSpacing = 8.dp,
+    verticalArrangement = Arrangement.spacedBy(8.dp),
     modifier = modifier
       .windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom))
       .padding(horizontal = 16.dp),
```
StylianosGakis, line 272: LazyVerticalGrid should have a `contentPadding` parameter where we should pass these insets to. It may be something like 
```suggestion
    contentPadding = PaddingValues(horizontal = 16.dp) + WindowInsets.safeDrawing
      .only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom)
```

We do something similar here https://github.com/HedvigInsurance/android/blob/85c852c395ff84b14e7010f0c3cba49d17138821/app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt#L328-L331

I am mentioning this since otherwise the screen is cut-off at the bottom where the navigation items are.

StylianosGakis (approved):
Looks very nice, and great usage of some of the things that existed for the chat which shared some of this functionality.

Just remember to add the URL to the backend service for staging/production too

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/data/UploadFileUseCase.kt
```diff
@@ -0,0 +1,87 @@
...
+  suspend fun invoke(url: String, uri: Uri): Either<ErrorMessage, UploadSuccess>
+}
+
+internal data class UploadFileUseCaseImpl(
+  private val uploadFileService: UploadFileService,
+  private val fileService: FileService,
+  private val contentResolver: ContentResolver,
+) : UploadFileUseCase {
+  override suspend fun invoke(url: String, uri: Uri): Either<ErrorMessage, UploadSuccess> = either {
+    val claimId = Uri.parse(url).getQueryParameter("claimId") ?: raise(ErrorMessage("No claim id found in url"))
```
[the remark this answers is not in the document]
hugokallstrom: You are right, this is currently the same as the claim id. I will keep it like this though since this URL is set on the claim, so it may change in the future.

PR #1939, opened 2023-12-18
Hook up unread chat message information to HomePresenter

StylianosGakis:
Will be used to render if there is an unread message next to the chat bubble


hugokallstrom (approved):
👍

PR #1954, opened 2024-01-19
Center quick links and common claims text

hugokallstrom:
Also open web url instead of defaulting to home deeplink.

### DIFF
--- a/app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
+++ b/app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
@@ -16,18 +16,22 @@ interface HedvigDeepLinkContainer {
   val helpCenterCommonTopic: String // A common topic inside the help center
   val helpCenterQuestion: String // A specific question inside the help center
 
-  fun toDeepLink(url: String) = when {
-    url.contains("home") -> home
-    url.contains("insurances") -> insurances
-    url.contains("forever") -> forever
-    url.contains("profile") -> profile
-    url.contains("eurobonus") -> eurobonus
-    url.contains("chat") -> chat
-    url.contains("connectPayment") -> connectPayment
-    url.contains("directDebit") -> directDebit
-    url.contains("payments") -> payments
-    url.contains("helpCenter") -> helpCenter
-    else -> home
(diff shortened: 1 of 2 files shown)


app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -16,18 +16,22 @@ interface HedvigDeepLinkContainer {
...
+      url.contains("chat") -> chat
+      url.contains("connectPayment") -> connectPayment
+      url.contains("directDebit") -> directDebit
+      url.contains("payments") -> payments
+      url.contains("helpCenter") -> helpCenter
+      else -> home
+    }
+  } else {
+    url
   }
```
StylianosGakis, line 35: Aha so this is a function which kinda converts an arbitrary string into our own deep link if it has "link" inside of it.

And the reason we can't just let it use the link as-is, is because in lokalise we get `[Make a claim](https://hedvig.page.link/home)` but we want to translate this into whatever it should be for dev/staging builds. Otherwise it would just work without having to do this check at all right?

I think I get this now if that's the case, so yeah sounds ok to default to the url that was passed in if that's how it's used.

One thing that kinda threw me off a bit in the beginning is that his is a function that is on the `HedvigDeepLinkContainer`, but we do not really need an instance of it in order to run this logic right? We could put this in a companion object and have it only try to map to a compatible Hedvig deep link if it matches, otherwise it can default to null.

Then on the call-site, we can call this, and if we get null back we can default to whatever it is that we want to default to. Otherwise `toDeepLink` does not really describe without looking at the implementation what it does if the url is not one that matches our deep links

```suggestion
fun hedvigLinkFromUrlOrNull(url: String): String? {
  if (!url.contains("link")) return null
  return when {
    url.contains("home") -> home
    url.contains("insurances") -> insurances
    url.contains("forever") -> forever
    url.contains("profile") -> profile
    url.contains("eurobonus") -> eurobonus
    url.contains("chat") -> chat
    url.contains("connectPayment") -> connectPayment
    url.contains("directDebit") -> directDebit
    url.contains("payments") -> payments
    url.contains("helpCenter") -> helpCenter
    else -> home
  }
}
```

This way we can call this like this or something
`HedvigDeepLinkContainer.hedvigLinkFromUrlOrNull(url) ?: url`

StylianosGakis (approved):

PR #1955, opened 2024-01-19
Remove zignsec and use generic auth for DK and NO.

hugokallstrom:


app/app/src/main/kotlin/com/hedvig/app/feature/genericauth/SsnInputScreen.kt
```diff
@@ -0,0 +1,176 @@
...
+  onSubmit: () -> Unit,
+  error: String?,
+) {
+  HedvigTextField(
+    value = input,
+    onValueChange = { newInput ->
+      val maxLengthAllowed = when (market) {
+        Market.NO -> 11
+        Market.DK -> 10
+        Market.SE -> 10
```
StylianosGakis, line 95: Do we not tend to use the full 12 character SSN lately for such places? Would this work with 12 instead?

Oh also actually we should never end up in this screen with market being SE right? Because we default to another screen instead? If that's the case, we should just `error()` here and let it crash if it's the case, so we actually know that we got an important bug to fix.

app/app/src/main/kotlin/com/hedvig/app/feature/genericauth/SsnInputScreen.kt
```diff
@@ -0,0 +1,176 @@
...
+      Text(
+        text = stringResource(R.string.zignsec_login_screen_title),
+        style = MaterialTheme.typography.headlineMedium,
+      )
+      Spacer(Modifier.height(20.dp))
+      SSNTextField(
+        input = emailInput,
+        market = market,
+        onInputChanged = onInputChanged,
+        onSubmit = onSubmitEmail,
```
StylianosGakis, line 66: We call `onSubitEmail` with an `ssn` here as I understand it. Is this WAI and is the name just a bit confusing? If so we might wanna rename so we do not fall into wrong assumptions when we see `onSubmitEmail` higher up and assume it's actually an email.

app/app/src/main/AndroidManifest.xml
```diff
@@ -54,10 +54,6 @@
           android:scheme="https" />
       </intent-filter>
     </activity>
-    <activity
-      android:name=".feature.zignsec.SimpleSignAuthenticationActivity"
```
StylianosGakis, line 58: Almost there, only a few to go!

StylianosGakis (approved):

app/app/src/main/kotlin/com/hedvig/app/feature/genericauth/SsnInputScreen.kt
hugokallstrom: Thats a good point. Will change

PR #1956, opened 2024-01-22
Parse markdown in help center to allow for better styling

hugokallstrom:
Its now no longer possible to open deep links in dev or staging, since we get the urls from lokalise. Im not sure how to get around this without introducing too much complexity on the client... I think we just have to be careful and have the correct deeplinks in lokalise.


app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -247,14 +251,18 @@ private fun HelpCenterHomeScreen(
...
           items = questions,
           itemText = { resources.getString(it.questionRes) },
           onClickItem = { onNavigateToQuestion(it) },
         )
-        Spacer(Modifier.height(24.dp))
+        Spacer(Modifier.weight(1f))
+        Spacer(Modifier.height(56.dp))
+        StillNeedHelpSection(openChat = {
+          openChat()
+        })
```
StylianosGakis, line 265: nit:
```suggestion
        StillNeedHelpSection(openChat = openChat)
```

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/ui/TextWithClickableDeepLink.kt
```diff
@@ -1,82 +0,0 @@
...
-  ),
-  linkMatcher: Regex = Regex("\\[([^\\[]+)](\\(.*\\))"),
-  onClick: (url: String) -> Unit = with(LocalContext.current) {
-    { url -> startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(url))) }
-  },
-) {
-  val annotatedText = buildAnnotatedString {
-    val offset = linkMatcher.findAll(text).fold(0) { offset: Int, match: MatchResult ->
-      val linkText = match.groups[1]?.value
-      val url = match.groups[2]?.value?.removeSurrounding("(", ")")
```
StylianosGakis, line 43: How is the link actually rendered now with the markdown library for the questions when the URL is there? Does it render just fine?

And yeah I see what you mean now, if we want to have the link works for dev/staging we'd have to take the string and manipulate it for those environments, probably with something very close to what we're doing here.
Ok to just let this be for now if this works anyway. We can sorta test that the links work in a way since the dev/staging build should deep link us into the real app on our device if we have it installed and that versions supports that link anyway.

StylianosGakis (approved):

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/ui/TextWithClickableDeepLink.kt
hugokallstrom: They render fine!

PR #1957, opened 2024-01-22
Fix claim flow loading button

hugokallstrom:


StylianosGakis (commented):
Aha, we rely on the button itself being fed the loading state, I see.
I do wonder if this should be a concern for the button to solve, and not for the VM to dictate. If we are in a situation where we do not use that button with the loading state, then we'd end up in a scenario where you can submit the thing as many times as you click the button really fast.
Perhaps the solution could be that the logic stays in the VM so that we are sure that the logic works as intended regardless of how the UI is implemented.

And then to fix the button looking in a disabled state while this is happening, we can make the button itself not look disabled when the loading state is on.
Basically something like `Button(disabled = state.disabled && !isLoading)`, to ensure that disabled is shown when it needs to be shown, but it is *not* rendered that way when it's loading, and it has the spinner inside.

What do you think?

hugokallstrom:
> Aha, we rely on the button itself being fed the loading state, I see. I do wonder if this should be a concern for the button to solve, and not for the VM to dictate. If we are in a situation where we do not use that button with the loading state, then we'd end up in a scenario where you can submit the thing as many times as you click the button really fast. Perhaps the solution could be that the logic stays in the VM so that we are sure that the logic works as intended regardless of how the UI is implemented.
> 
> And then to fix the button looking in a disabled state while this is happening, we can make the button itself not look disabled when the loading state is on. Basically something like `Button(disabled = state.disabled && !isLoading)`, to ensure that disabled is shown when it needs to be shown, but it is _not_ rendered that way when it's loading, and it has the spinner inside.
> 
> What do you think?

We have this in the button composable: 
```
    onClick = {
      if (enabled && !isLoading) {
        onClick()
      }
    }
```
so that you cannot submit if its loading. In my head a buttons enabled state is separate from the loading state - and looking at the button implementation in material3 I think it make sense: "enabled - controls the enabled state of this button. When false, this component will not respond to user input, and it will appear visually disabled and disabled to accessibility services.".

StylianosGakis:
Yeah we have that check in `HedvigContainedButton`, but what if we are not using that as the action for a particular UI?

In this line https://github.com/HedvigInsurance/android/blob/519e0c2b466f877d34076d2d680d8370b6665674/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/button/HedvigContainedButton.kt#L104 if we change this to `enabled = enabled && !isLoading,` then we get the functionality we're looking for here.

The UI state in any case does not really say anything about a button being disabled or not. What is says is that currently, this should not be able to be submitted, hence the name `canSubmit` instead of `isButtonDisabled`. Then the UI can interpret this as it wants. In our case it can interpret it as that the button is disabled in the scenarios where we are in an idle state, aka nothing is loading. But when one can not submit, but we're also loading, that means that the button should know that this means that it's now showing the spinner, and it does not need to look disabled and unclickable.

That's more of where I was going with this comment.

hugokallstrom:
I see your point. In this case we would have to do `enabled = enabled || isLoading` in `HedvigContainedButton`, right? Otherwise the button will be greyed in when we load.

StylianosGakis:
Yes exactly. Since HedvigContainedButton (And a couple more in there) are doing these extra checks on `onClick` themselves then they should be able to do this check too, it's [here](https://github.com/HedvigInsurance/android/blob/519e0c2b466f877d34076d2d680d8370b6665674/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/button/HedvigContainedButton.kt#L104) along with a couple more inside that file afaik.

PR #1958, opened 2024-01-22
Fix language and market

StylianosGakis:
Before, sometimes on startup there'd be no language properly set and
we'd default back to en_SE, despite someone already having chosen a
different language, but in a way that was not understood by the
"migration" function.

Now we always run a "migration" function on startup, checking the
current market and setting the right language directly, or defaulting to
SE market if none was selected, and consequently selecting a language
in the process.

This should now work for all API levels in a consistent way, which will
not be cleared after the app is closed and started again.


app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/marketing/MarketingDestination.kt
```diff
@@ -183,7 +185,7 @@ private fun MarketingScreen(
             modifier = Modifier.testTag("login_button"),
           )
           HedvigTextButton(
-            text = "Get a price quote",
```
hugokallstrom, line 186: Wops. Good find.

app/market/market-core/src/main/kotlin/com/hedvig/android/market/MarketStorage.kt
```diff
@@ -11,14 +11,14 @@ import kotlinx.coroutines.flow.map
...
   private val datastore: DataStore<Preferences>,
 ) {
-  val market: Flow<Market> = datastore.data
+  val market: Flow<Market?> = datastore.data
     .map { preferences ->
       val storedMarket: Market? = preferences[marketPreferencesKey]?.let { Market.valueOf(it) }
       if (storedMarket != null) {
         return@map storedMarket
       } else {
         logcat { "No market stored in datastore, falling back to SE" }
```
hugokallstrom, line 20: Change this log message

hugokallstrom (approved):
Not trivial :D Good job!

app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/marketing/MarketingDestination.kt
StylianosGakis: Yeah, I have no idea how this one slipped through, I see that I did some refactoring some months ago and I seem to have just never updated the localization, oops.

app/market/market-core/src/main/kotlin/com/hedvig/android/market/MarketStorage.kt
StylianosGakis: Nice catch, thanks!

PR #1960, opened 2024-01-23
Chat banner

hugokallstrom:


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatPresenter.kt
```diff
@@ -111,11 +115,12 @@ internal class ChatPresenter(
       },
     )
     LaunchPeriodicMessagePollsEffect(
-      reportNextUntilFromPolling = { nextUntil: Instant ->
+      reportNextUntilFromPolling = { result: ChatMessagesResult ->
```
StylianosGakis, line 118: Can we also rename this to help describe a bit better that it returns more than just the `nextUntil` after these changes?
```suggestion
      reportChatMessagesResultFromPolling = { result: ChatMessagesResult ->
```

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/ChatRepositoryImpl.kt
```diff
@@ -76,6 +76,7 @@ internal class ChatRepositoryImpl(
         messages = result.chat.messages.mapNotNull { it.toChatMessage() },
         nextUntil = result.chat.nextUntil,
         hasNext = result.chat.hasNext,
+        informationMessage = null, // TODO query from schema
```
StylianosGakis, line 79: Will backend return this message all the time? Is it something that is not dismissible in general? I suppose it's only to be used in situations where something is wrong right? Like a downtime, that the chat is closed at this time of day etc.?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatBanner.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.halilibo.richtext.ui.material3.RichText
+import com.hedvig.android.core.designsystem.component.card.HedvigInfoCard
+import com.hedvig.android.core.designsystem.material3.infoContainer
+import com.hedvig.android.core.designsystem.material3.infoElement
+import com.hedvig.android.core.designsystem.material3.onInfoContainer
+import com.hedvig.android.core.icons.Hedvig
+import com.hedvig.android.core.icons.hedvig.normal.InfoFilled
+
+@Composable
+fun ChatBanner(text: String, modifier: Modifier = Modifier) {
```
StylianosGakis, line 28:
```suggestion
internal fun ChatBanner(text: String, modifier: Modifier = Modifier) {
```

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatLoadedScreen.kt
```diff
@@ -142,6 +142,10 @@ internal fun ChatLoadedScreen(
           .clearFocusOnTap()
           .nestedScroll(topAppBarScrollBehavior.nestedScrollConnection),
       )
+      uiState.bannerText?.let {
+        Divider(Modifier.fillMaxWidth())
+        ChatBanner(text = it)
```
StylianosGakis, line 147:
```suggestion
        ChatBanner(text = it, Modifier.fillMaxWidth())
```
Could it be that this also needs to be told to go full width? How does it function if the content passed into it only takes some part of the width available. Does it expand already due to some other thing inside ChatBanner perhaps?

StylianosGakis (approved):
Looks super good, great work!

I'd double check with what happens if the markdown text passed into it is barely rendering anything (or if it's nothing at all) to make sure that if the background is still at least going all the way to the edges with full width, or just pass fillMaxWidth regardless to be safe in all cases.

I do wonder if we are ever in the risk that the text provided into the ChatBanner is not null but it's an empty String, which would result in a bit of an odd looking UI perhaps

PR #1962, opened 2024-01-23
Pass context to chat in order to tag conversations

hugokallstrom:


StylianosGakis (approved):
Really nice how it all turned out to be I think, I like the idea of adding the topic to the question itself, then calling it with the right context is really simple.

PR #1963, opened 2024-01-23
Fix chat scrolling to top bug

StylianosGakis:
Initially, when there is nothing in the lazy list, the messages are
empty, and what shows in the lazy list is the loading indicator and a
spacer here.
https://github.com/HedvigInsurance/android/blob/c1d80d089e9cf7af184e94503fddcc09c70f67f5/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatLoadedScreen.kt#L277-L308

Then after the messages show, what LazyList tries to do when new items
come into the list is to see the unique ID of each of the items that
were showing before, and still keep those items in view, because
otherwise it looks super odd to have new items come in and suddenly what
you were looking at just disappears.

Now what that means is that all the chat messages come first in the
list, but it tries to keep the loading indicator in the screen too, so
despite the fact that we have reverseScrolling set to true and it should
scroll from bottom-up, it just scrolls all the way to the top in order
to keep those two items at the top.

Not adding any items to the list before the first messages show up fixes
this issue.

To still show a loading state, we show the same loading indication as we
did when the chat is still initializing until the first messages come in
since we can not rely on the loading indicator to show the loading state

### DIFF
--- a/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/compose/LogCompositions.kt
+++ b/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/compose/LogCompositions.kt
@@ -1,10 +1,9 @@
-package com.hedvig.app.util.compose
+package com.hedvig.android.core.ui.compose
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.remember
 import com.hedvig.android.logger.logcat
-import com.hedvig.app.BuildConfig
 
 /**
  * Only for use with [LogCompositions], not defined as private because inline [LogCompositions] doesn't allow it.
@@ -19,9 +18,7 @@ class Ref(var value: Int)
 @Suppress("NOTHING_TO_INLINE")
 @Composable
 inline fun LogCompositions(message: String) {
-  if (BuildConfig.DEBUG) {
-    val ref = remember { Ref(0) }
-    SideEffect { ref.value++ }
-    logcat { "Debug Log Compositions: $message ${ref.value}" }
-  }
+  val ref = remember { Ref(0) }
+  SideEffect { ref.value++ }
+  logcat { "Debug Log Compositions: $message ${ref.value}" }
 }
--- a/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatDestination.kt
+++ b/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatDestination.kt
@@ -101,7 +101,9 @@ private fun ChatScreen(
       ChatTopAppBar(
         onNavigateUp = onNavigateUp,
         topAppBarScrollBehavior = topAppBarScrollBehavior,
-        onHeightChanged = { height -> with(density) { topAppBarHeight = height.toDp() } },
+        modifier = Modifier.onSizeChanged {
+          with(density) { topAppBarHeight = it.height.toDp() }
+        },
       )
       Box(
         modifier = Modifier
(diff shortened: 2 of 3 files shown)


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/compose/LogCompositions.kt
```diff
@@ -19,9 +18,7 @@ class Ref(var value: Int)
 @Suppress("NOTHING_TO_INLINE")
 @Composable
 inline fun LogCompositions(message: String) {
-  if (BuildConfig.DEBUG) {
```
StylianosGakis, line 22: Moved this to a common module so that I could use it in Chat too to test something.
In there we have no access to `com.hedvig.app.BuildConfig`, but that's fine, we should just not call this from production code as we've already been doing

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatDestination.kt
```diff
@@ -110,9 +113,10 @@ private fun ChatScreen(
           .consumeWindowInsets(PaddingValues(top = topAppBarHeight)),
         propagateMinConstraints = true,
       ) {
+        val loadingIndicator = remember { movableContentOf { HedvigFullScreenCenterAlignedProgress() } }
```
StylianosGakis, line 116: The alternative to this here would be to do
```
        when (uiState) {
          ChatUiState.Initializing -> {
            HedvigFullScreenCenterAlignedProgress()
          }

          is ChatUiState.Loaded -> {
	@@ -128,6 +132,12 @@ private fun ChatScreen(
              onSendMedia = onSendMedia,
              onFetchMoreMessages = onFetchMoreMessages,
            )
            val stillLoadingInitialMessages = uiState.messages.isEmpty() &&
              (uiState.fetchMoreMessagesUiState is ChatUiState.Loaded.FetchMoreMessagesUiState.StillInitializing ||
                uiState.fetchMoreMessagesUiState is ChatUiState.Loaded.FetchMoreMessagesUiState.FetchingMore)
            if (stillLoadingInitialMessages) {
              HedvigFullScreenCenterAlignedProgress()
            }
          }
        }
```

Without movableContentOf, since we call `HedvigFullScreenCenterAlignedProgress` in two different places, what would happen before is that if the loading indicators were already mid-animation (they always are of course) then changing from Initialized to Loaded it'd mean a *new* `HedvigFullScreenCenterAlignedProgress` would show on the screen so the loading indicator would start over.

With movableContentOf, the same composable is placed on the UI in both of those cases, and the animation just continues playing from where it was already called, despite the fact that it's called from two separate parts of the UI.

This was pretty cool tbh, I hadn't found much use of movableContentOf before but here it fits perfectly.

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatLoadedScreen.kt
```diff
@@ -267,37 +267,41 @@ private fun ChatLazyColumn(
...
-    ) {
-      Spacer(modifier = Modifier.height(8.dp))
-    }
-    if (
-      uiState.fetchMoreMessagesUiState is ChatUiState.Loaded.FetchMoreMessagesUiState.FailedToFetch ||
-      uiState.fetchMoreMessagesUiState is ChatUiState.Loaded.FetchMoreMessagesUiState.FetchingMore
-    ) {
+    // We want to show other items only when there are already some chat messages, to have first messages appear at the
+    // bottom of the UI when the screen is first opened
+    if (uiState.messages.isNotEmpty()) {
```
StylianosGakis, line 272: Not placing anything else in the list before the chat messages first show up is the core fix here.
This is the only change here, the rest is just indentation, despite GitHub rendering the diff in a weird way.

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatDestination.kt
hugokallstrom: Nice!

hugokallstrom (approved):

PR #1974, opened 2024-02-01
Add support for top-level logAction and logError

StylianosGakis:
Following the same pattern as logcat {} in order to be able to log datadog actions and datadog errors from any place.


app/tracking/tracking-core/src/main/kotlin/com/hedvig/android/core/tracking/ActionLogger.kt
```diff
@@ -0,0 +1,89 @@
+package com.hedvig.android.core.tracking
+
+import com.hedvig.android.logger.logcat
+
+interface ActionLogger {
```
hugokallstrom, line 5: Could we call this just "Logger"? Or ActionAndErrorLogger? Since its logging both actions and errors

hugokallstrom (approved):

app/tracking/tracking-core/src/main/kotlin/com/hedvig/android/core/tracking/ActionLogger.kt
StylianosGakis: Yeah, we also have `interface LogcatLogger` for the other logger and I didn't want to do something that would mix those two up.
`Logger` might be a tiny bit confusing as it might imply it's logcat too, since it's the default place we log to on Android, while it isn't really the case here.

What is the concept they use in Datadog in general for all this? Something about metrics perhaps?

PR #1975, opened 2024-02-01
Apollo interceptor with tracking module

StylianosGakis:
Adds the changes from https://github.com/HedvigInsurance/android/pull/1974 along with calling that new API from the Datadog interceptor that was added


hugokallstrom (approved):
Nice!

PR #1976, opened 2024-02-01
Feature/send memberid to unleash

panasetskaya:
add memberId to unleash context


app/auth/auth-core-public/src/main/kotlin/com/hedvig/android/auth/MemberIdService.kt
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  private fun extractMemberIdFromAccessToken(accessToken: String): String? {
+    return try {
+      val payload = accessToken.split(".").getOrNull(1) ?: return null
+      val decodedPayload = Base64.decode(payload, Base64.DEFAULT).decodeToString()
+      val payloadJsonObject: JsonObject = Json.parseToJsonElement(decodedPayload).jsonObject
+      val subContent: JsonElement = payloadJsonObject.getOrElse("sub") { return null }
+      val subText = subContent.jsonPrimitive.content
+      if (!subText.startsWith("mem_")) return null
```
hugokallstrom, line 40: We always add brackets for if statements, like 
```
if (x) {
  return y
}
```
Makes it a bit easier to read and modify :)

app/featureflags/feature-flags-public/src/main/kotlin/com/hedvig/android/featureflags/HedvigUnleashClient.kt
```diff
@@ -70,14 +79,17 @@ class HedvigUnleashClient(
...
-    .appName(APP_NAME)
-    .properties(
-      mutableMapOf(
-        "appVersion" to appVersion,
-        "appName" to APP_NAME,
-        "market" to market,
-      ),
-    )
-    .build()
+  private fun createContext(market: String, appVersion: String, memberId: String): UnleashContext {
```
hugokallstrom, line 82: You could pass a `String?` memberId here and pass it as a property if its non-null.

hugokallstrom (commented):
Just a couple of smaller nits, looks good overall! 👍

StylianosGakis (approved):
🔥

PR #1977, opened 2024-02-05
Chat banner from backend

StylianosGakis:
Adds a module which will contain just a simple Markdown type, which is necessary for GQL to know what to map the `Markdown` scalar that now exists on the supergraph's scehma.

A small change in ChatLoadedScreen composable so that it takes in a slot for the chat input composable. This is because that composable sets up a state object which requires an activity to be present, which made it impossible to create in a @Preview


hugokallstrom (approved):
👌

PR #1978, opened 2024-02-05
Analytics for chat fetching error

StylianosGakis:
This also sets it up so that the module which adds logError and logAction is available to *all* other modules


hugokallstrom (approved):
Nice!

PR #1983, opened 2024-02-06
Augment link handling for chat banner to logAction the navigation to HC

StylianosGakis:
This can be used to determine how often people go to the chat but end up first checking out the help center instead, or if they have chatted and then also decided to check out the help center

If a message has been sent is determined by the Presenter, exposing that information in the UI state. If one exits the chat and comes back into it again, that state is cleared, but that is working as intended, we do not plan to make this flag be tracked across app-starts or something more intricate than what we have in place here

### DIFF
--- a/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatPresenter.kt
+++ b/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatPresenter.kt
@@ -60,6 +60,8 @@ internal sealed interface ChatUiState {
     val messages: ImmutableList<UiChatMessage>,
     val fetchMoreMessagesUiState: FetchMoreMessagesUiState,
     val bannerText: String?,
+    // This is used to track if a message has been sent at least once in this chat "session"
+    val haveSentAtLeastOneMessage: Boolean,
   ) : ChatUiState {
     data class UiChatMessage(
       val chatMessage: ChatMessage,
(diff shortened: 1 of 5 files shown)


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatPresenter.kt
```diff
@@ -90,8 +92,10 @@ internal class ChatPresenter(
         *(lastState.safeCast<ChatUiState.Loaded>()?.messages?.map { it.chatMessage } ?: emptyList()).toTypedArray(),
       )
     }
-
-    var bannerText: String? by remember { mutableStateOf(null) }
+    var bannerText: String? by remember { mutableStateOf(lastState.safeCast<ChatUiState.Loaded>()?.bannerText) }
+    var haveSendAtLeastOneMessage: Boolean by remember {
```
hugokallstrom, line 96: haveSendAtLeastOneMessage -> haveSentAtLeastOneMessage

hugokallstrom (approved):

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatPresenter.kt
StylianosGakis: Oops, fixed now 👍

PR #1987, opened 2024-02-07
Feature/travel certificate list

panasetskaya:
add separate destination for displaying all travel certificates created earlier


app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/CheckTravelCertificateAvailabilityUseCase.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.hedvig.android.logger.logcat
+import octopus.CurrentContractsQuery
+
+interface CheckTravelCertificateAvailabilityUseCase {
+  suspend fun invoke(): Either<TravelCertificateAvailabilityError, Unit>
+}
+
+internal class CheckTravelCertificateAvailabilityUseCaseImpl(
+  private val getTravelCertificatesHistory: GetTravelCertificatesHistoryUseCase,
+  val apolloClient: ApolloClient,
```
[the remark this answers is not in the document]
panasetskaya: I don't think so, the di modules in the feature modules that need this UseCase cannot reach it then?

app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/CheckTravelCertificateAvailabilityUseCase.kt
```diff
@@ -0,0 +1,69 @@
...
+        }
+        .onLeft { errorMessage ->
+          logcat { "Could not fetch current contracts. Message:${errorMessage.message}" }
+        }
+      val historyResult = getTravelCertificatesHistory.invoke()
+        .onLeft { errorMessage ->
+          logcat(LogPriority.ERROR) { "Could not fetch travel certificates history: ${errorMessage.message}" }
+        }
+
+      if (contractsResult is Either.Left<ErrorMessage> && historyResult is Either.Left<ErrorMessage>) {
```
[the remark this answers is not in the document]
panasetskaya: we could probably cast it lower, not in the expression itself, but it's basically the same, a bit prettier though

app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/GetTravelCertificatesHistoryUseCase.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.core.common.ErrorMessage
+import kotlinx.datetime.LocalDate
+import octopus.TravelCertificatesQuery
+
+interface GetTravelCertificatesHistoryUseCase {
+  suspend fun invoke(): Either<ErrorMessage, List<TravelCertificate>>
+}
+
+internal class GetTravelCertificatesHistoryUseCaseImpl(val apolloClient: ApolloClient) :
```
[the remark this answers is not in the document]
panasetskaya: thanks, fixed that

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
```diff
@@ -97,18 +97,28 @@ internal class GetQuickLinksUseCase(
...
-        ?.let {
-          add(
-            QuickAction.QuickLink(
-              destination = AppDestination.GenerateTravelCertificate,
-              titleRes = R.string.HC_QUICK_ACTIONS_TRAVEL_CERTIFICATE,
-              displayName = null,
-            ),
-          )
-        }
+//      contracts
```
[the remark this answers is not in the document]
panasetskaya: thanks, fixed that

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
```diff
@@ -0,0 +1,153 @@
...
+    popEnterTransition = { MotionDefaults.sharedXAxisPopEnter(density) },
+    popExitTransition = { MotionDefaults.sharedXAxisPopExit(density) },
+  ) {
+    composable<TravelCertificateDestination.TravelCertificateHistory> {
+      val viewModel: CertificateHistoryViewModel = koinViewModel()
+      val localContext = LocalContext.current
+      TravelCertificateHistoryDestination(
+        viewModel = viewModel,
+        navigateUp = navController::navigateUp,
+        onContinue = { navController.navigate(TravelCertificateDestination.GenerateTravelCertificateDestinations) },
```
StylianosGakis, line 44: Hmm the `onContinue` here (while it's used in some other screens to) is often a bit too abstract, in this case since it always navigates to the certificate generation flow, I'd say we can rename to `onStartGenerateTravelCertificateFlow` or something like that? :D Despite it being quite long it's more clear about what `continue` is and I think it's nicer that way.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+import com.hedvig.android.feature.travelcertificate.CertificateHistoryUiState
+import com.hedvig.android.feature.travelcertificate.CertificateHistoryViewModel
+import com.hedvig.android.feature.travelcertificate.data.TravelCertificateUri
+import hedvig.resources.R
+import kotlinx.datetime.LocalDate
+
+@Composable
+internal fun TravelCertificateHistoryDestination(
+  viewModel: CertificateHistoryViewModel,
+  onContinue: () -> Unit,
```
StylianosGakis, line 55: As discussed here https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1707394659436749?thread_ts=1707390537.972059&cid=C03U9C6Q7TP I think we will soon have to hide this button all the way down where we call it if we do not want them to be able to create a new certificate right?

Probably something in the `uiState` itself actually can determine if this button should be shown or not.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+  if (showBottomSheet) {
+    TravelCertificateInfoBottomSheet(
+      onDismiss = { showBottomSheet = false },
+      sheetState = explanationSheetState,
+    )
+  }
+
+  when (uiState) {
+    CertificateHistoryUiState.FailureDownloadingCertificate -> {
+      ErrorDialog(
```
StylianosGakis, line 92: As discussed irl, let's turn the error scenario into something that shows inside the success screen.
That can then also not be a Dialog anymore, but a normal part of the screen.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+            modifier = Modifier.size(40.dp),
+          ) {
+            Icon(
+              imageVector = Icons.Hedvig.Info,
+              contentDescription = stringResource(R.string.REFERRALS_INFO_BUTTON_CONTENT_DESCRIPTION),
+              modifier = Modifier.size(24.dp),
+            )
+          }
+        },
+      ) {
```
StylianosGakis, line 118: I wonder if in the failure scenario we can just show an error in the middle of the screen with the ability to retry loading and not show the top app bar with the info button and so on? I know the design does not specify it itself, so I wonder how this screen looks like with the info at the top but with error state in the middle.


If we do in fact want to keep the scaffold here anyway, I wonder then if the scaffold can be specified above here, so that it always show, and only the content of the scaffold changes depending on if we got a success or a failed state and so on. That way no need to call the scaffold twice in each case.

But if you've tried smth like that and think it doens't work here for some reason I'd also be happy to discuss that.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+        onContinue = onContinue,
+        navigateUp = navigateUp,
+        historyList = uiState.certificateHistoryList,
+      )
+    }
+  }
+}
+
+@Composable
+private fun ShowHistory(
```
StylianosGakis, line 158: This composable being a verb can be a tiny bit confusing when looked at from the outside. How about `TravelCertificateSuccessScreen`?

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+    if (historyList.isEmpty()) {
+      Spacer(modifier = Modifier.weight(1f))
+      ShowInitialInfo()
+    } else {
+      ShowNotEmptyList(
+        list = historyList,
+        onCertificateClick = onCertificateClick,
+      )
+    }
+    Spacer(modifier = Modifier.weight(1f))
```
StylianosGakis, line 191:
```suggestion
    Spacer(modifier = Modifier.weight(1f))
    Spacer(modifier = Modifier.height(16f))
```

You may have to do this here, so that if the parts above and the parts below are tall enough that there is no room for the weighted spacer to exist, then there should still be at least 16.dp of space (picked a random randomly now, maybe smth else makes more sense, depends on design) between those two

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+    VectorInfoCard(
+      text = String.format(stringResource(id = R.string.travel_certificate_start_date_info), 45),
+      modifier = Modifier
+        .fillMaxWidth()
+        .padding(16.dp, 16.dp, 16.dp, 8.dp),
```
StylianosGakis, line 196: Looking at this especially outside of the IDE with the hints can be a bit hard to understand about what is what.
I'd suggest this to be `.padding(horizontal = 16.dp)` and then for vertical spacings they can be spacers in-between the items themselves.

I very strongly recommend reading this https://www.jetpackcompose.app/articles/space-hoisting-should-I-use-the-padding-modifier-or-the-spacer-composable article around the philosophy of why using spacers is easier to use than padding on individual items, when the padding's goal is to make the items have some space between each other.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+      text = stringResource(R.string.travel_certificate_get_travel_certificate_button),
+      onClick = onContinue,
+      modifier = Modifier.padding(16.dp, 8.dp, 16.dp, 0.dp),
+    )
+    Spacer(Modifier.height(32.dp))
+  }
+}
+
+@Composable
+private fun ShowInitialInfo() {
```
StylianosGakis, line 208:
```suggestion
private fun NoCertificateHistorySection() {
```
Perhaps? So that we go away from a composable which shows UI from being a verb

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+    verticalArrangement = Arrangement.Center,
+  ) {
+    HedvigInformationSection(
+      stringResource(id = R.string.travel_certificate_empty_list_message),
+    )
+  }
+}
+
+@Composable
+private fun ShowNotEmptyList(list: List<TravelCertificate>, onCertificateClick: (String) -> Unit) {
```
StylianosGakis, line 221:
```suggestion
private fun CertificateHistoryList(list: List<TravelCertificate>, onCertificateClick: (String) -> Unit) {
```
Again to not have a verb. Open to a better name btw as always with these suggestions. I am just giving the first thing that pops to my mind with the amount of context that I currently got.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+  Column(
+    horizontalAlignment = Alignment.CenterHorizontally,
+  ) {
+    for (i in list) {
+      CertificateCard(
+        certificate = i,
+        onCertificateClick = onCertificateClick,
+      )
+    }
```
StylianosGakis, line 230: Looking great here. I'd rename `i` to `travelCertificate`.
I always prefer descriptive names than things like `i`, as it can sometimes get quite confusing.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+  modifier: Modifier = Modifier,
+) {
+  Row(
+    verticalAlignment = Alignment.CenterVertically,
+    modifier = modifier
+      .fillMaxWidth()
+      .clickable(onClick = { onCertificateClick(certificate.signedUrl) })
+      .padding(16.dp),
+  ) {
+    val isExpired = isExpired(certificate.expiryDate)
```
StylianosGakis, line 247: Great opportunity here to think about if this should be happening in the UI or not.
How would you feel if this `isExpired` was exposed directly from the ViewModel, from inside the `UiState` for each item? So if each `TravelCertificate` already had this information instead of having to determine that in the UI itself? This is especially true considering that doing it in the UI moves logic like comparing the expiration date with the current year, and doing a calculation etc.
The ViewModel could instead be taking in a `clock`, exactly as we do [here](https://github.com/HedvigInsurance/android/blob/2dc8dd0e355e2036d888e6e81d35c30e61c676c4/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatViewModel.kt#L11) for example. That way we can even in the future test this behavior in the VM. Or do more through knowing the year like partitioning the list into a list of lists, each one being for a specific year. Which would again would be much nicer to happen in the VM itself. Then the UI can just take in the state and render it, instead of having to care about such logic.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+    val color = if (isExpired) Color.Red else Color.Unspecified
```
StylianosGakis, line 248: For error colors we got `MaterialTheme.colorScheme.error` you can use

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+    val expiryDateText =
+      "${certificate.expiryDate.dayOfMonth} ${certificate.expiryDate.month.name.lowercase().substring(0, 3)}"
```
StylianosGakis, line 250: Could we look into using this https://github.com/HedvigInsurance/android/blob/0baeb902323e5e6443bdf61aee2f1ae4e59e2147/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/HedvigDateTimeFormatter.kt#L65-L67 perhaps here to do the formatting? It's nice to try and be consistent in the codebase in how we do this.

Try to use that API and if it's hard to use LMK. Inside a composable you can get the locale with `getLocale()` here https://github.com/HedvigInsurance/android/blob/7e85f0db5781027e32700587c32c517454c95dac/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/GetLocale.kt#L12

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/TravelCertificateHistory.kt
```diff
@@ -0,0 +1,422 @@
...
+        {},
+        CertificateHistoryUiState.SuccessDownloadingHistory(listOf()),
+      )
+    }
+  }
+}
+
+@HedvigPreview
+@Composable
+private fun PreviewTravelCertificateHistoryScreenWithExpiredEarlier() {
```
StylianosGakis, line 301: To be able to in one preview get 2 versions out of it, we got this https://github.com/HedvigInsurance/android/blob/fbf254341f75ef46dc66514ee1ebe8627a94b0c9/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt#L39-L41 which allows you to get a boolean, once false and once true.
Then you can use that to in one preview composable define what the UI should look like with the boolean in both cases. Here it could be 
`@PreviewParameter(BooleanCollectionPreviewParameterProvider::class) hasCertificateHistory: Boolean,` and then check for `hasCertificateHistory` to show the history or not.

StylianosGakis: Aha and I see now that this has more previews than I first thought. 
In that case, you can go full out and create your own class that you provide all the possible cases for, and then you can keep one preview composable still I think.

Look here https://github.com/HedvigInsurance/android/blob/3b9a24743d46e0c5d527e8c5b16060ccf4121b38/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ui/ChatDestination.kt#L185-L241 for inspiration. LMK if this does not make much sense

StylianosGakis (commented):
Gonna just post this as comment since I know you're working on a few more things before we merge this in.
Super nice work with all this, there are a lot of comments since this is the first time I review a full PR with UI involved and everything, so naturally there are more things that I wanted to share with you, like a couple of articles anywyay.

PR #1988, opened 2024-02-08
Fix insurance details view not showing up to date data

StylianosGakis:
Make GetInsuranceContractsUseCase return a flow of results. Now by default hits the cache *and* the network afterwards, which means that when one goes back to that screen after they've terminated their insurance, the cached value will immediatelly show, but then the real value will also be fetched from the network.
This also allows us to properly use the values from unleash, through listening to the value changes, and not just getting the first value and stopping there.


hugokallstrom (approved):
Great

PR #1989, opened 2024-02-09
Hide chat icon depending on claims or conversations

hugokallstrom:
Only show chat button if member has sent or received messages, or have submitted a claim.


app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomePresenter.kt
```diff
@@ -101,12 +122,20 @@ internal class HomePresenter(
...
-          showChatIcon = showChatIcon,
+          showChatIcon = chatEnabled && (hasReceivedOrSentMessages || successData.hasClaims),
           hasUnseenChatMessages = hasUnseenChatMessages,
+          hasClaims = successData.hasClaims,
         )
       }
     }
   }
+
+  private fun hasReceivedOrSentMessages(messages: List<ChatMessage>): Boolean {
```
panasetskaya, line 133: the only thing that seems a bit excessive to me here - if we always have one message from hedvig, and there are only two sides participating in the chat which are hedvig and the member, should we just check for messages.size > 1 and that would be all?

hugokallstrom: Good question. The reason for this logic is that we always have 1 automated message from Hedvig which says something like "Hey, how can I help you?". When the member then writes something in the chat, that automated messages is replaced by the members message.

panasetskaya: Ah, it is replaced. Got it, thanks!

StylianosGakis (changes requested):
We'll need to not do a feature to feature module dependency before we merge this in. See more in the comments.

app/feature/feature-home/build.gradle.kts
```diff
@@ -50,6 +50,7 @@ dependencies {
   implementation(projects.featureFlagsPublic)
   implementation(projects.languageCore)
   implementation(projects.marketCore)
+  implementation(projects.featureChat)
```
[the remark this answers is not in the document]
hugokallstrom: Good point, I will fix 🙏

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
```diff
@@ -138,6 +139,7 @@ internal data class HomeData(
   val claimStatusCardsData: ClaimStatusCardsData?,
   val veryImportantMessages: ImmutableList<VeryImportantMessage>,
   val memberReminders: MemberReminders,
+  val hasClaims: Boolean,
```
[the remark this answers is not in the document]
hugokallstrom: Oh, ofc 👍

StylianosGakis (approved):

PR #1990, opened 2024-02-13
Feature/submit bug button

panasetskaya:
add submit bug button in the profile -> app info


panasetskaya:
> Code looks great. I didn't test the functionality myself, would just be nice to make sure this functions somewhat fine also in scenarios where there are more than one email client. Or how about when there are no email clients at all? We'd at least want to like not crash, but also if nothing happens at all that may be a bit weird too. Just good to check in order for us to know what happens in those scenarios.

If there is no email clients it shows system message "no apps can perform this action", if there are few of them , we have a chooser. Which I think is okay, what do you think?

PR #1993, opened 2024-02-14
Move chat button and help center logic into the UseCase itself, and remove the second non-invoke function from it

StylianosGakis:
Makes it all easier to test all scenarios by testing the use case itself, and the presenter can do less work this way.

Also removes the need of sharing the query between the home and the chat feature, since they are not interested in all of the same fields in the first place.

Also clears us querying `getTravelCertificateSpecificationsUseCase` and `Feature.MOVING_FLOW` since we no longer use this in home after removing the "other services" bottom sheet.


hugokallstrom (approved):
Much better :D

PR #1994, opened 2024-02-15
Feature/travel certificate choose contract

panasetskaya:
Members with more than 1 eligible insurance can choose which one to choose for travel certificate


app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/GetEligibleContractsWithAddressUseCase.kt
```diff
@@ -0,0 +1,37 @@
...
+  override suspend fun invoke(): Either<ErrorMessage, List<ContractEligibleWithAddress>> {
+    return apolloClient.query(EligibleContractsWithAddressQuery())
+      .fetchPolicy(FetchPolicy.NetworkOnly)
+      .safeExecute()
+      .toEither(::ErrorMessage)
+      .map {
+        it.currentMember.activeContracts.filter { contract ->
+          contract.supportsTravelCertificate
+        }.map { contract ->
+          ContractEligibleWithAddress(contract.exposureDisplayName.substringBefore("•"), contract.id)
```
StylianosGakis, line 28: I checked what iOS does to get the name for this. The check inside `travelCertificateSpecifications.contractSpecifications.location.street` and that one gives the right string without having to rely on there being a `•` to format it right, since that may not always really be true now or in the future.

Can you give this a shot. Change the query to
```
query CurrentMember {
	currentMember {
    activeContracts {
      id
      supportsTravelCertificate
    }
    travelCertificateSpecifications {
      contractSpecifications {
        contractId
        location {
          street
        }
      }
    }
	}
}
```
And then you gotta match the `activeContracts.id` with `contractId` from inside `contractSpecifications`, and get the `street` from inside there.

For reference, this is a sample response for this.
image

And one with the exposureName which shows the difference between the two
image

panasetskaya: Yes, sure!
I actually had a thought that probably it makes sense to show the number of insured people on this address too? Since on this depends who you can add to the certificate on the next step. Maybe just leave the whole exposureDisplayName there without the substrings?

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificateViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+internal class ChooseContractPresenter(
+  private val getEligibleContractsWithAddressUseCase: GetEligibleContractsWithAddressUseCase,
+) : MoleculePresenter<ChooseContractEvent, ChooseContractUiState> {
+  @Composable
+  override fun MoleculePresenterScope<ChooseContractEvent>.present(
+    lastState: ChooseContractUiState,
+  ): ChooseContractUiState {
+    var dataLoadIteration by remember { mutableIntStateOf(0) }
+    var currentState by remember {
+      mutableStateOf<ChooseContractUiState>(ChooseContractUiState.Loading)
```
StylianosGakis, line 34:
```suggestion
      mutableStateOf<ChooseContractUiState>(lastState)
```
Here we can use what the state was before (if there was anything before) for the initial value

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificateViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+      when (event) {
+        ChooseContractEvent.RetryLoadData -> {
+          currentState = ChooseContractUiState.Loading
+          dataLoadIteration++
+        }
+      }
+    }
+
+    LaunchedEffect(dataLoadIteration) {
+      getEligibleContractsWithAddressUseCase.invoke().fold(
```
StylianosGakis, line 47: My hunch here tells me that doing the loadIteration ++ and setting the state to loading is doing a bit too much inside the event collection itself.
I was thinking we could have the event just increment the iteration counter, and then inside the LaunchedEffect we can set it to loading, something like this
```
        ChooseContractEvent.RetryLoadData -> {
          dataLoadIteration++
        }
      }
    }

    LaunchedEffect(dataLoadIteration) {    
      currentState = ChooseContractUiState.Loading
      getEligibleContractsWithAddressUseCase.invoke().fold(
```

But then it would be true that if you for example come *back* into this screen, where your `lastState` would actually be `ChooseContractUiState.Success` already, then it would be jarring to have the state go into Loading mode, and then go back to the success as it was already, after the network request is finished.

So in the scenario of "I was one destination forward, then I came back into this screen and I want to just see what I saw before without any flickering into loading state and back into the success state", we got two options. We
1) Let the data fetching run again in this case, and let that happen in the background while we do not show that we are loading something to the user. If this succeeds, we get the success state back and we send it to the UI, but since it will be equivalent anyway most likely, the user will perceive nothing, all good. If the network request fails, then we do show the error state actually, despite not having shown that we are loading anything in the first place.
This could be done by doing this instead:
```
        ChooseContractEvent.RetryLoadData -> {
          dataLoadIteration++
        }
      }
    }

    LaunchedEffect(dataLoadIteration) {    
      if (currentState !is ChooseContractUiState.Success) {
        currentState = ChooseContractUiState.Loading
      }
      getEligibleContractsWithAddressUseCase.invoke().fold(
```

2) Could just not fetch any new data, aka never call `getEligibleContractsWithAddressUseCase.invoke()` if we just came back and we already have the success state available to us. Since to be honest the response from `getEligibleContractsWithAddressUseCase` should not have changed after going forward and back into this screen realistically.
This could be done like this:

```suggestion
        ChooseContractEvent.RetryLoadData -> {
          dataLoadIteration++
        }
      }
    }

    LaunchedEffect(dataLoadIteration) {
      if (currentState is ChooseContractUiState.Success) {
        return@LaunchedEffect
      }
      getEligibleContractsWithAddressUseCase.invoke().fold(
```
Note that here there is no way to fetch again even if the retry event comes in (it should never happen, since the retry button won't show in this scenario) if we are already in the `Success` state.





What do you think about all this? Do you prefer 1, 2, or something else?
Btw the current behavior with how you've set it up right now is more or less the same as 1), since coming back you will not set the state to loading, since that's only set in the event itself.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificateViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+          logcat(priority = LogPriority.INFO) { "Cannot load contracts eligible for travel certificates" }
+          currentState = ChooseContractUiState.Failure
+        },
+      )
+    }
+    return currentState
+  }
+}
+
+sealed interface ChooseContractEvent {
```
StylianosGakis, line 62: Could this also be internal?
```suggestion
internal sealed interface ChooseContractEvent {
```

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/history/CertificateHistoryViewModel.kt
```diff
@@ -137,12 +147,16 @@ private sealed interface ScreenContentState {
 
   data object Loading : ScreenContentState
 
-  data class Success(val certificateHistoryList: List<TravelCertificate>, val eligibleToCreateCertificate: Boolean) :
+  data class Success(
+    val certificateHistoryList: List<TravelCertificate>,
+    val eligibleToCreateCertificate: Boolean,
+    val hasChooseOption: Boolean,
```
StylianosGakis, line 153: Could we name this a bit more descriptive? If they have a choice to choose something if you are lacking the context of what that choice is does not tell the entire story.
Something like `mustChooseContractBeforeGeneratingTravelCertificate` is the first thing that pops to my mind, to kinda steer you more towards what I would like this to say to be super clear about its purpose.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/overview/TravelCertificateOverview.kt
```diff
@@ -0,0 +1,144 @@
...
+  travelCertificateUrl: TravelCertificateUrl,
+  onDownloadCertificate: (TravelCertificateUrl) -> Unit,
+  navigateBack: () -> Unit,
+  onShareTravelCertificate: (TravelCertificateUri) -> Unit,
+  uiState: TravelCertificateOverviewUiState,
+) {
+  when (uiState) {
+    TravelCertificateOverviewUiState.Failure -> {
+      HedvigScaffold(
+        navigateUp = { navigateBack() },
```
StylianosGakis, line 60:
```suggestion
        navigateUp = navigateUp,
```

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/overview/TravelCertificateOverview.kt
```diff
@@ -0,0 +1,144 @@
...
+        topAppBarActionType = TopAppBarActionType.CLOSE,
+        modifier = Modifier.clearFocusOnTap(),
```
StylianosGakis, line 62:
```suggestion
        topAppBarActionType = TopAppBarActionType.CLOSE,
```
I think in this scenario there is nothing that would show a text field input, therefore there would be nothing to clear the focus from.
With that said, I don't think just having the modifier there breaks anything, or is bad in any way, but it won't really do something in this scenario.
I would be okay either way, but I know if I was around here and was moving any code around I would probably instinctively just remove this.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/overview/TravelCertificateOverview.kt
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    is TravelCertificateOverviewUiState.Success -> {
+      LaunchedEffect(uiState.travelCertificateUri) {
+        uiState.travelCertificateUri?.let {
+          onShareTravelCertificate(it)
+        }
+      }
+      HedvigScaffold(
+        navigateUp = { navigateBack() },
```
StylianosGakis, line 79:
```suggestion
        navigateUp = navigateUp,
```

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/overview/TravelCertificateOverviewViewModel.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        is TravelCertificateOverviewEvent.OnDownloadCertificate -> {
+          travelCertificateUrl = event.travelCertificateUrl
+          dataLoadIteration++
+        }
+      }
+    }
+
+    LaunchedEffect(dataLoadIteration) {
+      val url = travelCertificateUrl
```
StylianosGakis, line 56:
```suggestion
    var travelCertificateUrl by remember { mutableStateOf<TravelCertificateUrl?>(null) }

    CollectEvents { event ->
      when (event) {
        TravelCertificateOverviewEvent.RetryLoadData -> {
          dataLoadIteration++
        }

        is TravelCertificateOverviewEvent.OnDownloadCertificate -> {
          travelCertificateUrl = event.travelCertificateUrl
        }
      }
    }

    LaunchedEffect(dataLoadIteration, travelCertificateUrl) {
      val url = travelCertificateUrl
```
Could this work here instead? So that the event handling still just does the simplest thing possible, and the rest of the presenter can figure out what that should do?

Now we are changing the `travelCertificateUrl` and then trigger the iteration as a side effect to run the LaunchedEffect down there because we are reading the value inside of there. Doing something like changing the order of setting the travelCertificateUrl and `dataLoadIteration++` might even break this functionality, and especially if we do more things in-between those two, so it feels a bit brittle this way.
If we instead key the `LaunchedEffect` also on the URL so that it re-runs the effect when it changes, we get the same behavior as far as I understand here.

Does this make sense? Am I perhaps missing something here?

StylianosGakis (commented):
Hey super great job in getting all this done here. I have written once again a boat load of comments, but a lot of those were things that I would've written at some point in any case, like stuff around naming conventions of composables, how to use .weight(1f), reading things from figma and so on.

I am gonna just write this as a comment since there are some things that I would like us to address over here before merging. Overall I am super happy, great job with this!
There are also a bunch of good learning opportunities in all the comments that I've left here, so do not hesitate to reach out to me so that we can discuss them more in-depth if you want.

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificate.kt
```diff
@@ -0,0 +1,156 @@
...
+    }
+    is ChooseContractUiState.Success -> {
+      HedvigScaffold(
+        navigateUp = navigateBack,
+        modifier = Modifier.clearFocusOnTap(),
+      ) {
+        Spacer(modifier = Modifier.weight(1f))
+        Text(
+          text = stringResource(id = R.string.travel_certificate_select_contract_title),
+          style = MaterialTheme.typography.headlineSmall,
```
StylianosGakis, line 79: This one looks like headlineMedium if I read Figma correctly
```suggestion
          style = MaterialTheme.typography.headlineMedium,
```

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificate.kt
```diff
@@ -0,0 +1,156 @@
...
+          Spacer(modifier = (Modifier.height(4.dp)))
+        }
+        Spacer(modifier = Modifier.height(12.dp))
+        HedvigContainedButton(
+          onClick = { selectedContractId?.let { onContinue(it) } },
+          modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp),
+        ) {
+          Text(text = stringResource(id = R.string.general_continue_button))
```
StylianosGakis, line 118:
```suggestion
          Text(
            text = stringResource(id = R.string.general_continue_button),
            style = MaterialTheme.typography.headlineSmall
          )
```
And this is this probably

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/FullScreenLoading.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@Composable
+internal fun FullScreenLoading() {
+  Box(modifier = Modifier.fillMaxSize()) {
+    CircularProgressIndicator(modifier = Modifier.align(Alignment.Center))
+  }
+}
+
+@Composable
+internal fun ColumnScope.SomethingWrongInfo(onButtonClick: () -> Unit) {
```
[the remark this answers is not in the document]
panasetskaya: Thanks! :)

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
```diff
@@ -71,18 +91,18 @@ fun NavGraphBuilder.travelCertificateGraph(density: Density, navController: NavC
...
           },
           onIncludeMemberClicked = viewModel::onIncludeMemberClicked,
           onTravelDateSelected = viewModel::onTravelDateSelected,
           onContinue = viewModel::onContinue,
           onSuccess = { travelCertificateUrl ->
             navController.navigate(TravelCertificateDestination.ShowCertificate(travelCertificateUrl)) {
-              popUpTo<TravelCertificateDestination.GenerateTravelCertificateDestinations> {
-                inclusive = true
+              popUpTo<TravelCertificateDestination.TravelCertificateHistory> {
+                inclusive = false
```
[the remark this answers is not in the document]
panasetskaya: I think I did, and took me back to the Help and Support :)

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
```diff
@@ -43,14 +47,29 @@ fun NavGraphBuilder.travelCertificateGraph(density: Density, navController: NavC
...
     }
+
+    composable<TravelCertificateDestination.TravelCertificateChooseContract> {
+      val viewModel: ChooseContractForCertificateViewModel = koinViewModel()
+      ChooseContractForCertificateDestination(
+        viewModel = viewModel,
+        onContinue = { contractId ->
+          navController.navigate(TravelCertificateDestination.GenerateTravelCertificateDestinations(contractId))
+        },
+        navigateBack = navController::navigateUp,
```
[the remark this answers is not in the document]
panasetskaya: Important difference, thanks!

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
```diff
@@ -48,9 +42,7 @@ internal class GetHomeDataUseCaseImpl(
         .fetchPolicy(if (forceNetworkFetch) FetchPolicy.NetworkOnly else FetchPolicy.CacheFirst)
         .safeFlow(::ErrorMessage),
       getMemberRemindersUseCase.invoke(),
-      flow { emit(getTravelCertificateSpecificationsUseCase.invoke().getOrNull()) },
-      flow { emit(featureManager.isFeatureEnabled(Feature.MOVING_FLOW)) },
-    ) { homeQueryDataResult, memberReminders, travelCertificateData, isMovingFlowEnabled ->
+    ) { homeQueryDataResult, memberReminders ->
```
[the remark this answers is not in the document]
panasetskaya: Yes, my bad, I removed travelCertificateData bc it's not used at home anymore, and removed unused  featureManager as well. I'll merge develop in feature now

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificateViewModel.kt
panasetskaya: I guess I like more number 1 in general, but, as you said, it is a kind of response that is not likely to be changed at all, so I think we better go with n.2 :)

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/choose/ChooseContractForCertificate.kt
```diff
@@ -0,0 +1,156 @@
...
+        }
+        Spacer(Modifier.height(32.dp))
+      }
```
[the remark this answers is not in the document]
panasetskaya: Got it, thanks!

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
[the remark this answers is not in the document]
StylianosGakis: Yeah overall it's encouraged to sometimes if it makes sense to do some slight cleanup even if it's not necessarily 100% related with what you are working on.
This is not a problem at all in this PR, I only mention it since I know that we fixed this on develop too anyway, so it gives the merge conflict, that's all 😊

StylianosGakis:
Merged here https://github.com/HedvigInsurance/android/pull/1999 instead, for future reference.

PR #1999, opened 2024-02-16
Feature/travel certificate choose contract fix 

panasetskaya:
Members with more than 1 eligible insurance can choose which one to choose for travel certificate - fixed after code review


app/data/data-travel-certificate/data-travel-certificate-public/src/main/kotlin/com/hedvig/android/data/travelcertificate/GetEligibleContractsWithAddressUseCase.kt
```diff
@@ -0,0 +1,43 @@
...
+      .map { data ->
+        data.currentMember.activeContracts.filter { contract ->
+          contract.supportsTravelCertificate
+        }.map { contract ->
+          val street =
+            data.currentMember.travelCertificateSpecifications.contractSpecifications.firstOrNull {
+              it.contractId == contract.id
+            }?.location?.street
+          val address = street ?: contract.exposureDisplayName.substringBefore("•")
+          // todo: add this check here bc location.street is nullable?
```
StylianosGakis, line 33: Yeah that's a fair fallback here, do the right thing by default and fall back to the substring method in the odd case.

Since I expect this to never be hit unless there is an actual error on our backend, how about we log this as an error so that we make sure to address it if it ever happens?
```suggestion
          val address = street ?: contract.exposureDisplayName.substringBefore("•").also {
            logcat(LogPriority.ERROR) { "Received a travel certificate eligible contract, but the contractSpecifications did not include a valid street name" }
          }
          // todo: add this check here bc location.street is nullable?
```
Something like that? And if this actually gets hit in production there might be smth wrong with backend.

app/data/data-travel-certificate/data-travel-certificate-public/src/test/kotlin/com/hedvig/android/data/travelcertificate/GetTravelCertificateSpecificationsUseCaseTest.kt
```diff
@@ -55,11 +55,109 @@ internal class GetTravelCertificateSpecificationsUseCaseTest {
...
+        listOf(TravelCertificateData.InfoSection("infoTitle", "infoBody")),
+      ),
+    )
+  }
+
+  @Test
+  fun `when the passed contractId is correct, but the contract is not eligible, we get not eligible`() = runTest {
+    val travelCertificateUseCase = GetTravelCertificateSpecificationsUseCaseImpl(
+      apolloClient,
+    )
```
StylianosGakis, line 129: Wow super, that's exactly what I meant! Very well done

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/ui/generate/GenerateTravelCertificateViewModel.kt
```diff
@@ -23,30 +19,31 @@ import kotlinx.datetime.TimeZone
...
       _uiState.update { it.copy(isLoading = true) }
       getTravelCertificateSpecificationsUseCase
-        .invoke()
+        .invoke(contractId)
         .fold(
           ifLeft = { travelCertificateError ->
             _uiState.update {
               when (travelCertificateError) {
+                // todo: probably should not only show error dialog, but also don't show the input ui at all,
+                // todo: otherwise you have an opportunity to fill email etc manually, press continue and crash the app
```
StylianosGakis, line 43: Yes, let's try and keep this in mind when we eventually work a bit more on cleaning up this feature.

StylianosGakis (approved):
This looks super good, I feel like this is ready now, I didn't run it locally myself but if you feel confident in how it functions merge it to develop.
Super good job on fixing all the review comments!

PR #2001, opened 2024-02-17
Feature delete account

StylianosGakis:


app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountDestination.kt
```diff
@@ -0,0 +1,199 @@
...
+      }
+
+      is DeleteAccountUiState.Success -> {
+        when (uiState) {
+          is DeleteAccountUiState.Success.CanDelete -> {
+            if (uiState.failedToPerformDeletion) {
+              HedvigErrorSection(retry = retryLoading, Modifier.weight(1f))
+            } else {
+              DeleteScreenContents(
+                title = stringResource(R.string.DELETE_ACCOUNT_DELETE_ACCOUNT_TITLE),
```
hugokallstrom, line 73: Could you add extension functions on `DeleteAccountUiState.Success` and do eg. `uiState.title`, `uiState.description` etc. to avoid having these when clauses?

app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountPresenter.kt
```diff
@@ -0,0 +1,113 @@
...
+    LaunchedEffect(isPerformingDeletion) {
+      if (!isPerformingDeletion) return@LaunchedEffect
+      if (deleteAccountState !is DeleteAccountState.CanDelete) {
+        isPerformingDeletion = false
+        return@LaunchedEffect
+      }
+      val mutationResult = requestAccountDeletionUseCase.invoke()
+      Snapshot.withMutableSnapshot {
+        if (mutationResult.isLeft()) {
+          failedToPerformAccountDeletion = true
```
hugokallstrom, line 62: Are we fine with showing a generic error here? Are we logging the type of error somewhere?

hugokallstrom (approved):
Very clean 👌

app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountDestination.kt
```diff
@@ -0,0 +1,199 @@
...
+                DeleteAccountUiState.Success.AlreadyRequestedDeletion -> stringResource(
+                  R.string.DELETE_ACCOUNT_PROCESSED_TITLE,
+                )
+                DeleteAccountUiState.Success.HasActiveInsurance -> stringResource(
+                  R.string.DELETE_ACCOUNT_YOU_HAVE_ACTIVE_INSURANCE_TITLE,
+                )
+                DeleteAccountUiState.Success.HasOngoingClaim -> stringResource(
+                  R.string.DELETE_ACCOUNT_YOU_HAVE_ACTIVE_CLAIM_TITLE,
+                )
+                is DeleteAccountUiState.Success.CanDelete -> error("Invalid state")
```
panasetskaya, line 101: Question: could DeleteAccountUiState.Success.CanDelete be separate from the others? E.g, could AlreadyRequestedDeletion, HasActiveInsurance, HasOngoingClaim be not DeleteAccountUiState.Success, but smth like DeleteAccountUiState.DeleteNotAllowed? Feels like CanDelete is very different from the other ones

StylianosGakis: I completely agree.
Resolved with https://github.com/HedvigInsurance/android/commit/d5db1776ed25455781ff0d2c8156c5f227ff2ae1

app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountDestination.kt
StylianosGakis: 100% agree. Addressed here
https://github.com/HedvigInsurance/android/commit/9d535744ffc17c0eafbfe2cbab3095df1b7ba13c

app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountPresenter.kt
StylianosGakis: My intention with this was that if the mutation itself fails, we show an error state which they can click retry on and click the button again.
It was also partly because we got no design for this, nor error case copy so I tried to make do with what was already available there.

But yes, let's try to log something here.

This is one of those places where I'd love to be able to log an error but not log again (since we do it in our `DatadogInterceptor`) if the error was network related.

It's one of those things I will fix when we bump to apollo-kotlin 4.x when that is out of beta.

So I will log as a Warning instead of an error to avoid pinging on-call people when normal network errors happen

StylianosGakis: Addressed:
https://github.com/HedvigInsurance/android/commit/9a0fb2db79b4cd08b58c2c6674d87722c391a863

PR #2008, opened 2024-02-23
Use the Hedvig icons everywhere

StylianosGakis:
We should basically never use anything other than `Icons.Hedvig`, to make sure that we always use our own icons and there's harmony across the entire app


hugokallstrom (approved):
👏

PR #2010, opened 2024-02-23
Fix Scaffold top app bar insets

StylianosGakis:
Fix claim flow scaffold TopAppBar going behind the camera area

With this change the X or the back button do not show behind the system
camera.
Filed issue:
https://issuetracker.google.com/issues/326526714

Also just remove a `HedvigTheme` wrapper that was probably there for historical reasons that no longer matter


hugokallstrom (approved):
LGTM!

PR #2012, opened 2024-02-26
Fix grid bug

StylianosGakis:
While at that screen, also fixed the chat section at the bottom to go under the system bar as per the design
The color difference under the bottom part of the screen is visible here

Before:
image

After:
image

### DIFF
--- a/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
+++ b/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
@@ -187,8 +187,12 @@ private fun HelpCenterHomeScreen(
                   HedvigCard(
                     onClick = {
                       when (quickLink) {
-                        is HelpCenterUiState.QuickLinkType.CommonClaimType -> onNavigateToCommonClaim(quickLink.commonClaim)
-                        is HelpCenterUiState.QuickLinkType.QuickActionType -> onQuickActionsSelected(quickLink.quickAction)
+                        is HelpCenterUiState.QuickLinkType.CommonClaimType -> onNavigateToCommonClaim(
+                          quickLink.commonClaim,
+                        )
+                        is HelpCenterUiState.QuickLinkType.QuickActionType -> onQuickActionsSelected(
+                          quickLink.quickAction,
+                        )
                       }
                     },
                     modifier = Modifier.fillMaxWidth(),
@@ -200,7 +204,9 @@ private fun HelpCenterHomeScreen(
                       Text(
                         text = when (quickLink) {
                           is HelpCenterUiState.QuickLinkType.CommonClaimType -> quickLink.commonClaim.title
-                          is HelpCenterUiState.QuickLinkType.QuickActionType -> stringResource(quickLink.quickAction.titleRes)
+                          is HelpCenterUiState.QuickLinkType.QuickActionType -> stringResource(
+                            quickLink.quickAction.titleRes,
+                          )
                         },
                         textAlign = TextAlign.Center,
                         modifier = Modifier.padding(16.dp),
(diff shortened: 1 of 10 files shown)


app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -166,78 +168,57 @@ private fun HelpCenterHomeScreen(
...
+          exit = fadeOut() + shrinkVertically(clip = false, shrinkTowards = Alignment.CenterVertically),
+        ) {
+          HelpCenterSection(
+            title = stringResource(R.string.HC_QUICK_ACTIONS_TITLE),
+            chipContainerColor = MaterialTheme.colorScheme.typeContainer,
+            contentColor = MaterialTheme.colorScheme.onTypeContainer,
+            content = {
               HedvigGrid(
-                insideGridSpace = InsideGridSpace.Companion.invoke(8.dp),
+                contentPadding = WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal).asPaddingValues(),
```
panasetskaya, line 182: The HedvigGrid layout itself is pretty complicated:) I guess I have one question here about WindowInsets here, I am not very familiar with them. Why do we apply WindowInsets.safeDrawing as contentPadding to the HedvigGrid, StillNeedHelpSection separately here, and not apply Modifier.windowInsetsPadding(WindowInsets.safeDrawing) to the whole Surface/Column on top? Wouldn't it do the same thing? I see that you got rid of Modifier.windowInsetsPadding(WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal)) for individual HedvigCards a bit lower.

hugokallstrom (approved):
LGTM!

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
StylianosGakis: Yes, applying them to the entire layout would indeed be the same thing as applying it to each item of the layout individually in many cases.
Now why do we need to do it on individual items in some layouts?

In this one in particular, if we did that, then we would not allow the `StillNeedHelpSection` would not actually be able to go edge to edge. Since it is also part of the outer column, if the column already had those insets, then the full area that the `StillNeedHelpSection` would be able to take would be everything *except* for the insets horizontally. Insets in this context being basically the navigation bar.

If the column does *not* have the padding on itself:
image

If the column does have the inset padding on itself and none of the children add their own insets:
image

Notice how on the left, the chat section is cut-off on the left, since that's where the camera lives on the real device (not visible in the screenshot)


----------------

Then for me getting rid of the insets from each card individually:
The thing which we want to move according to the insets is in this case the entire grid all together. So even by simplicity sake, this is in fact easier to read rather than having to go into the cards individually.
But most importantly, let's think of this scenario:
* We are in portrait mode, and on the right side of the app are the three-buttons for navigation, therefore we got some insets on the right side.
* We take those insets and we apply it on every card individually.
* Now since we are on a grid, the cards are laid out in a 2-column way, so we got cards on the left, and cards on the right
* Since we apply the insets on _all_ cards, the card on the right is moved a bit towards the left, to not be under the navigation bar (this is good). But then the card on the left side of the grid _also_ receives the same insets on the right side. So now it has some space on the right for no reason at all (this is the bad part)

So by moving the insets on the entire grid, we make it simpler while also fixing the bug where the insets are applied twice, in two different places while that is not what we want here.

Existing behavior:

image


Behavior after this PR:

image

panasetskaya: Thanks, that's very detailed!

PR #2022, opened 2024-02-29
Remove mockk

StylianosGakis:
I think we are no longer using this at all, wanna see if CI passes after removing it now

### DIFF
--- a/app/app/build.gradle.kts
+++ b/app/app/build.gradle.kts
@@ -226,7 +226,6 @@ dependencies {
   testImplementation(libs.assertK)
   testImplementation(libs.coroutines.test)
   testImplementation(libs.koin.test)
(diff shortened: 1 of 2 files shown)


panasetskaya:
> I think we are no longer using this at all, wanna see if CI passes after removing it now

what are we using for mocks now? :)

StylianosGakis:
We just do not use mocks, we use fake implementations of the classes we want to test instead if we can not use the real implementation in the first place.
This is https://testing.googleblog.com/2024/02/increase-test-fidelity-by-avoiding-mocks.html what I read which reminded me of the fact that we used to have many mocks in the past that we had to get rid of, and I wondered if we have in fact gotten rid of all of them. And it does seem like at this point we have gotten rid of all of them.

PR #2023, opened 2024-02-29
Feature/claim overview files audio

panasetskaya:
add files and audio record to display on ClaimSummary screen


app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
```diff
@@ -103,16 +137,112 @@ private fun ClaimSummaryScreen(
...
       Column(sideSpacingModifier.fillMaxWidth()) {
         for ((left, right) in detailPairs) {
-          Row {
-            Text(text = left, modifier = Modifier.weight(1f))
-            Text(text = right, modifier = Modifier.weight(1f), textAlign = TextAlign.End)
+          HorizontalItemsWithMaximumSpaceTaken(
+            startSlot = {
+              Text(text = left)
+            },
+            endSlot = {
```
StylianosGakis, line 144: I see the `modifier = Modifier.weight(1f)` was used in here before too, but since now we are inside `HorizontalItemsWithMaximumSpaceTaken` the two slots already get the right amount of space given to them. I think you can make do with just removing the `modifier = Modifier.weight(1f)` completely here? Do correct me if I am wrong here

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
```diff
@@ -103,16 +137,112 @@ private fun ClaimSummaryScreen(
...
+        0.dp
+      } else if (fileListSize <= 3) {
+        130.dp
+      } else if (fileListSize <= 6) {
+        260.dp
+      } else {
+        300.dp
+      }
+    LazyVerticalGrid(
+      modifier = Modifier.height(calculatedHeight),
```
StylianosGakis, line 185: Oh sorry I just realized I didn't mention that part. I was thinking we can just use `FilesLazyVerticalGrid` directly, not a custom one built for this place here. I feel like the way it's written it should be possible to re-use it, could you give that a try?

And then since that grid is more dynamic regarding how many columns it gives for the grid, we can instead not do any kind of height calculation ourselves here, and give it a sane default and let it be.
```suggestion
    LazyVerticalGrid(
      modifier = Modifier.height(300.dp),
```
I know this will mean on 1 file it will take a bit too much, and that on multiple files it will only show some files completely, but since we are in this scrollable-column inside a scrollable-column situation we gotta make some compromise I think.

It is also possible that I am thinking about this wrong, I think it might be a good idea for us to look at it together a bit so that I can also see how this visually looks like when we run it.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
```diff
@@ -126,6 +256,13 @@ private fun ClaimSummaryScreen(
   }
 }
 
+private fun getIconFromMimeType(mimeType: String) = when (mimeType) {
+  "image/jpg" -> Icons.Hedvig.Pictures
+  "video/quicktime" -> Icons.Hedvig.Play
+  "application/pdf" -> Icons.Hedvig.Document
+  else -> Icons.Hedvig.Document
+}
```
StylianosGakis, line 264: Using the existing component will also help us by avoiding creating this function a third time too.
I though we only had it in one place, I now realize we got it in 2 already, but I think that's another thing that we want to fix. It should be possible to have 1 unified way to render this that we can re-use from the now 3 places we want to show it.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
```diff
@@ -143,7 +280,7 @@ private fun PreviewClaimSummaryScreen() {
             ),
             itemType = ClaimSummaryInfoUiState.ItemType.Model(
               itemModel = ItemModel.Known(
-                displayName = "Apple iPhone 14 Pro Super Omega New Model",
+                displayName = "Apple iPhone 14 Pro",
```
StylianosGakis, line 283: I didn't realize I left that in 😂
I did it to test longer texts 😅

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
panasetskaya: Ok! I'll move it to core-icons then.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
panasetskaya: Yeah, it worked just fine with longer text, I just wanted to see how the normal-sized columns looked 😂

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
panasetskaya: I have changed it to FilesLazyVerticalGrid in the last commit, but I'm actually not sure it is a better decision than this custom grid here. Or maybe some other thing altogether? I still left checking for an empty list, since for this case it really looks bad with empty 300.dp and it's easy to fix, but otherwise it's not quite possible to make it look good with with for example only 2 files and fixed 300.dp height (see screenshot, the button is hidden below with no reason). I looked up how it's done on AddFilesScreen, but the situation is a bit different there:)

Screenshot_20240229_174953_Hedvig Dev

StylianosGakis: Yeahh I agree it's looking odd. Let's look at it together soon.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryDestination.kt
StylianosGakis: Hmm I had something else in mind actually here.
core-icons is a nice little module that can contain all of the `Icons.Hedvig.Foo` icons in there, I think we should keep it like that actually. Moving it there will mean that basically every single feature module that we have, since it needs to access our icons then automatically gets access to the `getIconFromMimeType` but they really do not need to know about it. I would optimally prefer this to be a private function inside the UI component which is made to render the grid with those icons.

So this `FilesLazyVerticalGrid` component can stay in core-ui, since this is a common UI component that we use througout our app, but it's not necessarily part of our design-system, it's a bit bigger than that, hence core-ui is the right place for it to be. And this `getIconFromMimeType` can stay private in there I think.

panasetskaya: Ok, got it! Moved it here, though still left it public because ClaimDetailsDestination also uses it outside FilesLazyVerticalGrid.

PR #2024, opened 2024-03-01
Fix the screen clipping its size when navigating away from UrlFetched

StylianosGakis:
When we were in the `UrlFetched` state of this screen, there was no UI emitted. This actually ends up making the content of the NavHost go down to showing no UI, so since the entire space is not needed, it only takes up 0 pixel and clips the rest of the content. This is done with a center alignment by default in the NavHost. This results in the when navigating away from this screen, the next screen as it animates into view, it also animates the content size accidentally, expanding from the center towards the edges, making it look like it's clipped. This wraps the entire screen in a simple box which fills max size to ensure that no matter which state the screen is in, it always takes up the entire screen size as it should, and this fixes the navigation clipping issue.

Also run ktlintFormat

### DIFF
--- a/app/datadog/datadog-core/src/main/kotlin/com/hedvig/android/datadog/core/memberid/DatadogMemberIdUpdater.kt
+++ b/app/datadog/datadog-core/src/main/kotlin/com/hedvig/android/datadog/core/memberid/DatadogMemberIdUpdater.kt
@@ -6,8 +6,6 @@ import com.hedvig.android.datadog.core.attributestracking.DatadogAttributesManag
 import com.hedvig.android.initializable.Initializable
 import com.hedvig.android.logger.LogPriority
 import com.hedvig.android.logger.logcat
-import kotlin.coroutines.CoroutineContext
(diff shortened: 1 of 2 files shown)


panasetskaya:
> When we were in the `UrlFetched` state of this screen, there was no UI emitted. This actually ends up making the content of the NavHost go down to showing no UI, so since the entire space is not needed, it only takes up 0 pixel and clips the rest of the content. This is done with a center alignment by default in the NavHost. This results in the when navigating away from this screen, the next screen as it animates into view, it also animates the content size accidentally, expanding from the center towards the edges, making it look like it's clipped. This wraps the entire screen in a simple box which fills max size to ensure that no matter which state the screen is in, it always takes up the entire screen size as it should, and this fixes the navigation clipping issue.
> 
> Also run ktlintFormat

So I can see that you've wrapped everthing in a Box to avoid that, is that right? I'm a bit confused with the code changes, seems to be a lot of things removed - but probably just moved, right?

PR #2025, opened 2024-03-04
Fix/separate audio player modules

panasetskaya:
Divide audio-player module into two: audio-player-ui and audio-player-data


app/audio-player-ui/src/main/kotlin/com/hedvig/android/audio/player/audioplayer/RememberAudioPlayer.kt
```diff
@@ -265,3 +225,18 @@ private class AudioPlayerImpl(
     }
   }
 }
+
+@Composable
+fun rememberAudioPlayer(playableAudioSource: PlayableAudioSource): AudioPlayer {
```
StylianosGakis, line 230: This file is now named `RemeberAudioPlayer.kt`.
Maybe we can name it `AudioPlayerImpl.kt` instead since that's the main class in here, while it just also has a `remember` composable function it it?

app/audio-player-data/src/main/kotlin/com/hedvig/audio/player/data/SIgnedAudioUrl.kt
```diff
@@ -1,9 +1,13 @@
...
+
+import kotlinx.serialization.Serializable
 
 /**
  * A class to hold the url coming from the backend, with an overridden equals in order to return true in case we're
  * dealing with the same audio file but with a different query string
  */
+
+@Serializable
 class SignedAudioUrl private constructor(
```
StylianosGakis, line 11: This was not @ Serializable before, I assume it is now because a new destination needs to access it right?

gradle/libs.versions.toml
```diff
@@ -194,6 +197,9 @@ kotlinter-gradlePlugin = { module = "org.jmailen.gradle:kotlinter-gradle", versi
 
 # lint dependencies
 lintApi = { module = "com.android.tools.lint:lint-api", version.ref = "lintApi" }
+androidx-test-ext-junit = { group = "androidx.test.ext", name = "junit", version.ref = "androidx-test-ext-junit" }
+espresso-core = { group = "androidx.test.espresso", name = "espresso-core", version.ref = "espresso-core" }
+material = { group = "com.google.android.material", name = "material", version.ref = "material" }
```
StylianosGakis, line 202: Were these added by accident perhaps? 👀
I don't see them being used

StylianosGakis (commented):
I think the changes inside libs.versions.toml are not needed right? Maybe just revert those before merging this?
Or am I missing something?

app/audio-player-data/src/main/kotlin/com/hedvig/audio/player/data/SIgnedAudioUrl.kt
panasetskaya: Yep, after I merge this I will need to use it somewhere where it has to be serializable, so I've decided to annotate it right away

PR #2026, opened 2024-03-05
Claim overview files audio grid fix

StylianosGakis:
Fix a custom layout which always ensures that the grid takes up at least
160.dp worth of height, while still taking up as much space as it can
when there is enough space available which the other elements in the
layout are not taking up by themselves.


panasetskaya:
We thoroughly went through it, I think it's perfect:)

PR #2035, opened 2024-03-07
Show app review dialog after completing co insured self serve

hugokallstrom:


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/CommitMidtermChangeUseCase.kt
```diff
@@ -35,6 +37,7 @@ internal class CommitMidtermChangeUseCaseImpl(
     when (val state = result.midtermChangeIntentCommit.intent?.state) {
       MidtermChangeIntentState.COMPLETED -> {
         networkCacheManager.clearCache()
+        selfServiceCompletedEventStore.onSelfServiceCompleted()
```
hugokallstrom, line 40: Will add this to moving flow and claim submission as well

app/core/core-app-review/src/main/kotlin/com/hedvig/android/core/appreview/ReviewDialogViewModel.kt
```diff
@@ -12,20 +11,19 @@ import kotlinx.coroutines.flow.receiveAsFlow
...
   val shouldOpenReviewDialog: Flow<Boolean> = _shouldOpenReviewDialog.receiveAsFlow()
 
   init {
     viewModelScope.launch {
-      chatClosedEventStore.observeChatClosedCounter()
-        .map { it % 4 == 0 && it != 0 }
-        .filter { it == true }
+      selfServiceCompletedEventStore.observeNumberOfCompletedSelfServices()
+        .map { it > 0 }
+        .filter { it }
```
panasetskaya, line 23:
```suggestion
        .filter { it > 0 }
```

panasetskaya: Just felt a bit confused :)

StylianosGakis (approved):
Super good idea.
I took the time to do this https://github.com/HedvigInsurance/android/pull/2041 on top of this just to simplify a bit the fact that we do not need the VM as we had it in the past. Also made it so that it will not try to call the review dialog whenever the app is opened and they may have already done a self service process in the past. This would happen before because it > 0 would be true on startup, and then we'd try to trigger it.

That PR makes it so that whenever a new observer is looking into the flow, the first emission from the Datastore is ignored, as that's the one that just emits whatever there was inside the DataStore itself. And then the *next* emission, which can only happen when a self-service flow is successfully made, then we try to launch the review dialog

PR #2036, opened 2024-03-07
Feature/new car claim flow

panasetskaya:
Add towing deflect screen with copy content and navigation to it.
Add eir screen without copy or navigation to it.


StylianosGakis (approved):
Super ⭐️

PR #2041, opened 2024-03-08
App review dialog additions

StylianosGakis:
Removes the ViewModel as it was only used for the purpose of holding the logic. That can just be a UseCase directly without it being a AAC VM and whatever else comes with that.

The WaitUntilAppReviewDialogShouldBeOpenedUseCase is a single stop of calling that, and knowing that when you are back from it you should show the review dialog. Modeling it as a suspend function also engrains the fact that we only want to call it once in one app lifetime.

Added a bunch of logs around the review launch itself, to be able to perhaps tell a bit more around how many times this actually is called.


app/app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -135,11 +134,9 @@ class LoggedInActivity : AppCompatActivity() {
       launch {
         lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
           authTokenService.authStatus.first { it is AuthStatus.LoggedIn }
-          reviewDialogViewModel.shouldOpenReviewDialog.collect { shouldOpenReviewDialog ->
-            if (shouldOpenReviewDialog) {
-              showReviewWithDelay()
-            }
-          }
+          waitUntilAppReviewDialogShouldBeOpenedUseCase.invoke()
+          delay(REVIEW_DIALOG_DELAY_MILLIS)
```
panasetskaya, line 138: why delay?

app/core/core-app-review/src/main/kotlin/com/hedvig/android/core/appreview/WaitUntilAppReviewDialogShouldBeOpenedUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+}
+
+internal class WaitUntilAppReviewDialogShouldBeOpenedUseCaseImpl(
+  private val dataStore: SelfServiceCompletedEventStore,
+) : WaitUntilAppReviewDialogShouldBeOpenedUseCase {
+  override suspend fun invoke() {
+    dataStore.observeNumberOfCompletedSelfServices()
+      // Drop the first number, whatever it was, to not show a dialog right as the app is launched. We are only
+      // interested in observing changes to this number after we've already started observing it
+      .drop(1)
```
panasetskaya, line 21: That drop would only be used if we by some reason haven't showed the review dialog last time the app was opened  and the member completed selfService, right?

app/app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
StylianosGakis: We have had this in the past too.
The decision was made as a result of basically what I am saying here https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1709909630305479
A combination of not wanting to show the Dialog immediately as we are done with a flow, but before the member actually realizes that the flow is done (confusion)
But also it has to be a number which if possible will not overlap with if a member at the same time tries to click something else, so they accidentally dismiss the review dialog. It's quite easy to happen surprisingly, I've done it many times in the past.

app/core/core-app-review/src/main/kotlin/com/hedvig/android/core/appreview/WaitUntilAppReviewDialogShouldBeOpenedUseCase.kt
StylianosGakis: Hmm not quite, it's a bit of a complicated situation but I can try to explain.
The place in which we are listening in on this flow is wrapped in a `repeatOnLifecycle(Lifecycle.State.STARTED)`, since we only want to ever call this if the app is in the foreground.
So that means that if someone goes to home, and comes back into the app, this will in fact be subscribed to again from scratch.
The DataStore, when a new observer is trying to observe a value from inside of it, will on the flow it gives you emit the first value that it has already in the disk, and then it will wait until any changes happen so that it emits again.
So in short, this `drop(1)` will simply ignore the first emission from the DataStore since that one is just what was on the disk itself, we do not care for what it already was. We only want to know that if it has *changed*, which may only happen if one was in the app already, and the `selfServiceCompletedEventManager.completedSelfServiceSuccessfully()` was called, so the counter has gone up, so the datastore will emit another value, which will continue further down in this flow chain, since `drop(1)` will have already happened.

StylianosGakis: Yup the in-app review APIs handle this automatically for us. https://developer.android.com/guide/playcore/in-app-review#quotas
"To provide a great user experience, Google Play enforces a time-bound quota on how often a user can be shown the review dialog. Because of this quota, calling the launchReviewFlow method more than once during a short period of time (for example, less than a month) might not always display a dialog."

PR #2044, opened 2024-03-11
close claim flow when opening the browser for Eir

panasetskaya:
Closing the claim flow when going to browser to open Eir link (other car damage claims) after 3 min delay (just decided). And small design changes, pointed out by William

### DIFF
--- a/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
+++ b/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
@@ -107,7 +107,7 @@ internal fun darkHedvigColorScheme(hedvigTonalPalette: HedvigTonalPalette) = Hed
   onContainedButtonContainer = hedvigTonalPalette.greyscale1000,
   secondaryContainedButtonContainer = hedvigTonalPalette.greyscale800,
   onSecondaryContainedButtonContainer = hedvigTonalPalette.greyscale100,
-  alwaysBlackContainer = hedvigTonalPalette.greyscale700,
+  alwaysBlackContainer = hedvigTonalPalette.greyscale900,
   onAlwaysBlackContainer = hedvigTonalPalette.greyscale25,
   // In the comments are the light mode colors, showing the equivalent and how it was chosen
   warningElement = hedvigTonalPalette.amber700,
--- a/app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/informdeflect/DeflectCarOtherDamageDestination.kt
+++ b/app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/informdeflect/DeflectCarOtherDamageDestination.kt
@@ -17,11 +17,6 @@ import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.material3.windowsizeclass.WindowSizeClass
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.getValue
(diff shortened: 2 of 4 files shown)


app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
```diff
@@ -107,7 +107,7 @@ internal fun darkHedvigColorScheme(hedvigTonalPalette: HedvigTonalPalette) = Hed
   onContainedButtonContainer = hedvigTonalPalette.greyscale1000,
   secondaryContainedButtonContainer = hedvigTonalPalette.greyscale800,
   onSecondaryContainedButtonContainer = hedvigTonalPalette.greyscale100,
-  alwaysBlackContainer = hedvigTonalPalette.greyscale700,
+  alwaysBlackContainer = hedvigTonalPalette.greyscale900,
```
StylianosGakis, line 110: This is now going to be greyscale900 on top of a greyscale1000 background. Is this what design wanted here? The contrast between the two is quite low is my concern, it barely looks like a card in high light environments or if the phone's brightness is not that high :D

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/informdeflect/DeflectCarOtherDamageDestination.kt
```diff
@@ -69,6 +76,17 @@ private fun DeflectCarOtherDamageScreen(
     closeClaimFlow = closeClaimFlow,
     topAppBarText = stringResource(id = R.string.SUBMIT_CLAIM_CAR_TITLE),
   ) {
+    var shouldCloseFlow by remember {
+      mutableStateOf(false)
+    }
+
+    LaunchedEffect(shouldCloseFlow) {
+      if (shouldCloseFlow) {
+        delay(3.minutes) // todo: on the test session decided to make a delay here
```
StylianosGakis, line 85: We do not get any guarantees that our app will be kept running in the background for this long in order to do this reliably. If the app is killed inside those 3 minutes this will simply never be called. I will bring this up in the slack channel instead to pick alternatives. I would prefer us not to do this and get unreliable results.

StylianosGakis (changes requested):
I see the design changes, it makes sense. I a bit skeptical about the 900 on 1000 color on the card but if that's what we want then that's fine.
I am marking this as "request changes" because I want to discuss the 3 minutes rule with the rest first.

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
panasetskaya: They said they want it the same as the expandable fields below, and the expandable fields have, if I'm not mistaken, Surface as container color - and it's this one. Actually, at the meeting William said that the expandables are too dark, but afterwards he sent me the right color, and it's #303030 - which is exactly what we had there, Surface color for dark theme

StylianosGakis: Alrighty then, sounds good to me. Then how about we remove the other change here with the delay before we close the flow and for now do nothing there so that we can merge the design changes at least.
Then when we make the decision on if the flow will be eagerly closed or not we can add that code later. It'd be a shame for the design changes to be postponed from being merged because of the other issue

panasetskaya: Removed the delay with the last commit. I guess it's okay to merge now?

PR #2045, opened 2024-03-11
Contract deep link + LocalUriHandler which tries to treat URIs as deep links in our app before delegating to the default UriHandler

StylianosGakis:
The `"$baseDeepLinkDomain/contract?contractId={contractId}"` deep link route is added to be able to deep link directly into a specific contract along with the contractId

A new NoContractFound case is added to differentiate in the UI between other retryable errors and if the deep link ID just did not match any contract.

Add DeepLinkFirstUriHandler as our global LocalUriHandler which should try to open uris as deep links before it tries to go outside of the app. This fixes the problem with opening our app again in a separate task as this issue https://issuetracker.google.com/issues/329046732 indicates which broke the behavior of pressing the back or the navigateUp buttons

A big part of this PR is me just renaming inside HedvigErrorSection the lambda for the button from `retry` to `onButtonClick` which I kinda regret for this PR, it could be another, but it is what it is. It now reflects better that this button does more than just retry.

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
@@ -14,11 +14,11 @@ internal class DeepLinkFirstUriHandler(
   private val delegate: UriHandler,
 ) : UriHandler {
   override fun openUri(uri: String) {
-    try {
+    if (navController.graph.hasDeepLink(uri.toUri())) {
       logcat { "DeepLinkFirstUriHandler will try to navigate to uri:$uri" }
       navController.navigate(uri.toUri())
       logcat { "DeepLinkFirstUriHandler did deep link to uri:$uri" }
-    } catch (e: IllegalArgumentException) {
+    } else {
       logcat { "DeepLinkFirstUriHandler falling back to default UriHandler for uri:$uri" }
       delegate.openUri(uri)
     }
--- a/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
@@ -9,7 +9,10 @@ import com.hedvig.android.logger.logcat
  * Tries to open a deep link with the [navController] and falls back to the [delegate] if the uri is not a valid deep
  * link to our app
  */
-internal class DeepLinkFirstUriHandler(private val navController: NavController, private val delegate: UriHandler) : UriHandler {
+internal class DeepLinkFirstUriHandler(
+  private val navController: NavController,
+  private val delegate: UriHandler,
+) : UriHandler {
   override fun openUri(uri: String) {
     try {
       logcat { "DeepLinkFirstUriHandler will try to navigate to uri:$uri" }
--- a/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
@@ -1,0 +1,23 @@
+package com.hedvig.android.app.ui
+
+import androidx.compose.ui.platform.UriHandler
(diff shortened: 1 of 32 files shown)


app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
```diff
@@ -0,0 +1,26 @@
...
+  private val delegate: UriHandler,
+) : UriHandler {
+  override fun openUri(uri: String) {
+    if (navController.graph.hasDeepLink(uri.toUri())) {
+      logcat { "DeepLinkFirstUriHandler will try to navigate to uri:$uri" }
+      navController.navigate(uri.toUri())
+      logcat { "DeepLinkFirstUriHandler did deep link to uri:$uri" }
+    } else {
+      logcat { "DeepLinkFirstUriHandler falling back to default UriHandler for uri:$uri" }
+      delegate.openUri(uri)
```
panasetskaya, line 23: So if I understood correctly, this delegate (which is LocalUriHandler.current in our case) when opening a link, creates a separate backstack, and this causes all the problems, right? So you avoid this by calling navController directly first, which should in reality cover all cases with our deep links?

app/app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
```diff
@@ -175,25 +178,29 @@ class LoggedInActivity : AppCompatActivity() {
...
-          clearInitialTab = {
-            intent.removeExtra(INITIAL_TAB)
-          },
-          shouldShowRequestPermissionRationale = ::shouldShowRequestPermissionRationale,
-          market = market,
-          imageLoader = imageLoader,
-          languageService = languageService,
-          hedvigBuildConstants = hedvigBuildConstants,
-        )
+      CompositionLocalProvider(
```
panasetskaya, line 181: and here I understood that I in fact do not fully understand the usage of CompositionLocalProvider :)) probably need to talk about that

StylianosGakis: Good idea, let's talk about it on Thursday.
In the meantime interestingly the way I learned about this like 3 years ago was by actually looking at what React Context is https://legacy.reactjs.org/docs/context.html. It's more or less the same concept 😅

I see that by now there are some nice docs https://developer.android.com/jetpack/compose/compositionlocal for compose too.
The gist is that we want to avoid using those in most cases, since they bring in implicit dependencies on this local being present but there's no guarantee that it's gonna be present and if it's not it's a runtime crash.

Here I felt like using it still since it's a composition local that is provided by Android already by default, we are only enchasing its capabilities with this change.

And by avoiding using those most of the time we've seen that problem before with some of our previews failing to work because some permission APIs are internally using a composition local which assumes some things that are not true for the previews. It's a super super easy API to misuse and bring maintainability problems, so that's why we do not use it for anything else besides what is already provided by the system (LocalUriHandler, LocalContext, LocalResources etc.) and what the m3 library wants in order to work (MaterialTheme.colorScheme... is also working off of composition locals)

app/app/src/main/kotlin/com/hedvig/android/app/ui/DeepLinkFirstUriHandler.kt
StylianosGakis: Yeap that is exactly the problem.
It kinda opens our app as if we were coming from a deep link from another app. I am not sure how this works exactly, but then when we end up in the deep link, we have no backstack behind us, we are just in our app in a new "Task" I think it's called, and we lose track of what was there before.
And yes exactly, if it's our own deep link this should just work instantly and it's all good. If not it will in any case go and try to open another app or a website. Exactly what we want here!

app/app/src/main/kotlin/com/hedvig/app/feature/loggedin/ui/LoggedInActivity.kt
panasetskaya: Great, I'll read it up by Thursday then

PR #2047, opened 2024-03-12
Add fallback for contract deep links that do not have a contractId

StylianosGakis:
Without this change, a link to `{baseUrl}/contract` crashed the app since it was unable to take a null string and turn it into the `String` type that the destination was expecting.


sladan-hedvig (approved):
I will just pretend I know whats going on here and press Approve 👍

PR #2051, opened 2024-03-14
change quick links design

panasetskaya:
in draft because we don't have copy translations yet here

### DIFF
--- a/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/commonclaim/CommonClaimsData.kt
+++ b/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/commonclaim/CommonClaimsData.kt
@@ -1,14 +1,17 @@
 package com.hedvig.android.feature.help.center.commonclaim
 
+import hedvig.resources.R
 import kotlinx.serialization.Serializable
 import octopus.CommonClaimsQuery
 
 internal sealed interface CommonClaim {
   val title: String
+  val hintTextRes: Int?
 
   @Serializable
   data class Emergency(
     override val title: String,
+    override val hintTextRes: Int?,
     val emergencyNumber: String,
   ) : CommonClaim {
     companion object {
@@ -19,6 +22,7 @@ internal sealed interface CommonClaim {
       ) = Emergency(
         title = title,
         emergencyNumber = layout.emergencyNumber,
+        hintTextRes = R.string.HC_QUICK_ACTIONS_SICK_ABROAD_SUBTITLE, // as we only have one emergency: sick abroad
       )
     }
   }
@@ -27,6 +31,7 @@ internal sealed interface CommonClaim {
   data class Generic(
     val id: String,
     override val title: String,
+    override val hintTextRes: Int?,
     val bulletPoints: List<BulletPoint>,
   ) : CommonClaim {
     @Serializable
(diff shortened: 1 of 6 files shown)


panasetskaya:
Have the copy now:)

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -190,27 +186,47 @@ private fun HelpCenterHomeScreen(
...
-                        modifier = Modifier.padding(16.dp),
+                        textAlign = TextAlign.Start,
+                        modifier = Modifier.padding(horizontal = 16.dp),
+                      )
+                      Spacer(modifier = Modifier.height(4.dp))
+                      Text(
+                        text = when (quickLink) {
+                          is HelpCenterUiState.QuickLinkType.CommonClaimType -> {
+                            val hintTextRes = quickLink.commonClaim.hintTextRes
+                            if (hintTextRes != null) stringResource(hintTextRes) else ""
```
StylianosGakis, line 218: If we do in fact get an empty string here, do we want the text to still show a bit off-center as it will, or should it become centered vertically in the card, or should the card itself shrink?
I assume design does not handle this case right?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -190,27 +186,47 @@ private fun HelpCenterHomeScreen(
...
+                          }
+                          is HelpCenterUiState.QuickLinkType.QuickActionType -> stringResource(
+                            quickLink.quickAction.hintTextRes,
+                          )
+                        },
+                        textAlign = TextAlign.Start,
+                        modifier = Modifier.padding(horizontal = 16.dp),
+                        color = MaterialTheme.colorScheme.onSurfaceVariant,
+                        fontSize = MaterialTheme.typography.titleSmall.fontSize,
```
StylianosGakis, line 227: I'd suggest you just pass the style completely here instead of only changing the font size.
Something like `style = MaterialTheme.typography.titleSmall`

StylianosGakis (approved):
Lgtm, just had a couple of small points

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -190,27 +186,47 @@ private fun HelpCenterHomeScreen(
...
                       )
+                      Spacer(modifier = Modifier.height(14.dp))
```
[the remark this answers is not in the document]
panasetskaya: Not sure, but that's what Figma shows me. It does look a bit asymetrical. I'll change it to 12 in both cases

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
panasetskaya: Nope, but again - there are only two common claims now, and both have hint text, so I think that's why they didn't make the design. I thought that in an unlikely case we don't have the hint, it would be more logical to have the title a bit off-center, so all cards look alike

StylianosGakis: Yeap sounds good to me

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -178,10 +176,8 @@ private fun HelpCenterHomeScreen(
             chipContainerColor = MaterialTheme.colorScheme.typeContainer,
             contentColor = MaterialTheme.colorScheme.onTypeContainer,
             content = {
-              HedvigGrid(
-                contentPadding = WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal).asPaddingValues(),
-                insideGridSpace = InsideGridSpace(8.dp),
-                modifier = Modifier.padding(horizontal = 16.dp),
+              Column(
+                verticalArrangement = Arrangement.spacedBy(4.dp),
```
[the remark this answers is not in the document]
panasetskaya: yep, forgot about landscape. fixed:)

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/commonclaim/CommonClaimsData.kt
```diff
@@ -43,16 +48,24 @@ internal sealed interface CommonClaim {
...
+          },
+        )
+      }
     }
   }
 }
+
+private fun checkIfFirstVet(id: String): Boolean {
+  return id == "30" || id == "31" || id == "32"
+}
```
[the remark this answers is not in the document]
panasetskaya: It's used in the static fun from(), but you're right, I totally shouldn't have it twice. I changed the local variable to call the same fun, would that be okay?

PR #2052, opened 2024-03-15
Forever in payments

StylianosGakis:
This is exactly what `feature/forever-in-payments` had as well, but with some things on top.


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsDestination.kt
```diff
@@ -146,31 +151,168 @@ internal fun DiscountsScreen(
...
+            .clip(MaterialTheme.shapes.squircleLarge)
+            .clickable { showForeverInfoBottomSheet = true }
+            .minimumInteractiveComponentSize(),
+        )
+      },
+    )
+    Spacer(modifier = Modifier.height(16.dp))
+    HorizontalItemsWithMaximumSpaceTaken(
+      startSlot = {
+        HedvigCard(
```
panasetskaya, line 211: Maybe also add copying the code on click here?

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/Discount.kt
```diff
@@ -9,9 +9,17 @@ internal data class Discount(
   val code: String,
   val displayName: String?,
   val description: String?,
-  val expiresAt: LocalDate?,
+  val expiredState: ExpiredState,
   val amount: UiMoney?,
   val isReferral: Boolean,
 ) {
-  fun isExpired(now: LocalDate) = expiresAt?.let { it < now } ?: true
+  sealed interface ExpiredState {
```
panasetskaya, line 16: yay

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsDestination.kt
```diff
@@ -146,31 +151,168 @@ internal fun DiscountsScreen(
...
+          colors = ButtonDefaults.buttonColors(
+            containerColor = MaterialTheme.colorScheme.containedButtonContainer,
+            contentColor = MaterialTheme.colorScheme.onContainedButtonContainer,
+          ),
+          textStyle = MaterialTheme.typography.bodyMedium,
+          text = stringResource(R.string.important_message_read_more),
+          onClick = { showForeverInfoBottomSheet = true },
+        )
+      },
+    )
```
panasetskaya, line 258: I think on the last design critique William said that the whole list of the referrals that the member already has should also be displayed here. But I think we're not doing it before they tell if it's okay from the legal perspective to show referrals' names here?

panasetskaya (approved):

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsDestination.kt
StylianosGakis: Yeah the names I am exactly postponing until we make a decision.
For the forever well since it turns out we're keeping it as a 5th tab I will try to see how the design evolves after that decision. I think I'll merge for now and do anything else on top of this

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsDestination.kt
StylianosGakis: Will do!

PR #2053, opened 2024-03-15
Make PaymentHistoryDestination use a local PaymentHistory

StylianosGakis:
This avoids doing list operations in-between the UI code mostly

### DIFF
--- a/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/preview/ConvenienceCollectionPreviewParameterProvider.kt
+++ b/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/preview/ConvenienceCollectionPreviewParameterProvider.kt
@@ -6,6 +6,15 @@ class BooleanCollectionPreviewParameterProvider : CollectionPreviewParameterProv
   listOf(true, false),
 )
 
+/**
+ * Models 3 distinct permutations
+ */
+class TripleBooleanCollectionPreviewParameterProvider : CollectionPreviewParameterProvider<TripleCase>(
+  listOf(TripleCase.FIRST, TripleCase.SECOND, TripleCase.THIRD),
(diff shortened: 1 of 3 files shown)


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/preview/ConvenienceCollectionPreviewParameterProvider.kt
```diff
@@ -6,6 +6,15 @@ class BooleanCollectionPreviewParameterProvider : CollectionPreviewParameterProv
   listOf(true, false),
 )
 
+/**
+ * Models 3 distinct permutations
+ */
+class TripleBooleanCollectionPreviewParameterProvider : CollectionPreviewParameterProvider<TripleCase>(
```
panasetskaya, line 12: Is this one to be used in Previews when we need 3 distinct states?

StylianosGakis: Yeahh, I am still not 100% happy with how this looks like now that I see it again, I maybe should make it an enum of three cases tbh. I was trying out some other things back then and it ended up looking as it does now. I'll try to fix that before merging though.

StylianosGakis: Oh 🤦 I already did that change. I actually forgot 😅
But yes that's exactly the scenario. So that you do not need to create another CollectionPreviewParameterProvider for those cases. I am not 100% convinced this is a better API than just creating a custom CollectionPreviewParameterProvider on each place, but we can always adjust this in the future accordingly

PR #2056, opened 2024-03-18
Payments bottom nav tab

StylianosGakis:
The history is a bit confusing as I first did the work to move Forever outside of the nav bars, but then we decided to keep it as a 5th tab, so I added it back in :D 

I also took the opportunity to cleanup some of the code around this. The optimal state of this is that `AppDestination` interface eventually goes away and all destinations exist only inside their perspective modules, no global destinations like this. So I got rid of some of them at least :D


app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigApp.kt
```diff
@@ -36,50 +35,39 @@ import androidx.compose.ui.platform.LocalDensity
...
   activityNavigator: ActivityNavigator,
-  getInitialTab: () -> TopLevelGraph?,
-  clearInitialTab: () -> Unit,
   shouldShowRequestPermissionRationale: (String) -> Boolean,
   market: Market,
   imageLoader: ImageLoader,
   languageService: LanguageService,
   hedvigBuildConstants: HedvigBuildConstants,
 ) {
-  LaunchedEffect(getInitialTab, clearInitialTab, hedvigAppState) {
```
panasetskaya, line 65: not entirely understand what this launchedEffect did before

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -187,27 +200,30 @@ private fun TopLevelDestinationNavigationSideEffect(
   DisposableEffect(navController, tabNotificationBadgeService, coroutineScope) {
```
panasetskaya, line 200: Why is it better to have a DisposableEffect here and not LaunchEffect if we have to pass a coroutineScope as a key here anyway, and it's default rememberCoroutineScope()? Can't find where it might be changed

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/TopLevelGraph.kt
```diff
@@ -1,39 +1,9 @@
...
-      return when (input) {
-        "HOME" -> HOME
-        "INSURANCE" -> INSURANCE
-        "REFERRALS" -> FOREVER
-        "PROFILE" -> PROFILE
-        else -> null
-      }
-    }
-  }
+enum class TopLevelGraph {
```
panasetskaya, line 3: I guess they shouldn't be serializable now?

StylianosGakis: They are no longer a `Destination` but just an enum to enumerate all the possible graphs we got. We use this as a way to be able to exhaustively go over them inside HedvigAppState for example. Like here
https://github.com/HedvigInsurance/android/blob/afb510b3ebdf48ebc9e9597d31fc5de13d327081/app/app/src/main/kotlin/com/hedvig/android/app/ui/IsTopLevelGraphInHierarchy.kt#L19-L23
or here
https://github.com/HedvigInsurance/android/blob/e6a57dc642bf66349834d4ed219cebe9b78ba4b8/app/navigation/navigation-ui/src/main/kotlin/com/hedvig/android/navigation/ui/TopLevelGraphExt.kt#L2-L1
or here
https://github.com/HedvigInsurance/android/blob/14eaa9cbb8cbe59d074790c62e754b5fc7738460/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt#L162
etc.

The graph destination itself which the navigation library is using is a different type, like here https://github.com/HedvigInsurance/android/blob/df388b81b92baa091a692146b0558a524424a767/app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/navigation/ProfileDestinations.kt#L9

This allows those to just stay inside their modules without having to be available everywhere like they used to in AppDestinations, but we still want to be able to exhaustively go over them, hence this enum.

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
StylianosGakis: It's because what we do inside here is we register a listener to the navController.
So we need to make 100% sure that whatever we regiester, we also _unregister_ once this leaves composition, or one of the keys change (and we then add the new listener in that case)
This is done here with the onDispose https://github.com/HedvigInsurance/android/blob/14eaa9cbb8cbe59d074790c62e754b5fc7738460/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt#L239-L241 which DispoableEffect forces you to call in the end.

Read more here https://developer.android.com/jetpack/compose/side-effects#disposableeffect 😊

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigApp.kt
StylianosGakis: When we had many more activities, a bunch of those would do something like
`LoggedInActivity.newInstance(..., initialTab = ForeverTab)`, so they'd navigate here but also inform this to also navigate to the right top level tab as soon as you've gone to LoggedInActivity.
In practice this now basically never gets called anymore, so it was time to get rid of the baggage.

PR #2057, opened 2024-03-18
Add support for "recents" links to our deep-linkable destinations

StylianosGakis:
Adding this makes all of our destinations which contain a deep link to show the link icon in the "Recent" screens in all phones that support that feature.
More Context here https://developer.android.com/guide/components/activities/recents#url-sharing

Some screenshots:
| home | insurances | Specific insurance with args |
| --- | --- | --- |
| image | image | image |

This just does nothing in screens where there are no deep links, like the settings one
image

And a video:

https://github.com/HedvigInsurance/android/assets/44558292/78d2a04f-eebf-4a40-9d6a-25a3e46fea79


gradle/libs.versions.toml
```diff
@@ -169,6 +170,7 @@ kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serializa
 ktor = { module = "io.ktor:ktor-client-android", version.ref = "ktor" }
 leakCanary = { module = "com.squareup.leakcanary:leakcanary-android", version.ref = "leakCanary" }
 moneta = { module = "org.javamoney.moneta:moneta-core", version.ref = "moneta" }
+navigationRecentsUrlSharing = { module = "com.stylianosgakis.navigation.recents.url.sharing:navigation-recents-url-sharing", version.ref = "navigationRecentsUrlSharing" }
```
panasetskaya, line 173: where is the lib to take a look at?:)

StylianosGakis: https://github.com/StylianosGakis/navigation-recents-url-sharing :D

The logic is here https://github.com/StylianosGakis/navigation-recents-url-sharing/blob/main/library/navigation-recents-url-sharing/src/main/java/com/stylianosgakis/navigation/recents/url/sharing/ProvideAssistContent.kt

And here is a test that passes that kinda showcases how it works when navigating to some deep-linkable destinations https://github.com/StylianosGakis/navigation-recents-url-sharing/blob/main/library/navigation-recents-url-sharing/src/androidTest/java/com/stylianosgakis/navigation/recents/url/sharing/ProvideAssistContentTest.kt

PR #2059, opened 2024-03-19
Custom bottom tab which gives more space to the text label to grow horizontally before wrapping

StylianosGakis:
HedvigAppState now exposes a much more convenient NavigationSuiteType to determine what should show. Inspired from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3-adaptive-navigation-suite/src/commonMain/kotlin/androidx/compose/material3/adaptive/navigationsuite/NavigationSuiteScaffold.kt;l=338?q=NavigationSuiteType&sq=
And extracted the logic of which navigation item is used, rail or bar inside `NavigationSuite` composable. This also made it simpler to write some previews for it.

The biggest part of the diff is just bringing in the entire `HedvigNavigationBar.kt` file from material3. We have to copy all of the source so that we can edit the details, otherwise it's not possible.

The navigation bar now gives the labels a bit more space to grow, as described in NavigationBarLabelHorizontalBreakoutPadding, and the font size is allowed to grow over 100%, but only up to NavigationBarFontScaleCap to give a sane maximum space taken

The following screenshot gives a worst case scenario example, where there is the smallest phone I could test with the maximum size letters possible
image


app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -190,6 +188,17 @@ internal class HedvigAppState(
     }
 }
 
+@JvmInline
+value class NavigationSuiteType private constructor(private val description: String) {
```
panasetskaya, line 192: Oh cool, didn't know about value class :) Have read about them now, handy
One question: why override String here? can't find usages. In other words, why do we need description?

app/navigation/navigation-activity/src/main/kotlin/com/hedvig/android/navigation/activity/ActivityNavigator.kt
```diff
@@ -10,14 +10,26 @@ import com.hedvig.android.core.common.android.tryOpenPlayStore
...
+  fun openAppSettings(context: Context)
+
+  fun navigateToLoggedInScreen(context: Context, clearBackstack: Boolean = true)
+
+  fun tryOpenPlayStore(context: Context)
+
+  fun openWebsite(context: Context, uri: Uri)
+}
+
+class ActivityNavigatorImpl(
```
panasetskaya, line 25: Would it be worth it, now that it's an impl make it internal, make a separate di module in navigation and expose out only the interface?

panasetskaya (approved):

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
StylianosGakis: Yeah you are right this looks a bit odd.
It ended up being like this as I used the code from here https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3-adaptive-navigation-suite/src/commonMain/kotlin/androidx/compose/material3/adaptive/navigationsuite/NavigationSuiteScaffold.kt;l=336-374?q=NavigationSuiteType and how they do it in the latest alphas of material 3 (it's experimental there atm) so I just kept it the same way. I started from their impl and stripped away stuff until we got what we had before too, only in a nice `NavigationSuite` composable.

The reason they do it this way from my understanding is so that they are able to add more cases in the future, without breaking any older callers. Which a sealed interface would not handle. While also not adding any overhead with an extra class, so they just go with a value class, which practically when the app runs this will look like a `String` in the code at runtime.
The description therefore is used exactly for that, so that after the value class is gone, at runtime these three cases will just be seen as something like "NavigationBar", "NavigationRail" and "None".

To be fair we could absolutely not do all this to get whatever minuscule benefits it gives us, but it's fine this way too I think :D

PR #2060, opened 2024-03-19
Feature/new cancellation screen for choosing insurance

panasetskaya:
Add new screen: to choose insurance to terminate from your list of insurances.
Navigation to this screen from Help Center -> QuickActions and from Insurance Details.


app/apollo/apollo-octopus-public/build.gradle.kts
```diff
@@ -36,7 +36,9 @@ apollo { // Octopus client
 
     // https://www.apollographql.com/docs/android/advanced/operation-variables/#make-nullable-variables-non-optional
     generateOptionalOperationVariables.set(false)
-
+    outputDirConnection {
+      connectToKotlinSourceSet("main") // main is by default but setting this explicitly fixed the warning "Duplicate content roots detected.
```
panasetskaya, line 40: Or not! Still have the warning, but that might be cache.

StylianosGakis: It's okay to keep this then since it shouldn't be changing anything in practice

app/apollo/apollo-octopus-public/src/main/graphql/com/hedvig/android/apollo/octopus/graphql/termination/QueryContractsToTerminate.graphql
```diff
@@ -0,0 +1,26 @@
+query ContractsToTerminate {
```
StylianosGakis, line 1: Since this is a new one, as I understand this is only used in that one screen.
Could we move this query inside that module instead of having it accessible by all feature modules in the app?
The naming would also not make sense querying it from inside :feature-home for example.
Putting it in there should fix this issue.

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -252,11 +250,39 @@ internal fun HedvigNavHost(
...
+            }
+            QuickLinkDestination.QuickLinkTermination -> {
+              backStackEntry.navigate(TerminateInsuranceFeatureDestination(null))
+            }
+            QuickLinkDestination.QuickLinkTravelCertificate -> {
+              backStackEntry.navigate(AppDestination.TravelCertificate)
+            }
+          }
+        }
+      },
```
StylianosGakis, line 279: Perfect!

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt
```diff
@@ -12,21 +12,23 @@ import androidx.compose.material3.Surface
...
 @Composable
-fun SelectIndicationCircle(selected: Boolean, modifier: Modifier = Modifier) {
+fun SelectIndicationCircle(selected: Boolean, modifier: Modifier = Modifier, customColor: Color? = null) {
+  val selectedColor = customColor ?: LocalContentColor.current
   Spacer(
     modifier
       .size(24.dp)
       .clip(CircleShape)
       .then(
         if (selected) {
```
StylianosGakis, line 30: We can make use of default parameters here to avoid having the color as a nullable when we can avoid it.
You can notice that the material3 library does this a lot in all places where they accept a color in too.

```suggestion
fun SelectIndicationCircle(selected: Boolean, modifier: Modifier = Modifier, selectedIndicationColor: Color = LocalContentColor.current) {
  Spacer(
    modifier
      .size(24.dp)
      .clip(CircleShape)
      .then(
        if (selected) {
          Modifier.background(selectedIndicationColor)
```
Something like this should work

app/data/data-contract/data-contract-public/src/main/kotlin/com/hedvig/android/data/contract/ContractGroup.kt
```diff
@@ -1,5 +1,8 @@
 package com.hedvig.android.data.contract
 
+import kotlinx.serialization.Serializable
+
+@Serializable
 enum class ContractGroup {
```
StylianosGakis, line 6: I thought that enums are just serializable by default without having to annotate them with the annotation.
Did this break before you added the annotation?
Did it perhaps break when the `alias(libs.plugins.serialization)` serialization gradle plugin was not applied, but then it starts working when you add it, without even having to add the annotation?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
```diff
@@ -90,13 +89,26 @@ internal class GetQuickLinksUseCase(
...
+        ?.let {
+          add(
+            QuickAction.QuickLink(
+              quickLinkDestination = QuickLinkDestination.QuickLinkTermination,
+              titleRes = R.string.HC_QUICK_ACTIONS_CANCELLATION_TITLE,
+              hintTextRes = R.string.HC_QUICK_ACTIONS_CANCELLATION_SUBTITLE,
+              displayName = null,
+            ),
+          )
+        }
```
StylianosGakis, line 103: The null `displayName` here, which part of the UI is that driving?
Also I wonder, besides hiding when the feature flag is off, do we also need to hide this quick action for people who do not have any active insurance right now? Like the same check we have to do in the termination flow itself which checks if they can select any of their insurances to terminate?

If the answer is yes, it may be worth it extracting that piece of logic in some sort of ":data-termination" module which we can use here, and then the termination flow screen can also share that same piece of logic. 

What do you think, am I missing something here?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/model/QuickLink.kt
```diff
@@ -16,6 +16,6 @@ sealed interface QuickAction {
     override val titleRes: Int,
     val displayName: String?,
     override val hintTextRes: Int,
-    val destination: Destination,
+    val quickLinkDestination: QuickLinkDestination,
```
StylianosGakis, line 19: So glad for this 😅

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/GetContractsToTerminateUseCase.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.core.common.ErrorMessage
+import com.hedvig.android.core.uidata.UiMoney
+import com.hedvig.android.data.contract.ContractGroup
+import com.hedvig.android.data.contract.toContractGroup
+import com.hedvig.android.logger.logcat
+import kotlinx.datetime.LocalDate
+import octopus.ContractsToTerminateQuery
+
+internal interface GetContractsToTerminateUseCase {
```
StylianosGakis, line 16: This is potentially the same piece of logic that the help-center may need to reuse in order to know when not to show the termination flow.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/GetContractsToTerminateUseCase.kt
```diff
@@ -0,0 +1,63 @@
...
+  suspend fun invoke(): Either<ErrorMessage, List<InsuranceForCancellation>>
+}
+
+internal class GetContractsToTerminateUseCaseImpl(
+  private val apolloClient: ApolloClient,
+) : GetContractsToTerminateUseCase {
+  override suspend fun invoke(): Either<ErrorMessage, List<InsuranceForCancellation>> {
+    logcat { "mariia: GetContractsToTerminateUseCaseImpl launched" }
```
StylianosGakis, line 24: Remember to remove before merge 😊

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/GetContractsToTerminateUseCase.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.hedvig.android.data.contract.toContractGroup
+import com.hedvig.android.logger.logcat
+import kotlinx.datetime.LocalDate
+import octopus.ContractsToTerminateQuery
+
+internal interface GetContractsToTerminateUseCase {
+  suspend fun invoke(): Either<ErrorMessage, List<InsuranceForCancellation>>
+}
+
+internal class GetContractsToTerminateUseCaseImpl(
```
StylianosGakis, line 20: Do you think the name `GetTerminatableContractsUseCase` would fit this class better?
My thought process is that the current name tells us that we are getting the contracts to terminate. That may imply that there's a decision of which contracts are supposed to be terminated, and not necessarily the contracts that are just *eligible* to be terminated.
Something like `GetContractsEligibleForTerminationUseCase` could work too, albeit while being quite more verbose.
Wdyt?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/GetContractsToTerminateUseCase.kt
```diff
@@ -0,0 +1,63 @@
...
+        .safeExecute()
+        .toEither(::ErrorMessage)
+        .bind()
+        .currentMember
+      member.toInsurancesForCancellation()
+    }
+  }
+}
+
+data class InsuranceForCancellation(
```
StylianosGakis, line 37: Same logic here, this could also be `TerminatableInsurance` perhaps?
I'd also just avoid adding the name `Cancel` instead of `terminate` because then both of those concepts exist in the same place making it confusing if one means something else compared to the other, while both in our context mean termination.

In the same logic as above, we call them `Insurance` here, while we call them `Contract` above. Perhaps sticking with `Contract` in both would lower the mental overhead of reading those, wdyt?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/data/GetContractsToTerminateUseCase.kt
```diff
@@ -0,0 +1,63 @@
...
+  val nextPaymentDate = futureCharge?.date
+  return activeContracts.map {
+    InsuranceForCancellation(
+      id = it.id,
+      displayName = it.currentAgreement.productVariant.displayName,
+      contractGroup = it.currentAgreement.productVariant.typeOfContract.toContractGroup(),
+      contractExposure = it.exposureDisplayName,
+      activateFrom = it.currentAgreement.activeFrom,
+      monthlyPayment = UiMoney.fromMoneyFragment(it.currentAgreement.premium),
+      nextPaymentDate = if (it.id in futureChargeIds) nextPaymentDate else null,
```
StylianosGakis, line 60: Nice, this looks good here.
Let's keep this in our minds to make sure we show it at the right times, because I am not sure what the design has in mind for this. But your logic here looks solid!

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceDestination.kt
```diff
@@ -3,15 +3,13 @@ package com.hedvig.android.feature.terminateinsurance.navigation
...
 
 @Serializable
 data class TerminateInsuranceFeatureDestination(
-  val contractId: String,
-  val insuranceDisplayName: String,
-  val exposureName: String,
-  val contractGroup: ContractGroup,
-  val activeFrom: LocalDate,
+  @SerialName("insuranceId")
+  val contractId: String?,
```
StylianosGakis, line 12: Now I know we use `insuranceId` for the route name, but why do we have this split here? Is there a world where these two could be named the same thing?
If it means that the route will also change to contractId the best time to do it is now before the deep link is out in production and it's not possible to change anymore without breaking old clients.
Otherwise, maybe the parameter should be named insuranceId as well? Not sure if those two imply any different meaning really.

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -49,6 +51,8 @@ internal class HedvigDeepLinkContainerImpl(
   override val contractWithoutContractId: String = "$baseDeepLinkDomain/contract"
   override val contract: String = "$baseDeepLinkDomain/contract?contractId={contractId}"
 
+  override val terminateInsurance: String = "$baseDeepLinkDomain/terminateInsurance?insuranceId={insuranceId}"
```
StylianosGakis, line 54: Here this `insuranceId` also shows that it's different from the `contractId` right 2 lines above and I wonder if these two actually are referring to the same kind of id or not. Which is what makes me wonder if perhaps this should also be `contractId` if it's the case.

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -18,6 +18,8 @@ interface HedvigDeepLinkContainer {
   val contractWithoutContractId: String
   val contract: String // A specific contract destination with a contractId. If none match, an empty screen is shown
 
+  val terminateInsurance: String
```
StylianosGakis, line 21: Try to add a little description here just like the other deep links. Just to make this file easy to read out of context, if you are just looking at the deep links and not where they're used.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    LaunchedEffect(loadIteration) {
+      currentPartialState = PartialUiState.Loading
+      combine(
+        flow { emit(getContractsToTerminateUseCase.invoke()) },
+        featureManager.isFeatureEnabled(Feature.TERMINATION_FLOW),
+      ) { contractsResult, isTerminationFlowEnabled ->
+        contractsResult to isTerminationFlowEnabled
+      }.collect { (contractsResult, isTerminationFlowEnabled) ->
+        logcat { "mariia: contractsResult: $contractsResult isTerminationFlowEnabled: $isTerminationFlowEnabled" }
```
StylianosGakis, line 89: Remember to remove the log. Or if you think the log itself will be useful just remove your name but keep the log and give it a better "title" at the start of it.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+          }
+        }
+      }
+    }
+
+    LaunchedEffect(currentSelectedId) {
+      val id = currentSelectedId
+      if (id != null) {
+        terminationStep = terminateInsuranceRepository.startTerminationFlow(InsuranceId(id)).getOrNull()
+        logcat { "mariia: LaunchedEffect(currentSelectedId) finished, terminationStep: $terminationStep" }
```
StylianosGakis, line 77: Same idea here with this log as I write below here, or any logs with your name really, the same applies, keep if you want the log to stay remove the name and give it a title, otherwise if the log was temporary only to debug stuff then feel free to just delete

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+
+    LaunchedEffect(loadIteration) {
+      currentPartialState = PartialUiState.Loading
```
StylianosGakis, line 82: Since we always seem to do this here, if we've just come *back* to this screen if we went forward, realized that we picked the wrong contract perhaps, and we come back to this screen, the state will go to loading again and then load the data and show the data again. Doing a little "flickering" effect.
Do you think there's room for us here to make it work in a way that the old state still stays on the screen while we still kick off the work of fetching the data in case it's stale for whatever reason?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+          currentPartialState = PartialUiState.NotAllowed
+        } else {
+          contractsResult.fold(
+            ifLeft = {
+              logcat(priority = LogPriority.INFO) { "Cannot load contracts for cancellation" }
+              logcat { "mariia: LaunchedEffect(loadIteration) Cannot load contracts for cancellation" }
+              currentPartialState = PartialUiState.Failure
+            },
+            ifRight = { eligibleInsurances ->
+              logcat { "mariia: LaunchedEffect(loadIteration) Successfully loaded contracts for cancellation" }
```
StylianosGakis, line 101: ^^ log comment

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+      }.collect { (contractsResult, isTerminationFlowEnabled) ->
+        logcat { "mariia: contractsResult: $contractsResult isTerminationFlowEnabled: $isTerminationFlowEnabled" }
+
+        if (!isTerminationFlowEnabled) {
+          currentPartialState = PartialUiState.NotAllowed
+        } else {
+          contractsResult.fold(
+            ifLeft = {
+              logcat(priority = LogPriority.INFO) { "Cannot load contracts for cancellation" }
+              logcat { "mariia: LaunchedEffect(loadIteration) Cannot load contracts for cancellation" }
```
StylianosGakis, line 97: ^^ log comment

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+              currentPartialState = PartialUiState.Failure
+            },
+            ifRight = { eligibleInsurances ->
+              logcat { "mariia: LaunchedEffect(loadIteration) Successfully loaded contracts for cancellation" }
+              logcat(priority = LogPriority.INFO) { "Successfully loaded contracts for cancellation" }
+              val selectedInsurance = if (insuranceId != null) {
+                eligibleInsurances.firstOrNull { it.id == insuranceId }
+              } else {
+                null
+              }
```
StylianosGakis, line 107: 👌

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+              currentPartialState = if (eligibleInsurances.isNotEmpty()) {
+                PartialUiState.Success(
+                  eligibleInsurances,
+                  selectedInsurance,
+                )
+              } else {
+                PartialUiState.NotAllowed
+                // todo: if we somehow get an empty list, I think we should show the same
+                // "you don't have eligible contracts blablabla" copy
```
StylianosGakis, line 116:
```suggestion
                PartialUiState.NotAllowed
```
Yep I agree, this feels like it can be the same fallback of telling them that they got nothing to cancel and to show the copy we have + a button to the chat perhaps.

So with that said since it's more of a scenario where they are not eligible, and not that they are not allowed, could we name the state as `PartialUiState.NotEligible`?

StylianosGakis (commented):
Had a comment for when we show this or not in the quick links. We might need to tweak that too. Might need to double check with other people.

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
panasetskaya: displayName is shown in the select dialog, if we have multiple insurances to edit co-ensured. we don't need it for cancellation.
Agree with the separate :data-termination module, will do that!

StylianosGakis: Sounds like an opportunity to perhaps make this easier to use by having those two cases be distinct classes inside some sort of sealed hierarchy perhaps?
So that one can't misuse this API by accident. Or if they provide a displayName and expect it to show, but it just doesn't for this case?

panasetskaya:
Not to forget: also check what navigation graph pops up to. Maybe use navigate back instead of navigate up.

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
panasetskaya: divided them now:)

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt
```diff
@@ -12,21 +12,26 @@ import androidx.compose.material3.Surface
...
+  selected: Boolean,
+  modifier: Modifier = Modifier,
+  selectedIndicationColor: Color = LocalContentColor.current,
+) {
   Spacer(
     modifier
       .size(24.dp)
       .clip(CircleShape)
       .then(
         if (selected) {
```
hugokallstrom, line 33: Would be nice to animate this color :)

hugokallstrom (approved):
Looks great!

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt
panasetskaya: added in the new PR:) not sure what our animation guidelines are, just used tween with 500 millis

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
```diff
@@ -42,15 +44,15 @@ internal class GetQuickLinksUseCase(
           if (it.size > 1) {
             val links = it.map { contract ->
               if (contract.coInsured?.any { it.hasMissingInfo } == true) {
```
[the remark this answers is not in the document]
panasetskaya: Thanks! added in the new PR:)

panasetskaya:
Copy not ready yet!

StylianosGakis (commented):
Will leave this here as it is for now since I got some actionable stuff already in here.
I left the presenters and the UI for last which I will do in a follow up review now soon, I just needed some more time there to make sure I understand everything right

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/ui/TerminationOverviewScreenScaffold.kt
```diff
@@ -26,144 +25,63 @@ import androidx.compose.runtime.mutableStateOf
...
-  InsuranceCard(
-    chips = buildList {
-      if (terminationDate != null) {
-        add(terminationDateText(terminationDate = terminationDate))
+    Box {
+      Column {
+        val topAppBarScrollBehavior = TopAppBarDefaults.pinnedScrollBehavior()
+        TopAppBarWithBackAndClose(
+          onNavigateUp = navigateUp,
+          onClose = closeTerminationFlow,
```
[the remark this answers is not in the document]
panasetskaya: I think we agreed on not having confirmation:)

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
```diff
@@ -77,79 +76,108 @@ fun NavGraphBuilder.terminateInsuranceGraph(
       navigateBack = navigator::popBackStack,
     )
   }
-  navigation<TerminateInsuranceFeatureDestination>(
+  navigation<AppDestination.TerminationFlow>(
```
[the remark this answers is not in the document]
panasetskaya: I think I did it for closing the flow, the same way the closing the claim flow is done, but I must admit I don't remember the exact reason. Was that one internal? Probably

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/presentation/InsurancePresenter.kt
```diff
@@ -107,7 +107,7 @@ internal class InsurancePresenter(
       loadInsuranceData(
         getInsuranceContractsUseCase = getInsuranceContractsUseCaseProvider.provide(),
         getCrossSellsUseCase = getCrossSellsUseCaseProvider.provide(),
-        forceNetworkFetch = isRetryingIteration,
+        forceNetworkFetch = true,
```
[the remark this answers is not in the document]
panasetskaya: Yes, that was exactly the problem here:)

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt
```diff
@@ -10,28 +17,54 @@ import androidx.compose.material3.LocalContentColor
...
-  )
+fun SelectIndicationCircle(
+  selected: Boolean,
+  modifier: Modifier = Modifier,
+  selectedIndicationColor: Color = LocalContentColor.current,
+) {
+  val selectedTransition = updateTransition(selected)
+  selectedTransition.AnimatedContent(
+    transitionSpec = {
+      fadeIn(tween(durationMillis = 500)) togetherWith fadeOut(tween(90))
```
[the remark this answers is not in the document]
panasetskaya: I can't find any specifications, but I asked Julia how they do it:)

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
```diff
@@ -32,32 +33,30 @@ import com.kiwi.navigationcompose.typed.popUpTo
...
       onSurveyClicked = { openUrl(surveyUrl) },
-      navigateUp = navigator::navigateUp,
+      navigateToInsurances = {
+        navigateToInsurances(backStackEntry)
+        /**
+         We clear everything except for Home destination and navigate to Insurances when we press Done on success screen.
+         If we go back (system back button) we return to screen we we've been before termination flow (Help Center or Insurance Details).
+         If we successfully delete insurance that was supposed to activate in the future and press system back button on Success screen,
+         We'll get a "can't find this insurance" error on Insurance Details screen, which I think is logical.
+         */
```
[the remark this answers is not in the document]
panasetskaya: Kind of not feeling too good towards deep links now, I must admit:) I'll learn to love them

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -138,14 +139,22 @@ internal fun HedvigNavHost(
...
           openChat = { backStackEntry ->
             with(navigator) {
               backStackEntry.navigate(AppDestination.Chat())
             }
           },
           openUrl = openUrl,
           openPlayStore = { activityNavigator.tryOpenPlayStore(context) },
+          hedvigDeepLinkContainer = hedvigDeepLinkContainer,
+          navigateToInsurances = { backStackEntry ->
+            val navOptions = NavOptions.Builder()
```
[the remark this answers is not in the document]
panasetskaya: I think the extra forever destination is for other branch, for untagle-payments:)

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/navigation/TerminateInsuranceGraph.kt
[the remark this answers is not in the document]
StylianosGakis: We can solve this in another PR where we take the effort to remove all AppDestinations, it's fine for now leave it as-is

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/SelectIndicationCircle.kt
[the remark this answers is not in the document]
StylianosGakis: If EaseInOut is used, we might actually want to use something like `MotionTokens.EasingStandardCubicBezier` instead of a `tween` which does it completely linerarly. We don't have an EaseInOut specifically ourselves but maybe we can add one?

Something like this https://cubic-bezier.com/#.6,0,.4,1 perhaps? Maybe William should have an opinion about these things too. We could get some motion tokens too to re-use across the app.

Btw read more here https://medium.com/androiddevelopers/easing-in-to-easing-curves-in-jetpack-compose-d72893eeeb4d about all this if you are not very familiar

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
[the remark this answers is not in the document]
StylianosGakis: Let's make sure we do that before we release then

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
```diff
@@ -0,0 +1,243 @@
...
+          HedvigCard(
+            onClick = { selectInsurance(insurance) },
+            modifier = Modifier
+              .fillMaxWidth()
+              .padding(horizontal = 16.dp),
+          ) {
+            Row(
+              verticalAlignment = Alignment.CenterVertically,
+              modifier = Modifier
+                .heightIn(72.dp)
```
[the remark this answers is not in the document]
StylianosGakis: Yeah I also remember these being 72 before. Let's hope the new design settles on what this should be

panasetskaya:
NB! 
- [ ] check all the copy
- [ ] remove all todos

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
```diff
@@ -0,0 +1,170 @@
...
+        }
+      }
+    }
+
+    LaunchedEffect(currentSelectedId) {
+      val id = currentSelectedId
+      if (id != null) {
+        terminationStep = terminateInsuranceRepository.startTerminationFlow(InsuranceId(id)).getOrNull()
+      }
+    }
```
StylianosGakis, line 78: Aha I see, you fetch the next step here on each change of the currentSelectedId (which is also what comes as a parameter here if we want to have something pre-selected).

Two thoughts here:
* If the user goes back and forth between options, on each one of them we do a request to start the termination flow as a side effect. This comes at the benefit that then when you click the button you already know where to go instead of loading.
* If this fails, so when we do `getOrNull()` it is null, there does not seem to be any way for the user to retry this besides picking another option and coming back to this again.
* This also makes it so that the button to move forward does not show as enabled as soon as you click the button, but this needs to run first.

Do you think we can kinda revert the order in which things are happening here to avoid the above issues?
So when the user picks one of the options (or one is pre-selected even), we just keep that selection locally in the state but without triggering any network request.
Then when the button is pressed (and it should be able to be pressed immediately as long as there is a selection in the first place) then we do the network request, and we show a loading spinner inside the button.
This way if this fails, the user can actually click the button again to try again in some way. We do not have design for what happens if that fails of course right now, but for now it can even be a Toast till we discuss this with design.

In general though this flip of the concerns needs to happen here so that we do not end up in these unrecoverable scenarios if `startTerminationFlow` fails.

That way I think you'll also be able to get rid of the partial state and keep one UiState. Probably, you can figure out if its still not possible yet as you do this

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/ui/TerminationOverviewScreenScaffold.kt
```diff
@@ -26,144 +24,63 @@ import androidx.compose.runtime.mutableStateOf
...
+              WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom,
+            ),
+          ),
       ) {
+        Spacer(modifier = Modifier.height(8.dp))
+        Text(
+          text = stringResource(id = R.string.TERMINATION_FLOW_CANCELLATION_TITLE),
+          fontSize = MaterialTheme.typography.headlineSmall.fontSize,
+          fontStyle = MaterialTheme.typography.headlineSmall.fontStyle,
+          fontFamily = MaterialTheme.typography.headlineSmall.fontFamily,
```
StylianosGakis, line 73:
```suggestion
          style = MaterialTheme.typography.headlineSmall,
```
I **think** this can be replaced just by this, can you give it a try?

StylianosGakis (commented):
I think the biggest change just now needs to happen on the picking screen regarding when the network request happens, give that comment a look

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateViewModel.kt
panasetskaya: Fixed in the last commit:)

PR #2061, opened 2024-03-21
Fix/design through all app

panasetskaya:
small ui fixes from here: https://www.figma.com/file/BbYyhe0vJe9Q8XTVMSdRPr/Design-QA?type=design&node-id=158-6890&mode=design&t=48D3fQsGJ1VeXoPf-0


panasetskaya:
A bit confused as to why it shows such a big conflict in HedvigMaterial3Theme.kt, there are only 5 changed lines there:)

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
```diff
@@ -149,7 +151,9 @@ internal fun lightHedvigColorScheme(hedvigTonalPalette: HedvigTonalPalette) = He
   onTypeElement = hedvigTonalPalette.greyscale1000,
   typeContainer = hedvigTonalPalette.green100,
   onTypeContainer = hedvigTonalPalette.green800,
-  typeHighlight = hedvigTonalPalette.green300,
+  lightTypeContainer = hedvigTonalPalette.green50,
+  onLightTypeContainer = hedvigTonalPalette.greyscale1000,
+  typeHighlight = hedvigTonalPalette.green100,
```
StylianosGakis, line 156: So it only looks like the differences with typeContainer/onTypeContainer are
green100 -> green50
green800 -> greyscale1000

And then there is no difference at all for dark mode.
This looks like it's super close to the existing behavior, where was it that this was needed? Also, the list here https://www.figma.com/file/qUhLjrKl98PAzHov9ilaDH/Hedvig-UI-Kit?type=design&node-id=91%3A1127&mode=design&t=EriUfdYxfXrM0HK7-1 does not seem to show those.

I am just curious if this is a case where we just got 1 place which is a bit different, and I wonder if we really need to add this overhead here for that one place. Maybe that component should follow what the rest of the design system uses as opposed to making this exception?

StylianosGakis: Or vice, versa, if everything else needs to be adjusted to this color. That might work too if the design needs that to happen instead.

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/infocard/InfoCardTextButton.kt
```diff
@@ -19,5 +20,6 @@ fun InfoCardTextButton(onClick: () -> Unit, text: String, modifier: Modifier = M
     ),
     textStyle = MaterialTheme.typography.bodyMedium,
     modifier = modifier,
+    elevation = ButtonDefaults.elevatedButtonElevation(defaultElevation = 1.dp),
```
StylianosGakis, line 23: Same worry here with the elevation, we have ~8 uses I think of this one

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -153,8 +152,7 @@ private fun ContractDetailScreen(
             contentPadding = WindowInsets
               .safeDrawing
               .only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom)
-              .asPaddingValues()
-              .plus(PaddingValues(top = 16.dp)),
+              .asPaddingValues(),
```
StylianosGakis, line 155: Will this touch the top now without this extra padding? Could it be that it needs to be 8.dp instead? We do that in some places I think, where it's 8

StylianosGakis (commented):
Always a bit scared with doing these sorts of changes :D
Wrote some comments especially on the changes that look like they'll affect a lot of places in the app

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/HedvigMaterial3Theme.kt
panasetskaya: It's button alt default here: https://www.figma.com/file/qUhLjrKl98PAzHov9ilaDH/Hedvig-UI-Kit?type=design&node-id=91-1139&mode=design&t=B7600qn0jU7A7Hxx-4 

so I'm guessing it's not one place? Like here: 
https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?type=design&node-id=7014-14100&mode=design&t=fb2WODrzjxlsu4G9-4
https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?type=design&node-id=1996-35666&mode=design&t=fb2WODrzjxlsu4G9-4
https://www.figma.com/file/WlwFCAWywXNcXVgPfYVmWu/Final-Design?type=design&node-id=9885-35843&mode=design&t=fb2WODrzjxlsu4G9-4

Maybe my dark mode is wrong though, I'm not sure here

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/button/HedvigContainedSmallButton.kt
```diff
@@ -38,13 +39,14 @@ fun HedvigContainedSmallButton(
...
   enabled: Boolean = true,
+  shape: Shape = MaterialTheme.shapes.squircleExtraSmall,
   colors: ButtonColors = ButtonDefaults.buttonColors(
     containerColor = MaterialTheme.colorScheme.primary,
     contentColor = MaterialTheme.colorScheme.onPrimary,
     disabledContainerColor = MaterialTheme.colorScheme.primary.copy(alpha = 0.12f),
     disabledContentColor = MaterialTheme.colorScheme.onPrimary.copy(alpha = 0.38f),
   ),
-  elevation: ButtonElevation? = ButtonDefaults.buttonElevation(),
+  elevation: ButtonElevation? = ButtonDefaults.elevatedButtonElevation(defaultElevation = 1.dp),
```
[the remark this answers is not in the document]
panasetskaya: I went through different screens in figma, and this button seems to be always on top of info section, and is always with shadow. I could also go through all the usages in the code to check if there are exceptions there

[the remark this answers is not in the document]
panasetskaya: Did it, see comment below

PR #2062, opened 2024-03-21
Payments screen without refreshing on going back

StylianosGakis:
Fix clearing the data on refresh when the presenter starts presenting
Remove unnecessary events and types that were only used in other screens
We still keep the `contentForOtherScreens` because of the way these
screens are structured for now. But in the future each screen should
fetch their own information instead.
Map the raw PaymentOverviewData into a screen-specific model to make the
code in the UI much simpler to go through rather than checking for all
the deep parameters inside a big object making the UI code harder to
read.

Also fixes the top app bar having a back button which does not fit the top level destination now.

### DIFF
--- a/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/information/HedvigInformationSection.kt
+++ b/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/information/HedvigInformationSection.kt
@@ -12,13 +12,11 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.safeDrawing
 import androidx.compose.material.icons.Icons
 import androidx.compose.material3.Icon
-import androidx.compose.material3.LocalContentColor
 import androidx.compose.material3.LocalTextStyle
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
@@ -69,16 +67,15 @@ fun HedvigInformationSection(
     )
     Spacer(Modifier.height(2.dp))
     if (subTitle != null) {
-      CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.onSurfaceVariant) {
-        Text(
-          text = subTitle,
-          textAlign = TextAlign.Center,
-          modifier = Modifier.fillMaxWidth(),
-        )
-      }
+      Text(
+        text = subTitle,
+        textAlign = TextAlign.Center,
+        color = MaterialTheme.colorScheme.onSurfaceVariant,
+        modifier = Modifier.fillMaxWidth(),
+      )
     }
-    Spacer(Modifier.height(24.dp))
     if (onButtonClick != null) {
+      Spacer(Modifier.height(24.dp))
       HedvigContainedSmallButton(
         text = buttonText,
         onClick = onButtonClick,
--- a/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/scaffold/ClaimFlowScaffold.kt
+++ b/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/scaffold/ClaimFlowScaffold.kt
@@ -5,18 +5,27 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
+import androidx.compose.foundation.layout.displayCutout
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.only
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.safeDrawing
+import androidx.compose.foundation.layout.union
 import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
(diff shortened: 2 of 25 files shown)


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/scaffold/ClaimFlowScaffold.kt
```diff
@@ -103,3 +114,53 @@ fun ClaimFlowScaffold(
     }
   }
 }
+
+@Composable
+private fun TopAppBarWithBackAndClose(
```
StylianosGakis, line 119: Moved in here since it was the only call site. This is not new code.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsViewModel.kt
```diff
@@ -1,16 +1,16 @@
-package com.hedvig.android.feature.payments.overview
+package com.hedvig.android.feature.payments.discounts
 
 import com.hedvig.android.core.demomode.Provider
 import com.hedvig.android.feature.payments.overview.data.AddDiscountUseCase
 import com.hedvig.android.feature.payments.overview.data.GetPaymentOverviewDataUseCase
 import com.hedvig.android.molecule.android.MoleculeViewModel
 
-internal class PaymentOverviewViewModel(
+internal class DiscountsViewModel(
```
StylianosGakis, line 8: The payments and the discounts screens were sharing the exact same ViewModel.
Here I just copy-pasted it to be used in the Discounts screen as-is, and later in the PR I made a new VM just for payments, since it was not using half of the stuff from in the VM and was making it hard for the UI to use this information, a lot of logic was delegated to the UI before.

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/information/HedvigInformationSection.kt
```diff
@@ -69,16 +67,15 @@ fun HedvigInformationSection(
     )
     Spacer(Modifier.height(2.dp))
     if (subTitle != null) {
-      CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.onSurfaceVariant) {
```
panasetskaya, line 72: That was something I wanted to ask you about. Is it womething that we don't want it to be used everywhere, only on top of the Composable tree?

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/PaymentConnection.kt
```diff
@@ -3,23 +3,19 @@ package com.hedvig.android.feature.payments.data
 import kotlinx.serialization.Serializable
 
 @Serializable
-internal data class PaymentConnection(
-  val connectionInfo: ConnectionInfo?,
-  val status: PaymentConnectionStatus,
-) {
+internal sealed interface PaymentConnection {
```
panasetskaya, line 6: Cool

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/payments/PaymentsViewModel.kt
```diff
@@ -0,0 +1,18 @@
...
+import kotlinx.coroutines.flow.WhileSubscribed
+
+internal class PaymentsViewModel(
+  getPaymentOverviewDataUseCase: Provider<GetPaymentOverviewDataUseCase>,
+) : MoleculeViewModel<PaymentsEvent, PaymentsUiState>(
+    PaymentsUiState.Loading,
+    PaymentsPresenter(
+      getPaymentOverviewDataUseCase = getPaymentOverviewDataUseCase,
+    ),
+    SharingStarted.WhileSubscribed(0.seconds),
```
panasetskaya, line 17: Could you explain a bit this 0.seconds here?

panasetskaya (approved):

app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/information/HedvigInformationSection.kt
StylianosGakis: Hmm, it's a bit of a nuanced discussion. And how interesting timing where I just made this PR as well https://github.com/HedvigInsurance/android/pull/2063
The short version is that when we can use things explicitly it's just better to do it this way instead of using a composition local. Especially here where the usage is literally just a couple of lines further down, it was a mistake from me to do it this way.
It only makes sense to provide such composition locals in scenarios like this one https://github.com/HedvigInsurance/android/blob/71f939aaf99c645e5a7c67ae980b6aca0717eaf1/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/textfield/HedvigTextFieldImpl.kt#L235-L241 where we do not actually know what the composable content has inside of it, and we can not pass the color in there instead. So there we pass it as a composition local so that the content can use it if they want. But that still is just that we provide it, the content itself may just decide not to use it.
More on locals here https://developer.android.com/jetpack/compose/compositionlocal
More on why they are often not the right choice here 

Some more thoughts on this from one of the compose engineers in Google:
https://slack-chats.kotlinlang.org/t/13866799/ambients-were-renamed-to-compositionlocal-keep-in-mind-that-#16dc0903-b0c6-437d-be88-9a1e060c4750

They got their place, the App's theme being one of them. But overall it's hard to justify using in normal code.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/payments/PaymentsViewModel.kt
StylianosGakis: I can definitely explain it by saying that this was a mistake 😂
Thanks for noticing!

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/PaymentConnection.kt
StylianosGakis: Indeed, I was pulling my hair (the little I got) when I couldn't understand how this model had null content on an active payment connection 😅

PR #2065, opened 2024-03-28
Feature/new cancellation flow design

panasetskaya:
design updates for the cancellation flow (without survey)


app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/scaffold/ClaimFlowScaffold.kt
```diff
@@ -116,7 +116,7 @@ fun ClaimFlowScaffold(
 }
 
 @Composable
-private fun TopAppBarWithBackAndClose(
+fun TopAppBarWithBackAndClose(
```
hugokallstrom, line 119: Can we move this to the appbar package now that its public?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/terminationdate/TerminationDateDestination.kt
```diff
@@ -49,100 +50,178 @@ import kotlinx.datetime.toLocalDateTime
...
+          style = MaterialTheme.typography.bodyLarge,
+        )
+        Text(
+          text = stringResource(id = R.string.TERMINATION_FLOW_IMPORTANT_INFORMATION_TEXT),
+          style = MaterialTheme.typography.bodyMedium,
+          color = MaterialTheme.colorScheme.onSurfaceVariant,
+        )
+        Spacer(Modifier.height(16.dp))
+        HedvigContainedButton(
+          onClick = {},
```
hugokallstrom, line 151: I would actually like to have `onCheckedChange()` here as well. Imo it makes sense that the button controls the checkbox, since the checkbox is "inside" the button.

hugokallstrom (approved):
LGTM :)

PR #2066, opened 2024-04-02
add visual transformation to ssn

panasetskaya:
Add visual transformation to ssn with a mask and check explicitly for ssn.length of 12

### DIFF
--- a/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
+++ b/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
@@ -254,7 +254,6 @@ private class PersonalNumberVisualTransformation(
       withStyle(SpanStyle(color = maskColor)) {
         append(mask.takeLast(mask.length - length))
       }
-      toAnnotatedString()
     }
 
     val personalNumberOffsetTranslator = object : OffsetMapping {
@@ -267,7 +266,10 @@ private class PersonalNumberVisualTransformation(
       }
 
       override fun transformedToOriginal(offset: Int): Int {
-        return text.length
+        return when {
+          offset <= 8 -> offset
+          else -> offset - 1
+        }.coerceAtMost(text.length)
       }
     }
     return TransformedText(annotatedString, personalNumberOffsetTranslator)
--- a/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
+++ b/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
@@ -28,9 +28,13 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.input.ImeAction
 import androidx.compose.ui.text.input.KeyboardCapitalization
 import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.text.input.OffsetMapping
+import androidx.compose.ui.text.input.TransformedText
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import com.hedvig.android.core.designsystem.component.button.HedvigContainedButton
@@ -186,7 +190,7 @@ private fun FetchFromSsnFields(
   onContinue: () -> Unit,
 ) {
   var ssnInput by remember { mutableStateOf("") }
-
+  val mask = stringResource(id = R.string.edit_coinsured_ssn_placeholder)
   Column {
     HedvigTextField(
       value = ssnInput,
@@ -198,6 +202,9 @@ private fun FetchFromSsnFields(
         ssnInput = it
       },
       errorText = errorMessage,
+      visualTransformation = visualTransformation@{ annotatedString: AnnotatedString ->
+        personNummerInputFormatter(annotatedString, mask)
+      },
       keyboardOptions = KeyboardOptions(
         keyboardType = KeyboardType.Number,
         imeAction = ImeAction.Done,
@@ -224,6 +231,37 @@ private fun FetchFromSsnFields(
       )
     }
   }
+}
+
+private fun personNummerInputFormatter(text: AnnotatedString, mask: String): TransformedText {
+  val trimmed = if (text.text.length >= 12) text.text.substring(0..11) else text.text
+
+  val annotatedString = AnnotatedString.Builder().run {
+    for (i in trimmed.indices) {
+      append(trimmed[i])
+      if (i == 7) {
+        append("-")
+      }
+    }
+    pushStyle(SpanStyle(color = Color.LightGray))
+    append(mask.takeLast(mask.length - length))
+    toAnnotatedString()
+  }
+
+  val personNummerOffsetTranslator = object : OffsetMapping {
+    override fun originalToTransformed(offset: Int): Int {
+      return when {
+        offset < 8 -> offset
+        offset <= 12 -> offset + 1
+        else -> 13
+      }
+    }
+
+    override fun transformedToOriginal(offset: Int): Int {
+      return text.length
+    }
+  }
+  return TransformedText(annotatedString, personNummerOffsetTranslator)
 }
 
 @Composable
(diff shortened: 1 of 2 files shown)


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -198,6 +202,9 @@ private fun FetchFromSsnFields(
         ssnInput = it
       },
       errorText = errorMessage,
+      visualTransformation = visualTransformation@{ annotatedString: AnnotatedString ->
```
StylianosGakis, line 205: Can just remove the `visualTranformation@` part since we do not use it here to do a `return@visualTransformation` here somewhere

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +233,37 @@ private fun FetchFromSsnFields(
   }
 }
 
+private fun personNummerInputFormatter(text: AnnotatedString, mask: String): TransformedText {
```
StylianosGakis, line 236: I think this can just be a private class here instead right? So it's just created once. Can also change the name to just be the English version. It'll help with if we want to search for it in the future.
`private class PersonalNumberInputFormatter`

StylianosGakis: Actually, can this be a class extending from `VisualTransformation` instead? And just pass an instance of that class at the call site above, without even a lambda? The name can also change to have `VisualTransformation` as a suffix instead

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +233,37 @@ private fun FetchFromSsnFields(
...
+  val trimmed = if (text.text.length >= 12) text.text.substring(0..11) else text.text
+
+  val annotatedString = AnnotatedString.Builder().run {
```
StylianosGakis, line 239: I think you can use the `val annotatedString = buildAnnotatedString { ...` builder here instead. Just a bit nicer syntax of the same thing.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +233,37 @@ private fun FetchFromSsnFields(
...
+      return when {
+        offset < 8 -> offset
+        offset <= 12 -> offset + 1
+        else -> 13
+      }
+    }
+
+    override fun transformedToOriginal(offset: Int): Int {
+      return text.length
+    }
```
StylianosGakis, line 262: You may need to do the inverse logic here as well. I think otherwise the thing that looks odd is the cursor is on the wrong index, off by one. Can you test that to see how it is?

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +233,37 @@ private fun FetchFromSsnFields(
...
+  val trimmed = if (text.text.length >= 12) text.text.substring(0..11) else text.text
+
+  val annotatedString = AnnotatedString.Builder().run {
+    for (i in trimmed.indices) {
+      append(trimmed[i])
+      if (i == 7) {
+        append("-")
+      }
+    }
+    pushStyle(SpanStyle(color = Color.LightGray))
```
StylianosGakis, line 246: Is there some `withStyle` dsl function you can use here to denote that only that one part wants this color?

Also, we probably want to use SurfaceVariant from our color scheme here to make sure it both looks the same as other parts of the app, but also dark and light mode look ok too.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +233,37 @@ private fun FetchFromSsnFields(
   }
 }
 
+private fun personNummerInputFormatter(text: AnnotatedString, mask: String): TransformedText {
+  val trimmed = if (text.text.length >= 12) text.text.substring(0..11) else text.text
```
StylianosGakis, line 237: One thing I worry about here. If someone does in fact enter more characters. Will they be added to the text field but simply not show in the UI? Like if they add 12, then press 10 more characters, will all those still be in the state? Will they then have to also delete 10 times to get back to the original state?
If yes, this is something to solve in the onChange lambda above, not in the visual transformation. But we should make it impossible to add more than 12 anyway. Unless it already is this way of course and I missed it here.

StylianosGakis (approved):
Love it, the rest of the characters always being there so they know what to enter is a huge usability win here. We should've always had it this way tbh.

Had some comments, lmk if any don't make sense

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
panasetskaya: Yep, didn't check the state, thanks! Fixed it now.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
```diff
@@ -226,6 +237,43 @@ private fun FetchFromSsnFields(
...
+      for (i in trimmed.indices) {
+        append(trimmed[i])
+        if (i == 7) {
+          append("-")
+        }
+      }
+      withStyle(SpanStyle(color = maskColor)) {
+        append(mask.takeLast(mask.length - length))
+      }
+      toAnnotatedString()
```
StylianosGakis, line 257: You can remove this here too, since internally the function calls it automatically
```kotlin
inline fun buildAnnotatedString(builder: (Builder).() -> Unit): AnnotatedString =
    Builder().apply(builder).toAnnotatedString()
```
It was part of why I wanted to use this dsl too, since it does this part for you and does not leave you room to make a mistake of not calling it otherwise

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
StylianosGakis: I should've written inheriting, what you did is exactly what I had in mind. It's my deep engrained Java brain here I think, where we used `extends` instead of `inherits` when inheriting from abstract classes 😅

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
StylianosGakis: If you try to highlight the text (by long pressing it, to copy it for example) and then you try to move the little handles around in order to select parts of the text, do those handles work as expected as well? That's what breaks usually I think if that mapping is not correct both back and forth.

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
StylianosGakis: Yeah, my point was mostly that if we then add another thing below it would also get this style, but it may not be what we actually want to happen. With the `withStyle` and the block as you did it now it is more clear which parts need to have that style, so adding something else would be a clearer decision of it it goes inside or outside of that lambda

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/AddCoInsuredBottomSheetContent.kt
panasetskaya: Got it:) I thought about extension function and couldn't figure it out:))

PR #2067, opened 2024-04-03
Feature/on back press fix

StylianosGakis:
Fallback to system back if popBackStack does nothing

Enhance the behavior inside the `Navigator` which all screens are using
to pop the backstack when a button which does not do `navigateUp()` but
still wants to go `back` one screen is pressed.
This fixes the problem as described here
https://issuetracker.google.com/issues/329046732 which makes it so that
if you end up in a destination without anything else on your backstack,
which typically happens when you are in the app from a deep link. Now
instead of that button doing nothing, and even breaking future
`navigateUp()` calls, it will fall back to whatever behavior the
system's back button has, which in the deep link scenarios it is to go
back to the app which they came from.

---

Fix the terminate insurance terminal destination done button

This fixes the bug as shown here
https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1712150517177999
where trying to press the "OK" button would not actually remove the
result destination from the backstack in all scenarios. Specifically in
deep link scenarios.


StylianosGakis:
Did this PR myself because this is a very complicated interaction involving a bunch of things and what I think is a bug in the nav library.
https://issuetracker.google.com/issues/329046732

I honestly need to bump them on that issue or create a new one which is more clear now that I understand the issue better compared to when I opened this issue back then.

StylianosGakis:
Also this kinda nerdsniped me and I got carried away with trying to understand and fix this as opposed to finishing the PR review. Working on that now 😅

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/terminationsuccess/TerminationSuccessDestination.kt
```diff
@@ -35,7 +35,7 @@ import kotlinx.datetime.toJavaLocalDate
 internal fun TerminationSuccessDestination(
   terminationDate: LocalDate?,
   onSurveyClicked: () -> Unit,
-  navigateToInsurances: () -> Unit,
```
panasetskaya, line 38: Cool! my only concern here is that we go back, but not to the insurances, I mentioned it here: https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1712163628112979?thread_ts=1712150517.177999&cid=C03HT2JRDPG

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -377,7 +374,8 @@ private fun NavGraphBuilder.nestedHomeGraphs(
 }
 
 @Composable
-private fun rememberNavigator(navController: NavController): Navigator {
+private fun rememberNavigator(navController: NavController, onActivityBackPressed: () -> Unit): Navigator {
+  val updatedOnActivityBackPressed by rememberUpdatedState(onActivityBackPressed)
```
panasetskaya, line 378: I don't entirely understand why it's wrapped in rememberUpdatedState here:)

StylianosGakis: I am gonna go on a trip down the memory lane and share with you this 2 year old thread https://slack-chats.kotlinlang.org/t/507148/i-have-a-question-regarding-compose-and-some-differences-bet#7ab510bc-7b86-4ce7-a8a1-75ab75c30a27
It has some extra stuff that are not exactly that, but the comment from Alex specifically is the one that explains what can go wrong if the lambda is not used this way.

So there is two right ways to write this function, and one wrong. One correct way is the one in the code right now, which is in a way "the more correct" way to do it.

The other correct way would be this
```
private fun rememberNavigator(navController: NavController, onActivityBackPressed: () -> Unit): Navigator {
  return remember(navController, onActivityBackPressed) {
    object : Navigator {
      override fun NavBackStackEntry.navigate(
        destination: Destination,
        navOptions: NavOptions?,
        navigatorExtras: androidx.navigation.Navigator.Extras?,
      ) {
        if (lifecycle.currentState == Lifecycle.State.RESUMED) {
          navigateUnsafe(destination, navOptions, navigatorExtras)
        }
      }

      override fun navigateUnsafe(
        destination: Destination,
        navOptions: NavOptions?,
        navigatorExtras: androidx.navigation.Navigator.Extras?,
      ) {
        navController.navigate(destination, navOptions, navigatorExtras)
      }

      override fun navigateUp() {
        navController.navigateUp()
      }

      override fun popBackStack() {
        if (!navController.popBackStack()) {
          onActivityBackPressed()
        }
      }
    }
  }
}
```

And the wrong way would be like this
```
private fun rememberNavigator(navController: NavController, onActivityBackPressed: () -> Unit): Navigator {
  return remember(navController) {
    object : Navigator {
      override fun NavBackStackEntry.navigate(
        destination: Destination,
        navOptions: NavOptions?,
        navigatorExtras: androidx.navigation.Navigator.Extras?,
      ) {
        if (lifecycle.currentState == Lifecycle.State.RESUMED) {
          navigateUnsafe(destination, navOptions, navigatorExtras)
        }
      }

      override fun navigateUnsafe(
        destination: Destination,
        navOptions: NavOptions?,
        navigatorExtras: androidx.navigation.Navigator.Extras?,
      ) {
        navController.navigate(destination, navOptions, navigatorExtras)
      }

      override fun navigateUp() {
        navController.navigateUp()
      }

      override fun popBackStack() {
        if (!navController.popBackStack()) {
          onActivityBackPressed()
        }
      }
    }
  }
}
```

The difference is this. If this function takes in a new instance of the lambda:

With existing solution: The latest lambda is then stored inside the `updatedOnActivityBackPressed`. So when the function in the interface is called, it will go inside `updatedOnActivityBackPressed`, get that latest version of the lambda and call it.

With the second correct solution: When a new lambda comes, the remember is re-triggered since one of the keys has changed. Then the entire `Navigator` object is destroyed and a new one is constructed, with the new instance referencing the right lambda. Then this new `Navigator` object is returned from this function, and it will trigger a recomposition basically everywhere where it is used in the hierarchy, which is more or less the entire app.
This will give the correct behavior, albeit with some extra recompositions that are not necessary which solution number 1 above solves.

With the wrong solution: The new lambda will come to this function, but since the `remember` does not have it as a key it will not be re-run, and the `Navigator` object which was created will just stay as it is. Even if you then pass in a completely new lambda that does something very different, the `Navigator` will keep calling the first instance it ever saw. Or rather, whatever the latest instance of that lambda was at the moment the last `NavController` was passed in. Since the remember **will** be re-run when that changes since it's a key in the remember.

panasetskaya: That's great, thanks! I think that the second right option is actually more readable - but that's probably just the question of getting used to it. And of course that one gives the extra recompositions, so. I'll read that link more thoroughly today, maybe will come up with more things to ask:)

PR #2068, opened 2024-04-04
delete unused infosections

panasetskaya:
the infosections in the travelCertificateSpecifications are deprecated now


StylianosGakis (approved):
Was this not used at all? We were grabbing it but no longer using it somewhere then?

panasetskaya:
> Was this not used at all? We were grabbing it but no longer using it somewhere then?

Yep, it was used in the old travel certificates. It was not clear back if we would potentially use it, so we left it there for the time being. Seems that we don't need after all, iOs neither:)

PR #2072, opened 2024-04-08
Fix/payments untangle

panasetskaya:
Divide one big payment viewmodel into 3.

The part that is not working well - navigation to Forever from payments-discount (no going back, we get to Home tab on backPressed). Will add that in a different PR via putting Forever ui into shared-forever module that is called from 2 destinations at two different bottom tabs.


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/di/PaymentsModule.kt
```diff
@@ -38,41 +57,64 @@ val paymentsModule = module {
...
+    )
+  }
+  single<GetUpcomingPaymentUseCaseDemo> {
+    GetUpcomingPaymentUseCaseDemo(
+      clock = get<Clock>(),
     )
   }
 }
+
+// todo: should I do this Provider thing for the new usecases?
```
StylianosGakis, line 120: So the provider thing is for demo mode. What you can do is open the app in demo mode and see what happens. If it's an error case it's a good to have to just return at least some sample data which might just render a minimum amount of data to the UI.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/payments/PaymentsDestination.kt
```diff
@@ -526,13 +516,13 @@ private class PaymentsStatePreviewProvider : CollectionPreviewParameterProvider<
         upcomingPayment = PaymentsUiState.Content.UpcomingPayment(
           UiMoney(100.0, CurrencyCode.SEK),
           Clock.System.now().toLocalDateTime(TimeZone.UTC).date,
+          "w345423t6",
```
StylianosGakis, line 519: I love how these get more weird as you scroll down 😂

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/history/PaymentHistoryDestination.kt
```diff
@@ -35,47 +38,82 @@ import kotlinx.datetime.LocalDate
...
+                  )
+                },
+              )
+            },
+            showInfoAboutOlderCharges = uiState.paymentHistory.size > 11,
+          ) to { chargeId: String ->
+            updatedOnChargeClicked(uiState.paymentHistory.first { it.id == chargeId }.id)
+            // todo: I must say I'm a bit confused here :) but if understand correctly,
+            // we don't really need to update it anymore now that we have String as the parameter,
+            // we can just pass down onChargeClicked?
```
StylianosGakis, line 102: we need to do `rememberUpdatedState` because it's inside a `remember` which does not itself include the lambda in its keys.
I think just leave it as it is, it should be fine :D

app/feature/feature-payments/src/main/graphql/QueryShortPaymentsHistory.graphql
```diff
@@ -0,0 +1,13 @@
+
+query ShortPaymentHistory {
```
StylianosGakis, line 2:
```suggestion
query ShortPaymentHistory {
```

StylianosGakis (approved):
You are right this is a lot of stuff moved around and the git diff is not doing me any favors 😅
As I looked around it should all be fine. Some testing on the feature itself before release should be fine 😊

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/di/PaymentsModule.kt
panasetskaya: So as it is now, we can go to Payments page in the demo case, but if we try to go deeper to the history it should show an error, right?
The demo is for Google Play only, right? I forgot what was the trick to launch it:)

StylianosGakis: Yeah that’s exactly it. Someone could theoretically end up there too, which is totally fine, so we don’t need it to be 100% pristine.
The way to enter it is to go in the SE login page, it will automatically open bankID, just go back from there, don’t login.
Then retry the login, and in the QR code which will show up long press it. Then press Yes in the dialog that shows up

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/history/PaymentHistoryDestination.kt
StylianosGakis: Yeah in terms of correctness they should be giving the same end result

The one slight difference is that moving it as a key to the remember it will then trigger running the entire code inside the remember block, and it will produce the new values, and then potentially re-compose an extra time if those values are not stable etc etc. Nothing too important to be honest in this case here.

PR #2074, opened 2024-04-09
help center shimmering effect and GetMemberActionUseCase

panasetskaya:
Add shimmer to loading quickLinks in Help Center. Add data logic for fetching new memberActions (but don't use it anywhere yet, will have it in a separate PR to make it smaller).

### DIFF
--- a/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/HelpCenterPresenter.kt
+++ b/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/HelpCenterPresenter.kt
@@ -18,6 +18,9 @@ import com.hedvig.android.molecule.public.MoleculePresenter
 import com.hedvig.android.molecule.public.MoleculePresenterScope
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.persistentListOf
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.flow
 
 internal sealed interface HelpCenterEvent {
   data class OnQuickActionSelected(val quickAction: QuickAction) : HelpCenterEvent
@@ -29,6 +32,7 @@ internal data class HelpCenterUiState(
   val topics: ImmutableList<Topic>,
   val questions: ImmutableList<Question>,
   val quickLinks: ImmutableList<QuickLinkType>,
+  val isLoadingQuickLinks: Boolean,
   val selectedQuickAction: QuickAction?,
 ) {
   sealed interface QuickLinkType {
@@ -55,6 +59,7 @@ internal class HelpCenterPresenter(
       )
     }
     var selectedQuickAction by remember { mutableStateOf<QuickAction?>(null) }
+    var isLoadingQuickLinks by remember { mutableStateOf(lastState.quickLinks.isEmpty()) }
 
     CollectEvents { event ->
       selectedQuickAction = when (event) {
@@ -64,15 +69,21 @@ internal class HelpCenterPresenter(
     }
 
     LaunchedEffect(Unit) {
-      getQuickLinksUseCase.invoke().fold(
-        ifLeft = { quickActions = persistentListOf() },
-        ifRight = { quickActions = it },
-      )
-
-      getCommonClaimsUseCase.invoke().fold(
-        ifLeft = { commonClaims = persistentListOf() },
-        ifRight = { commonClaims = it },
-      )
+      isLoadingQuickLinks = true
+      combine(
+        flow { emit(getQuickLinksUseCase.invoke()) },
+        flow { emit(getCommonClaimsUseCase.invoke()) },
+      ) { quickLinksResult, commonClaimsResult ->
+        quickLinksResult.fold(
+          ifLeft = { quickActions = persistentListOf() },
+          ifRight = { quickActions = it },
+        )
+        commonClaimsResult.fold(
+          ifLeft = { commonClaims = persistentListOf() },
+          ifRight = { commonClaims = it },
+        )
+        isLoadingQuickLinks = false
+      }.collectLatest {}
     }
 
     return HelpCenterUiState(
@@ -83,6 +94,7 @@ internal class HelpCenterPresenter(
         *commonClaims.map { HelpCenterUiState.QuickLinkType.CommonClaimType(it) }.toTypedArray(),
       ),
       selectedQuickAction = selectedQuickAction,
+      isLoadingQuickLinks = isLoadingQuickLinks,
     )
   }
 }
--- a/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
+++ b/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
@@ -1,0 +1,106 @@
+package com.hedvig.android.feature.help.center.data
+
+import arrow.core.Either
+import arrow.core.raise.either
+import com.apollographql.apollo3.ApolloClient
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.apollo.toEither
+import com.hedvig.android.core.common.ErrorMessage
+import com.hedvig.android.featureflags.FeatureManager
+import com.hedvig.android.featureflags.flags.Feature
+import com.hedvig.android.logger.logcat
+import kotlinx.coroutines.flow.first
+import kotlinx.serialization.Serializable
+import octopus.MemberActionsQuery
+
+internal interface GetMemberActionsUseCase {
+  suspend fun invoke(): Either<ErrorMessage, MemberAction>
+}
+
+internal class GetMemberActionsUseCaseImpl(
+  private val apolloClient: ApolloClient,
+  private val featureManager: FeatureManager,
+) : GetMemberActionsUseCase {
+  override suspend fun invoke(): Either<ErrorMessage, MemberAction> {
+    return either {
+      val result = apolloClient.query(MemberActionsQuery())
+        .safeExecute().toEither(::ErrorMessage).onLeft {
+          logcat { "Cannot load memberActions" }
+        }.bind()
+      val isCoInsuredFeatureOn = featureManager.isFeatureEnabled(Feature.EDIT_COINSURED).first()
+      val isMovingFeatureOn = featureManager.isFeatureEnabled(Feature.MOVING_FLOW).first()
+      val isConnectPaymentFeatureOn = featureManager.isFeatureEnabled(Feature.PAYMENT_SCREEN).first()
+      MemberAction(
+        isCancelInsuranceEnabled = result.currentMember.memberActions?.isCancelInsuranceEnabled ?: false,
+        isConnectPaymentEnabled = isConnectPaymentFeatureOn && result.currentMember.memberActions?.isConnectPaymentEnabled ?: false,
+        isEditCoInsuredEnabled = isCoInsuredFeatureOn && result.currentMember.memberActions?.isEditCoInsuredEnabled ?: false,
+        isMovingEnabled = isMovingFeatureOn && result.currentMember.memberActions?.isMovingEnabled ?: false,
+        isTravelCertificateEnabled = result.currentMember.memberActions?.isTravelCertificateEnabled ?: false,
+        sickAbroadAction = result.currentMember.memberActions?.sickAbroadAction.toSickAbroadAction(),
+        firstVetAction = result.currentMember.memberActions?.firstVetAction?.toVetAction(),
+      )
+    }
+  }
+}
+
+internal data class MemberAction(
+  val isCancelInsuranceEnabled: Boolean?,
+  val isConnectPaymentEnabled: Boolean?,
+  val isEditCoInsuredEnabled: Boolean?,
+  val isMovingEnabled: Boolean?,
+  val isTravelCertificateEnabled: Boolean?,
+  val sickAbroadAction: MemberActionWithDetails.SickAbroadAction?,
+  val firstVetAction: MemberActionWithDetails.FirstVetAction?,
+)
+
+internal sealed interface MemberActionWithDetails {
+  data class SickAbroadAction(
+    val partners: List<DeflectPartner>?,
+  ) : MemberActionWithDetails
+
+  data class FirstVetAction(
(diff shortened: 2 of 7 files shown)


app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -166,32 +167,17 @@ private fun HelpCenterHomeScreen(
...
+        HelpCenterSection(
+          title = stringResource(R.string.HC_QUICK_ACTIONS_TITLE),
+          chipContainerColor = MaterialTheme.colorScheme.typeContainer,
+          contentColor = MaterialTheme.colorScheme.onTypeContainer,
+          content = {
+            Column(
+              verticalArrangement = Arrangement.spacedBy(4.dp),
+            ) {
+              if (isLoadingQuickLinks) {
+                List(5) {
```
StylianosGakis, line 179: Also do you think we can extract everything from this `List(5) {` to the end of it to a private `QuickActionPlaceholders` composable just to keep things a bit easier to read here

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -201,40 +187,83 @@ private fun HelpCenterHomeScreen(
...
                       )
                       Spacer(modifier = Modifier.height(12.dp))
                     }
                   }
                 }
               }
-            },
-          )
-        }
+              for (quickLink in quickLinks) {
```
StylianosGakis, line 210: Since now we have this scenario where technically quickLinks and placeholders can show at the same time, since the state is possible to be in that combination, the change to a sealed interface will help with that too

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -166,32 +167,17 @@ private fun HelpCenterHomeScreen(
...
-            content = {
-              Column(
-                verticalArrangement = Arrangement.spacedBy(4.dp),
-              ) {
-                for (quickLink in quickLinks) {
+        HelpCenterSection(
+          title = stringResource(R.string.HC_QUICK_ACTIONS_TITLE),
+          chipContainerColor = MaterialTheme.colorScheme.typeContainer,
+          contentColor = MaterialTheme.colorScheme.onTypeContainer,
+          content = {
```
StylianosGakis, line 174: Now with the content here not being inside any sort of animation container, if we have those 5 entries shimmering here, and the real response comes back and we then only have 1 item this change will happen as well in a jumpy manner instead of smoothly animating to there.
Could we look into (with the sealed interface as discussed in the other parts of this PR) wrapping that `when` with `AnimatedVisibility` again perhaps? And then we can figure out the best way to animate between those two states. Or even the "no quick actions" state.

StylianosGakis (commented):
This far these are for the first commit about the shimmers, some things we can improve

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+
+internal interface GetMemberActionsUseCase {
+  suspend fun invoke(): Either<ErrorMessage, MemberAction>
+}
+
+internal class GetMemberActionsUseCaseImpl(
+  private val apolloClient: ApolloClient,
+  private val featureManager: FeatureManager,
+) : GetMemberActionsUseCase {
+  override suspend fun invoke(): Either<ErrorMessage, MemberAction> {
```
StylianosGakis, line 24: For this, can you try to use parZip here from arrow.fx?
Basically we want to find all those 4 things, but it would be nice to kick all of those off in parallel, so that we do not ask for one, wait for it to return, ask the next one, wait for that too, and so on.

I just did this locally on my machine to see how it would compile. And I created a git patch out of this which I have sent to you in Slack, check it out there.
Try to copy that and in your IDE find the "Apply patch..." action to apply it. Did this since it requires you to add a dependency too so I couldn't just give an inline suggestion here.

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+
+@Serializable
+data class FirstVetSection(
+  val buttonTitle: String?,
+  val description: String?,
+  val title: String?,
+  val url: String?,
+)
+
+private fun MemberActionsQuery.Data.CurrentMember.MemberActions.FirstVetAction.toVetAction(): MemberActionWithDetails.FirstVetAction {
```
StylianosGakis, line 82: ktlint will probably complain about the long line here right?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+
+@Serializable
+data class DeflectPartner(
+  val id: String,
+  val imageUrl: String?,
+  val phoneNumber: String?,
+  val url: String?,
+)
+
+@Serializable
```
StylianosGakis, line 74: Do these two need to be serializable for some reason? I don't see any such usages of it

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+    val partners: List<DeflectPartner>?,
+  ) : MemberActionWithDetails
+
+  data class FirstVetAction(
+    val sections: List<FirstVetSection>,
+  ) : MemberActionWithDetails
+}
+
+@Serializable
+data class DeflectPartner(
```
StylianosGakis, line 67: `internal` here perhaps?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+  val id: String,
+  val imageUrl: String?,
+  val phoneNumber: String?,
+  val url: String?,
+)
+
+@Serializable
+data class FirstVetSection(
```
StylianosGakis, line 75: `internal` here perhaps?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/di/HelpCenterModule.kt
```diff
@@ -1,13 +1,19 @@
...
+import com.hedvig.android.feature.help.center.data.GetMemberActionsUseCase
+import com.hedvig.android.feature.help.center.data.GetMemberActionsUseCaseImpl
 import com.hedvig.android.feature.help.center.data.GetQuickLinksUseCase
+import com.hedvig.android.featureflags.FeatureManager
 import org.koin.androidx.viewmodel.dsl.viewModel
 import org.koin.dsl.module
 
 val helpCenterModule = module {
+
+  // todo remove from here
```
StylianosGakis, line 16: Perfect, this will then mean the HelpCenter presenter can just use this new use case instead right? Will there be a follow up commit for that in this PR?

StylianosGakis (commented):
Some more comments on the MemberActions commit

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
panasetskaya: I'll figure it out later, but for now I left them serializable bc they will maybe be used for further navigation

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
```diff
@@ -0,0 +1,106 @@
...
+
+internal sealed interface MemberActionWithDetails {
+  data class SickAbroadAction(
+    val partners: List<DeflectPartner>?,
+  ) : MemberActionWithDetails
+
+  data class FirstVetAction(
+    val sections: List<FirstVetSection>,
+  ) : MemberActionWithDetails
+}
```
StylianosGakis, line 64: What does it mean for all these to be nullable here?
Is there something stopping us from doing this instead

```suggestion
internal data class MemberAction(
  val isCancelInsuranceEnabled: Boolean,
  val isConnectPaymentEnabled: Boolean,
  val isEditCoInsuredEnabled: Boolean,
  val isMovingEnabled: Boolean,
  val isTravelCertificateEnabled: Boolean,
  val sickAbroadAction: MemberActionWithDetails.SickAbroadAction?,
  val firstVetAction: MemberActionWithDetails.FirstVetAction?,
)

internal sealed interface MemberActionWithDetails {
  data class SickAbroadAction(
    val partners: List<DeflectPartner>,
  ) : MemberActionWithDetails

  data class FirstVetAction(
    val sections: List<FirstVetSection>,
  ) : MemberActionWithDetails
}
```
So that the booleans are either true or false.
And the other two are either null, which should show nothing, or be non-null with content. We might even want to use NonEmptyList in there instead of List? Perhaps?

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/di/HelpCenterModule.kt
panasetskaya: Yep, but this use case it's not finished yet -you just said to open PR with what I have for shimmer, so I did, and this unfinished useCase just tagged along:) It's not used anywhere yet, so I think this could be merged after I fix the shimmer thing, and I will continue to work on the useCase usagefor a separate PR.

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -201,40 +187,83 @@ private fun HelpCenterHomeScreen(
...
-                            quickLink.quickAction.titleRes,
-                          )
-                        },
-                        textAlign = TextAlign.Start,
-                        modifier = Modifier.padding(horizontal = 16.dp),
+                      Spacer(
+                        modifier = Modifier
+                          .height(24.dp)
+                          .padding(horizontal = 16.dp)
+                          .width(80.dp)
```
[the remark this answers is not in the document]
panasetskaya: Been wondering about this exact thing! Great, I'll do that then

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/HelpCenterPresenter.kt
```diff
@@ -29,6 +32,7 @@ internal data class HelpCenterUiState(
   val topics: ImmutableList<Topic>,
   val questions: ImmutableList<Question>,
   val quickLinks: ImmutableList<QuickLinkType>,
+  val isLoadingQuickLinks: Boolean,
```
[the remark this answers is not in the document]
panasetskaya: ok! 😄

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetMemberActionsUseCase.kt
panasetskaya: was thinking of moving these two out of the feature module actually to some data layer.

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/home/HelpCenterHomeDestination.kt
```diff
@@ -201,40 +187,83 @@ private fun HelpCenterHomeScreen(
...
+                          .placeholder(true, highlight = PlaceholderHighlight.shimmer()),
```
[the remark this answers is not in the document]
panasetskaya: I actually think that now that it's fade and not shimmer + fade animation, it looks good even with the default color:)

StylianosGakis (approved):
Just some suggestions for the presenter to make it a bit simpler to read, but with that and the animation fix PR I made I think we can merge this.

I could not 100% tell the logic from the stuff coming from the backend if all of that works as it should, but I'll just assume that we do get all the right info there :D

app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/data/GetQuickLinksUseCase.kt
```diff
@@ -22,73 +20,86 @@ import octopus.AvailableSelfServiceOnContractsQuery
...
+            titleRes = R.string.HC_QUICK_ACTIONS_SICK_ABROAD_TITLE,
+            hintTextRes = R.string.HC_QUICK_ACTIONS_SICK_ABROAD_SUBTITLE,
+          ),
+        )
+      }
+      if (memberActionOptions.isEditCoInsuredEnabled) {
+        val contracts = apolloClient.query(AvailableSelfServiceOnContractsQuery())
+          .safeExecute()
+          .toEither(::ErrorMessage)
+          .onLeft { logcat(LogPriority.ERROR) { it.message ?: "Could not fetch common claims" } }
```
StylianosGakis, line 91: A little trick here which I've found useful
```suggestion
          .onLeft { logcat(LogPriority.ERROR) { "Could not fetch common claims ${it.message}" } }
```

This may look odd at first glance but the idea is that if the message is in fact not null, then on the logs in DDog we'll get the error message but we'll not be able to easily track it back to which log in the app it actually was. With the string always being there it's a quick "search all" and we're exactly at the right spot!

PR #2075, opened 2024-04-10
Small fixes for release

StylianosGakis:
Just used the app a bit, tested various things before release.
Changes:

* Add back enableOnBackInvokedCallback since it works well now with material sheets
* Some alignment fixes which I tested on a big font phone
* Remove a bunch of duplicated code in all of our `Vector*Card` composables, all delegating to VectorInfoCard now.
* Fix always navigating to the claim file upload screen after selecting to pick from the phone's pictures despite the fact that a picture was sometimes not taken
* And some version bumps


app/app/src/main/AndroidManifest.xml
```diff
@@ -23,7 +23,7 @@
   <application
     android:name=".HedvigApplication"
     android:allowBackup="true"
-    android:enableOnBackInvokedCallback="false"
+    android:enableOnBackInvokedCallback="true"
```
panasetskaya, line 26: That's to do with the back gesture, right? Could you briefly explain here what problem was there with the sheets?

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/layout/LayoutWithoutPlacement.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.compose.ui.layout.layout
+
+/**
+ * A [Layout] which places [sizeAdjustingContent] first in a box, making that box take *at least* its size but does not
+ * draw [sizeAdjustingContent] on the screen, and then puts [content]
+ */
+@Composable
+fun LayoutWithoutPlacement(
+  modifier: Modifier = Modifier,
+  sizeAdjustingContent: @Composable BoxScope.() -> Unit,
```
panasetskaya, line 18: So we need this for aligning according to the user's custom text size, right? Are there other scenarios where this could be adjusting to something other than text size? I was just thinking that if it's only text in all scenarios, we could probably change it to smth like ```fun LayoutCenteredAdjustedToTextSize``` and pass just textSize size there instead of sizeAdjustingContent, would be a bit easier to understand its purpose

panasetskaya (approved):

app/app/src/main/AndroidManifest.xml
StylianosGakis: Yes it's this https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture

So back some releases ago, material3 sheets were somehow unable to be dismissed when you did a back gesture when this option was on.

This was the issue https://issuetracker.google.com/issues/271303558
And this https://android-review.googlesource.com/c/platform/frameworks/support/+/2885946 commit seems to have been the fix, which was released in `material3:1.2.0-beta02`, which we have been using for quite some time now.

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/layout/LayoutWithoutPlacement.kt
StylianosGakis: Yeah the name of this composable is also kinda horrible, I didn't find something better.
Our current use case is in fact only texts.
I did not do it like that because then it becomes impossible to adjust the text's font size, which is a very important thing to do from some callers. And if I expose the font size as a parameter too, then as we use it more we will simply keep adding parameters here only to forward them to the text.
The composable slot avoids any of this problem all together. I think I may have shared this before https://chrisbanes.me/posts/slotting-in-with-compose-ui/ but it really is a great way to introduce flexibility to such layouts.

app/app/src/main/AndroidManifest.xml
panasetskaya: Great, thanks

PR #2076, opened 2024-04-10
Fix/small cleanup on the logout use case and screen

StylianosGakis:
Cleaned up LogoutUseCase to only just call `authTokenService.logoutAndInvalidateTokens` and let the side effects that come with logging out go into `AuthEventListener`s as we've done already for everything else

This allows it also to be moved outside of the app module. A lot of cleanup here all in preparation to me working with the logged out state cleaning.


app/core/core-app-review/src/main/kotlin/com/hedvig/android/core/appreview/di/CoreAppReviewModule.kt
```diff
@@ -18,4 +21,7 @@ val coreAppReviewModule = module {
   single<WaitUntilAppReviewDialogShouldBeOpenedUseCase> {
     WaitUntilAppReviewDialogShouldBeOpenedUseCaseImpl(get<SelfServiceCompletedEventStore>())
   }
+  single<SelfServiceCompletedAuthEventListener> {
+    SelfServiceCompletedAuthEventListener(get<SelfServiceCompletedEventManager>())
+  } bind AuthEventListener::class
```
panasetskaya, line 26: the fact that bind is used here - is it due to the fact that we have more than one implementation of AuthEventListener? It doesn't matter that the other implementation is in a different module?

panasetskaya (approved):

app/core/core-app-review/src/main/kotlin/com/hedvig/android/core/appreview/di/CoreAppReviewModule.kt
StylianosGakis: That is exactly it.
Here https://github.com/HedvigInsurance/android/blob/a9eac5e9e9cb9f4a40eab9e12691932eaebdfdbb/app/auth/auth-core-public/src/main/kotlin/com/hedvig/android/auth/di/AuthModule.kt#L45 you can see how in the place where we want all of those we do a getAll() to simply grab all the implementations of this interface.

You should give this https://adambennett.dev/2021/03/intoset-dagger-multibindings-and-architecture/ a read, it's for dagger but it's where I was inspired to do this.

We do this exact same trick for Initilizable https://github.com/HedvigInsurance/android/blob/4c68e5ad4723206aa1d16e79ebbc2006f7149011/app/app/src/main/kotlin/com/hedvig/android/app/di/AppModule.kt#L9 in our app, as described in the article

panasetskaya: I think I used somewhat similar multibinding, but intoMap for custom viewmodel factory, now that I read that article:) smth like:

```
    @Binds
    @IntoMap
    @ViewModelKey(RecordsViewModel::class)
    fun bindRecordsViewModel(impl: RecordsViewModel): ViewModel
```
and
```
    @Binds
    @IntoMap
    @ViewModelKey(AuthViewModel::class)
    fun bindAuthViewModel(impl: AuthViewModel): ViewModel
```
and 
```
class ViewModelFactory @Inject constructor(
    private val viewModelProviders: @JvmSuppressWildcards Map<Class<out ViewModel>, Provider<ViewModel>>
): ViewModelProvider.Factory {
    override fun <T : ViewModel> create(modelClass: Class<T>): T {
        return viewModelProviders[modelClass]?.get() as T
    }
}
```

but here's a whole new interesting usage with multi-moduling and sets. 
I wonder how this works with di-subcomponents or separate Dagger components that the main di-component depends on, here the examples are all with just di-modules. Will try it sometime later

PR #2078, opened 2024-04-11
Login in one activity

StylianosGakis:
First steps into getting rid of it.
Now move all login related screens into one activity with normal compose nav between them.
Follow up PR will be removing the entire activity as well


app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/otpinput/OtpInputDestination.kt
```diff
@@ -81,26 +121,25 @@ fun OtpInputScreen(
...
-          inputValue,
-          onInputChanged,
-          onSubmitCode,
-          networkErrorMessage,
-          onResendCode,
-          loadingResend,
-          onOpenEmailApp,
-          Modifier.padding(paddingValues),
-        )
+    Box(Modifier.fillMaxSize(), propagateMinConstraints = true) {
```
panasetskaya, line 124: Could you tell me why is this better here than AnimatedContent?

panasetskaya (approved):

app/app/src/main/kotlin/com/hedvig/app/feature/marketing/MarketingActivity.kt
```diff
@@ -67,52 +72,43 @@ class MarketingActivity : AppCompatActivity() {
       }
       lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
         demoManager.isDemoMode().first { it == true }
-        activityNavigator.navigateToLoggedInScreen(this@MarketingActivity, false)
+        activityNavigator.navigateToLoggedInScreen(this@MarketingActivity, true)
```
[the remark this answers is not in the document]
StylianosGakis: Not sure why it was not clearing the state before and what that actually meant practically. In any case however this doesn't matter here since I remove this completely anyway in a future PR

app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/otpinput/OtpInputDestination.kt
StylianosGakis: This was adding an animation between the two cases which wasn't quite necessary here. Practically it meant that when it was loading the content was flickering out and in again quickly if I remember correctly, which was quite jarring.
I just simplified this a bit to just show the loading on top of everything to remove the flickering. In general this one screen is not super well maintained anyway, but I don't think it will stick around for too long regardless.

PR #2080, opened 2024-04-12
Move the language migration from Application to LIActivity

StylianosGakis:
These sources talk a bit more about this.
https://developer.android.com/guide/topics/resources/app-languages#androidx-impl https://cs.android.com/androidx/platform/frameworks/support/+/336a1f14a9e8e6729e833b3022ad3ffa8a3f0433:appcompat/appcompat/src/main/java/androidx/appcompat/app/AppCompatDelegate.java;l=733-734

I suspect us doing it in the Application is the reason it simply didn't properly select the language for people who did not go ahead and do a language selection manually which has confused me for a very long time.

This along with getting rid of the other activities should meant that this happens just fine for all members soon.


panasetskaya (approved):
cool thing

PR #2083, opened 2024-04-15
moved forever to separate module

panasetskaya:
- moved forever ui and logic to :shared-foreverui module that :feature-forever and :feature-payments call when nedeed;
- moved related graphql files to the same module from :apollo-octopus-public


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountsDestination.kt
```diff
@@ -221,12 +221,11 @@ private fun ForeverSection(
     HorizontalItemsWithMaximumSpaceTaken(
       startSlot = {
         HedvigCard(
-          shape = MaterialTheme.shapes.squircleSmall,
           colors = CardDefaults.cardColors(
             containerColor = MaterialTheme.colorScheme.secondaryContainedButtonContainer,
             contentColor = MaterialTheme.colorScheme.onSecondaryContainedButtonContainer,
           ),
-          modifier = Modifier.wrapContentSize(Alignment.TopStart).clickable {
+          modifier = Modifier.wrapContentSize(Alignment.TopStart).clip(MaterialTheme.shapes.squircleSmall).clickable {
```
StylianosGakis, line 228: HedvigCard has an overload which takes in a onClick https://github.com/HedvigInsurance/android/blob/a93b5fe072e9aafe0f516c2409d136e16c988098/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/card/HedvigCard.kt#L50 which is applied at the right place relevant to the clip always I think.
Could you take a look if you can use that one instead here? that way you won't have to specify some shape either, it can just use the default one

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/navigation/PaymentsGraph.kt
```diff
@@ -91,13 +96,33 @@ fun NavGraphBuilder.paymentsGraph(
       )
     }
 
+    composable<PaymentsDestinations.Forever>(
+      deepLinks = listOf(
+        navDeepLink { uriPattern = hedvigDeepLinkContainer.forever },
+      ),
```
StylianosGakis, line 102:
```suggestion
```
I think we should not have the deep link on this one for forever. Instead the deep link should still just take the member to the top-level destination which shows the Forever information

Having the same deep link at two spots will most likely result in undefined behavior, like the one winning being the one first specified in the graph or something silly like that, which we definitely do not want

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/navigation/PaymentsGraph.kt
```diff
@@ -91,13 +96,33 @@ fun NavGraphBuilder.paymentsGraph(
...
+      enterTransition = { MotionDefaults.fadeThroughEnter },
+      exitTransition = { MotionDefaults.fadeThroughExit },
```
StylianosGakis, line 104:
```suggestion
```
These can be removed as well, so that way the normal x-axis animation can play here, like any other screen

These fade transitions are specifically made for those top level destinations only.

It is a bit simple how we've set this all up, but due to lack of more specific design guidance on this we decided on this approach. Fade for top level, x axis for everything else.

So "top-level" here https://m3.material.io/styles/motion/transitions/transition-patterns#8327d206-2a7d-423a-abf0-2be86130535b

And forward and backward here https://m3.material.io/styles/motion/transitions/transition-patterns#df9c7d76-1454-47f3-ad1c-268a31f58bad for the x-axis that I am talking about

app/shared/forever-ui/src/main/graphql/MutationMemberReferralInformationCodeUpdate.graphql
StylianosGakis, line 1: Is this an empty .graphql file or is GH acting weird here? If it's completely empty could we just delete it perhaps?

hedvig-lint/lint-baseline/lint-baseline-forever-ui.xml
```diff
@@ -0,0 +1,18 @@
...
+        <location
+            file="build.gradle.kts"/>
+    </issue>
+
+    <issue
+        id="UnknownIssueId"
+        message="Unknown issue id &quot;UnsafeOptInUsageError&quot;">
+        <location
+            file="build.gradle.kts"/>
+    </issue>
```
StylianosGakis, line 16: Interesting that these are here in the first place. Usually these are empty, especially for new modules. If you delete this baseline file and run `./gradlew lint` again does it tell you where these are originating from?

StylianosGakis (approved):
Hey you killed it, perfect, this is exactly what we want here!

Two comments on some things that were copied over from the top-level forever destination which we do not want here regarding deep links and animations. Otherwise this is perfect! Great job.

These sort of changes certainly blur the lines a bit about what is a feature module and what is not, but this is a thing for us to consider going forward anyway. Just noting it down to keep it in our minds

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/navigation/PaymentsGraph.kt
panasetskaya: shoud've thought of that:)

StylianosGakis: Dw about it, that's what the PR review is for 😊

app/shared/forever-ui/src/main/graphql/MutationMemberReferralInformationCodeUpdate.graphql
panasetskaya: It's not empty actually, I wonder why it showed that way to you:(

hedvig-lint/lint-baseline/lint-baseline-forever-ui.xml
panasetskaya: empty now:)

PR #2091, opened 2024-04-16
GEN-1925 Fix/payout changes  - can merge now

panasetskaya:
Changes to payout screen with the new repair cost.

- [x] Do not merge, waiting for the copy
- [x] Breaking schema CHANGE!


StylianosGakis:
SingleItemCheckoutViewModelTest tests seems to be failing for this.
Btw still waiting for copy for this right?

panasetskaya:
> SingleItemCheckoutViewModelTest tests seems to be failing for this. Btw still waiting for copy for this right?

yep, it's not for merging now. I'll look into the tests

StylianosGakis:
I'll keep it as draft until translations come because I get confused otherwise thinking I forgot to address this :D

app/apollo/apollo-octopus-public/src/main/graphql/com/hedvig/android/apollo/octopus/graphql/claimflow/FragmentClaimFlowStepFragment.graphql
```diff
@@ -90,21 +90,48 @@ fragment FlowClaimSingleItemStepFragment on FlowClaimSingleItemStep {
...
+  availableCheckoutMethods {
+    id
+    ...CheckoutMethodFragment
+  }
+  compensation {
+    ...CompensationFragment
+  }
+}
+
+fragment CompensationFragment on FlowClaimSingleItemCheckoutCompensation {
```
StylianosGakis, line 105: This fragment only seems to be used in this one place. I don't think we have any rules or something like that on how to usually approach this, but do you think it'd make sense here to inline this at the one place where it's called? So that we do not introduce a new fragment here which might come with some mental overhead of what this new type is about?
Or do you think that it helps with keeping it easier to work with in the code?

I just kinda remember sometime in the past getting rid of a bunch of fragments which were used only in one place and it typically made things a bit easier to reason about.
But I do not have a very strong opinion about it I think, I just believe that if I wrote it I would try out both and try to see which one feels nicer. 

Or do we perhaps expect this to in the future be used in more than one of these steps?

app/data/data-claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowDestination.kt
```diff
@@ -128,13 +128,39 @@ sealed interface ClaimFlowDestination : Destination {
...
+      @Serializable
+      data class RepairCompensation(
+        val repairCost: UiMoney,
+        override val deductible: UiMoney,
+        override val payoutAmount: UiMoney,
+      ) : Known
+    }
+
+    @Serializable
+    data object UnKnown : Compensation
```
StylianosGakis, line 161:
```suggestion
    data object Unknown : Compensation
```

app/apollo/apollo-octopus-public/src/main/graphql/com/hedvig/android/apollo/octopus/graphql/claimflow/FragmentClaimFlowStepFragment.graphql
```diff
@@ -90,21 +90,48 @@ fragment FlowClaimSingleItemStepFragment on FlowClaimSingleItemStep {
...
+  id
+  ...CheckoutRepairCompensationFragment
```
StylianosGakis, line 107: Don't know if I am being too extra here 😅 But these two also are used in one place only. And I also wonder if we can defer creating them until we have a use case where we do in fact want to use this type in more than one place?

app/data/data-claim-flow/src/main/kotlin/com/hedvig/android/data/claimflow/ClaimFlowStepExt.kt
```diff
@@ -171,6 +178,28 @@ private fun CheckoutMethodFragment.toCheckoutMethod(): CheckoutMethod {
   }
 }
 
+private fun CompensationFragment.toCompensation(): ClaimFlowDestination.Compensation {
```
StylianosGakis, line 181: If those framents were not here, we'd then be checking here on
is FlowClaimSingleItemCheckoutRepairCompensationCompensationFragment
is FlowClaimSingleItemCheckoutValueCompensationCompensationFragment
Instead of those two fragments. Which is definitely more verbose, but we get type safety here it's not a problem of mistyping something :D

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -89,56 +105,175 @@ private fun SingleItemCheckoutScreen(
   navigateUp: () -> Unit,
   closeClaimFlow: () -> Unit,
 ) {
+  var bottomSheetText by remember { mutableStateOf<Pair<String, String>?>(null) }
+  val explanationSheetState = rememberModalBottomSheetState(true)
+  if (bottomSheetText != null) {
+    SingleItemCheckoutInfoBottomSheet(
+      onDismiss = { bottomSheetText = null },
+      sheetState = explanationSheetState,
+      explanationTitleAndText = bottomSheetText ?: Pair("", ""), // todo: do I like it here?
```
StylianosGakis, line 114: You can write it like this
```kotlin
  var bottomSheetText by remember { mutableStateOf<Pair<String, String>?>(null) }
  val explanationSheetState = rememberModalBottomSheetState(true)
  val bottomSheetTextValue = bottomSheetText
  if (bottomSheetTextValue != null) {
    SingleItemCheckoutInfoBottomSheet(
      onDismiss = { bottomSheetText = null },
      sheetState = explanationSheetState,
      explanationTitleAndText = bottomSheetTextValue,
    )
  }
```
And this way you will get the right null safe check in there

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -89,56 +105,175 @@ private fun SingleItemCheckoutScreen(
...
+    HorizontalItemsWithMaximumSpaceTaken(
+      startSlot = {
+        Row(verticalAlignment = Alignment.CenterVertically) {
+          Text(
+            text = stringResource(R.string.CLAIMS_CHECKOUT_COUNT_TITLE),
+            modifier = sideSpacingModifier,
+          )
+        }
+      },
+      endSlot = {
```
StylianosGakis, line 147: Do we want some space between here just in case?

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -89,56 +105,175 @@ private fun SingleItemCheckoutScreen(
...
     CompositionLocalProvider(LocalContentColor provides MaterialTheme.colorScheme.onSurfaceVariant) {
       Column(sideSpacingModifier) {
-        val pairs = listOf(
-          stringResource(R.string.KEY_GEAR_ITEM_VIEW_VALUATION_PAGE_TITLE) to uiState.price.toString(),
-          stringResource(R.string.claims_payout_age_deduction) to "-" + uiState.depreciation.toString(),
-          stringResource(R.string.claims_payout_age_deductable) to "-" + uiState.deductible.toString(),
-        )
+        val pairs = when (uiState.compensation) {
+          is ClaimFlowDestination.Compensation.Known.RepairCompensation,
+          -> listOf(
```
StylianosGakis, line 183:
```suggestion
          is ClaimFlowDestination.Compensation.Known.RepairCompensation -> listOf(
```

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -89,56 +105,175 @@ private fun SingleItemCheckoutScreen(
...
         for ((left, right) in pairs) {
-          Row {
-            Text(left, Modifier.weight(1f))
-            Text(right, Modifier.weight(1f), textAlign = TextAlign.End)
-          }
+          HorizontalItemsWithMaximumSpaceTaken(
+            startSlot = {
+              Text(left)
+            },
+            endSlot = {
```
StylianosGakis, line 202: We probably want some sane space between here too. I know it's not really evident in the design for all these scenarios, since they rarely show the edge cases, but even a sane 8.dp as space between is more or less always the right choice when having texts side-by-side like that.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -204,9 +341,81 @@ private fun ColumnScope.CheckoutMethods(
   }
 }
 
+@Composable
+internal fun SingleItemCheckoutInfoBottomSheet(
+  onDismiss: () -> Unit,
+  sheetState: SheetState,
+  explanationTitleAndText: Pair<String, String>,
+) {
+  ModalBottomSheet(
```
StylianosGakis, line 350: Can you check if you could use HedvigBottomSheet here?
It contains some important changes around the dismiss click and letting the sheet slide away before it actually disappears here https://github.com/HedvigInsurance/android/blob/73696b860d2a86faee18d4b89f56f9877714d182/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/bottomsheet/HedvigBottomSheet.kt#L77-L83
Otherwise if we call dismiss directly it just disappears instantly without animating away. Not sure if it is applicable here but worth a look I think.

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
```diff
@@ -148,11 +283,13 @@ private fun SingleItemCheckoutScreen(
     )
     Spacer(Modifier.height(16.dp))
     Spacer(Modifier.weight(1f))
-    VectorInfoCard(stringResource(R.string.CLAIMS_CHECKOUT_NOTICE), sideSpacingModifier)
+    if (uiState.compensation is ClaimFlowDestination.Compensation.Known.ValueCompensation) {
+      VectorInfoCard(stringResource(R.string.CLAIMS_CHECKOUT_NOTICE), sideSpacingModifier)
+    }
     Spacer(Modifier.height(16.dp))
```
StylianosGakis, line 289: If the vector card does not show here, then we got a 16 height space, then the 1f weight then a 16 height space again. Making this button be 32 away from whatever is above, while if the card is there, the card is 16 away from both. Is this expected here?

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
StylianosGakis: It also adds the dismiss button in the right spot automatically I think, as a bonus :D

StylianosGakis (approved):
Some more thoughts, check out the comments 😊

app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/singleitemcheckout/SingleItemCheckoutDestination.kt
panasetskaya: maybe we put it as default for HorizontalItemsWithMaximumSpaceTaken, no?

StylianosGakis: I've thought of it before too, but we don't always use this just for texts being laid out side-by-side, and in those scenarios it would be confusing to have some padding that you did not specify, where you have to pass `spaceBetween = 0.dp`. It feels odd to pass the 0.dp explicitly in a way.

panasetskaya:
Done and the code review comments are fixed, but waiting for copy still.

panasetskaya:
A per message in Slack here: https://hedviginsurance.slack.com/archives/C06JR8EHX6E/p1715154145688379
Added purchasePriceApplicable (new field in schema) to hide/show purchase price on SingleItemStep.

PR #2096, opened 2024-04-17
New hedvig design system icons

StylianosGakis:
Just the 24 ones for now.
For this I also added our own LocalContentColor and Icon composables
Will look into the colored and the special ones later, and which ones we can support simply by using the existing ones and adding colors, size etc on top of them

image


panasetskaya (approved):
13000 lines of changes sound scary 😄

StylianosGakis:
Yeah but it's all icons, you can see the files changed too :D

PR #2098, opened 2024-04-18
Bump splash screen dependency to fix status bar flicker

StylianosGakis:
This one *finally* fixes this super annoying problem where the status bar would quickly flicker to dark mode when the app was launching after the splash screen left the view and the app started showing its UI https://developer.android.com/jetpack/androidx/releases/core#core-splashscreen-1.2.0-alpha01

https://android-review.googlesource.com/c/platform/frameworks/support/+/3018911

| before | after |
| --- | --- |
| https://github.com/HedvigInsurance/android/assets/44558292/5044c5b5-59ec-438a-9b6d-e565601b1894 | https://github.com/HedvigInsurance/android/assets/44558292/64307876-8640-47f8-a4a2-85015320ecba |


StylianosGakis:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-splashscreen/build.gradle;l=41-43
alpha is fine here since they do not depend on nay other alpha library for us to accidentally bring in more alphas with it.

panasetskaya (approved):

PR #2100, opened 2024-04-18
Feature/placeholders in insurances

panasetskaya:
- Added placeholders for Insurances, InsuranceDetails, rows in the Profile
- Changed placeholder impl from composed {} to ModifierNodeElement
- Also put animation preview on click to core-ui function, bc it's very useful:) it's HedvigPreviewWithProvidedParametersAnimation now


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
```diff
@@ -145,59 +152,62 @@ private fun InsuranceScreen(
...
-        true -> HedvigFullScreenCenterAlignedProgressDebounced(Modifier.fillMaxSize())
-        false -> {
-          Column(
-            Modifier
-              .fillMaxSize()
-              .pullRefresh(pullRefreshState)
-              .verticalScroll(rememberScrollState())
-              .windowInsetsPadding(WindowInsets.safeDrawing),
-          ) {
-            Spacer(Modifier.height(16.dp))
```
panasetskaya, line 166: This one is to bring "Insurances" title so it would be at the same place as other titles in the other bottom tabs (Forever etc.), and at the same place as "Insurance Details" title one screen deeper

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
```diff
@@ -322,6 +334,85 @@ private fun CrossSellItem(crossSell: CrossSell, onCrossSellClick: (String) -> Un
...
+    List(5) {
+      CrossSellItemPlaceHolder(
+        modifier = Modifier.padding(horizontal = 16.dp),
+      )
+    }
+  }
+}
+
+@Composable
+private fun CrossSellItemPlaceHolder(modifier: Modifier = Modifier) {
```
panasetskaya, line 364: the placeholder does not strictly follow the shape of Composable, it adds a bit of overlap. That's why if we add two lines of text for placeholder it looks a bit ugly; and it still moves around a bit when it loaded, bc some crossCells have 1 row of text, and some two.

app/ui/placeholder/src/main/kotlin/com/hedvig/android/placeholder/PlaceholderHightlightMaterial3.kt
```diff
@@ -39,7 +40,10 @@ fun PlaceholderHighlight.Companion.shimmer(
   animationSpec: InfiniteRepeatableSpec<Float> = PlaceholderDefaults.shimmerAnimationSpec,
   @FloatRange(from = 0.0, to = 1.0) progressForMaxAlpha: Float = 0.6f,
 ): PlaceholderHighlight = PlaceholderHighlight.shimmer(
-  highlightColor = PlaceholderDefaults.shimmerHighlightColor(),
+  highlightColor = PlaceholderDefaults.shimmerHighlightColor(
+    backgroundColor = MaterialTheme.colorScheme.surface,
+    alpha = 1f,
```
StylianosGakis, line 45: How did the default of 0.75f look like here? did you feel like it was way too subtle?

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -228,24 +299,14 @@ private fun ColumnScope.ProfileItemRows(
     icon = Icons.Hedvig.ContactInformation,
     onClick = showMyInfo,
   )
-  AnimatedVisibility(
-    visible = profileUiState.travelCertificateAvailable,
-    enter = fadeIn() + expandVertically(expandFrom = Alignment.CenterVertically, clip = false),
-    exit = fadeOut() + shrinkVertically(shrinkTowards = Alignment.CenterVertically, clip = false),
-    label = "Travel",
-  ) {
+  if (profileUiState.travelCertificateAvailable) {
```
StylianosGakis, line 302: Do we want these to jump in instead of animate in now? Did the existing animation interfere with something else? In any case, this would make these items jump more suddenly in or out of the view right? if yes, is this what we want here?

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -160,13 +162,13 @@ private fun ProfileScreen(
...
-        navigateToEurobonus = navigateToEurobonus,
-        navigateToTravelCertificate = navigateToTravelCertificate,
+      ProfileRowsWithPlaceholders(
+        uiState,
+        navigateToMyInfo,
+        navigateToSettings,
+        navigateToAboutApp,
+        navigateToEurobonus,
+        navigateToTravelCertificate,
       )
```
StylianosGakis, line 172: Since all of those here have the same signature of just `() -> Unit` it's super easy to flip one by accident and there won't be anything that would indicate this mistake.
Having the named parameter at least would make it more obvious if we did for example `showSettings = navigateToAboutApp,` to spot such problems. So I think we should make sure to add the named parameters back here.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -214,6 +218,73 @@ private fun ProfileScreen(
   }
 }
 
+@Composable
+private fun ColumnScope.ProfileRowsWithPlaceholders(
```
StylianosGakis, line 222: The `ColumnScope` here doesn't do something since it only has one composable top-level anyway, the AnimatedContent, so it can be removed.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -214,6 +218,73 @@ private fun ProfileScreen(
...
+      text = title,
+      style = MaterialTheme.typography.bodyLarge,
+      modifier = Modifier
+        .placeholder(true, highlight = PlaceholderHighlight.shimmer()),
+    )
+  }
+}
+
 @Composable
 private fun ColumnScope.ProfileItemRows(
```
StylianosGakis, line 289:
```suggestion

@Suppress("UnusedReceiverParameter")
@Composable
private fun ColumnScope.ProfileItemRows(
```
Here I would suggest this, since we do in fact want to know that we are inside a ColumnScope, since we emit mutltiple composables without explicitly wrapping them in some container like Column.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
StylianosGakis: Or is it because we have the placeholders anyway initially, so the moment that this would run here it would be at the same time that we go from placeholder -> real content, so it wouldn't show any jumpiness?

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
StylianosGakis, line 222: I would dare say that we do not need to make this have in the name the `WithPlaceholders` part. That should just be a decision that it can make internally on how to render the profile rows according to the ProfileUiState passed into it.

And for the other `ProfileRows` further down which would then have the same signature, I would say that one can instead receive `ProfileUiState.Loaded` or something like that, indicating that there is no loading stuff anymore.

Since I know that the UiState does not have this kind of shape right now, perhaps it's the time to change it. If the UI wants those two states to be distinct enough as to what it will render, the presenter can help out by providing those states exactly like that instead.

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/tab/ProfileDestination.kt
```diff
@@ -214,6 +218,73 @@ private fun ProfileScreen(
...
+@Composable
+private fun ColumnScope.ProfileItemRowsPlaceholders() {
+  ProfileRowPlaceholder(stringResource(R.string.PROFILE_MY_INFO_ROW_TITLE))
+  ProfileRowPlaceholder(stringResource(R.string.PROFILE_ROW_TRAVEL_CERTIFICATE))
+  ProfileRowPlaceholder(stringResource(R.string.PROFILE_ABOUT_ROW))
+  ProfileRowPlaceholder(stringResource(R.string.profile_appSettingsSection_row_headline))
+}
+
+@Composable
+private fun ProfileRowPlaceholder(title: String) {
```
StylianosGakis, line 264: This function and 
`private fun ProfileRow(title: String, icon: ImageVector, onClick: () -> Unit, modifier: Modifier = Modifier) {`
function each having their own UI logic means that they can also go out of sync quite easily. Could you instead look into providing another private composable in this file, which will contain the core logic of how the UI should be laid out. It may need to have some composable slots in there, I don't have something specific in mind unless I sit down and try to do it myself. And then have the placeholder and the non-placeholder composables call *that* instead. This way there will be better guarantees that these two do not grow out of sync.

In general as we try to do this more and more I am sure we will pick up better patters on how to do this without complicating our code too much and without making it possible to have the placeholder and non-placeholder composables go out of sync.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -151,11 +173,6 @@ private fun ContractDetailScreen(
               .only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom)
               .asPaddingValues()
               .plus(PaddingValues(top = 16.dp)),
-            modifier = Modifier
-              .fillMaxSize()
-              .consumeWindowInsets(
-                WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom),
-              ),
```
StylianosGakis, line 158: Did these break something?
They were here to consume those insets since the insets are applied to the lazyColumn using `asPaddingValues()`, but that function does not consume the insets for the children, it only uses them and leaves them untouched, so we had to do this manually here.
It may need to stay here for this reason.
Now that we are consuming them higher up, I am not sure if this modifier is actually affected 👀
Because I know that it does not set the consumer modifier local, but I am not sure if it at least reads from it only.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -127,14 +140,23 @@ private fun ContractDetailScreen(
         }
       },
       label = "contract detail screen fade animated content",
-      modifier = Modifier.weight(1f),
+      transitionSpec = { fadeIn() togetherWith fadeOut() },
```
StylianosGakis, line 143: I think we still want the weight here so that the layout still always tries to take up the entire height. I remember the problem was that for example if you are only in the documents tab, which is quite short and often does not have enough content to cover the entire screen, then going back to the Coverage tab you'd see clipping for a bit until it animated back up again.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
StylianosGakis: I think the same logic applies here as what I describe in the profile screen.
For these placeholders to still be able to follow the changes of the real UI if they ever come, they need to share more of their UI code.
If that is by having a common parent composable, or by both states being actually the exact same composable but with having `placeholder(visible = {state is Loading})` is probably something that should be decided on a case by case basis.
But if they are just copy pasted UI it is bound to have slight differences at least down the road.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt
```diff
@@ -145,59 +152,62 @@ private fun InsuranceScreen(
...
+        },
+        label = "uiState is Loading",
+      ) { isLoading ->
+        Column(
+          Modifier
+            .fillMaxSize(),
+        ) {
+          if (uiState.hasError) {
+            HedvigErrorSection(onButtonClick = reload)
+          } else {
```
StylianosGakis, line 187: With this there is no animation to and from the error state. Would we perhaps optimally want to have something there too? So that the AnimatedContent was on the uiState in general instead of just the isLoading case?

app/feature/feature-insurances/build.gradle.kts
```diff
@@ -63,6 +63,7 @@ dependencies {
   testImplementation(projects.loggingTest)
   testImplementation(projects.moleculeTest)
   testImplementation(projects.notificationBadgeDataFake)
+  implementation(project(":placeholder"))
```
StylianosGakis, line 66:
```suggestion
  implementation(projects.placeholder)
```

StylianosGakis (changes requested):
I really like the way things start to look doing this! The UI is also looking great in general, and since in this case we have a bit less control over what the real data will show it's fine that some things aren't replaced just in-place directly.

I have some thoughts around:
* Not changing the animation between screens yet
* Sharing the core layouts behind the placeholder and the real counterparts so that there's no shift between the two with any introduced change
* Some things around previews and the new functions you've added and how to improve them in some ways

And other small things that would be nice to be addressed before we merge this in 😊

app/ui/placeholder/src/main/kotlin/com/hedvig/android/placeholder/Placeholder.kt
```diff
@@ -81,117 +76,227 @@ internal object PlaceholderDefaults {
...
- * @param placeholderFadeTransitionSpec The transition spec to use when fading the placeholder
+ * @param placeholderFadeAnimationSpec The transition spec to use when fading the placeholder
  * on/off screen. The boolean parameter defined for the transition is [visible].
- * @param contentFadeTransitionSpec The transition spec to use when fading the content
+ * @param contentFadeAnimationSpec The transition spec to use when fading the content
  * on/off screen. The boolean parameter defined for the transition is [visible].
  */
-fun Modifier.placeholder(
+
+fun Modifier.placeHolder(
```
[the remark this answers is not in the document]
panasetskaya: Yep:) changed that. I wanted to distinguish between two Modifier.placeholder, one of them used inside the other. made this one internal then.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceGraph.kt
```diff
@@ -96,3 +104,17 @@ fun NavGraphBuilder.insuranceGraph(
...
+    initialAlpha = 0.99f,
+  )
+}
+
+private val noExitTransition: AnimatedContentTransitionScope<NavBackStackEntry>.() -> ExitTransition = {
+  fadeOut(
+    animationSpec = tween(durationMillis = 300),
+    targetAlpha = 0.99f,
+  )
+}
```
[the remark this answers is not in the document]
panasetskaya: these were done to avoid a quick flashing that is caused by the default navigation behaviour.

With noExitTransition:

https://github.com/HedvigInsurance/android/assets/90948269/94304589-aae2-4677-9b14-36d9d3b4cf83

With default:

https://github.com/HedvigInsurance/android/assets/90948269/a615e1c7-8603-4d8d-a4fd-fbe66871fd4e

app/ui/placeholder/src/main/kotlin/com/hedvig/android/placeholder/PlaceholderHightlightMaterial3.kt
panasetskaya: a bit, but I've put it back to default now:) it's not that much of difference

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceGraph.kt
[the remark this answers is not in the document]
StylianosGakis: Yeah but this is fine. We will fix the card looking like it's in-place with the shared element apis.
Now with removing the animation the rest of the screen also fails to animate, meaning the items at the bottom and this will be different compared to every other screen in the app.
It also changes the animation for if you are not coming from this screen, if that ever happens in some way, instead of only the transition between these two screens.

StylianosGakis (changes requested):
Some more thoughts on the placeholder stuff. Namely concerns about using the preview models in the production composables and what implications those may have.

Also a note on using https://developer.android.com/develop/ui/compose/animation/composables-modifiers#animatedcontent AnimatedContent and using the value provided by its lambda which may be a problem here

app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/preview/ConvenienceCollectionPreviewParameterProvider.kt
```diff
@@ -40,16 +40,15 @@ class DoubleBooleanCollectionPreviewParameterProvider : CollectionPreviewParamet
...
-  surfaceColor: Color,
   content: @Composable (parameterState: T) -> Unit,
+  modifier: Modifier = Modifier,
 ) {
   var parameterStateIndex by remember { mutableIntStateOf(0) }
   val interactionSource = remember { MutableInteractionSource() }
-  Surface(
-    color = surfaceColor,
-    modifier = Modifier.clickable(
+  Column(
```
StylianosGakis, line 50: Box should probably be best here as we do not need a column behavior

PR #2105, opened 2024-04-24
show net amount instead of gross

panasetskaya:
Use "net" amount for upcoming payment info section on Payments screen instead of gross (that's how it's done on iOs too).


StylianosGakis (approved):
Perfect 🤩

PR #2106, opened 2024-04-25
Design: Tokens and Button component

StylianosGakis:
known issues:
- Dark mode incoming
- The loading indicators are wrong for the primary button https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1714039563189999

Open this page https://www.figma.com/file/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?type=design&node-id=15694%3A15728&mode=design&t=Rm5zSWhquVdnRBOi-1 and then look at the showcase composable here:

https://github.com/HedvigInsurance/android/assets/44558292/fd02efdb-94f0-44a0-a5f9-b2a444a764df

I definitely encourage you to try and check this branch out and run the sample app to get an idea of how these showcase pages can look like.

From there you can look into the implemenation as well.
It took me a bunch of do-overs to be in the state it is right now. I feel like this as a first step is good to move forward to adding the next components as well.

I will as a first step add dark mode tokens too, since now I feel a bit more confident in how I've mapped everything from the DS.

The token structure may look a bit daunting, and in a way it is. But it is meant as a way for us to be able to do the hard work once for each new component, by adding a token list for each one of them, which should just refer to our [ColorSchemeTokens](https://github.com/HedvigInsurance/android/blob/a04a6b8e0c4ab5399d1f4689c301dff07c93260c/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/ColorSchemeKeyTokens.kt#L4), and then everything must just fall into place.
This is the same structure m3 uses internally more or less.


StylianosGakis:
Another important caveat here, for the ripple implementation I used an alpha dependency on foundation or something like that because it made copying over the implementation much simpler. I wanted to make sure things work before I did more work there.
Before merging I will likely just go and bring in all the code, the deprecated code too which m3 is also using so that we do not need to bring in the alpha dependency

panasetskaya:
Is it supposed that the black button has no dots when loading?

StylianosGakis:
It's part of the known issues with a link to slack discussion above.
I will change that to something else soo

panasetskaya:
Ah, alright, I thought from our conversation in slack [here](https://hedviginsurance.slack.com/archives/D06FUJP5JRL/p1714050129323149?thread_ts=1714046361.395189&cid=D06FUJP5JRL) for now you decided to pick a different color:)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
```diff
@@ -0,0 +1,366 @@
...
+  ) {
+    val loadingTransition = updateTransition(isLoading, label = "loading transition")
+    loadingTransition.AnimatedContent(
+      transitionSpec = {
+        fadeIn(tween(durationMillis = 220, delayMillis = 90)) togetherWith fadeOut(tween(90))
+      },
+      contentAlignment = Alignment.Center,
+    ) { loading ->
+      if (loading) {
+        LayoutWithoutPlacement(
```
panasetskaya, line 72: I know it's not new already, but I love it

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
```diff
@@ -0,0 +1,366 @@
...
+  internal fun hoverColor(isHovered: Boolean): Color = when {
+    isHovered -> hoverContainerColor
+    else -> Color.Unspecified
+  }
+}
+
+private sealed interface Size {
+  val contentPadding: PaddingValues
+
+  @get:Composable
```
panasetskaya, line 201: that a new annotation for me

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
```diff
@@ -0,0 +1,382 @@
...
+        modifier = Modifier.padding(buttonSize.size.contentPadding),
+        horizontalArrangement = Arrangement.Center,
+        verticalAlignment = Alignment.CenterVertically,
+        content = content,
+      )
+    }
+  }
+}
+
+object ButtonDefaults {
```
panasetskaya, line 132: I like the whole system quite a lot! It seems to be quite logical. Maybe something will come up with actual implementation, but now it at least seems very easy-to-use

panasetskaya (approved):

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
StylianosGakis: Yep, it's the Kotlin way to say that the getter has to be annotated with the annotation that you provide after the get: part. It's interesting indeed, if you tried to put the annotation normally on the val it just doesn't work here 😊

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
StylianosGakis: Absolutely, these are for us to figure out how we like them or not. I definitely borrowed this entire structure from material3, where they do more or less the same. I adjusted it to our needs where we gave this 2x2 matrix of possibilities. Size and style. And the public API reflects that exactly. I hope it proves to be as easy to use as I think it will. It will 100% depend also on our designs following the design system properly 😅

panasetskaya: I think next week you'd have to teach me a bit on how to use it with figma. I guess, for example, surface/opaque/primary is just surfacePrimary in our system, right? But for another example, in figma size: Medium with corner radius/l - this one is not very clear to me yet, how it would be translated to our android.

PR #2107, opened 2024-04-25
GEN-2034 Migrate InsuranceDetails and TerminatedInsurances VM to molecule

panasetskaya:
- ContractDetailViewModel and TerminatedContractsViewModel become molecule.
- Moved QueryInsuranceContracts inside the module
- Add tests for both new presenters


panasetskaya:
should `cancelAndIgnoreRemainingEvents()` be added at the end of the tests here? sometimes I see it added, sometimes not

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/terminatedcontracts/TerminatedContractsViewModel.kt
```diff
@@ -1,75 +1,87 @@
...
-    SharingStarted.WhileSubscribed(5.seconds),
-    TerminatedContractsUiState.Loading,
-  )
+    }
 
-  fun retry() {
-    if (uiState.value is TerminatedContractsUiState.Error) {
-      retryChannel.retry()
+    LaunchedEffect(dataLoadIteration) {
+      if (lastState !is TerminatedContractsUiState.Success) {
```
StylianosGakis, line 51: If the `lastState` was in fact success.
In the scenario where after the success, we still end up going into an error state, and then we do a retry after that, we will notice that the `lastState` which does not change was success, we will not see the state go from error to loading and then to the result of the use case. It would go from error to staying as error while we're loading, and then we'd get back the result, we wouldn't see the loading in-between.

I wrote this test
```kotlin
  @Test
  fun `with an initial success state, if there is an error, we are able to retry and get back in success state again`() = runTest {
    val getInsuranceContractsUseCase = FakeGetInsuranceContractsUseCase()
    val presenter = TerminatedContractsPresenter { getInsuranceContractsUseCase }
    presenter.test(TerminatedContractsUiState.Success(getInsuranceContractsUseCase.getTerminatedInsurances())) {
      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Success>()
      getInsuranceContractsUseCase.addErrorToResponse()
      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Error>()
      sendEvent(TerminatedContractsEvent.Retry)
      // assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Loading>()
      getInsuranceContractsUseCase.addTerminatedInsurancesToResponse()
      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Success>()
    }
  }
```
And notice how it passes without there being a Loading state after the restart, despite us being in the error case and then doing restart. Specifically because of the bug I mention here. If you uncomment the line that checks for the error state the test will fail. Optimally it should pass. Perhaps you have to check on the `currentState` instead of `initialState` inside the LaunchedEffect.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailViewModel.kt
```diff
@@ -1,54 +1,75 @@
...
-  fun retryLoadingContract() {
-    retryChannel.retry()
+    CollectEvents { event ->
+      when (event) {
+        is ContractDetailsEvent.RetryLoadingContract -> dataLoadIteration++
+      }
+    }
+
+    LaunchedEffect(dataLoadIteration) {
+      if (lastState !is ContractDetailsUiState.Success) {
```
StylianosGakis, line 48: The same bug would apply here too, see the comment in the other presenter

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailPresenterTest.kt
```diff
@@ -0,0 +1,222 @@
...
+import org.junit.Test
+
+class ContractDetailPresenterTest {
+  @get:Rule
+  val testLogcatLogger = TestLogcatLoggingRule()
+
+  @Test
+  fun `if termination flow enabled and no termination date show cancel insurance button`() = runTest {
+    val getContractForContractIdUseCase = FakeGetContractForContractIdUseCase()
+    val featureManager = FakeFeatureManager(featureMap = { mapOf(Feature.TERMINATION_FLOW to true) })
```
StylianosGakis, line 35: That's totally my bad, but can you see if you can use FakeFeatureManager2 here instead?
The first iteration did not allow for use to do stuff like wait and assert states before the feature manager has managed to return something yet and so on.
I just have never taken the time to remove the FakeFeatureManager from the tests that use it and just use "2" everywhere. And obviously remove the "2" part from them 😂

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailPresenterTest.kt
```diff
@@ -0,0 +1,222 @@
...
+import com.hedvig.android.molecule.test.test
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.LocalDate
+import org.junit.Rule
+import org.junit.Test
+
+class ContractDetailPresenterTest {
```
StylianosGakis, line 28: Super nice here as well.
I would suggest writing some tests with a different initial state too, bring in the test that would repro the bug of having success as the initial state as well and adjust accordingly.

StylianosGakis (approved):
Amazing job with all these, I really like your initiative. I think you've done a great job, I have a lot of comments because there were a lot of things I wanted to share around this :D 

There's only 1 "real" bug where I describe a test case which would fail right now, around separate initialState scenarios. I think it's good to have some tests that test different initial states too, as that is something that does happen sometimes.


About `cancelAndIgnoreRemainingEvents`. It should optimally not be used indeed in these tests.
It is used in scenarios where we want to test the presenter only up to a certain point, and from that point on we might not care for the rest of the states as those scenarios may be covered from other tests.
So in normal tests where you want to ensure that all emitted UI states are accounted for you definitely do not want to ignore the follow-up states.

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/terminatedcontracts/TerminatedContractsPresenterTest.kt
```diff
@@ -0,0 +1,241 @@
...
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.toPersistentList
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.LocalDate
+import org.junit.Rule
+import org.junit.Test
+
+class TerminatedContractsPresenterTest {
```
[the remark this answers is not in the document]
panasetskaya: Thanks! That's very helpful to know. I think one thing that I don't like here in the second approach is that there are two things on the different levels being tested at the same time, but sometimes this is needed.

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailPresenterTest.kt
```diff
@@ -0,0 +1,340 @@
...
+      featureManager = featureManager,
+      getContractForContractIdUseCase = getContractForContractIdUseCase,
+    )
+    presenter.test(
+      ContractDetailsUiState.Error,
+    ) {
+      assertThat(awaitItem()).isInstanceOf<ContractDetailsUiState.Error>()
+      getContractForContractIdUseCase.addInsuranceWithNoTerminationDateToResponseTurbine()
+      assertThat(awaitItem()).isInstanceOf<ContractDetailsUiState.Loading>()
+      assertThat(awaitItem()).isInstanceOf<ContractDetailsUiState.Success>()
```
panasetskaya, line 114: I don't understand why it shows loading state in here, if we didn't emit retry event and didn't trigger dataLoadIteration. Shouldn't we just go directly from error to success here?

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/terminatedcontracts/TerminatedContractsPresenterTest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+  @Test
+  fun `with an initial error state, if there comes a successful response in flow, show success state`() = runTest {
+    val getInsuranceContractsUseCase = FakeGetInsuranceContractsUseCase()
+    val presenter = TerminatedContractsPresenter { getInsuranceContractsUseCase }
+    presenter.test(TerminatedContractsUiState.Error) {
+      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Error>()
+      getInsuranceContractsUseCase.addTerminatedInsurancesToResponse()
+      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Loading>()
+      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Success>()
```
panasetskaya, line 82: same question: why Loading state?

StylianosGakis: If you move `      assertThat(awaitItem()).isInstanceOf<TerminatedContractsUiState.Loading>()
` to the line above the `addTerminatedInsurancesToResponse` the test still passes right?

From the code it looks like when the LaunchedEffect starts, it it's not success it will automatically set the state to loading. And here we start with error, so this does happen.
The key concept here is that LaunchedEffect will in fact run by itself on the first composition, and here we are coming to a fresh presenter just with an error last state. That's still a fresh composition, which is the same behavior people will experience using the app for real.

So practically, the loading state is emitted to the UI, it's just that in the test we have not looked at it yet, it's just there waiting since every emission in our presenter is put in the turbine, which is more or less a channel with all the emissions, one by one as they come.

So this is great actually! This test caches the fact that if we come back to this screen and we have had an error state before, what we try is immediately set it to loading and we try to load the state again.

Now the question is if we want to keep this behavior or not. I'd expect if someone comes *back* into an error screen it's okay to implicitly trigger the reload for them and put them to the loading state, so this test will in fact keep this behavior here, so if we ever break this we will know by the test failing.

One thing I would absolutely recommend is to move the assertion to above the `addTerminatedInsurancesToResponse` to properly indicate that it is not the one which triggers this behavior, but it happens by itself!


Once again, amazing with how well you've picked up writing all these tests already, and that such behaviors start surfacing by adding them, I am very impressed, great job!

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailPresenterTest.kt
StylianosGakis: Same applies here as well as far as I can tell from the presenter code. Also make sure to move the is Loading check above the `addInsuranceWithNoTerminationDateToResponseTurbine` function here too 😊

StylianosGakis (approved):
Added the explanation for the loading "weirdness" you experienced, hope it's clear!

Great job! 🤩

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/terminatedcontracts/TerminatedContractsPresenterTest.kt
panasetskaya: Thanks! And thanks for the explanation, I somehow missed the fact that it's not only the initial state that is Loading originally, but also the first dataIteration = 0 that triggers the same Loading state.

But it's interesting how it works with turbine and all

PR #2108, opened 2024-04-26
Fix/payment details ripple

panasetskaya:
maybe overdid it with 1000.dp :)

### DIFF
--- a/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/details/PaymentDetailExpandableCard.kt
+++ b/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/details/PaymentDetailExpandableCard.kt
@@ -3,6 +3,8 @@ package com.hedvig.android.feature.payments.details
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.expandVertically
 import androidx.compose.animation.shrinkVertically
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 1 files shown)


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/details/PaymentDetailExpandableCard.kt
```diff
@@ -54,16 +58,24 @@ internal fun PaymentDetailExpandableCard(
...
     Column(
       modifier = Modifier
         .padding(horizontal = 16.dp, vertical = 12.dp)
-        .fillMaxWidth(),
+        .fillMaxWidth()
+        .clickable(
+          interactionSource = remember { MutableInteractionSource() },
+          indication = rememberRipple(
+            bounded = false,
+            radius = 1000.dp),
```
StylianosGakis, line 72: Yeah no problem with it being big if it does what we want.
Perhaps since it will be looking quite odd out of context write a little comment about it.
```suggestion
            // This fixes the problem of the ripple not properly resizing as the card expands
            radius = 1000.dp
          ),
```

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/details/PaymentDetailExpandableCard.kt
```diff
@@ -54,16 +58,24 @@ internal fun PaymentDetailExpandableCard(
 ) {
   val dateTimeFormatter = rememberHedvigMonthDateTimeFormatter()
   HedvigCard(
-    onClick = onClick,
+    onClick = null,
```
StylianosGakis, line 61: Maybe can just remove this one completely? I think we have a HedvigCard without a clickable lambda at all?

StylianosGakis (approved):

PR #2111, opened 2024-04-29
Home screen top bar icons

panasetskaya:
So about CrossSells - we use them in 2 places now (Insurances and Home). Makes sense to put the CrossSellsSection somewhere in one of ui modules, but neither of the ui modules are really suitable. And to create a separate module for only one screen also doesn't seem right somehow. So I've put them in ui-emergency, and renamed this module to "ui-screens" (there are 3 screens in there now: Emergency, Firstvet and CrossSells section). What do you think?


app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/bottomsheet/HedvigBottomSheet.kt
```diff
@@ -70,19 +73,23 @@ fun HedvigBottomSheet(
...
-          sheetState.hide()
-        }.invokeOnCompletion {
-          onDismissed()
-        }
-      },
-      modifier = Modifier.padding(start = 24.dp, end = 24.dp, bottom = 16.dp),
-    )
+    Column(
+      modifier = Modifier.verticalScroll(rememberScrollState()),
+    ) {
```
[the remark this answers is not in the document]
panasetskaya: Not with this one, I don't think so. Maybe for ModalBottomSheet somewhere.

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -200,14 +223,35 @@ private fun HomeScreen(
...
+                onClick = { navigateToFirstVet(sections) },
+              )
+            }
+          }
+          if (uiState.topBarActions.indexOf(action) != uiState.topBarActions.lastIndex) {
+            Spacer(modifier = Modifier.width(8.dp))
+          }
         }
+      }
+      if (uiState.topBarActions.contains(HomeTopBarAction.ChatAction)) {
```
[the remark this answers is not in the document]
panasetskaya: I still made a list of actions (spacers are still needed to be added properly), but now it's built in the destination, while the state carries only properties:)

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -229,32 +229,40 @@ private fun HomeScreen(
...
+              is HomeTopBarAction.FirstVetAction -> {
+                val sections = action.sections
+                ToolbarFirstVetIcon(
+                  onClick = { navigateToFirstVet(sections) },
+                )
+              }
+            }
+            if (actionsList.indexOf(action) != actionsList.lastIndex) {
+              Spacer(modifier = Modifier.width(8.dp))
             }
```
StylianosGakis: I've found it more convenient to do this check the other way around lately, by doing what is done here https://github.com/HedvigInsurance/android/blob/0ec4aaee4da0c6552868e06c462ab30c2536de5b/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountRow.kt#L39 of iterating over the list with an index, and if the index is not 0 to add the "in-between" item then.
When doing it on a for-loop you can do it this way https://github.com/HedvigInsurance/android/blob/638e6dc11eacef74ff4ed95815805c87d7d498a0/app/feature/feature-help-center/src/main/kotlin/com/hedvig/android/feature/help/center/ui/HelpCenterSectionWithClickableRows.kt#L43 (ignore the `> 0` there, I think `!= 0` is more readable), so the withIndex() function. The IDE actually helps you by suggesting it if you do Option Enter while hovering on the `in` keyword in the `for ... in ...` I think.

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -229,32 +229,40 @@ private fun HomeScreen(
...
-                crossSellsForBottomSheet = action.crossSells
-              },
-            )
+        val currentState = uiState as? HomeUiState.Success
+        if (currentState != null) {
+          val actionsList = buildList {
+            if (currentState.crossSellsAction != null) add(currentState.crossSellsAction)
+            if (currentState.firstVetAction != null) add(currentState.firstVetAction)
+            if (currentState.chatAction != null) add(currentState.chatAction)
+          }
```
StylianosGakis, line 238: Yeap that looks great 👍

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -310,10 +318,10 @@ private fun HomeScreenSuccess(
...
   Box(
     modifier = modifier
-      .fillMaxSize()
-      .onSizeChanged { fullScreenSize = it }
-      .pullRefresh(pullRefreshState)
-      .verticalScroll(rememberScrollState()),
+        .fillMaxSize()
+        .onSizeChanged { fullScreenSize = it }
+        .pullRefresh(pullRefreshState)
+        .verticalScroll(rememberScrollState()),
```
StylianosGakis: Might just need to re-run ktlintFormat to fix all these

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomePresenter.kt
```diff
@@ -175,23 +178,18 @@ private data class SuccessData(
...
-          )
-        }
-        if (homeData.showChatIcon) add(HomeTopBarAction.ChatAction)
-      }
-
+      val crossSellsAction = if (homeData.crossSells.isNotEmpty())
+        HomeTopBarAction.CrossSellsAction(homeData.crossSells) else null
+      val chatAction = if (homeData.showChatIcon) HomeTopBarAction.ChatAction else null
+      val firstVetAction = if (homeData.firstVetSections.isNotEmpty())
+        HomeTopBarAction.FirstVetAction(homeData.firstVetSections) else null
```
StylianosGakis: Great!

StylianosGakis (approved):
🔥

PR #2115, opened 2024-05-06
delete not needed baseline file

panasetskaya:


StylianosGakis (approved):
It's my fault

image

Perhaps some day I will spend the time to also cleanup old files from there too :D Some day

PR #2119, opened 2024-05-11
GEN-2067: login after process death

StylianosGakis:


panasetskaya:
I don't fully know how Auth goes in general, but this SavedStateHandle seems like a interesting thing for a very narrow range of usecases. I tried to think of another cases where we could use it in the project, but this seems the only one where we need to handle process death?

StylianosGakis:
In general, the way we handle process death is that we rely on the fact that the navigation library retains the backstack and the parameters passed to it, so when we come back from process death we end up in the same place, and since we typically pass the right information to each screen so that it knows how to fetch the right information from GQL, it just does so.

There are right now with this PR 2 places where we are doing something a bit more special. In the chat screen, we make sure to properly store the ID of a photo we may be taking, so that when we come back to the killed app, the ID is retained (through `rememberSaveable` there actually).

And in this case it's another one where we can't rely on just the destination, we want to retain the ID of the ongoing login attempt. And in this case, since we get this in the VM and not the UI itself, we can't go with a rememberSaveable (you can't do that inside the molecule presenters), so we gotta use the SavedStateHandle.
More info on this btw here https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-savedstate

StylianosGakis:
If you find more places where you feel like using the SSH to save some data on process death of course bring them up, it's a tool we haven't made use of too much, but it's quite useful.

PR #2121, opened 2024-05-13
Fix lokalise translations

StylianosGakis:
This also stops downloading the FR translations, and removes the workaround we had in there.
The important change comes from the POST body we send with `put("escape_percent", true)` here, which is described here https://developers.lokalise.com/reference/download-files. It turns `[%]` from lokalise translations to `%%` in our xml result

### DIFF
--- a/lokalise-gradle-plugin/lokalise/src/main/kotlin/com/hedvig/android/lokalise/task/DownloadStringsTask.kt
+++ b/lokalise-gradle-plugin/lokalise/src/main/kotlin/com/hedvig/android/lokalise/task/DownloadStringsTask.kt
@@ -5,6 +5,8 @@ import java.io.File
 import java.net.URI
 import javax.inject.Inject
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.add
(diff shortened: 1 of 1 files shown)


lokalise-gradle-plugin/lokalise/src/main/kotlin/com/hedvig/android/lokalise/task/DownloadStringsTask.kt
```diff
@@ -132,6 +111,30 @@ abstract class DownloadStringsTask @Inject constructor(
...
+      .filter { it.name == """strings.xml""" }
+      .toList()
+
+    val fileSystem = FileSystem.SYSTEM
+    for (currentLanguageTranslationXmlFilePath in allTranslationXmlFilePaths) {
+      val content = fileSystem.read(currentLanguageTranslationXmlFilePath) { readUtf8() }
+      val updatedContent = content
+        .replace(
+          oldValue = """%%""",
+          newValue = """\u0025""",
```
panasetskaya, line 130: 🔥

panasetskaya (approved):

PR #2134, opened 2024-06-04
popUp termination flow backstack when redirecting to moving flow

panasetskaya:


app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -155,7 +155,15 @@ internal fun HedvigNavHost(
           },
           navigateToMovingFlow = { backStackEntry ->
             with(navigator) {
-              backStackEntry.navigate(AppDestination.ChangeAddress)
+              val options = navOptions {
```
StylianosGakis, line 158: In places where the options are really complex or are used in many cases, we do this https://github.com/HedvigInsurance/android/blob/78a70c0c5217575ed261965784f3a7da5489267e/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt#L183-L196
But I do think that in simpler cases like this you can omit the `option` variable and just do it inline here, wdyt?

StylianosGakis (approved):
🌟

PR #2137, opened 2024-06-10
fix incorrect ssn crash

panasetskaya:
The bug was: incorrectly written SSN with impossible date (like 31st of April), which of course empty info when fetching personal info, then we try to convert it to the date, but it crashes bc of the date being impossible.

### DIFF
--- a/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
+++ b/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
@@ -51,7 +51,7 @@ private fun convertSsnToBirthDateOrNull(ssn: String): LocalDate? {
   if (stringMonth !in 1..12) return null
   val result = try {
     LocalDate(year = stringYear, month = Month(stringMonth), dayOfMonth = stringDate)
-  } catch (e: Exception) {
+  } catch (e: IllegalArgumentException) {
     null
   }
   return result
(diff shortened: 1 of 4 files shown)


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
```diff
@@ -49,5 +49,10 @@ private fun convertSsnToBirthDateOrNull(ssn: String): LocalDate? {
   val stringDate = ssn.substring(6, 8).toIntOrNull()
   if (stringYear == null || stringMonth == null || stringDate == null) return null
   if (stringMonth !in 1..12) return null
-  return LocalDate(year = stringYear, month = Month(stringMonth), dayOfMonth = stringDate)
+  val result = try {
+    LocalDate(year = stringYear, month = Month(stringMonth), dayOfMonth = stringDate)
+  } catch (e: Exception) {
```
StylianosGakis, line 54:
```suggestion
  } catch (e: IllegalArgumentException) {
```

app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/FetchCoInsuredPersonalInformationUseCase.kt
```diff
@@ -49,5 +49,10 @@ private fun convertSsnToBirthDateOrNull(ssn: String): LocalDate? {
   val stringDate = ssn.substring(6, 8).toIntOrNull()
   if (stringYear == null || stringMonth == null || stringDate == null) return null
   if (stringMonth !in 1..12) return null
-  return LocalDate(year = stringYear, month = Month(stringMonth), dayOfMonth = stringDate)
+  val result = try {
```
StylianosGakis, line 52: Perhaps along with this change we can look if we can also get rid of that `formatShortSsn` function as discussed here https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1718007153791809 .
If we do not use it anywhere else of course and what I am saying there makes sense.

StylianosGakis (approved):

PR #2138, opened 2024-06-10
GEN-2041 Refactor/change address vm to molecule

panasetskaya:
- [x] tests;
- [x] some annoying UI un-smoothness with extraBuilding bottomSheet and keyboard focus, both were there before too, but would be good to get rid of them
- [x] checking about maxCoInsured number - is it 5, why is it 5 and should we add(1) to it

So it was one viewModel here, I divided it in 4, changed navParameters a bit, and wrote tests.


panasetskaya:
So I fixed the bottom padding, but cannot make the bottom sheet adjust to the keyboard (it goes up to the bottom of textInput, not to the bottom of the whole column). Maybe it because of the bottom sheet in combination with vertical scroll - not sure. Maybe I'm missing something.

So the dismissing keyboard with bottom sheet problem remains here:)

panasetskaya:
As one solution for dismissing the bottomSheet - maybe get rid of bottomSheet here if we won't be able to make it work? Make a separate screen, like we do e.g. for full screen free text in survey?

StylianosGakis:
> As one solution for dismissing the bottomSheet - maybe get rid of bottomSheet here if we won't be able to make it work? Make a separate screen, like we do e.g. for full screen free text in survey?

Honestly feel free to do that if you feel like it's gonna improve the experience. Or you could also try to give the modalsheet library a shot and if it proves to work quite well (especially around the keyboard stuff) then you could also do that instead. For now as you say this is fine, especially considering the bug was there before anyway.

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/enternewaddress/EnterNewAddressDestination.kt
```diff
@@ -147,19 +169,28 @@ private fun ChangeAddressEnterNewAddressScreen(
...
     uiState.datePickerUiState?.let {
       Spacer(modifier = Modifier.height(8.dp))
       MovingDateButton(
-        onDateSelected = { onMoveDateSelected(it) },
+        onDateSelected = {
+          onMoveDateSelected(it)
+          focusManager.clearFocus()
+        },
+        // todo: Scaffold's modifier does not clear focus in this case,
+        //  nor does MovingDateButton modifier, so it's annoying when keyboard opens again
```
StylianosGakis, line 193: Sounds good to add the clearFocus() here then 👍

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/enternewaddress/EnterNewAddressViewModel.kt
```diff
@@ -0,0 +1,282 @@
...
+      movingDate = movingDate.copy(
+        errorMessageRes = if (!movingDate.isPresent) {
+          hedvig.resources.R.string.CHANGE_ADDRESS_MOVING_DATE_ERROR
+        } else {
+          null
+        },
+      ),
+      numberInsured = numberInsured.copy(
+        errorMessageRes = if (!numberInsured.isPresent || isNumberInsuredTooLow(numberInsured.input!!.toIntOrNull())) {
+          hedvig.resources.R.string.CHANGE_ADDRESS_CO_INSURED_ERROR // todo: I think the copy may be wrong here. It is the number of all the insured (incl.the main insured, not only co-insured)
```
StylianosGakis, line 201: Can we ask iOS too what they do?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/offer/ChangeAddressOfferViewModel.kt
```diff
@@ -0,0 +1,152 @@
...
+    var currentState by remember { mutableStateOf(lastState) }
+
+    var dataLoadIteration by remember { mutableIntStateOf(0) }
+
+    var quotesLoadIteration by remember { mutableIntStateOf(0) }
+
+    CollectEvents { event ->
+      when (event) {
+        is ConfirmMove -> {
+          dataLoadIteration++
```
StylianosGakis, line 56: Since this is a bit more than just refreshing the most up-to-date data to show in the UI. Could we name it something a bit more descriptive. Like the confirmMoveIteration or something like that?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/selecthousingtype/SelectHousingTypeViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+          currentState = currentState.copy(housingType = ValidatedInput(input = event.housingType))
+        }
+
+        SubmitHousingType -> {
+          if (!currentState.housingType.isPresent) {
+            currentState = currentState.copy(
+              errorMessageRes = R.string.CHANGE_ADDRESS_HOUSING_TYPE_ERROR,
+            )
+          } else {
+            dataLoadIteration++
```
StylianosGakis, line 60: Same concept here, since this is in fact submitting something, I would prefer to have a name that describes that

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/selecthousingtype/SelectHousingTypeViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+                currentState.housingType.input != HousingType.VILLA,
+              maxSquareMeters = when (currentState.housingType.input) {
+                HousingType.APARTMENT_RENT,
+                HousingType.APARTMENT_OWN,
+                -> moveIntent.maxApartmentSquareMeters
+
+                HousingType.VILLA -> moveIntent.maxHouseSquareMeters
+                null -> null
+              },
+              maxNumberCoInsured = when (currentState.housingType.input) { // todo: is this number of coInsured or just Insured (incl.main insured?) bc later it's used as if it is the later one.
```
StylianosGakis, line 96: I don't know myself either actually, probably worth running with a sample memeber to see what it really is to make sure we're not misusing it

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingBottomSheet.kt
```diff
@@ -56,6 +62,8 @@ internal fun ExtraBuildingBottomSheet(
     shape = MaterialTheme.shapes.squircleLargeTop,
     sheetState = sheetState,
     tonalElevation = 0.dp,
+    windowInsets = WindowInsets.ime,
```
StylianosGakis, line 65: For this btw, could a solution instead be that the content itself is the one that takes the IME insets?

Because now I think that the entire sheet will have the IME insets, and the content will *also* receive them, getting them double applied.

So the sheet probably is offset up by the IME height, and then inside the content itself there is also a space at the bottom equal to the IME height, am I correct in my assumption here?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/offer/QuoteCard.kt
```diff
@@ -127,9 +127,9 @@ private fun QuoteDetailsAndPrice(isExpanded: Boolean, quote: MoveQuote) {
         Spacer(Modifier.width(8.dp))
         val angle = animateFloatAsState(
           targetValue = if (isExpanded) {
-            0f
-          } else {
             -180f
+          } else {
+            0f
           },
```
StylianosGakis, line 133: Was it flipped the wrong way? 😅

I think we want this to point down when the content is shrinked, and to point up when it's expanded. Sort of to have the arrow point towards where the content will go when clicked.

app/feature/feature-changeaddress/src/test/kotlin/com/hedvig/android/feature/changeaddress/destination/enternewaddress/EnterNewAddressPresenterTest.kt
```diff
@@ -0,0 +1,352 @@
...
+import com.hedvig.android.language.test.FakeLanguageService
+import com.hedvig.android.logger.TestLogcatLoggingRule
+import com.hedvig.android.molecule.test.test
+import java.util.Locale
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.LocalDate
+import org.junit.Rule
+import org.junit.Test
+
+class EnterNewAddressPresenterTest {
```
StylianosGakis, line 22: 🌟

StylianosGakis (approved):
A herculean effort, I am very impressed, great job!

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/offer/QuoteCard.kt
panasetskaya: Yep, was flipped and felt strange:)

panasetskaya: I was also thinking of adding a gigantic ripple that covers all the card there (we had it in payments somewhere - smaller round ripple in the center of card looking a bit strange), but it's a smaller thing

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingBottomSheet.kt
panasetskaya: I'll check this out. I tried to add ime to the content in one of my attempts, but maybe it was not done correctly

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/enternewaddress/EnterNewAddressViewModel.kt
panasetskaya: They actually have a completely different field here:) which I think is better (easier to understand)

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/selecthousingtype/SelectHousingTypeViewModel.kt
panasetskaya: I did, it seems to be overall for main+co-insured. 
Also, Sladan says in iOs they can go as high as up to 6 co-insured. I wonder how it corresponds with this maxNumber

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/enternewaddress/EnterNewAddressViewModel.kt
StylianosGakis: Do you we don't have this at all in Android I think right?
New design I assume has this component?
Also, do we have a new design for moving flow anyway?
Maybe this could be fixed while those are being worked on so we don't make smth now only to be replaced soon. Otherwise if new design does not have moving flow, perhaps we should just make a quick impl for this which fixes it.

panasetskaya: Kinda did a quick impl without checking:))) but we do in fact have this component in the new design: https://www.figma.com/design/5kmmDdh6StpXzbEfr7WevV/Hedvig-UI-Kit?node-id=15724-23348&t=uKXQD6goM9XLlf5C-4

panasetskaya: But I say it's a whole new component, so would take some time to implement. Maybe a separate PR for this?

PR #2145, opened 2024-06-14
New design/radio button

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,454 @@
...
+object RadioOptionDefaults {
+  internal val radioOptionStyle: RadioOptionStyle = Default
+  internal val radioOptionSize: RadioOptionSize = RadioOptionSize.Large
+
+  sealed interface RadioOptionStyle {
+    data object Default : RadioOptionStyle
+
+    data class Label(val labelText: String) : RadioOptionStyle
+
+    data class Icon(val iconResource: IconResource) : RadioOptionStyle // we need another solution here. Content?
```
StylianosGakis, line 206: Another solution for what, if there is a scenario where we need both an icon and a label?

I think there's two approaches to consider here.
If the component really does not mix and match those this sealed interface should be a simple API which lets us easily just do `RadioOptionStyle.` and get autocompletion for all the relevant styles.

If we figure out that we want both, we could have the icon not be part of RadioOptionStyle, but be a normal parameter in the public `RadioOption` function. Like `iconResource: IconResource? = null`.
And then on the caller, if you pass nothing you get nothing, if you pass non-null the icon should show and the UI should adjust accordingly.

It's a question of how it's gonna be used really to know what the best option is. We can also redesign that later if we find that we need that shape instead.

StylianosGakis: Oh I do see now that the left-aligned one would simply not work with an Icon really. Maybe we should not allow the icon in all scenarios then.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/RadioOptionTokens.kt
```diff
@@ -0,0 +1,75 @@
...
+      else -> TopPadding
+    }
+  }
+
+  fun bottomPadding(style: RadioOptionStyle): Dp {
+    return when (style) {
+      is Label -> LabelBottomPadding
+      else -> BottomPadding
+    }
+  }
```
StylianosGakis, line 41: I have a hunch that the tokens here shouldn't really need to know anything about the style and should hopefully just be raw tokens ready to be used in whatever way.
At the same time we don't want to repeat ourselves over at the place where we call these.
I think I had the same scenario over here https://github.com/HedvigInsurance/android/blob/ee04408d62eba72d61de7391b832fcb4d2513fab/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigTextField.kt#L347-L361 where I added this piece of logic inside the style itself, and the implementation of each size just knew about both LabelTopPadding and TopPadding and that function in there picked the right one for all three sizes.

StylianosGakis (approved):
⭐️⭐️⭐️
Amazing job, got some comments and a bunch of them are discussion starters, if you wanna expand on some of them let's do that over at slack.
The overall approach you've gone with is perfect I'd say, I feel like doing things this way we'll get a really good end result for our DS!

Lmk how doing this PR felt like over all, if anything was too confusing, if anything felt like it needed improvement on the approach etc.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,454 @@
...
+  }
+}
+
+private class RadioOptionStyleProvider :
+  CollectionPreviewParameterProvider<RadioOptionStyle>(
+    listOf(
+      Default,
+      Label("Label"),
+      RadioOptionStyle.Icon(IconResource.Painter(hedvig.resources.R.drawable.pillow_hedvig)),
+      //   RadioOptionStyle.Icon(IconResource.Vector(HedvigIcons.FlagSweden)),
```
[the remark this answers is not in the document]
panasetskaya: Too many, I think, 4 is the limit :)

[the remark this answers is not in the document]
panasetskaya: Yes, could be - it often acts out with previews in weird ways

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,454 @@
...
+      }
+    }
+  }
+}
+
+sealed interface IconResource {
+  data class Vector(val imageVector: ImageVector) : IconResource
+
+  data class Painter(val painterResId: Int) : IconResource
+}
```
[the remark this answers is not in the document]
panasetskaya: I don't think m3 uses this, no - just something I thought would be nice to not to limit ourselves with ImageVectors. I'll move it!

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
panasetskaya: I say let's keep it like this here, and if the need for this combination arises, later add 
`data class LabelAndIcon(val labelText: String, val iconResource: IconResource) : RadioOptionStyle`

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,454 @@
...
+    }
+  }
+}
+
+@Composable
+private fun CheckItemAnimation(state: RadioOptionState, content: @Composable (state: RadioOptionState) -> Unit) {
+  val selectedTransition = updateTransition(state)
+  selectedTransition.AnimatedContent(
+    transitionSpec = {
+      fadeIn(tween(durationMillis = 200, easing = EaseInOut)) togetherWith fadeOut(tween(200, easing = EaseInOut))
```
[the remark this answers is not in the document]
panasetskaya: made TweenAnimationTokens for that :
`
internal abstract class TweenAnimationTokens {

  abstract val durationMillis: Int
  abstract val easing: Easing

  data object FastAnimationTokens: TweenAnimationTokens() {
    override val durationMillis: Int = 200
    override val easing: Easing = EaseInOut
  }
}`

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,562 @@
...
+                when (lockedState) {
+                  Locked -> Modifier.border(8.dp, radioOptionColors.disabledIndicatorColor, CircleShape)
+                  NotLocked -> Modifier.border(8.dp, radioOptionColors.chosenIndicatorColor, CircleShape)
+                }
+              }
+
+              NotChosen -> {
+                Modifier.border(2.dp, radioOptionColors.notChosenIndicatorColor, CircleShape)
+                // disabled color is actually brighter than notChosenIndicatorColor,
+                // makes no sense here, so I left only one
```
StylianosGakis, line 237: Did the design purposefully chose it to be that way?
Does it look confusing if we have a disabled and a non-disabled but unselected option side by side?
Could it be a bit confusing if perhaps in the future the color for disabled is adjusted, but the not-selected is not, and we make the change in the tokens but the implementation changes for the not-selected option too, which might be a bit confusing say 1 year in the future.

Or is the part that does not add up something that should be brought up with design perhaps?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,562 @@
...
+              }
+            },
+          ),
+      )
+    }
+  }
+}
+
+@Composable
+internal fun CheckItemAnimation(state: ChosenState, content: @Composable (state: ChosenState) -> Unit) {
```
StylianosGakis, line 247: Is this internal function used in other files too or could it be private?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Checkbox.kt
```diff
@@ -0,0 +1,456 @@
...
+import com.hedvig.android.design.system.hedvig.tokens.CheckboxColorTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeCheckboxTokens.LargeSizeCheckboxTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeCheckboxTokens.MediumSizeCheckboxTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeCheckboxTokens.SmallSizeCheckboxTokens
+
+@Composable
+fun Checkbox(
+  data: OptionData,
+  checkboxStyle: CheckboxStyle,
+  groupLockedState: LockedState,
```
StylianosGakis, line 58: This is probably leftover from copy-pasting the API from the group one. Just `lockedState: LockedState` might be what we want here just like you got below

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+          groupLockedState = groupLockedState,
+          onOptionClick = {
+            onOptionClick(i.id)
+          },
+          modifier = Modifier
+            .weight(1f)
+            .width(Min)
+            .padding(itemPadding),
+        )
+      }
```
StylianosGakis, line 169: I think to avoid having to double call this RadioOption twice, you can make use of the [chunked](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/chunked.html) function to chunck the list in groups of two, and inside the lambda, if the list is actually 1 item then add the spacer instead. You can see it being used here https://github.com/HedvigInsurance/android/blob/c92623cf36c37592aee1772f78b974e8b9303ab9/app/core/core-ui/src/main/kotlin/com/hedvig/android/core/ui/grid/HedvigGrid.kt#L67 too.
I think it would make this a bit easier to write and read perhaps

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+  modifier: Modifier = Modifier,
+) {
+  FlowRow(
+    modifier,
+    maxItemsInEachRow = 2,
+    overflow = Companion.Visible,
+    verticalArrangement = Arrangement.spacedBy(4.dp, Alignment.CenterVertically),
+  ) {
+    for (i in data) {
+      val itemPadding = if (data.indexOf(i) % 2 != 0) {
```
StylianosGakis, line 133: Since we seem to do `indexOf()` every time inside the loop, and often more than once too, I'd suggest we turn this into a loop where we have the index already like this https://github.com/HedvigInsurance/android/blob/12a8a8a3c450bc1274782baa551fc3b5ff20d4b2/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurance/InsuranceDestination.kt#L259
The IDE provides an auto-complete for this if you "option + enter" while having the cursor on `in` or perhaps `data` here I think.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+        PaddingValues()
+      } else {
+        PaddingValues(end = 4.dp)
```
StylianosGakis, line 136: A bit confused here how there is a need for 4dp on the right only for the items that are on the right side in this flowrow, do the left items get this from somewhere else that I am not seeing?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+    verticalAlignment = Alignment.CenterVertically,
+    modifier = modifier
+      .clickable(
+        enabled = enabled,
+        role = Role.RadioButton,
+        interactionSource = interactionSource,
+        indication = ripple(
+          bounded = false,
+          radius = 50.dp,
+          // todo: pls check out this ripple. I like this one, but maybe it's too much
```
StylianosGakis, line 262: We can take a look at this within some test session etc. to see what others think too

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+    color = radioOptionColors.containerColor,
+  ) {
+    Column(
+      modifier,
+    ) {
+      val labelTextColor = radioOptionColors.labelTextColor(groupLockedState)
+      HedvigText(
+        modifier = Modifier.padding(contentPaddingValues),
+        text = radioGroupStyle.groupLabelText,
+        style = MediumSizeRadioOptionTokens.LabelTextFont.value, // same for all sizes in figma
```
StylianosGakis, line 336: I think if it happens not to be too much of a hassle to add it inside the RadioGroupSize object, it can be there so that if the impl changes in the future and we only go and edit the tokens we don't get behavior that we do not expect by having only the medium tokens drive the value of this text.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
```diff
@@ -0,0 +1,905 @@
...
+              onClick = {
+                onOptionClick(radioOptionData.id)
+              },
+            )
+          val radioOptionStyle = when (radioGroupStyle) {
+            is VerticalWithGroupLabel.Default -> Default
+            is VerticalWithGroupLabel.Icon -> RadioOptionStyle.Icon(
+              radioOptionData.iconResource ?: Painter(R.drawable.pillow_hedvig),
+            ) // todo: how is that for placeholder
+            is VerticalWithGroupLabel.Label -> Label(radioOptionData.labelText ?: "") // todo: no placeholder
```
StylianosGakis, line 359: In which scenarios would it be possible to be in the label style but have no labelText? Would it make sense for that to be impossible in the first place?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/CheckboxTokens.kt
```diff
@@ -0,0 +1,129 @@
...
+import com.hedvig.android.design.system.hedvig.CheckboxDefaults.CheckboxStyle
+
+internal object CheckboxColorTokens {
+  val ContainerColor = ColorSchemeKeyTokens.SurfacePrimary
+  val OptionTextColor = ColorSchemeKeyTokens.TextPrimary
+  val LabelTextColor = ColorSchemeKeyTokens.TextSecondaryTranslucent
+  val DisabledOptionTextColor = ColorSchemeKeyTokens.TextDisabledTranslucent
+  val DisabledLabelTextColor = ColorSchemeKeyTokens.TextDisabledTranslucent
+  val ChosenIndicatorColor = ColorSchemeKeyTokens.SignalGreenElement
+  val NotChosenIndicatorColor = ColorSchemeKeyTokens.FillDisabled // todo: if i put borderSecondary here as it is in figma, it's almost invisible
```
StylianosGakis, line 15: Aha the problem then is that the contrast with the background is just so low that it's unreasonably hard to see? Some of our "disabled" colors are like that, but is this even more so than the others?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
```diff
@@ -140,43 +137,18 @@ private fun ChooseInsuranceToTerminateScreen(
...
         }
-        for (insurance in uiState.insuranceList) {
-          HedvigCard(
-            onClick = { selectInsurance(insurance) },
-            colors = CardDefaults.outlinedCardColors(
-              containerColor = MaterialTheme.colorScheme.surface,
-            ),
+        val radioOptionData = uiState.insuranceList.toRadioOptionDataList(uiState.selectedInsurance?.id)
+        com.hedvig.android.design.system.hedvig.HedvigTheme {
+          // todo: where do we apply the theme now that we're still on both old and new theme?
```
StylianosGakis, line 142: So at the top-level where we provide it for the entire app, we need to provide both themes.
Take a look at this https://github.com/HedvigInsurance/android/blob/5001fea1697470018ffd9a7014c3a8a24acc14b7/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigApp.kt#L131-L139 I seem to already have done that

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
```diff
@@ -200,6 +172,21 @@ private fun ChooseInsuranceToTerminateScreen(
...
+      OptionData(
+        id = i.id,
+        optionText = i.displayName,
+        chosenState = if (selectedInsuranceId == i.id) Chosen else NotChosen,
+        labelText = i.contractExposure,
+      ),
+    )
+  }
+  return result
+}
```
StylianosGakis, line 188:
```suggestion
private fun List<TerminatableInsurance>.toRadioOptionDataList(selectedInsuranceId: String?): List<OptionData> {
  return this.map { i ->
    OptionData(
      id = i.id,
      optionText = i.displayName,
      chosenState = if (selectedInsuranceId == i.id) Chosen else NotChosen,
      labelText = i.contractExposure,
    )
  }
}
```

I am a `mutableList` hater, I must admit 😅

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/survey/TerminationSurveyDestination.kt
```diff
@@ -183,34 +184,17 @@ private fun TerminationSurveyScreen(
...
-              containerColor = MaterialTheme.colorScheme.surface,
-            ),
-            modifier = Modifier
-              .fillMaxWidth()
-              .padding(horizontal = 16.dp),
-          ) {
-            Row(
-              verticalAlignment = Alignment.CenterVertically,
+          com.hedvig.android.design.system.hedvig.HedvigTheme {
+            // todo: where do we apply the theme now that we're still on both old and new theme?
```
StylianosGakis, line 188: Same as above, all these can be removed since we got the top-level theme set at the application level.
For previews of course it means we gotta provide the right theme, especially since in the DS module itself we only get access to one of the two themes and it makes the live template a bit trickier to use for now

StylianosGakis (approved):
Looking good, left some comments. Lmk if I missed smth that you'd like me to not miss :D

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
```diff
@@ -0,0 +1,562 @@
...
+import com.hedvig.android.design.system.hedvig.RadioOptionDefaults.RadioOptionStyle.LeftAligned
+import com.hedvig.android.design.system.hedvig.icon.HedvigIcons
+import com.hedvig.android.design.system.hedvig.icon.flag.FlagSweden
+import com.hedvig.android.design.system.hedvig.tokens.RadioOptionColorTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeRadioOptionTokens.LargeSizeRadioOptionTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeRadioOptionTokens.MediumSizeRadioOptionTokens
+import com.hedvig.android.design.system.hedvig.tokens.SizeRadioOptionTokens.SmallSizeRadioOptionTokens
+import com.hedvig.android.design.system.hedvig.tokens.TweenAnimationTokens
+
+data class OptionData(
```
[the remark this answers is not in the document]
panasetskaya: Yep, will rename it back :)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
panasetskaya: Yep, it is used in checkbox too

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
panasetskaya: Ah, alright! Maybe it was because I've seen it in HedvigPreview, which is the old theme, and it was not behaving well there

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/CheckboxTokens.kt
panasetskaya: I think this is something I better show you. It's a bit confusing for me here

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioOption.kt
panasetskaya: _Does it look confusing if we have a disabled and a non-disabled but unselected option side by side?_ - yes, basically this. Maybe it won't ever be the case, or very rarely, so maybe I would just put notChosenIndicatorColor there

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
panasetskaya: it's actually the other way round, with items on the left (index 0, 2 etc) they need a space between it and the right item, and items on the right (index 1,3) are at the end of two-items row, so they don't need extra padding there

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/RadioGroup.kt
panasetskaya: I'll try one thing!

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Checkbox.kt
```diff
@@ -0,0 +1,460 @@
...
+          if (onCheckedChange != null) {
+            onCheckedChange()
+          }
+        },
+      contentAlignment = Alignment.Center,
+    ) {
+      if ((currentState == Chosen)) {
+        Icon(HedvigIcons.Checkmark, contentDescription = null, tint = checkIconColor)
+        // todo: I am not entirely sure this is the right Checkmark icon. It looks differently in figma
+      }
```
panasetskaya, line 251: for Stelios

PR #2151, opened 2024-06-27
Stop querying for unused status

StylianosGakis:


StylianosGakis:
Holy formatting, what happened here 🥸

PR #2153, opened 2024-06-27
Bump ktlint+kotlinter

StylianosGakis:
Okay not *that* bad after disabling 
```
# https://pinterest.github.io/ktlint/latest/rules/standard/#class-signature
ktlint_standard_class-signature = disabled
# https://pinterest.github.io/ktlint/latest/rules/standard/#chain-method-continuation
ktlint_standard_chain-method-continuation = disabled
```


StylianosGakis:
Ah, build is failing, will look at it some time, maybe it was not that easy

PR #2155, opened 2024-07-08
GEN-2431: Update termination survye

hugokallstrom:
* Query description and button text for the suggestion card 
* Disable continue button if selected option contains a suggestion

Screenshot_20240708_135159


StylianosGakis:
🔥

PR #2158, opened 2024-07-15
Remove chat reply in notification

hugokallstrom:


StylianosGakis (approved):
Perfect, thanks a lot!

PR might fail on lint or whatever due to some other things I am trying to fix RN. Let me just fix those first and then I'll merge so you don't have to worry about it

PR #2160, opened 2024-07-15
Show button text for important messages dynamically

hugokallstrom:


StylianosGakis:
Already merged here https://github.com/HedvigInsurance/android/commit/de491864c5dd5df31efb67e52fab6ac7bd1fd9ed GH does not recognize it because I think I rebased by accident

PR #2167, opened 2024-08-05
GEN-2353 New design BottomSheet

panasetskaya:
I think now bottomSheet is working fine with onBack, thanks to you talking to the lib's dev:)  pls take a look.
I also added HintArrowDown if the bottom sheet content is too long to scroll to see the bottom - works once. Just for fun, will remove it, if you don't like it there.

I used this new HedvigBottomSheet for ChangeAddress feature's Villa ExtraBuilding bottom sheet. Should I change it back or can we leave the new one there?


app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingTypeContainer.kt
```diff
@@ -39,13 +32,15 @@ internal fun ExtraBuildingTypeContainer(
...
-    Column(Modifier.nestedScroll(nestedScrollConnection).verticalScroll(rememberScrollState())) {
+//    val nestedScrollConnection = remember {
+//      object : NestedScrollConnection {
+//        // to intercept scrolling gesture and do not close the sheet while scrolling the list
+//        override fun onPostScroll(consumed: Offset, available: Offset, source: NestedScrollSource) = available
+//      }
+//    }
+    Column(
+      // Modifier.nestedScroll(nestedScrollConnection).verticalScroll(rememberScrollState())
+    ) {
```
StylianosGakis, line 43: We can just remove this now since it works well right?

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/destination/entervillainfo/EnterVillaInformationDestination.kt
```diff
@@ -173,12 +174,14 @@ private fun ChangeAddressEnterVillaInformationScreen(
...
-      onClick = { onIsSubletSelected(!uiState.isSublet.input) },
-    )
+    Row(modifier = Modifier.padding(horizontal = 16.dp)) {
+      ChangeAddressSwitch(
+        label = stringResource(id = R.string.CHANGE_ADDRESS_SUBLET_LABEL),
+        checked = uiState.isSublet.input,
+        onCheckedChange = onIsSubletSelected,
+        onClick = { onIsSubletSelected(!uiState.isSublet.input) },
+      )
+    }
```
StylianosGakis, line 184: Since this only has one item inside the row, could the change here just be that we add the padding modifier on the `ChangeAddressSwitch` and we do not introduce a wrapping Row?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+        shape = ShapeDefaults.CornerSmall,
+        color = bottomSheetColors.chipColor,
+      ).clip(ShapeDefaults.CornerSmall),
+  ) {}
+}
+
+@Composable
+private fun BoxScope.HintArrowDown(onClick: () -> Unit) {
+  val infiniteTransition = rememberInfiniteTransition(label = "arrowDownAlpha")
+  val arrowAlpha by infiniteTransition.animateFloat(
```
StylianosGakis, line 137: I like the idea, and that it only shows when there is in fact something to scroll to.
I think it would be nicer without the pulsating alpha actually, we could just keep it at 100% alpha all the time and instead only fade it in/out when it shows/does not show, possibly with a CrossFade where the if statement is on line 111.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+        HintArrowDown {
+          scrollDown = true
+        }
+      }
+    }
+  }
+}
+
+@Composable
+private fun ColumnScope.LittleUpperChip() {
```
StylianosGakis, line 121: Same concept here. The CenterHorizontally part can come from the parent, and this composable can take in a `modifier` and not be an extension on `ColumnScope`.
The width/height etc can stay in here since this composable should just render the handle itself standalone without having to know how the parent will use it.

StylianosGakis, line 121: https://m3.material.io/components/bottom-sheets/overview#2cce5bae-eb83-40b0-8e52-5d0cfaa9b795
They call this a `DragHandle` I think in the material docs, we can call it the same too

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+        HedvigButton(
+          onClick = onBottomButtonClick,
+          text = bottomButtonText,
+          enabled = true,
+          buttonStyle = Ghost,
+          modifier = Modifier.fillMaxWidth(),
+        )
+        Spacer(modifier = Modifier.height(32.dp))
+      }
+      if (scrollState.canScrollForward && !scrollDown) {
```
StylianosGakis, line 111: Here this could be a CrossFade and that'd be good enough as far as animations go imo, as I mention further below as well.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+    animationSpec = infiniteRepeatable(
+      animation = tween(
+        durationMillis = 2000,
+        easing = LinearEasing,
+      ),
+      repeatMode = RepeatMode.Reverse,
+    ),
+    label = "arrowDownAlpha animation",
   )
+  Row(Modifier.align(Alignment.BottomEnd).padding(horizontal = 32.dp, vertical = 16.dp)) {
```
StylianosGakis, line 149: This Row could also just not exist I think.
The Icon composable itself can take in the modifier which we can pass into HintArrowDown.

This would also allow this to not have to be an extension on BoxScope.

I like to think of it as that the HintArrowDown composable is responsible for drawing the hint arrow down on the screen, and does not concern itself on where it may be placed.
The fact that it's placed on the bottom right and padded by some value is more of a concern that the parent has

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+    Icon(
```
StylianosGakis, line 150: I have a feeling that I'd like for such an arrow to also have a little bit of a circle background behind it to be a bit more visible.
Could we take a look at how other apps do this perhaps?
At minimum I'd just add a circle behind it that is the opposite color
Also probably good to make it a
```
IconButton(onClick = onClick) {
  Icon()
}
```
To have it be the right touch size for sure.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -25,38 +56,111 @@ import eu.wewox.modalsheet.ModalSheet
...
+      scrollState.animateScrollTo(scrollState.maxValue)
+    }
+  }
   val isImeVisible = WindowInsets.isImeVisible
   val defaultPadding = if (isImeVisible) {
     WindowInsets.ime.asPaddingValues()
   } else {
     WindowInsets.safeDrawing
-      .only(WindowInsetsSides.Bottom).asPaddingValues()
+      .only(WindowInsetsSides.Bottom + WindowInsetsSides.Top).asPaddingValues()
```
StylianosGakis, line 78: If you are scrolled down, and you focus on the "Size" entry (in extra buildings screen) you will notice that now this makes the content render behind the top app bar temporarily, and it does not do that when you then dismiss the keyboard.

What are we trying to do here with this check here? Could we just do without any if/else check here?
The `safeDrawing` btw does include the ime padding itself too internally, see here https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/androidMain/kotlin/androidx/compose/foundation/layout/WindowInsets.android.kt;l=400?q=safeDrawing
The doc reads:
"""
The insets that include areas where content may be covered by other drawn content. This includes
all [system bars][systemBars], [display cutout][displayCutout], and [soft keyboard][ime].
"""

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingTypeContainer.kt
panasetskaya: I was not sure you'd want to use the new element right away:)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
panasetskaya: I think the imePadding is one of the things that were not working properly previously, I'll check this one out!

app/feature/feature-changeaddress/src/main/kotlin/com/hedvig/android/feature/changeaddress/ui/extrabuildings/ExtraBuildingTypeContainer.kt
StylianosGakis: Yeah especially since this is more of a foundational layout component and not so much a ui design thing, I think we should start using it asap so that we can start making sure that it works well in the first place.

PR #2172, opened 2024-08-07
GEN-2332 New design/toggle

panasetskaya:


StylianosGakis (commented):
Had some thoughts about the crossfade between the switch states and adding the ability to also drag the switch around.

I think before we start using the toggle in the app we will have to add that functionality to be similar with what we've had before too.

Also some questions about the animation APIs used, but that's not as important.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Toggle.kt
```diff
@@ -211,33 +210,45 @@ private fun Toggle(enabled: Boolean, onClick: (Boolean) -> Unit, modifier: Modif
       },
     )
   Crossfade(
-    // todo: looks fine without animating too, bc the is this container color change anyway
     targetState = enabled,
     animationSpec = tween(AnimationTokens().pulsatingAnimationDuration),
+    modifier = modifier
   ) { animatedEnabled ->
     Box(modifierNoIndication) {
```
StylianosGakis: Could we inline this `modifierNoIndication` so that the val does not exist now after the latest changes?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Toggle.kt
```diff
@@ -198,75 +214,113 @@ private fun DetailedToggle(
 
 @Composable
 private fun Toggle(enabled: Boolean, onClick: (Boolean) -> Unit, modifier: Modifier = Modifier) {
+  val density = LocalDensity.current
+  val positionalThreshold = { distance: Float -> distance * 0.3f }
+  val velocityThreshold = { with(density) { 100.dp.toPx() } }
+  val animationSpec = tween<Float>()
+  val decayAnimationSpec = rememberSplineBasedDecay<Float>()
+  val state = rememberSaveable(
```
StylianosGakis, line 201: Great job with all this. I've never used this API myself, only seen it exist. What is tricky to get it all working? Are the docs good around it?

The fact that you need to kind duplicate passing everything to the saver and the state itself is a bit odd, perhaps we could even make our own `rememberAnchorDraggableState()` which takes in these four and calls the right rememberSaveable thing. For example like this https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/text/input/TextFieldState.kt;l=636-642?q=rememberTextFieldState which does the right remember with the right saver internally.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Toggle.kt
```diff
@@ -71,35 +76,9 @@ fun HedvigToggle(
...
-      }
-    } else {
-      launch { containerColor.animateTo(initialContainerColor, animationSpec = animationSpec) }
-      launch { labelColor.animateTo(initialLabelColor, animationSpec = animationSpec) }
-      launch { descriptionColor.animateTo(initialDescriptionColor, animationSpec = animationSpec) }
-    }
-  }
+  val containerColor = toggleColors.containerColor(turnedOn)
+  val labelColor = toggleColors.labelColor(turnedOn)
+  val descriptionColor = toggleColors.descriptionColor(turnedOn)
```
StylianosGakis, line 81: Yup this is looking great, I feel like the color animation being an impl detail inside the `labelColor` etc is a nice thing. That way if it's animated or not does not matter really here.

We could also use the `by` syntax here too, to avoid the `.value` call further down, if possible.

StylianosGakis (approved):
The changes are looking super!

PR #2174, opened 2024-08-07
set up arrow colors

panasetskaya:
forgot to set up actual arrow colors :)


StylianosGakis (approved):
We seem to use the always black and always white colors for this.
Do we perhaps want the arrow to be inverted depending on the phone's theme?
Perhaps:
White background with black arrow for dark mode
Black background with white arreo for light mode

Otherwise with the always white background, how does this look like on light mode, is the contrast of the floating circle good enough to look different from whatever may be in the sheet itself?

panasetskaya:
> Otherwise with the always white background, how does this look like on light mode, is the contrast of the floating circle good enough to look different from whatever may be in the sheet itself?
you're probably right:)
Screenshot 2024-08-08 at 09 10 15

PR #2176, opened 2024-08-08
Bump compile sdk 35 and all the libraries that renovate tried to bump at the same time

StylianosGakis:
Also update the nav related typemaps to not have to reference primitive nullable types, and remove some unused fields from ClaimSummaryStep which were not used at all in the VM or the ui later


build-logic/convention/src/main/kotlin/com/hedvig/android/ConfigureCompose.kt
```diff
@@ -6,8 +6,6 @@ import org.gradle.kotlin.dsl.assign
 import org.jetbrains.kotlin.compose.compiler.gradle.ComposeCompilerGradlePluginExtension
 
 internal fun ComposeCompilerGradlePluginExtension.configureComposeCompiler(project: Project) {
-  enableStrongSkippingMode = true
```
StylianosGakis, line 9: It's now enabled by default

panasetskaya (approved):

PR #2177, opened 2024-08-09
GEN-2329 New design/stepper

panasetskaya:
two things here I'd be grateful to talk about: 

- animation (I think you mentioned animation api in the other pr comments - didn't yet go through it properly, the case where there are two states simultaneously  - I have this here (Stepper file, line 79)) - is this okay or a bit messy to read and  better to write it differently?

- padding. The padding that is there in Figma somehow does not make sense to me (e.g. if I put the paddings that are in Figma with the textStyle that is in Figma, medium size container gets higher than Large one), so I've put a minHeight in there (same as the fixedHeight in Figma) and padding that would allow the container to be of this minHeight if it's a one-line text. Is this okay? Now it looks like Figma though.


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Stepper.kt
```diff
@@ -0,0 +1,693 @@
...
+  private val disabledSymbolColor: Color,
+  private val pulsatingContainerColor: Color,
+  private val pulsatingContentColor: Color,
+  val errorTextColor: Color,
+) {
+  @Stable
+  fun containerColor(pulsating: Boolean): Color = when {
+    pulsating -> pulsatingContainerColor
+    else -> containerColor
+  }
```
StylianosGakis, line 529: Yeah so for the animated colors/values, can you try to see if you could make it work by adding the animations here instead? The exact same way I mentioned in the other PR, like we do here for example https://github.com/HedvigInsurance/android/blob/b5bf4698ac78da1f9753cb9e834467901210caaf/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/component/navigation/HedvigNavigationBarItem.kt#L320-L331

This would avoid the code which starts adding up quite a lot at the beginning of the component with having to manage all the animated values, especially when it's a lot of them.

Lmk if what I am saying does not make sense.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/StepperTokens.kt
```diff
@@ -0,0 +1,88 @@
+package com.hedvig.android.design.system.hedvig.tokens
+
+import androidx.compose.ui.unit.dp
+
+internal object StepperColorTokens {
+  val ContainerColor = ColorSchemeKeyTokens.SurfacePrimary
+  val EnabledSymbolColor = ColorSchemeKeyTokens.FillPrimary
+  val DisabledSymbolColor = ColorSchemeKeyTokens.FillDisabled
+  val TextColor = ColorSchemeKeyTokens.TextPrimary
+  val LabelColor = ColorSchemeKeyTokens.TextSecondaryTranslucent // todo: should we use Translucent?
```
StylianosGakis, line 10: Yeah you are right, I doubt it's what we actually want here. I would start off with the normal opaque color and if it turns out to be a problem we'll reconsider

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Stepper.kt
```diff
@@ -0,0 +1,693 @@
...
+  minusColor: Color,
+  showError: Boolean,
+  errorText: String?,
+  modifier: Modifier = Modifier,
+) {
+  Column(modifier) {
+    Surface(
+      shape = stepperSize.size.shape,
+      color = containerColor,
+      modifier = Modifier.defaultMinSize(minHeight = stepperSize.size.minHeight(Default)),
```
StylianosGakis, line 176: Yeah I think I understand what you mean regarding the padding being hard to get right without the minHeight.

I faced the exact same thing when working on the text field.
So the only way to properly make this work without having to resort to minHeight, which imo works a bit weird when the text font size changes, is to have different paddings for when there is a label and where there is no label.

So it looks like for the Large component (first column) for no label it's: top = 16, bottom = 18
For the one with label: top = 10, bottom = 9.

Also a weird thing is that the two texts when there *is* a label are actually overlapping a bit. In that first it's an overlap of 3.dps for example. You can see how this was done in the text field here https://github.com/HedvigInsurance/android/blob/ad43926deb13033b11d98ecd22a19f8f85fcebf0/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/internal/HedvigTextFieldImpl.kt#L346 It would be added to the column in line 253 if I am reading this correctly.

So after making these adjustments we'd be able to remove the minHeight modifier completely, and the component would look relatively the same regardless of text sizes. And you should continue to apply this padding to the entire content as you are already doing, meaning that if the text font is in fact small enough that the stepper symbols are bigger, then that's fine, those will then be the ones that determine the size of the container.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Stepper.kt
```diff
@@ -0,0 +1,693 @@
...
+) {
+  Row(modifier) {
+    IconButton(onClick = onMinusClick) {
+      Icon(HedvigIcons.Minus, null, tint = minusColor)
+    }
+    IconButton(onClick = onPlusClick) {
+      Icon(HedvigIcons.Plus, null, tint = plusColor)
+    }
+  }
+}
```
StylianosGakis, line 225: These in figma seem to be like this:
24x24icon - 16space - 24x24icon

I think with this impl we get (since the min touch target is making them 48.dp)
48x48 - 48x48,
Where if we have the icons be 24x24, it ends up having 12 and 12 spacing on each side, so the end result is
24x24 - 12 - 12 - 24x24
So our space inbetween ends up being 24dp instead of 16 as Figma describes.

Could we perhaps be explicit about the inner icons being 24.dp, so like
```
IconButton(onClick = onPlusClick) {
  Icon(HedvigIcons.Plus, null, tint = plusColor, modifier = Modifier.size(24.dp))
}
```
And then see if we can place them closer to each other if my assumption above is true?
Perhaps with the negative `Arrangement.spacedBy` as I mention in another comment, not sure.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Stepper.kt
StylianosGakis: And I just realize that you are doing the label/no label paddings separately exactly as I say already, so lmk if there's something else that was tricky.

StylianosGakis (approved):
Looking great, some comment on the stepper -+ icon spacing and the animation usage, like in the other PR. I think there is some room to look into the other place we do this which I linked in here and see if it would fit our needs here too.

PR #2182, opened 2024-08-14
Make sure to always fetch from network as well to avoid stale info

StylianosGakis:
Ref: https://hedviginsurance.slack.com/archives/C075NGQ600Z/p1723556945645839?thread_ts=1723555145.651149&cid=C075NGQ600Z

### DIFF
--- a/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/cbm/CbmChatRepository.kt
+++ b/app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/cbm/CbmChatRepository.kt
@@ -113,30 +113,40 @@ internal class CbmChatRepositoryImpl(
   }
 
   override fun bannerText(conversationId: Uuid): Flow<BannerText?> {
+    fun ConversationStatusMessageQuery.Data.toBannerText(): BannerText? {
+      return when {
+        conversation == null -> null
+        conversation.isOpen == false -> BannerText.ClosedConversation
+        conversation.statusMessage != null -> BannerText.Text(conversation.statusMessage)
+        else -> {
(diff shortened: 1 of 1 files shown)


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/cbm/CbmChatRepository.kt
```diff
@@ -113,30 +113,40 @@ internal class CbmChatRepositoryImpl(
   }
 
   override fun bannerText(conversationId: Uuid): Flow<BannerText?> {
```
panasetskaya, line 115: could you pls explain why 2 separate ones and not cacheAndNetwork?

panasetskaya (approved):

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/cbm/CbmChatRepository.kt
StylianosGakis: Yeah, I could not find a way to actually use CacheAndNetwork, and also make sure that if I get a cache miss I just emit nothing at all, and only continue with the network request, while also making sure that if I get a failure then in the network request, I do the delay and retry. While also not retrying the cache since we do not need to do that, only retry the network request. While also when I get a success response *from the network* I stop the loop completely and break out of it.

Basically there's no clear way to differentiate between getting a success or a failure from the cache or the network for me to then make these decisions :D 

Discussion here: https://slack-chats.kotlinlang.org/t/22917253/i-wanted-to-achieve-something-like-this-have-a-query-which-d#4aa48516-3a46-428f-92e8-58326207c0cf

PR #2184, opened 2024-08-16
GEN-2337 New design/free text

panasetskaya:
That's basically 2 separate elements. Interactive full screen FreeTextOverlay which is also a container for the content that it is shown above of, and it can be triggered by anything, ex. by some button or by the second element, FreeTextDisplay.
Basically what we done in the Termination Survey

### DIFF
--- a/app/design-system/design-system-hedvig/build.gradle.kts
+++ b/app/design-system/design-system-hedvig/build.gradle.kts
@@ -7,6 +7,7 @@ plugins {
 
 dependencies {
   implementation(libs.androidx.compose.foundation)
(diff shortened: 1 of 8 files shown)


StylianosGakis (approved):
Fixed couple of small things as I was looking at it, feel free to look at the diff here https://github.com/HedvigInsurance/android/compare/b65db4bb2b8c1c33ad263825076b683998e210fb...4d245073f2162345dd5292ca8460d177d2c52275
So I don't waste your time with you having to check this out, read my comments, change etc :D 
Will merge when CI is green, great work!

PR #2190, opened 2024-08-21
New design/tooltip

panasetskaya:
to do:

- [ ] uncomment line 72 transientShowTooltip in Tooltip.kt
- [ ] check offsets

### DIFF
--- a/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
+++ b/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
@@ -1,0 +1,475 @@
+package com.hedvig.android.design.system.hedvig
+
+import androidx.compose.animation.Crossfade
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.layout.wrapContentWidth
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Outline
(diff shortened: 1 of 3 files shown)


StylianosGakis:
Oh I just realized this was a draft, sorry about that :D
I made the preview just use the composable without the animation to make it easier to preview as I wanted to test it myself too.
But feel free to continue working on this if there's stuff left with the offsets, sorry for interfering 😅

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
```diff
@@ -158,49 +162,55 @@ private fun InnerChatTooltip(
...
+        TopEnd -> Offset(x = squircleSize.width - offsetFromEdgeWhenOffCentered, y = arrowHeight)
+        Start -> Offset(x = arrowHeight, y = squircleSize.height / 2)
+        End -> Offset(x = squircleSize.width, y = squircleSize.height / 2)
       }
       return Outline.Generic(
-        Path().apply {
-          addPath(path = squirclePath, offset = squircleOffset)
-          addPath(path = beakPath, offset = beakOffset)
-        },
+        Path.combine(
```
panasetskaya, line 209: Could you please explain here?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
```diff
@@ -102,52 +105,53 @@ private fun InnerChatTooltip(
...
+        modifier = Modifier
+          .widthIn(
+            min = TooltipDefaults.defaultMinWidth,
+            max = maxWidth,
+          )
+          .height(knownHeight)
+          .wrapContentHeight(unbounded = true)
+          .minimumInteractiveComponentSize()
+          .clickable(onClick = onClick)
+          .onPlaced { knownHeight = with(density) { it.size.height.toDp() } },
```
panasetskaya, line 131: some magic here:))

StylianosGakis: Some magic, which may absolutely not be needed after all.
image

Here's what I was trying to do, and the answer I received. I will try without it too. It seems I may have had a wrong understanding of what makes the clickable modifier make the touch size extend the bounds of the item itself. I thought minimumInteractiveComponentSize was needed, but perhaps not. Which would make all this weird height stuff go away!

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
StylianosGakis: Takes two standalone paths, and does a Union on them to merge them together.
Union as in boolean algebra union, this picture might help
image

It simply adds both areas together. If there is any overlap, it is simply also added to the final area

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
panasetskaya: Okay! So it has the right interactive size by default, and minimumInteractiveComponentSize is only needed if we have more than 1 interactive component close to each other?

StylianosGakis: Yup, totally unneeded, I was just wrong to try and use `minimumInteractiveComponentSize`.

I replied there too
image

PR #2193, opened 2024-08-23
GEN-2331 New design/ NotificationCard and Snackbar

panasetskaya:


app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
```diff
@@ -0,0 +1,158 @@
...
+  Row(
+    horizontalArrangement = Arrangement.Start,
+    verticalAlignment = Alignment.CenterVertically,
+    modifier = modifier,
+  ) {
+    icon()
+    Spacer(Modifier.width(8.dp))
+    Text(snackbarText, style = textStyle)
+  }
+}
```
StylianosGakis, line 152: These two seem to be identical, is it on purpose?

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
```diff
@@ -0,0 +1,158 @@
...
+          action()
+        }
+        showedSnackbar()
+      }
+    }
+  }
+  SnackbarHost(snackbarHostState, modifier.padding(16.dp)) { _ ->
+    Snackbar(
+      action = {
+        if (actionLabel != null) {
```
StylianosGakis, line 61: I don't know if this is relevant here or not, but the M3 API for this looks like it accepts a `action: @Composable (() -> Unit)?,` which means it would take `null` too,
Now, in the case where we have no actionLabel, we are passing it an empty lambda instead of a null lambda.
This *may* have implications on how it measures some stuff internally.
Do we perhaps want to try passing null here instead when we do not want to have an action and make sure things look OK like that too?

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
```diff
@@ -0,0 +1,158 @@
...
+
+      ActionPerformed -> {
+        if (action != null) {
+          action()
+        }
+        showedSnackbar()
+      }
+    }
+  }
+  SnackbarHost(snackbarHostState, modifier.padding(16.dp)) { _ ->
```
StylianosGakis, line 58: So as I understand the m3 api, this has a built-in queue so that we can't have many snackbars showing at the same time. In our case, we may have more than one of these since we may have separate instances of this. I suppose let's see how we end up using it to know if we must optimize for this in any way. We might not have more than one type of snackbars showing in some particular screen anyway.

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
```diff
@@ -0,0 +1,158 @@
...
+    ) {
+      if (actionLabel != null) {
+        SnackContentForAction(
+          icon = icon,
+          snackbarText = snackbarText,
+          textStyle = textStyle,
+          modifier = Modifier.padding(
+            top = 16.dp,
+            bottom = 16.dp,
+            end = 8.dp,
```
StylianosGakis, line 89: From the design it looks like it's still 16 for actions too. Is this to account for the action button taking up more space than it itself needs due to the fact that it's clickable?

StylianosGakis (approved):
Just had some questions in there but nothing too much. We can figure out if there's anything else as we actually try to use this

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
StylianosGakis, line 58: The 16.dp padding here looks to be added on top of some other padding too. On my phone, when I turned this to 0.dp it looked even more correct I'd say. perhaps the component has a padding itself applied for the insets + some sane default.
I'd suggest you try removing this padding modifier alltogether and see how that looks to you too

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
panasetskaya: The horizontalArrangement is different, but you're right, it's better be one function:)

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
panasetskaya: Nope, it is for the content that is to the left of the action, so it's a padding between the content and the action text. I've changed it now a little bit for it to be more readable.

app/design-system/design-system-internals/src/main/kotlin/com/hedvig/android/design/system/internals/SnackBar.kt
panasetskaya: On my phone it goes then to the bounds of the screen. I'd say let's leave it to be sure:)

PR #2194, opened 2024-08-24
GEN-2347 New design/progress bar

panasetskaya:
I am not entirely sure if that is the animation that was meant for the circular progress bar in the figma, but seems like it's close


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/ProgressBar.kt
```diff
@@ -0,0 +1,185 @@
...
+import com.hedvig.android.design.system.hedvig.ProgressDefaults.progressAnimationDuration
+import com.hedvig.android.design.system.hedvig.ProgressDefaults.stroke
+import com.hedvig.android.design.system.hedvig.ProgressDefaults.strokeCap
+import com.hedvig.android.design.system.hedvig.tokens.ColorSchemeKeyTokens.FillPrimary
+import com.hedvig.android.design.system.hedvig.tokens.ColorSchemeKeyTokens.SurfaceSecondary
+import kotlin.math.abs
+
+@Composable
+fun HedvigLinearProgressBar(modifier: Modifier = Modifier) {
+  val infiniteTransition = rememberInfiniteTransition(label = "rememberInfiniteTransition")
```
StylianosGakis, line 39: Yeah I do not expect us to use this one, as we typically do not have things that we know have a determinate end and we also know the progress of it. If we have an indeterminate (one that keeps flowing and does not have a start -> finish progress) version of this perhaps we should use that one.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/ProgressBar.kt
```diff
@@ -0,0 +1,185 @@
...
+      )
+    }
+  }
+}
+
+private fun DrawScope.drawLinearIndicatorTrack(color: Color, strokeWidth: Float, strokeCap: StrokeCap) =
+  drawLinearIndicator(0f, 1f, color, strokeWidth, strokeCap)
+
+@Composable
+fun HedvigCircularProgressBar(modifier: Modifier = Modifier) {
```
StylianosGakis, line 121: Yeah I see what you mean, it is close, but one thing that these progress bars typically do is that they are not a repeatable thing which does 1 loop on repeat, but their "starting point" moves along the circle a bit as the bar continues moving.
I think we need to either copy-paste the entire way they animate this from inside m3, or just use it as-is if it allows us to do the necessary adjustments to it, otherwise it does not "feel" right in a way.

micro-apps/design-showcase/src/main/kotlin/com/hedvig/android/sample/design/showcase/progress/ProgressBarShowcase.kt
```diff
@@ -0,0 +1,33 @@
...
+@Composable
+fun ProgressBarShowcase() {
+  Surface(color = Color.White, modifier = Modifier.fillMaxSize()) {
+    Column(
+      horizontalAlignment = Alignment.CenterHorizontally,
+      verticalArrangement = Arrangement.Center,
+    ) {
+      HedvigLinearProgressBar(Modifier.padding(32.dp))
+      Spacer(Modifier.height(32.dp))
+      ThreeDotsLoading()
```
StylianosGakis, line 28: I gotta admit sometimes I also find this bar a bit unnatural in terms of how it animates the circles. I feel like with these things either we need to find a spec online and follow it 100%, or it will just "feel" a bit off no matter what :D

StylianosGakis (commented):
Some thoughts, the linear bar we probably want it to be one without a start and a stop.
And for the circular we should probably steal the m3 things whatever it is that they do.

I could pick this up as well if you're in the middle of something else btw

PR #2195, opened 2024-08-26
Stop relying on isLast

StylianosGakis:
After feedback by Martin here
https://github.com/apollographql/apollo-kotlin/issues/6129 Replace the implementation which relied on the isLast to a much more robust implementation which should cover what the previous implementation did plus more.

### DIFF
--- a/app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
+++ b/app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
@@ -14,10 +14,8 @@ import com.hedvig.android.apollo.ApolloOperationError.CacheMiss
 import com.hedvig.android.apollo.ApolloOperationError.OperationError
 import com.hedvig.android.apollo.ApolloOperationError.OperationException
 import com.hedvig.android.core.common.ErrorMessage
-import com.hedvig.android.logger.LogPriority
-import com.hedvig.android.logger.logcat
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.filterNot
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 
 sealed interface ApolloOperationError {
@@ -51,19 +49,25 @@ suspend fun <D : Operation.Data, ErrorType> ApolloCall<D>.safeExecute(
 }
 
 fun <D : Operation.Data> ApolloCall<D>.safeFlow(): Flow<Either<ApolloOperationError, D>> {
-  return toFlow()
-    .filterNot { response ->
-      val isTemporaryCacheMiss = !response.isLast && response.exception is CacheMissException
-      // If we get a cache miss, but we know that we are awaiting a new response after that anyway, we do not need to
-      // emit the cache miss at all. We can just wait for the next response to come instead.
-      if (isTemporaryCacheMiss) {
-        logcat(LogPriority.DEBUG) {
-          "Apollo call:${this.operation.name()} emitted a cache miss, but another response is coming"
-        }
+  return flow<ApolloResponse<D>> {
+    var hasEmitted = false
+    var errorResponse: ApolloResponse<D>? = null
+    toFlow().collect {
+      if (it.exception != null) {
+        // Some errors may be followed by valid responses.
+        // In that case, wait for the next response to come instead.
+        errorResponse = it
+      } else {
+        hasEmitted = true
+        emit(it)
       }
-      isTemporaryCacheMiss
     }
-    .map { either { parseResponse(it) } }
+    val errorResponseValue = errorResponse
+    if (!hasEmitted && errorResponseValue != null) {
+      // Flow has terminated without a valid response, emit the error one if it exists.
+      emit(errorResponseValue)
+    }
+  }.map { either { parseResponse(it) } }
 }
 
 fun <D : Operation.Data, ErrorType> ApolloCall<D>.safeFlow(


app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
```diff
@@ -51,19 +49,25 @@ suspend fun <D : Operation.Data, ErrorType> ApolloCall<D>.safeExecute(
...
+      } else {
+        hasEmitted = true
+        emit(it)
       }
-      isTemporaryCacheMiss
     }
-    .map { either { parseResponse(it) } }
+    val errorResponseValue = errorResponse
+    if (!hasEmitted && errorResponseValue != null) {
+      // Flow has terminated without a valid response, emit the error one if it exists.
```
panasetskaya, line 67: but we're not logging the first errorResponse this way, only the last one will be emitted, right? I guess we don't really need it, if it's cache error?

panasetskaya (approved):

app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
StylianosGakis: On an app-wide scale, the logs will always happen due to this https://github.com/HedvigInsurance/android/blob/9bebe557f541337f9802a7b941594006db177504/app/app/src/main/kotlin/com/hedvig/android/app/apollo/LoggingInterceptor.kt#L19 interceptor.
That interceptor is added at the top of the list of interceptors, particularly here https://github.com/HedvigInsurance/android/blob/0e99875de4616d396158c723fcbf1fe7847095e0/app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt#L143 even before the normalizedCache interceptors are added.
That means that this one sees *everything*, and will in turn log everything as it needs to.

Regardless, yes, this one will "save" the error in a var until we know if we need to emit it or not to the caller. If the called receives a proper response afterwards, the error happening wasn't important anyway, we got the data anyway, so we get what we need.

Btw I would love to link you to the discussion inside https://slack-chats.kotlinlang.org/c/apollo-kotlin but it seems like it's having some troubles syncing the latest discussions. I will just paste it here, but I understand that the readability of this is not optimal. If they fix it I will link you to it properly.












Slack conversation follows:

---

Stylianos Gakis
  Thursday at 19:18
I got an interceptor like this:
```
internal class SomeLoggingInterceptor : ApolloInterceptor {
    override fun <D : Operation.Data> intercept(request: ApolloRequest<D>, chain: ApolloInterceptorChain): Flow<ApolloResponse<D>> {
        return chain.proceed(request).onEach { response ->
            logResponse(response)
        }
    }
}
```
Which I've atatched to my entire ApolloClient.
Then in a query with fetchPolicy(FetchPolicy.CacheAndNetwork), judging from the internals of the [CacheAndNetworkInterceptor](https://github.com/apollographql/apollo-kotlin/blob/de588a13bf73a4f85b82f8bac4d17be3f4b60d9b/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/FetchPolicyInterceptors.kt#L124-L127), I see two emits here, one for the cache emit, and one for the network emit.
I expect my logResponse(response) to be hit twice, since I am doing an onEach on the incoming flow.
However it looks like I am only receiving the network hit, since the cache fetch is failing and it actually emits
Thing is on the place where I do this query itself I get both emissions perfectly fine, I get one response with the CacheMiss exception as expected, and then the data response, but my interceptor does not get to see the cache miss emission at all.
What am I missing here? I can't figure out at all what exactly is going on here. Any ideas of something obvious I am missing, or what I can do to debug this?

---

Benoit 'BoD' Lubek
:apollo:  [Thursday at 19:25](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724347544343809?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
Did you setup your interceptor after the normalized cache? If yes, since the FetchPolicyInterceptor doesn't call chain.proceed when it's getting the value from the cache, your interceptor is not called in that case

---

Stylianos Gakis
  [Thursday at 19:29](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724347758224859?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
You know what, I did in fact setup my interceptor after I did .normalizedCache(...). So, I suppose I want my normalized cache to go at the end of whatever interceptor chain I got, never thought of that to be honest!
So how does the value actually end up being received at the caller which should be at the end of this entire chain? Or am I misunderstanding how this chain is setup?
I will try to move my interceptor above now btw to see how that works
[19:30](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724347852544669?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
Yup, switching them around I properly get both emissions there, and they even got the right isLast, the cache says isLast = false and the network request then returns isLast = true, as I was expecting in the first place! Quite interesting, thanks for this!

---

Benoit 'BoD' Lubek
:apollo:  [Thursday at 19:36](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724348162826209?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
So, at the end of the chain there is always the network call, which [is added](https://github.com/apollographql/apollo-kotlin/blob/f1d607c7e5a745a0bac976f2ff053190cb4af2e9/libraries/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo/ApolloClient.kt#L316) after all the other interceptors.
Since each interceptor can call the next one down the chain or not, the order in which you add them can impact on whether they are called or not.
[19:38](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724348310763689?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
(notice we already have a special one, retryOnErrorInterceptor which should always be in front of the network)

---

Stylianos Gakis
  Thursday at 19:41
Yup, debugged inside the proceed and I see the chain now:
```
0 = {LoggingInterceptor@44117} com.hedvig.android.app.apollo.LoggingInterceptor@4ca57ea
1 = {LogoutOnUnauthenticatedInterceptor@44118} com.hedvig.android.app.apollo.LogoutOnUnauthenticatedInterceptor@9eceadb
2 = {WatcherInterceptor@44119} com.apollographql.apollo.cache.normalized.internal.WatcherInterceptor@9d0e978
3 = {FetchPolicyRouterInterceptor@44120} com.apollographql.apollo.cache.normalized.FetchPolicyRouterInterceptor@fe62851
4 = {ApolloCacheInterceptor@44121} com.apollographql.apollo.cache.normalized.internal.ApolloCacheInterceptor@4ee08b6
5 = {DefaultRetryOnErrorInterceptorImpl@44122} com.apollographql.apollo.interceptor.DefaultRetryOnErrorInterceptorImpl@2c559b7
6 = {NetworkInterceptor@44123} com.apollographql.apollo.interceptor.NetworkInterceptor@9e72524
```
So this may be my flow knowledge failing me, but how does the result end up on the call site then? I imagine this tree as a long flow with many chained operations where at the end of it is the caller which does .collect and gets the emissions. How can one interceptor in this chain not receive the emission, but the caller does in fact receive that emission with the exception being there? I feel like I must have a faulty understanding of this somehow

---

Benoit 'BoD' Lubek
:apollo:  [Thursday at 19:50](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724349053822939?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
it's not exactly that but in essence:
you call toFlow()
  toFlow calls the first interceptor
    1st interceptor calls the 2nd interceptor
      2nd interceptor calls the network which returns a Flow
      2nd interceptor returns the Flow
    1st interceptor returns the Flow
  toFlow returns the Flow
you have a Flow, you call collect() on it
and in each step the interceptor can call e.g. onEach and add their own stuff. Also in each step maybe the interceptor doesn't call the next one (returns their own Flow instead)

---

Stylianos Gakis
  [Thursday at 19:52](https://kotlinlang.slack.com/archives/C01A6KM1SBZ/p1724349141043789?thread_ts=1724347099.024249&cid=C01A6KM1SBZ)
Aha wait, any interceptor can return a flow, but can just abruptly "end the chain" for that one emission by not allowing it to proceed further down first, but instead they emit something and can just end the chain. If they want as a follow up they can do an emitAll(...proceed) to "continue" the chain for the follow up emissions.
So in a way, the higher up you are in the chain, the more power you got to allow or not allow things to fall downwards to the rest of the interceptors, is that an okay understanding?

---

Benoit 'BoD' Lubek
:apollo:  Thursday at 19:52
exactly :white_check_mark:

PR #2198, opened 2024-08-27
GEN-2348 New design/tab

panasetskaya:
For our (only) use-case in the insurance details just a Row is fine too - so idk, just a matter of preference?


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,396 @@
...
+  tabSize: TabSize = TabDefaults.defaultSize,
+  tabStyle: TabStyle = TabDefaults.defaultStyle,
+) {
+  val density = LocalDensity.current
+  val currentWidthMap = remember { mutableStateMapOf<Int, Int>() }
+  val currentHeightMap = remember { mutableStateMapOf<Int, Int>() }
+  val currentOffsetMap = remember { mutableStateMapOf<Int, Offset>() }
+  val indicatorOffsetX: Dp by animateDpAsState(
+    targetValue = calculateIndicatorOffsetX(currentOffsetMap, selectedTabIndex, density),
+    animationSpec = tween(easing = LinearEasing),
```
StylianosGakis, line 73: This may be a bit tricky, but can you try and look at perhaps Sladan's phone where they have some tab indicator like that on some iPhone app and see how the indicator offset moves? I *suspect* that there is room for something which feels a bit smoother than a LinearEasing, but I definitely can't tell just from the top of my head.
You could also try at random some from here https://github.com/HedvigInsurance/android/blob/7e85f0db5781027e32700587c32c517454c95dac/app/core/core-design-system/src/main/kotlin/com/hedvig/android/core/designsystem/material3/motion/MotionTokens.kt#L24-L33 and see if any of them look nicer?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,396 @@
...
+          },
+        )
+      }
+    }
+  }
+}
+
+private fun calculateIndicatorWidth(map: SnapshotStateMap<Int, Int>, selectedTabIndex: Int, density: Density): Dp {
+  return with(density) {
+    map[selectedTabIndex]?.toDp() ?: 5.dp
```
StylianosGakis, line 141: The 5f fallback, do we ever expect to hit that? is it for the first frame where the map may not have been populated yet?
Does this actually have any visual effect in any case or is it just so that it does not crash? Also if we do just `!!` does it crash? I am mostly curious, I don't really suggest we do `!!`

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,396 @@
...
+  return with(density) {
+    result.toDp()
+  }
+}
+
+object TabDefaults {
+  internal val defaultSize: TabSize = Mini
+  internal val defaultStyle: TabStyle = Default
+  internal const val maxItemsInTheRow = 3
+  internal const val maxLines = 1
```
StylianosGakis, line 169: Are these used anywhere else or can they be private?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,396 @@
...
+  contentPadding: PaddingValues,
+  onTextMeasured: (width: Int) -> Unit,
+  onHeightMeasured: (height: Int) -> Unit,
+  modifier: Modifier = Modifier,
+) {
+  HedvigText(
+    modifier = modifier
+      .clickable(
+        interactionSource = remember { MutableInteractionSource() },
+        indication = null, // as the sliding indicator is enough, I think
```
StylianosGakis, line 349: Yeah since this is in reality an iOS component, I don't think they got an indicator there at all, so keep it like this and we can remove the comment too

StylianosGakis (approved):
🍏

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: I've looked at it on iPhone, and it feels like FastOutSlowInEasing. So I'm trying FastOutSlowInEasing with duration = 600 millis, I think it's close enough

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: `!!` just crashes. 5f (or 0f, doesn't matter) does give initial animation (the indicator expands to the element width and height from the top left corner), but I don't see an easy way to avoid that. Wdyt?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: They could be out of the object and private in the file, then. I'll do that

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
StylianosGakis: Hmmmm alright so it is more tricky than I had hoped for initially.
So if we did something like we've done in places like these https://github.com/HedvigInsurance/android/blob/7739da9bd86dc664f8e479b4e11e0de17f79327c/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt#L72-L80 do you imagine this working?
So in a box, layout the indicator as-is, but also layout the first entry in the flowrow too, but without placing it at all, only to have it take the same size.
so that the default size starts as the same as the first item anyway.

And you can then have this return null if it does not find anything in the map, and if that is the case just do not pass the width/height modifier at all, smth like
```
.then(
  if (resultFromInsideTheMap != null) {
    Modifier.width(resultFromInsideTheMap)
  } else {
    Modifier
  },
)
```

Does this make any sense at all?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
StylianosGakis: Could also be as part of the tokens too tbh. I put all of those in the tokens for the loading indicators https://github.com/HedvigInsurance/android/blob/4f8858ba42fb4f99b6cafd774d7d2e36cfb232d7/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/tokens/ProgressIndicatorTokens.kt#L8-L60 and it felt just fine tbh. A central spot to be able to adjust the component if need be.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: Great, will try that!

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+              textStyle = tabSize.textStyle,
+              tabTextColor = tabStyle.colors.textColor,
+              contentPadding = tabSize.tabPadding,
+            )
+          }
+        },
+      ),
+    ) { allMeasurables: List<List<Measurable>>, constraints ->
+      val tabItemsMeasurables = allMeasurables[1]
+      if (tabItemsMeasurables.size <= 1) {
```
StylianosGakis, line 114: Can do `isEmpty()` I think.
```suggestion
      if (tabItemsMeasurables.size.isEmpty()) {
```
Not important at all, just a bit easier on the eye when looking at this without having to look at the numbers

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+        layout(0, 0) {}
+      } else {
+        // calculate the size
+        val fullWidth = constraints.maxWidth
+        val desiredItemWidth =
+          if (tabItemsMeasurables.any { it.minIntrinsicWidth(oneLineHeight) > fullWidth / 3 }) {
+            fullWidth / 2
+            // this is the maximum width we give the item, if the text is too big for the basic 1/3 of the full width.
+            // If it's still not enough, the text itself gets eclipsed later on
+          } else if (tabItemsMeasurables.size == 2) {
```
StylianosGakis, line 124: Since we "short-circuit" into just splitting it half-half if we got specifically 2 items here, could we move the second check as the first check in this if/else chain?
It would be kinda like getting that "good scenario" out of the way first, and only after that is not true we iterate over the measurables to check their intrinsic width etc.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+        val howManyLines = (ceil(tabItemsMeasurables.size.toDouble() * desiredItemWidth / fullWidth)).toInt()
+        val howManyItemsInEachLine = fullWidth / desiredItemWidth
+        val fullLayoutCapacity = howManyLines * howManyItemsInEachLine
+        val isLastRowFull = tabItemsMeasurables.size == fullLayoutCapacity
+        val howManyItemsToCenter = if (isLastRowFull || tabItemsMeasurables.size == 2) {
+          0
+        } else {
+          howManyItemsInEachLine - (fullLayoutCapacity - tabItemsMeasurables.size)
+        }
+        val (indicesToCenter, _) = tabItemsMeasurables.indices.partition {
```
StylianosGakis, line 142: Since we do not use the second part, do you think this could be a `filter` instead of a `partition`?
It should give the same result as I read it

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+          it >
+            tabItemsMeasurables.size - 1 - howManyItemsToCenter
```
StylianosGakis, line 144: I think it'd be the same if you replaced `tabItemsMeasurables.size - 1` with `tabItemsMeasurables.lastIndex`, and it would show a bit better that the intent of the `-1` is to get the last index, and not some other kind of calculation.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+        val mapOfOffsets = mutableMapOf<Int, IntOffset>()
+
+        tabItemsMeasurables.forEachIndexed { index, _ ->
+          val currentRow = if (index <= howManyItemsInEachLine - 1) {
+            0
+          } else if (index <= (2 * howManyItemsInEachLine - 1)) {
+            1
+          } else {
+            2
+          }
```
StylianosGakis, line 156: Should this ever be hit? Since we only want to handle up to 6, I expect us to never have another row here right?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+    Layout(
+      contents = listOf(
+        {
+          TabIndicator(
+            indicatorColor = tabStyle.colors.chosenTabBackground,
+            indicatorShape = tabSize.tabShape,
+          )
+        },
+        {
+          tabTitles.forEachIndexed { index, title ->
```
StylianosGakis, line 97: Since we want to have this limit of 6 items, could we do this here
```suggestion
          tabTitles.take(6).forEachIndexed { index, title ->
```
To actually enforce it too?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+            // last line not full, centered
+            calculateHorizontalOffsetForCentered(
+              desiredItemWidth = desiredItemWidth,
+              fullWidth = fullWidth,
+              howManyItemsToCenter = howManyItemsToCenter,
+              currentIndex = index,
+              lastIndex = tabItemsMeasurables.lastIndex,
+            )
+          } else { // full line
+            val adjustedIndex = index - currentRow * howManyItemsInEachLine
```
StylianosGakis, line 170:
```suggestion
            val adjustedIndex = index - (currentRow * howManyItemsInEachLine)
```
Call me old-fashioned, but I like my parentheses around these just to again make it painfully obvious about what's going on

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
```diff
@@ -0,0 +1,443 @@
...
+      easing = FastOutSlowInEasing,
+    ),
+    label = "",
+  )
+  var oneLineHeight by remember { mutableIntStateOf(0) }
+  Box(
+    modifier = modifier
+      .clip(tabSize.rowShape)
+      .background(tabStyle.colors.tabRowBackground)
+      .height(intrinsicSize = IntrinsicSize.Min)
```
StylianosGakis, line 73: Does something change with/without it?
Always tricky to visualize things when intrinsic size is involved for me.

StylianosGakis (approved):
Alright, I was definitely doing a bunch of hand gestures trying to wrap all the calculations on my mind, but I think it all checks out now 😅

A small part of me also wonders if there's a way without having to do the extra calculcation with `oneLineHeight` which we grab from that other box, but I am not sure.
I know [TextMeasurer](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-text/src/commonMain/kotlin/androidx/compose/ui/text/TextMeasurer.kt;l=83?q=textMeasurer) exists, and here is some sample usage of it https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-text/src/androidInstrumentedTest/kotlin/androidx/compose/ui/text/TextMeasurerTest.kt;l=225-236?q=textMeasurerTest
And I wonder if we could inside the layout lambda grab a text measurer, give it the same sample text, and pass it in the constraints of the layout itself, and if we can get that information just directly there, without having to wait for first layout. The `TextLayoutResult` it returns back _should_ contain the information about its preferred height I think.
It's just an idea, do give it a try if you find the time. If not merge this in allright, and maybe I will see if I can play with it at a future date.

But this is great, I like the way we set the offset for the indicator too and how it animates around. Definitely looks like a hard component to write, and I believe this is your first complex Layout thingy you do, great job!!

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: Nope, there is in fact a third row if the size is too big for the screen and you pass 5-6 items there:)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: I'd like to not layout anything if size==1 too:)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
panasetskaya: nope, nothing changes:)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
StylianosGakis: Ah, right! A tab bar with one option certainly would be quite silly wouldn't it 😅

PR #2201, opened 2024-08-28
GEN-2507 SOS url button

panasetskaya:
- [x] wait till the copy SUBMIT_CLAIM_GLOBAL_ASSISTANCE_URL_LABEL is ready

ticket: [ticket](https://www.notion.so/hedviginsurance/a4c35052b6794b36a4b0d58b537cc325?v=c74a8cb958ab4ba6ba4c904d8540b106&p=6c50bcf6955f477588039ccb61ad2641&pm=s)
figma: [figma](https://www.figma.com/design/WlwFCAWywXNcXVgPfYVmWu/Final-Design?node-id=13270-61431&t=2iQkzOUKKfWnH5LV-0)

Also, as a thought - this seems to be a good place to start implementing the new DS. It's a small separate module and it's hectic with the old DS

### DIFF
--- a/app/ui/ui-emergency/src/main/kotlin/com/hedvig/android/ui/emergency/EmergencyScreen.kt
+++ b/app/ui/ui-emergency/src/main/kotlin/com/hedvig/android/ui/emergency/EmergencyScreen.kt
@@ -9,6 +9,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
+import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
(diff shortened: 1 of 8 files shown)


app/ui/ui-emergency/src/main/kotlin/com/hedvig/android/ui/emergency/EmergencyScreen.kt
```diff
@@ -91,14 +99,36 @@ fun EmergencyScreen(emergencyNumber: String?, navigateUp: () -> Unit, modifier:
...
+                emergencyUrl,
+              ),
+              onClick = {
+                openUrl(emergencyUrl)
+              },
+              modifier = Modifier.fillMaxWidth(),
+            )
+          }
+          if (emergencyNumber != null) {
+            val colors = if (emergencyUrl != null) {
```
StylianosGakis, line 116: So the color depends on if there are two buttons? So they don't look the same?

StylianosGakis (approved):

app/ui/ui-emergency/src/main/kotlin/com/hedvig/android/ui/emergency/EmergencyScreen.kt
panasetskaya: the way I see it, the top primary button is filled, and the bottom one is transparent, bc we want the member go the top button way (url in this case). But if there is only transparent button (for the call), it looks too bleak alone, and it should become our "primary way" if it is alone. So I did it like this.

StylianosGakis: I agree, it's a good idea

PR #2202, opened 2024-08-28
GEN-2334 Accordion

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Accordion.kt
```diff
@@ -0,0 +1,145 @@
...
+        onClick = {
+          isExpanded = !isExpanded
+        },
+        title = perilItem.title,
+        description = perilItem.description,
+        size = size,
+      )
+      if (index != items.lastIndex) {
+        Spacer(Modifier.height(4.dp))
+      }
```
StylianosGakis, line 39: More of a stylistic change, but I've found it much more convenient lately to just do the flipped of this, like we do here
https://github.com/HedvigInsurance/android/blob/0ec4aaee4da0c6552868e06c462ab30c2536de5b/app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/discounts/DiscountRow.kt#L39-L41

I like it for these two reasons
* The divider is at the start of the foreach, so it's easy to know it's there and does not get pushed further down when there is a lot of code in-between.
* The check is always done when `!= 0`, so no need to reference the list itself

No need to change it here, just thought I'd mention it :D

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Accordion.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+}
+
+@Composable
+fun AccordionItem(
```
StylianosGakis, line 45: Do we expect to use this item as an individual composable without it being in a list?
If not we could have this be private I think

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Accordion.kt
StylianosGakis: With that said, especially in this case where it is not a divider which we want to add in-between items, but it is only a spacer, this is usually much easier to do with doing spacedBy on the Column instead https://github.com/HedvigInsurance/android/blob/a042b5c5562b532699424ab27894b0f7534be519/app/audio-player-ui/src/main/kotlin/com/hedvig/android/audio/player/internal/FailedAudioPlayerCard.kt#L46
That way there's not even any need for extra code, index checks etc. And you can turn the .withIndex() into just a normal `for (perilItem in items) {` instead.

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Accordion.kt
```diff
@@ -0,0 +1,145 @@
...
+  isExpanded: Boolean,
+  onClick: () -> Unit,
+  title: String,
+  description: String,
+  modifier: Modifier = Modifier,
+  size: Size = Size.Small,
+) {
+  ExpandablePlusCard(
```
StylianosGakis, line 53: I wonder if `ExpandablePlusCard` is gonna be used anywhere else now outside of the accordion component.
If not, it should be moved here and be private, and we never refer to it from anywhere else. And we can update all call-sites to use the Accordion component instead. If this is the case, we can always do it at a later PR

StylianosGakis (approved):
👍

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Accordion.kt
StylianosGakis: Yep

PR #2213, opened 2024-09-04
Home screen shenanigans

StylianosGakis:
Log EVERYTHING around GQL requests, and make home query keep the ApolloOperationResponse signature so when it's logged it should hopefully contain the information of what went wrong.


panasetskaya (approved):
Wild! Do you want to make a new release with that?

PR #2215, opened 2024-09-05
GEN-2326 New design/dropdown

panasetskaya:


StylianosGakis (approved):
Not for this PR, but I wonder if we want to also explore the idea in the future where if we have two distinct styles of a component, like we do here with DropdownStyle where two of them take the simple item, and the other one takes the item with an icon, where we split it in two composables instead, which take in:
```
item: SimpleDropdownItem,
label: String?,
```
And the other
```
item: DropdownItemWithIcon
defaultIcon: IconResource
```

So we just have 2 functions as public API for this composable instead of a style parameter.

I am not saying it would be better or worse, but I think I've seen that pattern a bit more often in other stuff online.

Again, not for this PR, nor this component, but something to think of whenever we start using these more perhaps.

PR #2216, opened 2024-09-05
Remove Feature.ENABLE_CBM and all related code

StylianosGakis:
Removes all of the pre-CBM chat related code which makes the bulk of the PR


panasetskaya (approved):
Bit hard to review due to volume, so approving with pure belief:)

StylianosGakis:
yeah it's removing all the if else branches more or less, and then trying to prune out all newly unreachable code.
Worst case scenario is we still have some dead code which I failed to find, which we'll find at some point anyway

PR #2218, opened 2024-09-06
Feature/chat unread message from backend

StylianosGakis:
Offload the "read messages" bookkeeping entirely to the backend

Use the new `unreadMessageCount` field in `conversation`.
Use this information in home screen, claim details screen and inbox.


app/data/data-chat/src/main/kotlin/com/hedvig/android/data/chat/di/DataChatModule.kt
```diff
@@ -19,7 +18,6 @@ val dataChatModule = module {
     val applicationContext = get<Context>()
     Room
       .databaseBuilder<AppDatabase>(applicationContext, dbFile.absolutePath)
-      .fallbackToDestructiveMigration(true)
```
StylianosGakis, line 22: Not needed if we do in fact provide the right migration plan in annotation with `autoMigrations` over the `AppDatabase` class.

app/database/schemas/com.hedvig.android.data.chat.database.AppDatabase/2.json
```diff
@@ -0,0 +1,102 @@
+{
```
StylianosGakis, line 1: Autogenerated by Room after deleting the other table

panasetskaya (approved):

PR #2219, opened 2024-09-09
GEN-2552 Emergency-ui module with new DS + Scaffold + TopAppBar

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Notification.kt
```diff
@@ -60,7 +60,7 @@ import com.hedvig.android.design.system.internals.InternalSnackBar
 import com.hedvig.android.design.system.internals.NotificationColors
 
 @Composable
-fun HedvigNotificationCard(
+fun HedvigInfoCard(
```
StylianosGakis, line 63: I named it that since the Figma tab is named that, and I figured it'd then be hard to look at figma, see it named one way, and then in the code we have it named some other way.
Of course "notification" feels completely unrelated in our code, so I can definitely see why you renamed it, but I wonder if we should also ask design to consider renaming the component there too?

StylianosGakis (approved):
Another thought I was having, here we rename one thing from
`HedvigScaffold` to `Scaffold`
and then one thing from
`Text` to `HedvigText.

I think optimally, we should just not even have to have the `Hedvig...` prefix on anything, but it kinda helps when we are in the in-between state where everything is available when you start typing and you get every option as the import.

Just smth to have in mind so in the far future we go with no prefix anywhere.

PR #2220, opened 2024-09-09
Fix Android <= 8.1 not rendering clipped content properly

StylianosGakis:
Done by a hacky composition local which is just provided by the :app module through 
`UseSimplerShapesForOldAndroidVersions provides (hedvigBuildConstants.buildApiVersion <= 27),`
Then in the token resolution path, where we still have access to composition locals we grab it and branch to the right path.

Don't love it but I hope they do just fix it anyway at some point so we can just remove this local completely.

This has some unrelated material3 changes as I bumped to latest compose BOM to make sure they didn't just fix it in the latest release anyway.


panasetskaya (approved):
seems great!

PR #2221, opened 2024-09-10
Fix/outline generic paths

StylianosGakis:


panasetskaya (approved):
seems great:) I've tested it too, works on old ones for me

PR #2223, opened 2024-09-12
GEN-2638 nullify coinsured info

panasetskaya:
That bug affects not only the ssn, name also stays if you switch back from manual. I added some if-else at first, but then it seems that the most reliable thing to do would be just to clean the state ssn, birthday and name entirely. Tell me wdyt:)

### DIFF
--- a/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/RemoveCoInsuredBottomSheetContent.kt
+++ b/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/RemoveCoInsuredBottomSheetContent.kt
@@ -2,9 +2,7 @@ package com.hedvig.android.feature.editcoinsured.ui
 
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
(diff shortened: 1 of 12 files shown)


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/ui/RemoveCoInsuredBottomSheetContent.kt
```diff
@@ -24,14 +21,12 @@ import hedvig.resources.R
 @Composable
 internal fun RemoveCoInsuredBottomSheetContent(
   onRemove: (CoInsured) -> Unit,
-  onDismiss: () -> Unit,
   isLoading: Boolean,
-  errorMessage: String?,
+  errorMessage: String?, // todo: not used??
```
StylianosGakis, line 25: Yeah if it was not used before let it be I suppose. You can also remove it if you want. It may be that it had error messages that are not useful for the member to see so we stopped showing it?

StylianosGakis (approved):

StylianosGakis (approved):
Tested around, worked nicely. Filling a PN and switching around does not miss the input, having both filled still works by submitting the input from the one which was selected at the moment. And overall everything I tried looked good 👍

One thing which is not related to this PR but on horizontal position, both the insets are not looking right, but also the "Confirm changes" button along with the "cancel" and the price change are not part of the scrollable content, making the scrollable content have just a few pixels left to it.
image

PR #2230, opened 2024-09-18
Start potentially supporting partial gql responses

StylianosGakis:
Very important note here, that this is just done internally, but everything is still returned as Either, to keep the app running just as it used to. We can consider changing this at a later point. This PR is just to make it possible in the first place.

Use iorNel which instead of `Either<Error, Data>` returns `Ior<Nel<Error>, Data>`
Ior is basically an `Either` which can have *both* left and right at the same time, which is exactly what we want when we get partial responses.

This is for problems like when we've had the entire home screen error out on some members when one field failed, when we could've just had that one field be null and still showed all the rest of the data

See also articles like:
https://adambennett.dev/2024/09/using-feature-flags-in-graphql/ About how such partial responses break features and some other options for working around it. While the @include directive works for what they show there, just being able to work with partial responses properly is the more solid long-term approach without having to add @include directives everywhere.

### DIFF
--- a/app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
+++ b/app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
@@ -1,10 +1,16 @@
 package com.hedvig.android.apollo
 
 import arrow.core.Either
-import arrow.core.raise.Raise
-import arrow.core.raise.either
+import arrow.core.Ior
+import arrow.core.IorNel
+import arrow.core.Nel
+import arrow.core.nel
+import arrow.core.raise.IorRaise
+import arrow.core.raise.iorNel
+import arrow.core.toNonEmptyListOrNull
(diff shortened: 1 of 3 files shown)


app/apollo/apollo-core/src/main/kotlin/com/hedvig/android/apollo/ApolloCallExt.kt
```diff
@@ -117,21 +123,44 @@ private fun <D : Operation.Data> Raise<ApolloOperationError>.parseResponse(respo
...
+    )
+  }.toEither()
+}
+
+private fun <D : Operation.Data> iorFromErrorsAndData(
+  errors: Nel<ApolloOperationError>?,
+  data: D?,
+): Ior<Nel<ApolloOperationError>, D> {
+  return when {
+    errors != null && data != null -> Ior.Both(errors, data)
```
panasetskaya, line 159: cool

panasetskaya (approved):

PR #2232, opened 2024-09-24
Use new DS in login module

StylianosGakis:


app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/genericauth/SsnInputScreen.kt
```diff
@@ -56,9 +56,9 @@ fun SSNInputScreen(
         .verticalScroll(rememberScrollState()),
     ) {
       Spacer(Modifier.height(60.dp))
-      Text(
+      HedvigText(
         text = stringResource(R.string.zignsec_login_screen_title),
-        style = MaterialTheme.typography.headlineMedium,
+        style = HedvigTheme.typography.headlineMedium,
```
panasetskaya, line 61: aren't these two very different in size?

gradle/libs.versions.toml
```diff
@@ -217,6 +218,7 @@ kotlinter-gradlePlugin = { module = "org.jmailen.gradle:kotlinter-gradle", versi
 
 # lint dependencies
 lintApi = { module = "com.android.tools.lint:lint-api", version.ref = "lintApi" }
+androidx-core = { group = "androidx.core", name = "core", version.ref = "core" }
```
panasetskaya, line 221: it's new?

panasetskaya (approved):

app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/genericauth/SsnInputScreen.kt
StylianosGakis: 28 vs 24 sp, yeah they are different.
These screens are all already getting like absolutely no traffic, I thought it's fine to take such shortcuts tbh

app/feature/feature-login/src/main/kotlin/com/hedvig/android/feature/login/marketing/MarketingDestination.kt
```diff
@@ -171,24 +157,28 @@ private fun MarketingScreen(
...
             enabled = uiState is MarketingUiState.Success,
             onClick = {
               (uiState as? MarketingUiState.Success)?.run {
                 navigateToLoginScreen(market)
               }
             },
-            modifier = Modifier.testTag("login_button"),
+            modifier = Modifier
+              .fillMaxWidth()
+              .testTag("login_button"),
```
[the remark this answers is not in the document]
StylianosGakis: This was here for 2 reasons.
When we give instructions to play store on how to login to the app as demo mode, I also gave them information about the resource id of the things they must press
When we tried out some UI tests using Maestro, I needed the items to have a test id in order to tell it what to press.

I am keeping these around since it may be that play store is actually using this

gradle/libs.versions.toml
StylianosGakis: It does look suspect, I am not sure if I added that myself or if the IDE did when I accidentally pressed Option Enter to auto-resolve some import. And it instead brought in some dependency.
I will take a look at what happens if I remove it

PR #2248, opened 2024-10-09
Feature/tier summary submit

panasetskaya:


app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/stepsummary/SummaryDestination.kt
```diff
@@ -1,18 +1,345 @@
...
-  onNavigateToNewConversation: () -> Unit,
-  openUrl: (String) -> Unit,
+  onSuccess: () -> Unit,
+  onFailure: () -> Unit,
+  sharePdf: (File) -> Unit,
 ) {
-  Box(contentAlignment = Alignment.Center) {
-    HedvigText("Summary")
+  val uiState: SummaryState by viewModel.uiState.collectAsStateWithLifecycle()
+  when (uiState) {
```
StylianosGakis, line 67: Can we just do the `collectAsStateWithLifecycle` here and then delegate to another `ChangeTierSummaryScreen` composable which we can then `@Preview` all together for all possible SummaryState states?

app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/stepsummary/SummaryDestination.kt
```diff
@@ -1,18 +1,345 @@
...
+    HorizontalItemsWithMaximumSpaceTaken(
+      startSlot = {
+        Row(verticalAlignment = Alignment.CenterVertically) {
+          HedvigText(
+            color = HedvigTheme.colorScheme.textSecondary,
+            text = buildAnnotatedString {
+              append(name)
+              withStyle(
+                SpanStyle(
+                  baselineShift = BaselineShift(0.3f),
```
StylianosGakis, line 335: We've had to do this in like 3-4 places by now. Perhaps we should extract it somewhere like here https://github.com/HedvigInsurance/android/tree/442df7af55ca6431553c5473c1ccdec30f67dd02/app/compose/compose-ui/src/main/kotlin/com/hedvig/android/compose/ui in the future

app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/stepsummary/SummaryDestination.kt
```diff
@@ -1,18 +1,345 @@
...
+          )
+        }
+      },
+      endSlot = {
+        IconButton(
+          modifier = Modifier.size(24.dp),
+          onClick = {
+            downloadFromUrl()
+          },
+        ) {
```
StylianosGakis, line 352: It may be a good idea to make the entire row here clickable, along with the icon on the right. Did it feel like the entire thing should be clickable perhaps?

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/survey/TerminationSurveyDestination.kt
```diff
@@ -149,6 +165,40 @@ private fun TerminationSurveyScreen(
         navigateUp = navigateUp,
         closeTerminationFlow = closeTerminationFlow,
       ) {
+        AnimatedVisibility(uiState.showEmptyQuotesDialog) {
```
StylianosGakis, line 168: I'd expect the dialog itself to not take any effect from the AnimatedVisibility around it really, since the dialog is not a normal UI but it's a new window on top of everything.
I think we can replace this with an `if statement` instead.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/survey/TerminationSurveyViewModel.kt
```diff
@@ -136,20 +140,33 @@ internal class TerminationSurveyPresenter(
...
+                it.key.suggestion is SurveyOptionSuggestion.Action.DowngradePriceByChangingTier ||
+                  it.key.suggestion is SurveyOptionSuggestion.Action.UpgradeCoverageByChangingTier
+              }.keys
+              optionsToDisable.forEach {
+                if (currentReasonsWithFeedback.contains(it)) {
+                  currentReasonsWithFeedback.remove(it)
+                  val newOption = it.copy(isDisabled = true)
+                  currentReasonsWithFeedback[newOption] = null
+                }
+              }
```
StylianosGakis, line 159:
```suggestion
              Snapshot.withMutableSnapshot {
                  val optionsToDisable = currentReasonsWithFeedback.filter {
                    it.key.suggestion is SurveyOptionSuggestion.Action.DowngradePriceByChangingTier ||
                      it.key.suggestion is SurveyOptionSuggestion.Action.UpgradeCoverageByChangingTier
                  }.keys
                  optionsToDisable.forEach {
                    if (currentReasonsWithFeedback.contains(it)) {
                      currentReasonsWithFeedback.remove(it)
                      val newOption = it.copy(isDisabled = true)
                      currentReasonsWithFeedback[newOption] = null
                    }
                  }
              }
```
Probably a good idea to make sure that since we are doing many mutations here inside this one mutable set, we do it all and submit it in one go through the use of `Snapshot.withMutableSnapshot` instead of the state being mutated in various ways in sequence, meaning that while we're in the middle of this process we do not want to accidentally have it show half-done to the UI

Not sure how likely this is to happen and be a problem, but I still would suggest we do this anyway.

StylianosGakis (approved):

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/DownloadPdfUseCase.kt
```diff
@@ -35,26 +34,26 @@ internal class DownloadPdfUseCaseImpl(
...
 
-      try {
         val response = okHttpClient.newCall(request).await()
         downloadedFile.sink().buffer().use { fileSink ->
           fileSink.writeAll(response.body!!.source())
         }
         downloadedFile
-      } catch (exception: IOException) {
-        logcat(LogPriority.ERROR, exception) { "Could not download pdf" }
+      } catch (exception: Exception) {
```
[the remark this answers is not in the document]
panasetskaya: we were having a crush here because for some quotes the urls that we're getting in staging now are something like "MISSING TRANSLATIONS url for terms and conditions", which does not start with http:// and just crash immediately:)

app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/comparison/ComparisonViewModel.kt
```diff
@@ -36,16 +35,10 @@ private class ComparisonPresenter(
 
     LaunchedEffect(Unit) {
       // TODO: add error state!!! and either!
-      val result = tierRepository.getQuotesById(quoteIds)
+      val result = tierRepository.getQuotesById(quoteIds).sortedBy { it.tier.tierLevel }
```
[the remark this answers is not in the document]
panasetskaya: well we are getting them from our db here - so sorting is on our side. but that's what we have tierLevel for anyway

app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/stepsummary/SummaryDestination.kt
panasetskaya: extracted stringWithShiftedLabel

PR #2255, opened 2024-10-15
check if there is not already a change of Tier in the upcoming agreement

panasetskaya:
We display current tier/deductible by default. Now I've changed a little bit what this"current" is: if we have the upcoming change (for example, we have already gone through Tier change flow, now we want to do it again, while the first change is not yet activated), the current will be the tier and deductible for the upcoming agreement (because it's what BE counts as the base to give us available tiers/ deductibles).

Converting it to draft now; discussing possible API change

### DIFF
--- a/app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/data/GetCurrentContractDataUseCase.kt
+++ b/app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/data/GetCurrentContractDataUseCase.kt
@@ -16,6 +16,7 @@ import com.hedvig.android.logger.LogPriority.ERROR
 import com.hedvig.android.logger.logcat
 import kotlinx.coroutines.flow.first
 import octopus.CurrentContractsForTierChangeQuery
+import octopus.fragment.GeneralAgreementFragment
 
 internal interface GetCurrentContractDataUseCase {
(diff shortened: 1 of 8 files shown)


app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/data/GetCurrentContractDataUseCase.kt
```diff
@@ -44,19 +45,49 @@ internal class GetCurrentContractDataUseCaseImpl(
...
+            currentDisplayPremium = UiMoney.fromMoneyFragment(dataResult.currentAgreement.premium),
+            deductible = deductible,
+            productVariant = dataResult.currentAgreement.productVariant.toProductVariant(),
+            displayItems = dataResult.currentAgreement.displayItems.map {
+              ChangeTierDeductibleDisplayItem(
+                displayTitle = it.displayTitle,
+                displaySubtitle = it.displaySubtitle,
+                displayValue = it.displayValue,
+              )
+            }
```
StylianosGakis, line 88: Is it possible at all to perhaps here merge these two branches together by doing
```
val agreement = dataResult.upcomingChangedAgreement ?: dataResult.currentAgreement
```
And letting everything below it stay the same?
Do they share the same type for this to be possible? If not ignore this.

StylianosGakis (approved):

PR #2258, opened 2024-10-23
Replace tier db with datastore

StylianosGakis:
This is what I had in mind.

It relies on the model which we used to map the DB model to being Serializable itself, and we store its String representation inside DataStore.
DataStore lets us store a `Set<String>` in there, so we only need to do some small operations on that to turn the `Set<String>` into a `Map<String, TierDeductibleQuote>`, where the keys to the map are the `ID` of the quote.
This way we can also ensure uniqueness when we insert in there a new list of quotes, by overriding all the old quotes with the same ID.
Making `ProductVariant` serializable was actually a commit I copied over from my branch, since I had to do this anyway as well, so that part you can disregard in this change.


app/data/data-changetier/src/main/kotlin/com/hedvig/android/data/changetier/data/ChangeTierQuoteStorage.kt
```diff
@@ -0,0 +1,77 @@
...
+  suspend fun getQuotesById(ids: List<String>): List<TierDeductibleQuote>
+}
+
+internal class ChangeTierQuoteStorageImpl(
+  private val datastore: DataStore<Preferences>,
+) : ChangeTierQuoteStorage {
+  override suspend fun insertAll(quotes: List<TierDeductibleQuote>) {
+    datastore.edit { preferences ->
+      val existingQuotes = preferences[tierDeductibleQuotesKey]?.deserialize() ?: emptyMap()
+      val mergedQuotes = existingQuotes.plus(quotes.toQuoteIdKeyedMap())
```
panasetskaya, line 29: so .plus here overrides the values if there are any, right?

panasetskaya (approved):

PR #2260, opened 2024-10-25
Moving flow + Tiers

StylianosGakis:


app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -326,9 +342,12 @@ internal fun HedvigNavHost(
     helpCenterGraph(
       hedvigDeepLinkContainer = hedvigDeepLinkContainer,
       navigator = navigator,
-      onNavigateToQuickLink = { backStackEntry, quickLinkDestination ->
+      onNavigateToQuickLink = onNavigateToQuickLink@{ backStackEntry, quickLinkDestination ->
         val destination: Destination = when (quickLinkDestination) {
-          QuickLinkChangeAddress -> ChangeAddress
+          QuickLinkChangeAddress -> {
+            navigateToMovingFlow()
+            return@onNavigateToQuickLink
```
panasetskaya, line 349: this is a bit new for me :)

app/feature/feature-movingflow/src/main/graphql/MutationMoveIntentRequest.graphql
```diff
@@ -0,0 +1,62 @@
...
+    startDate
+    displayItems {
+      ...MoveQuoteDisplayItemFragment
+    }
+    exposureName
+    productVariant {
+      ...ProductVariantFragment
+      # These are not in the fragment, why?
+      #          termsVersion
+      #          displayNameSubtype
```
panasetskaya, line 53: I believe I've added termsVersion in here: https://github.com/HedvigInsurance/android/pull/2257 (comparison api) but it's still a draft

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/MovingFlowGraph.kt
```diff
@@ -0,0 +1,146 @@
...
+        viewModel = koinViewModel<ChoseCoverageLevelAndDeductibleViewModel>(),
+        navigateUp = navController::navigateUp,
+        onNavigateToSummaryScreen = { homeQuoteId ->
+          navController.navigate(MovingFlowDestinations.Summary(moveIntentId, homeQuoteId))
+        },
+      )
+    }
+    navdestination<MovingFlowDestinations.CompareCoverage> {
+      // todo moving flow, add shared compare coverage screen
+      CompareCoverageDestination(koinViewModel<CompareCoverageViewModel>())
```
panasetskaya, line 119: I suppose comparison should be in shared, where forever-ui module is? We'd pass list of termsVersion strings to there as parameters

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowQuotes.kt
```diff
@@ -0,0 +1,108 @@
...
+  data class MoveHomeQuote(
+    val id: String,
+    override val premium: UiMoney,
+    override val startDate: LocalDate,
+    override val displayItems: List<DisplayItem>,
+    override val exposureName: String,
+    override val productVariant: ProductVariant,
+    val tierName: String,
+    val tierLevel: Int,
+    val tierDescription: String?,
```
panasetskaya, line 38: would it maybe be convenient to put everything tier-related in one Tier entity (incl. productVariant.displayNameTier)?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowQuotes.kt
```diff
@@ -0,0 +1,108 @@
...
+import kotlinx.serialization.Serializable
+import octopus.feature.movingflow.fragment.MoveIntentQuotesFragment
+import octopus.feature.movingflow.fragment.MoveQuoteDisplayItemFragment
+
+@Serializable
+internal data class MovingFlowQuotes(
+  val homeQuotes: List<MoveHomeQuote>,
+  val mtaQuotes: List<MoveMtaQuote>,
+) {
+  interface Quote {
```
panasetskaya, line 20: would we want to have this general Quote interface for TierDeductibleQuote to inherit from too later?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/storage/MovingFlowStorage.kt
```diff
@@ -0,0 +1,52 @@
...
+   * Returns the edited [MovingFlowState]
+   */
+  suspend inline fun editMovingFlowState(crossinline block: (MovingFlowState) -> MovingFlowState): MovingFlowState? {
+    dataStore.edit { preferences ->
+      val oldState = preferences[movingFlowStateKey]?.let { Json.decodeFromString<MovingFlowState>(it) }
+      if (oldState != null) {
+        val newState = block(oldState)
+        preferences[movingFlowStateKey] = Json.encodeToString(newState)
+      } else {
+        logcat(LogPriority.ERROR) { "Trying to `editMovingFlowState` a non-existing moving flow state" }
```
panasetskaya, line 42: should we mb return ErrorMessage in this case?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/StartViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+            .bind()
+          val moveIntent = ensureNotNull(moveIntentCreate.moveIntent) {
+            val userError = moveIntentCreate.userError?.message
+            if (userError == null) {
+              StartUiState.StartError.GenericError(ErrorMessage("Unknown MoveIntentV2CreateMutation error"))
+            } else {
+              StartUiState.StartError.UserPresentable(userError)
+            }
+          }
+          movingFlowRepository.initiateNewMovingFlow(moveIntent, submittingHousingTypeValue)
```
panasetskaya, line 77: Should we check for the quotes list not to be empty too? is it a possible scenario here?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/StartViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+        is SelectHousingType -> selectedHousingType = event.housingType
+        SubmitHousingType -> submittingHousingType = selectedHousingType
+        NavigatedToNextStep -> moveIntentId = null
+        DismissStartError -> error = null
+      }
+    }
+
+    val submittingHousingTypeValue = submittingHousingType
+    if (submittingHousingTypeValue != null) {
+      LaunchedEffect(submittingHousingTypeValue) {
```
panasetskaya, line 60: we won't show loading here?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
```diff
@@ -0,0 +1,488 @@
...
+    topAppBarText = stringResource(R.string.CHANGE_ADDRESS_SUMMARY_TITLE),
+    modifier = Modifier.fillMaxSize(),
+  ) {
+    when (uiState) {
+      Loading -> HedvigFullScreenCenterAlignedProgress()
+      MissingOngoingMovingFlow -> HedvigErrorSection(
+        onButtonClick = navigateBack,
+        subTitle = null,
+        buttonText = stringResource(R.string.general_back_button),
+      )
```
panasetskaya, line 133: these two states look the same - will it change in the future?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/chosecoveragelevelanddeductible/ChoseCoverageLevelAndDeductibleViewModel.kt
```diff
@@ -0,0 +1,194 @@
...
+        val homeQuotes = movingFlowState.movingFlowQuotes.homeQuotes
+        if (homeQuotes.isEmpty()) {
+          tiersInfo = null.some()
+          return@collectLatest
+        }
+        val initiallySelectedHomeQuote = homeQuotes.firstNotNullOfOrNull { moveHomeQuote ->
+          if (moveHomeQuote.defaultChoice) moveHomeQuote else null
+        }
+        val uniqueCoverageOptions = homeQuotes
+          .groupBy { it.tierLevel }
```
panasetskaya, line 84: I think it should be tierName here

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/chosecoveragelevelanddeductible/ChoseCoverageLevelAndDeductibleViewModel.kt
```diff
@@ -0,0 +1,194 @@
...
+          .map { (_, moveHomeQuotes) ->
+            moveHomeQuotes.minByOrNull { it.premium.amount }
+          }
+          .filterNotNull()
+          .map { moveHomeQuote ->
+            CoverageInfo(moveHomeQuote.id, moveHomeQuote.tierName, moveHomeQuote.tierDescription, moveHomeQuote.premium)
```
panasetskaya, line 90: moveHomeQuote.productVariant.displayNameTier should be here

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/chosecoveragelevelanddeductible/ChoseCoverageLevelAndDeductibleViewModel.kt
```diff
@@ -0,0 +1,194 @@
...
+      when (allOptionsWithDeductible.size) {
+        0 -> DeductibleOptions.NoOptions
+        1 -> {
+          val onlyOption = allOptionsWithDeductible.first()
+          DeductibleOptions.OneOption(DeductibleOption(onlyOption.id, onlyOption.premium, onlyOption.deductible!!))
+        }
+
+        else -> DeductibleOptions.MutlipleOptions(
+          allOptionsWithDeductible.map { moveHomeQuote ->
+            DeductibleOption(moveHomeQuote.id, moveHomeQuote.premium, moveHomeQuote.deductible!!)
```
panasetskaya, line 166: I wonder if it's possible to have no deductible here, and if we should filter it out earlier, making it not-nullable by putting zero or smth

panasetskaya (approved):

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/chosecoveragelevelanddeductible/ChoseCoverageLevelAndDeductibleViewModel.kt
StylianosGakis: Yup, it was part of the notes I kept when we were testing yesterday, I've fixed it in a separate commit now

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
```diff
@@ -99,6 +101,10 @@ internal fun HedvigNavHost(
     }
   }
 
+  fun navigateToNewConversation(builder: (NavOptionsBuilder.() -> Unit)? = null) {
+    hedvigAppState.navController.navigate(ChatDestinations.Chat(Uuid.randomUUID().toString()), builder ?: {})
+  }
```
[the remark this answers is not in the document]
StylianosGakis: It's so we can do
```
navigate {
  popUpTo<SomeDestination>()
}
```
Basically the exact same lambd that the `navigate` function from NavController takes in

app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt
StylianosGakis: Haha yeah I know it's a bit special. it's so we can short-circuit out of the lambda, despite the fact that we are trying to do an assignment to `val destination: Destination = when...`
I don't necessarily love this too much, but I know it's temporary while the feature flag exists, so I thought it's okay.

app/feature/feature-movingflow/src/main/graphql/MutationMoveIntentRequest.graphql
StylianosGakis: Yeah, I'll let this be for now while comparison stuff is still WIP, no need to do anything about it for now

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/MovingFlowGraph.kt
StylianosGakis: Yeah something like that, I kept the comment vague as it will be easier to know the details once we actually know how we should approach this

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowQuotes.kt
StylianosGakis: I ended up grabbing the displayNameTier by doing this
```
val tierDisplayName = productVariant.displayTierName ?: tierName
```
Which I looked at Ugglan and they do the exact same thing.

So I am not sure, would the tier also include the ProductVariant in there? Since it needs it to grab the real displayable name.
But then, the ProductVariant exists in both type of quotes here, so extracting one Tier data class out of this would be tricky.

I think I won't do anything in particular for this yet, and as we work more on this if we find some good way to make this even more cohesive I'd be glad to do it

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowQuotes.kt
StylianosGakis: On one hand it feels that way, on another I do see there are quite some differences between the three.
The reason I added an interface in this particular scenario was so that I could present them in one common way on the UI later. Other than that there are a bunch of things that are different between the three that I don't know if a common interface would work.
We may need to do it if we want to in the future share the UI between these two screens in moving and change tier flows, but we can look into that when we get there

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/storage/MovingFlowStorage.kt
StylianosGakis: If anything I was even thinking of throwing an Exception here, but I figured doing nothing and showing an Error would be okay.
I can not possibly imagine a scenario where we are in the screens where we are editing a moving flow, but without having created a moving flow in the first place in the start screen.
I suspect we are never going to see this log, and if we do it means something is very very off

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/StartViewModel.kt
StylianosGakis: We should only be receiving quotes after we've done the `moveIntentRequest` mutation, otherwise they do not have enough information about what we want to move, besides it being a house/brf etc.
I do not think we need to consider it a possibility, if anything it would be a backend bug if that ever somehow happened.

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/StartViewModel.kt
StylianosGakis: Yeah if `val submittingHousingType: HousingType?,` is not null it should be interpreted as that we are loading the next step here.
I will add a `isLoading: Boolean` value in the data class which is derived from the value of `submittingHousingType` to make this more clear since now it's not really clear at all.

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
StylianosGakis: Yeah they both should be "I don't even know how it's possible for this state to have happened" states tbh.
I merged the two here https://github.com/HedvigInsurance/android/commit/74f690c294bda2371e9c1db918ca86f46f0022dd to remove this unnecessary duplication here

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/chosecoveragelevelanddeductible/ChoseCoverageLevelAndDeductibleViewModel.kt
StylianosGakis: I am not sure when this happens either tbh, and I would be very afraid of assuming a `0` there, as it may somehow be wrong and we might then be showing wrong information instead of having a lack of information.
I just tried to make this cover as much as possible all scenarios that the deductible being nullable makes possible for now. If we figure out there's something wrong with this we can always come back and fix it later

PR #2275, opened 2024-11-07
show "change address" in insurance detail only if it allowed

panasetskaya:
Check for memberAction isMovingAllowed before showing the "change address" button.
So if member has >1 home insurance they won't be able to start moving flow at all, both from HC and from ins.detail.

### DIFF
--- a/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
+++ b/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
@@ -36,19 +36,19 @@ internal class GetInsuranceContractsUseCaseImpl(
       featureManager.isFeatureEnabled(Feature.EDIT_COINSURED),
       featureManager.isFeatureEnabled(Feature.MOVING_FLOW),
       featureManager.isFeatureEnabled(Feature.TIER),
-    ) { insuranceQueryResponse, isEditCoInsuredEnabled, isMovingFlowEnabled, isTierEnabled ->
+    ) { insuranceQueryResponse, isEditCoInsuredEnabled, isMovingFlowFlagEnabled, isTierEnabled ->
       either {
         val insuranceQueryData = insuranceQueryResponse.bind()
         val contractHolderDisplayName = insuranceQueryData.getContractHolderDisplayName()
         val contractHolderSSN = insuranceQueryData.currentMember.ssn?.let { formatSsn(it) }
-
+        val isMovingEnabledForMember = insuranceQueryData.currentMember.memberActions?.isMovingEnabled ?: false
         val terminatedContracts = insuranceQueryData.currentMember.terminatedContracts.map {
           it.toContract(
             isTerminated = true,
             contractHolderDisplayName = contractHolderDisplayName,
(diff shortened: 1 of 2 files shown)


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
```diff
@@ -36,19 +36,19 @@ internal class GetInsuranceContractsUseCaseImpl(
...
       featureManager.isFeatureEnabled(Feature.MOVING_FLOW),
       featureManager.isFeatureEnabled(Feature.TIER),
-    ) { insuranceQueryResponse, isEditCoInsuredEnabled, isMovingFlowEnabled, isTierEnabled ->
+    ) { insuranceQueryResponse, isEditCoInsuredEnabled, isMovingFlowFlagEnabled, isTierEnabled ->
       either {
         val insuranceQueryData = insuranceQueryResponse.bind()
         val contractHolderDisplayName = insuranceQueryData.getContractHolderDisplayName()
         val contractHolderSSN = insuranceQueryData.currentMember.ssn?.let { formatSsn(it) }
-
+        val isMovingEnabledForMember = insuranceQueryData.currentMember.memberActions?.isMovingEnabled ?: false
```
StylianosGakis, line 44: We can probably do all of this in one go here, so we can use just `isMovingEnabledForMember` in the two places below, without risking accidentally changing one of them but forgetting to change the other, for example.
```suggestion
        val isMovingEnabledForMember = insuranceQueryData.currentMember.memberActions?.isMovingEnabled == true && isMovingFlowFlagEnabled
```

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
```diff
@@ -36,19 +36,19 @@ internal class GetInsuranceContractsUseCaseImpl(
...
         val terminatedContracts = insuranceQueryData.currentMember.terminatedContracts.map {
           it.toContract(
             isTerminated = true,
             contractHolderDisplayName = contractHolderDisplayName,
             contractHolderSSN = contractHolderSSN,
             isEditCoInsuredEnabled = isEditCoInsuredEnabled,
-            isMovingFlowEnabled = isMovingFlowEnabled,
+            isMovingFlowEnabled = isMovingFlowFlagEnabled && isMovingEnabledForMember,
```
StylianosGakis, line 51:
```suggestion
            isMovingFlowEnabled = isMovingEnabledForMember,
```

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
```diff
@@ -59,7 +59,7 @@ internal class GetInsuranceContractsUseCaseImpl(
...
```
StylianosGakis, line 62:
```suggestion
            isMovingFlowEnabled = isMovingEnabledForMember,
```

StylianosGakis (approved):
Great to have this be done eagerly here!

PR #2281, opened 2024-11-11
New design/claim details

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Dialog.kt
```diff
@@ -36,26 +36,26 @@ import com.hedvig.android.design.system.hedvig.tokens.DialogTokens
 import hedvig.resources.R
 
 @Composable
-fun HedvigDialogError(
-  titleText: String,
-  descriptionText: String,
-  buttonText: String,
-  onButtonClick: () -> Unit,
-  onDismissRequest: () -> Unit,
+fun ErrorDialog(
```
StylianosGakis, line 39: The one below is named `HedvigAlertDialog`, it might be a bit confusing to have them named differently perhaps?
What if we also named this `HedvigErrorDialog`?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/FileContainer.kt
```diff
@@ -0,0 +1,63 @@
...
+import coil.ImageLoader
+import coil.compose.AsyncImage
+import coil.compose.AsyncImagePainter
+import coil.request.ImageRequest
+import com.hedvig.android.placeholder.PlaceholderHighlight
+import com.hedvig.android.placeholder.fade
+import com.hedvig.android.placeholder.placeholder
+
+@Composable
+fun FileContainer(model: Any, imageLoader: ImageLoader, modifier: Modifier = Modifier, cacheKey: String? = null) {
```
StylianosGakis, line 23: If this is only used from inside the DS module for now, we should keep it internal

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/DynamicFilesGridBetweenOtherThings.kt
```diff
@@ -0,0 +1,425 @@
...
+import com.hedvig.android.design.system.hedvig.icon.Play
+
+/**
+ * Makes a column with content above and below a dynamic lazy grid of files thumbnails. The column
+ * takes the whole height, both the column and the inner grid are scrollable and the below content is always at the
+ * bottom (as if the lazy grid was with weight(1f), but without predefined height) no matter what the size
+ * of lazy grid and screen orientation are. Can be used for just the grid alone without above and below content too.
+ */
+@Composable
+fun ColumnScope.DynamicFilesGridBetweenOtherThings(
```
StylianosGakis, line 60: I assume all of this is just a copy-paste right? Was there anything tricky with moving it here?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/ShapedColorPainter.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.ColorFilter
+import androidx.compose.ui.graphics.Outline
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.graphics.drawscope.DrawScope
+import androidx.compose.ui.graphics.painter.ColorPainter
+import androidx.compose.ui.graphics.painter.Painter
+
+class ShapedColorPainter(
```
StylianosGakis, line 14: Same with this, if it's only used in the DS, we should keep it internal.

StylianosGakis (approved):
🤩

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Dialog.kt
panasetskaya: The thing is, those are named ErrorDialog in the old design. For the sake of easier transition, I'd say let's leave the naming as it was int he old one maybe, and then rename after the migration?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/DynamicFilesGridBetweenOtherThings.kt
panasetskaya: nope

PR #2282, opened 2024-11-11
Fix/table

StylianosGakis:
Surely this time is the last time
image

I caved and just made a custom layout for the table + cells + overlay to have full control of what goes where.

Basically with the implementation before, the left column would cap at 40% of the screen, even if the cells did not actually need the rest 60% of the screen, which was very common for wider screens/landscape mode.

This makes use of textMeasurer to calculate all the widths eagerly (textMeasurer has an internal cache, so we only measure each text exactly once), and we can make more well informed decisions about the table.

That way we have all the information we need in order to
achieve all of our requirements at the same time:
* If there is enough space for everything, render everything without
clipping any text
* If there is not enough space for everything, place the row titles
first and then the cells right next to them
* If the row titles end up being smaller than 40% of the screen, do
nothing else about it, and give the rest of the screen to the cells
* If the row titles were to try to take more than 40% of the screen,
force them to take 40% of the screen instead, and then give the cells
the rest of the screen

This also makes sure that since the interaction source of the cells comes from a different X axis compared to where the overlay sits in, that we offset the interaction source `pressPosition` by exactly the width of the left column width.


StylianosGakis:
Merging to facilitate further work on this, a PR review is still welcome

PR #2284, opened 2024-11-12
Make table use tokens from DS

StylianosGakis:
With this, add the shadow color which is transparent for dark mode.
Also fixes the text font accidentally changing to label in the bottom sheet too
It also uses LocalMinimumInteractiveComponentSize to make all rows tall enough for comfortable clicking

image
This surely is the last table PR

### DIFF
--- a/app/shared/tier-comparison/src/main/kotlin/com/hedvig/android/shared/tier/comparison/ui/ComparisonDestination.kt
+++ b/app/shared/tier-comparison/src/main/kotlin/com/hedvig/android/shared/tier/comparison/ui/ComparisonDestination.kt
@@ -61,6 +61,7 @@ import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.IntOffset
+import androidx.compose.ui.unit.coerceAtLeast
 import androidx.compose.ui.unit.coerceAtMost
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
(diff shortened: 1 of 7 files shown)


app/shared/tier-comparison/src/main/kotlin/com/hedvig/android/shared/tier/comparison/ui/ComparisonDestination.kt
```diff
@@ -194,6 +195,7 @@ private fun ComparisonScreen(uiState: Success, navigateUp: () -> Unit) {
     Spacer(Modifier.height(24.dp))
     Table(
       uiState = uiState,
+      tableStyle = TableStyle(),
```
panasetskaya, line 198: I have a question - what is the preference of data class with a private constructor over data object? Are we planning to use .copy and change the colors, for example?

StylianosGakis: This is a good question.
With a data object we'd need to immediately construct the object with the initial values. We can't however really do that since we need to be in a @Composable function in order to be able to call `.value` on the tokens. Here is that function https://github.com/HedvigInsurance/android/blob/6d659926dacaa16c0b2c70ac76bb45887f1e7d29/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/ColorScheme.kt#L462-L465 which needs to access the colorScheme composition local to grab the right colors.

Then made a data class as we needed a container to put all of the colors in from some composable function which called those .value methods, and that data class doubles as a nice place to keep some of the color decision logic, while keeping the colors themselves private. That way those decisions don't have to happen inline in the UI, possibly many times too, but just happen once inside this one data class.

Those were my thoughts around this and how it shaped up to be this way. If you have some idea on how to change this I would be more than happy to discuss it of course!

StylianosGakis: Oh and finally, I made this a `data` class so that it implements the `equals` and `hashcode` functions properly, since I think that is important for compose to know how to properly compare this with other versions of the same class and properly be able to determine if it has or has not changed, so it can skip recompositions when it knows that it was in fact the same as before, since the entire thing should be immutable this way.
This goes over this in greater detail https://developer.android.com/develop/ui/compose/performance/stability#immutable-objects 
I gotta admit I am not 100% sure on the details of this, and if compose could infer this as stable even if it were not a `data` class, since all of its params are immutable `val`s, but better safe than sorry :D

PR #2286, opened 2024-11-12
GEN-2534 Fix/connect payment new ds

StylianosGakis:
I thought I'd do it since I was here anyway


app/feature/feature-connect-payment-trustly/build.gradle.kts
```diff
@@ -12,26 +12,36 @@ hedvig {
 dependencies {
   api(libs.androidx.navigation.common)
 
-  implementation(libs.androidx.compose.material3)
+  implementation(libs.androidx.compose.animation)
```
panasetskaya, line 15: are you sure we need them all, androidx.compose? androidx.compose.foundation with the design system is usually enough

panasetskaya (approved):

app/feature/feature-connect-payment-trustly/build.gradle.kts
```diff
@@ -12,26 +12,36 @@ hedvig {
...
+  implementation(libs.androidx.compose.foundation)
+  implementation(libs.androidx.compose.foundationLayout)
+  implementation(libs.androidx.compose.runtime)
+  implementation(libs.androidx.compose.uiCore)
+  implementation(libs.androidx.compose.uiGraphics)
+  implementation(libs.androidx.compose.uiText)
+  implementation(libs.androidx.compose.uiUnit)
```
panasetskaya, line 22: are you sure we need them all, androidx.compose? androidx.compose.foundation with the design system is usually enough

StylianosGakis: I ran `./gradlew feature-connect-payment-trustly:fixDependencies` on this module and it added them all.
This comes from here https://github.com/autonomousapps/dependency-analysis-gradle-plugin?tab=readme-ov-file#fix-dependency-issues-automatically

The reasoning behind this is that this module was in fact using all of these dependencies, but they were hidden behind the fact that some other dependencies were transitively pulling these along as well.

This is not typically a problem until this one scenario happens: https://slack-chats.kotlinlang.org/t/23211614/i-stumbled-upon-the-artifact-https-central-sonatype-com-arti#b96f2b01-eb98-4020-a0af-ebd3fac3c169

* You have a module with some code, and you depend on some artifact
* You realize that one of the dependencies you have is no longer needed, so you remove that dependency
* You sync your project and you realize you got red code, despite the fact that you removed a dependency which you were not using in the first place.
* You look into it, and it turns out that dependency you were not using was transitively bringing in (aka it was doing `api(libs.that.dependency)`) some *other* dependency which you *were* using, however you were not declaring yourself

This fix just makes sure that we don't get any such surprises in any future.

I know it's not a very common occurrence, but it's done by the tool automatically, so I am not worried about it.

PR #2288, opened 2024-11-15
New design/forever feature shared

panasetskaya:
Forever UI to new DS. 
Also I didn't really like how it looked while initially loading, so I added placeholder there. 
And changed the structure quite a lot, since there were many workaround old bottom sheets, I think? Let me know what you think about it.

### DIFF
--- a/app/shared/forever-ui/src/main/kotlin/com/hedvig/android/shared/foreverui/ui/ui/ForeverDestination.kt
+++ b/app/shared/forever-ui/src/main/kotlin/com/hedvig/android/shared/foreverui/ui/ui/ForeverDestination.kt
@@ -257,33 +257,22 @@ internal fun ForeverContent(
   openEditCodeBottomSheet: () -> Unit,
   closeEditCodeBottomSheet: () -> Unit,
 ) {
-  var textFieldValue by remember(uiState.foreverData?.campaignCode) {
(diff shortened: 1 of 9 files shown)


app/shared/forever-ui/src/main/kotlin/com/hedvig/android/shared/foreverui/ui/ui/ForeverDestination.kt
```diff
@@ -114,159 +143,177 @@ private fun ForeverScreen(
...
 
-  LaunchedEffect(Unit) {
-    snapshotFlow { textFieldValueState }.collectLatest {
-      showedReferralCodeSubmissionError() // clear error after the member edits the code manually
+  LaunchedEffect(textFieldValue) {
+    showedReferralCodeSubmissionError() // Clear error on new referral code input
+  }
+  LaunchedEffect(uiState.showEditReferralCodeBottomSheet) {
+    if (!uiState.showEditReferralCodeBottomSheet) {
+      delay(500) // to avoid extra animation effects on closing bottom sheet
```
panasetskaya, line 276: // todo: okay, so this delay is ugly, I admit, but it's the only way to restore the value back on closing the BS and don't have a weird animation effect when closing the bottom sheet
while having submit error/changed textFieldValue in the BS

PR #2289, opened 2024-11-15
GEN-2542 Ds/insurances

StylianosGakis:
I changed one thing which might affect other screens which was the default padding around the entire bottom sheet.
With previous implementation, if we passed nothing in the `sheetPadding` value it would also move the sheet up above the bottom navigation bar, which is never what we want to do anyway.
This might have affected other callers as well, but I will try to go through them to make sure they look OK

### DIFF
--- a/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
+++ b/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
@@ -111,6 +111,10 @@ interface HedvigBottomSheetState<T> {
   fun dismiss()
 }
 
+fun HedvigBottomSheetState<Unit>.show() {
+  show(Unit)
+}
+
 private class HedvigBottomSheetStateImpl<T>() : HedvigBottomSheetState<T> {
   override var isVisible: Boolean by mutableStateOf(false)
     private set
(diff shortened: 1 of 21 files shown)


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -172,14 +177,24 @@ private fun ContractDetailScreen(
         }
 
         is ContractDetailsUiState.Success -> {
+          var consumedWindowInsets by remember { mutableStateOf(WindowInsets(0.dp)) }
           LazyColumn(
             contentPadding = WindowInsets
               .safeDrawing
               .only(WindowInsetsSides.Horizontal + WindowInsetsSides.Bottom)
+              .exclude(consumedWindowInsets)
```
panasetskaya, line 185: what was the problem with the insets here?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -92,6 +96,64 @@ fun HedvigBottomSheet(
...
+      if (!it) {
+        hedvigBottomSheetState.dismiss()
+      }
+    },
+    onSystemBack = {
+      hedvigBottomSheetState.dismiss()
+    },
+  ) {
+    if (hedvigBottomSheetState.data != null) {
+      content(hedvigBottomSheetState.data!!)
```
panasetskaya, line 152: why do we need data here?

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
```diff
@@ -127,42 +124,39 @@ private fun CrossSellItem(
...
-        containerColor = MaterialTheme.colorScheme.lightTypeContainer,
-        contentColor = MaterialTheme.colorScheme.onLightTypeContainer,
-      ),
+      buttonSize = Small,
+      buttonStyle = PrimaryAlt,
       modifier = Modifier.placeholder(
         visible = isLoading,
         highlight = PlaceholderHighlight.shimmer(),
-        shape = MaterialTheme.shapes.squircleLarge,
+        shape = HedvigTheme.shapes.cornerXLarge,
```
panasetskaya, line 159: maybe diff corner here?

panasetskaya (approved):

PR #2300, opened 2024-11-16
Bump to target kotlin2.0 and use ksp2

StylianosGakis:
Do a workaround for https://issuetracker.google.com/issues/379337774 since room plugin does not know how to read the schema location when provided through gradle when we're using ksp2.
We instead provide it both to the gradle plugin, since otherwise it fails some validation it has internally, but also to ksp as a generic arg which is how it then manages to actually read the location dir


settings.gradle.kts
```diff
@@ -32,6 +32,17 @@ dependencyResolutionManagement {
   }
 }
 
+plugins {
+  id("com.gradle.develocity") version "3.18.2"
+}
+
+develocity {
```
panasetskaya, line 39: vad?:)

panasetskaya (approved):

PR #2304, opened 2024-11-18
Migration/remove queries from core apollo module

StylianosGakis:
Move them into the modules in which they belonged in the first place


app/notification/notification-firebase/src/main/kotlin/com/hedvig/android/notification/firebase/FCMTokenUploadWorker.kt
```diff
@@ -43,9 +43,17 @@ internal class FCMTokenUploadWorker(
...
+          if (response.memberDeviceRegister) {
+            logcat { "NotificationRegisterDeviceMutation success with token:$storedToken" }
+            Result.success()
+          } else {
+            logcat {
+              "NotificationRegisterDeviceMutation failed with token:$storedToken. " +
+                "Will retry later. memberDeviceRegister:${response.memberDeviceRegister}"
+            }
+            Result.retry()
+          }
```
panasetskaya, line 56: did you encounter a problem with that?

StylianosGakis: No, however I realized that we do not take care of the scenario where backend may actually respond with a real response which says `false`.
Before we were assuming any response means that it went well, but the API returns a boolean to determine if it went well or not. So this just makes sure that if the backend returns `false`, that we actually do not consider this process as successfully done

PR #2310, opened 2024-11-18
Fix bottom sheet universal inset handling

StylianosGakis:
Make sure that modal bottom sheets don't give focus on appearance

Due to how he modal sheet is implemented
https://github.com/oleksandrbalan/modalsheet/blob/519938d25af1b9f41a409c5514f36a6cce0f4cb0/modalsheet/src/main/java/eu/wewox/modalsheet/FullscreenPopup.kt#L135
[ModalSheet] automatically requests focus on appearance. This means that if there is a focusable child available,
it gets focus immediately. This commonly is a TextField, which also brings the keyboard up. The keyboard ends up
coming up way too early, before the sheet manages to show, which results in a very awkward and janky animation
where the keyboard and the sheet are coming up in different timings, hiding each other while the insets are not
quick enough to catch up to make them go up in sync.
Making this box focusable means that it itself grabs the focus instead, not showing the keyboard on appearance.
@[StylianosGakis](https://github.com/HedvigInsurance/android/commits?author=StylianosGakis)
StylianosGakis committed 2 hours ago

Remove overload with a built-in bottom button

This ended up making it harder to work with insets, since any insets
added on the content slot would end up showing *above* that button,
which is not what we expect here.
Remove the `sheetPadding` parameter for everyone, since we do not want
to inset the sheet itself anything other than around the top and
horizontally, as we want the sheet to always go all the way to the
bottom of the screen.

Update all callers accordingly

Also just change all callers to have a 8.dp bottom space instead of 16, for some consistency between the sheets.
There may be room for embedding part of this logic in the sheet itself, but I am hesitant to do that as we're probably gonna at some point in the near future try to use the m3 sheet and I wanna make sure it'd work there too.
So for now, 8.dp + bottom insets are added on each caller individually, and we do not let callers change the insets of the sheet itself, since we never want the sheet to clip its contents at the bottom, yet we always want it to be inset at all other sides so that the sheet does not go "out of bounds" of the entire screen

### DIFF
--- a/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
+++ b/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
@@ -2,6 +2,7 @@ package com.hedvig.android.design.system.hedvig
 
 import androidx.compose.animation.Crossfade
 import androidx.compose.foundation.background
+import androidx.compose.foundation.focusable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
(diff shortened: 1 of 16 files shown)


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -181,20 +144,25 @@ private fun InternalHedvigBottomSheet(
...
     shape = bottomSheetShape.shape,
   ) {
-    Box(modifier) {
+    // [ModalSheet] automatically requests focus on appearance. This means that if there is a focusable child available,
+    // it gets focus immediately. This commonly is a TextField, which also brings the keyboard up. The keyboard ends up
+    // coming up way too early, before the sheet manages to show, which results in a very awkward and janky animation
+    // where the keyboard and the sheet are coming up in different timings, hiding each other while the insets are not
+    // quick enough to catch up to make them go up in sync.
+    // Making this box focusable means that it itself grabs the focus instead, not showing the keyboard on appearance.
+    Box(Modifier.consumeWindowInsets(sheetPadding).focusable()) {
```
panasetskaya, line 165: love this

panasetskaya (approved):

PR #2311, opened 2024-11-19
Migrate chat to new DS

StylianosGakis:


app/feature/feature-chat/build.gradle.kts
```diff
@@ -45,13 +44,10 @@ dependencies {
   implementation(projects.coreCommonAndroidPublic)
   implementation(projects.coreCommonPublic)
   implementation(projects.coreDemoMode)
-  implementation(projects.coreDesignSystem)
   implementation(projects.coreFileUpload)
```
panasetskaya, line 47: I think this one leaks material, no?

panasetskaya: and placeholder below should be removed

PR #2317, opened 2024-11-20
delete acc to new ds

panasetskaya:


app/feature/feature-delete-account/src/main/kotlin/com/hedvig/android/feature/deleteaccount/DeleteAccountDestination.kt
```diff
@@ -102,20 +99,16 @@ private fun DeleteScreenContents(
...
-    disabledContentColor = MaterialTheme.colorScheme.onPrimary.copy(alpha = 0.38f),
-  ),
 ) {
   Column(modifier) {
     Spacer(Modifier.height(16.dp))
-    Text(
+    HedvigText(
       text = title,
-      style = MaterialTheme.typography.headlineMedium,
+      textAlign = TextAlign.Center,
```
StylianosGakis, line 110: A lot of our newer screens that have this top attached text now are left-aligned. Do we want this one to be center aligned here? It seems like it was not as such before.

StylianosGakis (approved):

PR #2334, opened 2024-11-28
Fix deleting input coming back from comparison table

StylianosGakis:
Persist the selected choice on disk on each selection

This ensures that we persist this choice even for scenarios when we go
to the comparison screen and come back to this after the ViewModel's
StateFlow already turned cold


panasetskaya (approved):

panasetskaya:
it'll stay pre-selected even if we go back further than ChoseCoverageLevelAndDeductibleDestination this way, right? But I guess it's maybe even better.

StylianosGakis:
Yeah exactly, but it should be fine

PR #2335, opened 2024-11-28
Clear the backstack of all top level graphs on logout

StylianosGakis:
Clearing the backstack does not by itself clear the saved stack on the
top level graphs. Trying to navigate back to them with 
`restoreState = true` still does try to reset the backstack and as a
consequence the ViewModels as well.
This has the unfortunate side effect that for a moment the old UI state
is shown which could be the information for the wrong member.
It only happens for a frame or two, but we might as well fix this this
way for now


panasetskaya (approved):
Cool so cool

PR #2342, opened 2024-12-06
fix checking travel certificate availability locally

panasetskaya:
We checked if travel certificates "create certificate" button is enabled by checking locally. We have now BE for that (for quite some time, I guess)), and now I stumbled upon it when the button was not there for tiered contracts

### DIFF
--- a/app/feature/feature-travel-certificate/src/main/graphql/QueryCurrentContracts.graphql
+++ b/app/feature/feature-travel-certificate/src/main/graphql/QueryCurrentContracts.graphql
@@ -1,11 +1,7 @@
 query CurrentContracts {
   currentMember {
     activeContracts {
-      currentAgreement {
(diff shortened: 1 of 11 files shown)


app/feature/feature-travel-certificate/src/main/graphql/QueryCurrentContracts.graphql
```diff
@@ -1,11 +1,7 @@
 query CurrentContracts {
   currentMember {
     activeContracts {
-      currentAgreement {
-        productVariant {
-          typeOfContract
-        }
-      }
+      supportsTravelCertificate
```
StylianosGakis, line 4: Probably can also rename this file and query to show that we only are interested in this field now, not the rest of em

StylianosGakis (approved):

PR #2343, opened 2024-12-09
Feature/addon purchase UI

panasetskaya:
first iteration of addon flow with mocks


app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/success/AddonSuccessScreen.kt
```diff
@@ -0,0 +1,75 @@
...
+      buttonStyle = NoButton,
+    )
+    Spacer(Modifier.weight(1f))
+    HedvigTextButton(
+      stringResource(R.string.general_close_button),
+      onClick = dropUnlessResumed { popBackStack() },
+      buttonSize = Large,
+      modifier = Modifier.fillMaxWidth(),
+    )
+    Spacer(Modifier.height(32.dp))
```
StylianosGakis, line 67: 32 + insets may feel a bit much, perhaps 16 would be enough here.

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/summary/AddonSummaryDestination.kt
```diff
@@ -0,0 +1,287 @@
...
+    premium = uiState.summaryParameters.quote.price.toString(),
+    displayItems = uiState.summaryParameters.quote.addonVariant.displayDetails.map {
+      QuoteDisplayItem(
+        it.first,
+        null,
+        it.second,
+      )
+    },
+    underTitleContent = {
+      // todo: check here!
```
StylianosGakis, line 221: Perhaps just fine for it to be nothing here if design has nothing under the price

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/summary/AddonSummaryDestination.kt
```diff
@@ -0,0 +1,287 @@
...
+
+@Composable
+private fun SummaryCard(uiState: Content, modifier: Modifier = Modifier) {
+  val locale = getLocale()
+  val formattedDate = remember(uiState.summaryParameters.activationDate, locale) {
+    HedvigDateTimeFormatterDefaults.dateMonthAndYear(
+      locale,
+    ).format(uiState.summaryParameters.activationDate.toJavaLocalDate())
+  }
+  QuoteCard(
```
StylianosGakis, line 210: I tested here, and I got in a scenario where my hard had "Show details" but there was nothing in the details itself. I wonder if this is just something which should never happen, or if we should guard against this state and not show the button at all and not make this clickable if that state is passed in.
image

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
```diff
@@ -27,6 +27,7 @@ fun NavGraphBuilder.travelCertificateGraph(
   navigator: Navigator,
   applicationId: String,
   onNavigateToCoInsuredAddInfo: (String) -> Unit,
+  onNavigateToAddonPurchaseFlow: (List<String>) -> Unit,
```
StylianosGakis, line 30: If we need this to only be represented by a non-empty list of IDs, as you note in some comment else in the code, could this be a NonEmptyList instead?
If yes, can we still pass that as expected through Koin etc?
Just an idea, not necessary to do it if it complicates things too much.

StylianosGakis (approved):
Great work Maria! Left some comments, most are just ideas though. Feel free to merge when you want to

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/customize/CustomizeTravelAddonDestination.kt
```diff
@@ -0,0 +1,555 @@
...
+    Spacer(Modifier.height(32.dp))
+    HedvigTextButton(
+      text = stringResource(id = R.string.general_close_button),
+      buttonSize = Large,
+      onClick = { sheetState.dismiss() },
+      modifier = Modifier.fillMaxWidth(),
+    )
+    Spacer(Modifier.height(16.dp))
+    Spacer(Modifier.windowInsetsBottomHeight(WindowInsets.safeDrawing))
+// todo: here, check the padding!
```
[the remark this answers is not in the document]
panasetskaya: sorry for the"!", these todos were the notes to myself mostly:)

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/customize/CustomizeTravelAddonViewModel.kt
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+internal sealed interface CustomizeTravelAddonState {
+  data object Loading : CustomizeTravelAddonState
+
+  data class Success(
+    val travelAddonOffer: TravelAddonOffer,
+    val currentlyChosenOption: TravelAddonQuote,
+    val currentlyChosenOptionInDialog: TravelAddonQuote?,
```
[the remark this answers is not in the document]
panasetskaya: The bug there was just me using the wrong property yesterday, so no big thing there:) I think I did it like you're suggesting the first time with tiers, but I did not like how it turned out, don't remember exactly, but it sometimes behaved weird (was one recomposition late? or maybe smth else). This may be more reliable in the end - though more code as well. Wdyt?

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/summary/AddonSummaryDestination.kt
panasetskaya: Thought about this as well, but I think we should always get at least one of the documents there - TnA. So probably not worth safe-guarding, it's just a temporary staging thing now with the docs.
But may be we should log error if the docs list is empty?

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
panasetskaya: will try that!

panasetskaya: I think what may be better to do instead is to make TravelAddonBannerInfo eligibleInsurancesIds property a NonEmptyList, since this will be the source of banner everywhere. Then we won't show the banner at all and log the empty list as error if it would come from be

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/summary/AddonSummaryDestination.kt
StylianosGakis: Yeah a log could be good, but yeah let's keep it this way then

app/feature/feature-travel-certificate/src/main/kotlin/com/hedvig/android/feature/travelcertificate/navigation/TravelCertificateGraph.kt
StylianosGakis: Yeah that sounds good, do give it a try if you find the time :)

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
```diff
@@ -82,7 +83,7 @@ fun HedvigButton(
           )
         }
       } else {
-        HedvigText(text = text)
+        HedvigText(text = text, textAlign = TextAlign.Center)
```
panasetskaya, line 86: here

PR #2345, opened 2024-12-10
Feature/t&a termination

StylianosGakis:
| Before | After | After with sheet open |
| --- | --- | --- |
| image | image | image |

Screenshots don't show the "Extra coverage" but it's there too if backend returns it :D
From figma:
image


panasetskaya:
Just have one question, do you think there could be a situation where BE returns some extra coverage items, but we would want to not show them - with FF off? Or would it never be a case?

panasetskaya (approved):

PR #2071, opened 2024-04-07
Shared element for the top app bar in claim flow. Offpiste, playing with shared elements, do not merge.

StylianosGakis:

PR #2350, opened 2024-12-11
Make claim-flow navigation bar a shared element

StylianosGakis:
This makes this top app bar stick around as the video showcases

https://github.com/user-attachments/assets/0ca796cc-5d68-40b0-86ce-f55e7b3e17ab

This just wraps the entire app with a `SharedTransitionLayout`, and sets up the local.
And then uses the AnimatedContent local from the navigation itself to sync up the bar.

It's a successor to https://github.com/HedvigInsurance/android/pull/2071 but much cleaner since we use the locals to make it all work instead of polluting the entire codebase with scopes and such.


panasetskaya (approved):

StylianosGakis:
I'll merge this once I'm back next year, so that I can monitor better if anything comes up due to it, even though I don't expect it to be problematic

PR #2351, opened 2024-12-11
add go to chat button to error screen in the addon purchase flow

panasetskaya:


app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/customize/CustomizeTravelAddonDestination.kt
```diff
@@ -161,10 +175,16 @@ private fun FailureScreen(errorMessage: String?, reload: () -> Unit, popBackStac
         ),
     ) {
       Spacer(Modifier.weight(1f))
+      val buttonText = if (errorMessage == null) {
```
StylianosGakis, line 178: Since we for now only get this from a mock, there's no way to do it right yet, but it will be important in the real impl to only get here when we get a UserError specifically, and not any error. We might populate the `ErrorMessage` with some text in the error itself even in non `UserError` scenarios, just as a way to say what went wrong, and in those cases we would also accidentally show the open chat here instead.
So when we add the real impl, let's try to be careful with this

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/navigation/AddonPurchaseNavGraph.kt
```diff
@@ -72,6 +77,10 @@ fun NavGraphBuilder.addonPurchaseNavGraph(navigator: Navigator, navController: N
         navigateToSummary = { summaryParameters: SummaryParameters ->
           navigator.navigateUnsafe(Summary(summaryParameters))
         },
+        onNavigateToNewConversation = {
+          navController.typedPopBackStack<AddonPurchaseGraphDestination>(inclusive = true) // todo: is this the right thing?
```
StylianosGakis, line 81: This looks correct to me, can you think of cases where we don't want this to happen?

StylianosGakis (approved):

PR #2353, opened 2024-12-12
GEN-2988: T&A in moving flow summary screen

StylianosGakis:
| Before | After |
| --- | --- |
| image | image |

I re-used most of what was there to make it easier to work with.
When backend exists for this, we can populate this list.
If it turns out that what we get from backend does not match the shape of a `Quote` that can change.
If it turns out that it needs additional information, they can be added to `AddonQuote` as needed.
I made sure to use the QuoteCard overload where I pass in a null `contractGroup` so that the pillow image does not show.

All the saving/restoring from Datastore still should work since we take the object and store it as a Json, and the new addon is also @Serializable as it should.

The order in which they appear is also persistent since the list of addons are added to the quote itself, so they go right after the home quote itself, and before any other unrelated quotes.

### DIFF
--- a/app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
+++ b/app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
@@ -60,10 +60,12 @@ import com.hedvig.android.design.system.hedvig.HedvigText
 import com.hedvig.android.design.system.hedvig.HedvigTheme
 import com.hedvig.android.design.system.hedvig.HorizontalItemsWithMaximumSpaceTaken
 import com.hedvig.android.design.system.hedvig.NotificationDefaults.NotificationPriority.Info
+import com.hedvig.android.design.system.hedvig.NotificationDefaults.NotificationPriority.InfoInline
 import com.hedvig.android.design.system.hedvig.Surface
 import com.hedvig.android.design.system.hedvig.datepicker.HedvigDateTimeFormatterDefaults
 import com.hedvig.android.design.system.hedvig.datepicker.getLocale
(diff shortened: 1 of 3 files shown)


app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
```diff
@@ -193,6 +197,9 @@ private fun SummaryScreen(
       Spacer(Modifier.height(16.dp))
       Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
         QuoteCard(content.summaryInfo.moveHomeQuote)
+        for (addonQuote in content.summaryInfo.moveHomeQuote.relatedAddonQuotes) {
+          AddonQuoteCard(addonQuote)
+        }
         for (mtaQuote in content.summaryInfo.moveMtaQuotes) {
```
panasetskaya, line 203: could there also be addons for mta quotes in the future?

panasetskaya (approved):

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
StylianosGakis: Good question, let's ask I'd say. If the answer is yes, adding them here shouldn't be a big pr, probably just resting the same card here in the UI.

StylianosGakis: Okay, leaving this as-is for now according to the discussion we had in standup. If this comes up in the backend API, we'll do it then

PR #2369, opened 2025-01-10
add perils for addons

panasetskaya:
sorry for the commit's name :))


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/coverage/CoverageTab.kt
```diff
@@ -93,6 +98,33 @@ internal fun CoverageTab(
       )
       Spacer(Modifier.height(16.dp))
     }
+    if (addons?.isNotEmpty() == true) {
+      addons.forEach { addon ->
```
StylianosGakis, line 102:
```suggestion
    for (addon in addons.orEmpty()) {
```
I think this could be just this in short, since if it's null or empty it will simply not loop at all.

StylianosGakis (approved):

PR #2370, opened 2025-01-13
GEN-3119: Add support for disabling addons in moving flow behind feature flag

StylianosGakis:
Missing the connection with backend, this only takes care of the local handling we need to do in the clients

We need to propagate this information to the backend for this PR to make sense to merge, otherwise the buttons will not do what they should, and it would be super wrong behavior to do so. We wait with the merge of this until then

[remove_addon.webm](https://github.com/user-attachments/assets/93de59ba-6c98-4d29-9f68-7f7f8d7b0f1e)


app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
```diff
@@ -200,71 +208,16 @@ private fun SummaryScreen(
...
-            val bottomSheetTitle = addon.addonVariant.displayName
-            AddonQuoteCard(
-              quote = addon,
-              onInfoIconClick = {
-                infoFoBottomSheet =
-                  bottomSheetTitle to description
-              },
-            )
+          for (addonQuote in mtaQuote.relatedAddonQuotes) {
+            AddonQuoteCard(quote = addonQuote, toggleHomeAddonExclusion = { toggleHomeAddonExclusion(addonQuote) })
```
panasetskaya, line 220: here we won't be able to toggle mta quotes' addons, I think. so maybe for them we should better use the old type of AddonQuoteCard, with info button? Or even no button at all, because they are not really replaced, those insurances, they are just mid-term adjusted, so their addons stay as they are

panasetskaya (approved):

panasetskaya (approved):

sladan-hedvig:
@StylianosGakis according to the [design](https://www.figma.com/design/SNVLvLq7ztclXiDqqMNXgY/Tiers-%26-Addons-(App)?node-id=3907-42728&t=XJnRO6TZJQiz8pHC-0) we don't have `Add coverage` and `Show/Hide details` in the same screen. I was following that one when implementing. Not sure should we include it on the iOS also?

PR #2373, opened 2025-01-15
Fix/language default wrong

panasetskaya:


app/language/language-core/src/main/kotlin/com/hedvig/android/language/Language.kt
```diff
@@ -47,20 +53,30 @@ enum class Language {
...
-      BCP_47_EN_DK -> EN_DK
+    fun from(value: String): Language = when {
+      value == BCP_47_SV_SE -> SV_SE
+      value == BCP_47_EN_SE -> EN_SE
+      value == BCP_47_NB_NO -> NB_NO
+      value == BCP_47_EN_NO -> EN_NO
+      value == BCP_47_DA_DK -> DA_DK
+      value == BCP_47_EN_DK -> EN_DK
+      value == BCP_47_EN || value.startsWithSubstring(BCP_47_EN_FOR_SUBSTRING) -> EN_GLOBAL
+      value == BCP_47_SV || value.startsWithSubstring(BCP_47_SV_FOR_SUBSTRING) -> SV_GLOBAL
```
StylianosGakis, line 74: I see yeah, I think this is a great idea, let's see how this fares!
Should we also add a log over here to catch the cases when this happens, to see what other values we're getting and how often? I am mostly curious, and we can then see if it aligns with the people who have had issues with this.

app/market/market-set/src/main/kotlin/com/hedvig/android/market/set/SetMarketUseCase.kt
```diff
@@ -28,6 +28,7 @@ internal class SetMarketUseCaseImpl(
       languageService.setLanguage(
         when {
           existingLanguage == null -> market.localLanguage()
+          existingLanguage.isSwedishLanguage() -> Language.SV_SE
```
StylianosGakis, line 31: Was this case not covered well by the `else` below? Did we ever get wrong results from it perhaps?

app/language/language-android/src/main/kotlin/com/hedvig/android/language/AndroidLanguageService.kt
```diff
@@ -39,9 +39,9 @@ internal class AndroidLanguageService() : LanguageService {
   }
 
   private fun getLocaleFromAppCompat(): Locale? {
-    val localeList = AppCompatDelegate.getApplicationLocales()
+    val localeList = AppCompatDelegate.getApplicationLocales() // here: weirdly depends on cache
     return if (localeList.isEmpty) {
-      null
+      Locale.getDefault()
```
StylianosGakis, line 44: Changing this would mean that this function never returns null, so `getSelectedLanguage()` would also never return null, which it was doing so in some scenarios so that it would fallback to some other behavior.
Like here: https://github.com/HedvigInsurance/android/blob/ca10ff163cc89d50a307431eefce3e83cf09b811/app/market/market-set/src/main/kotlin/com/hedvig/android/market/set/SetMarketUseCase.kt#L28-L36
Which would then call `LanguageService.setLanguage()` which in turn would call `AppCompatDelegate.setApplicationLocales(LocaleListCompat.forLanguageTags(language.toString()))` which would set the locale for future app launches too.
Otherwise I suppose we may never call `AppCompatDelegate.setApplicationLocales` unless they explicitly change the marker or the language itself, and I am not sure if we want to not call that. I think we do want to call that to set the language choice for real, so future calls to `AppCompatDelegate.getApplicationLocales()` will return the language we chose.

StylianosGakis, line 44: This would also make the distinction between `getLocale` and `getSelectedLocale` be more weird. Since before, "selected" variant returned null if we did not know what it was. Now it would return the system default while the other variant would default to "en-SE" but it'd be quite unclear regarding why the behavior is the way it is and why would one use one over the other

StylianosGakis (commented):
My question is mostly around the defaulting inside `getLocaleFromAppCompat` since I think it has a bit bigger implications other than just picking the language better from other non-standard locales. It'd be nice to hear what the idea is behind that.

app/app/src/main/kotlin/com/hedvig/android/app/MainActivity.kt
```diff
@@ -244,3 +247,11 @@ private fun Activity.tryShowAppStoreReviewDialog() {
     }
   }
 }
+
+private fun getSystemLocale(config: android.content.res.Configuration): Locale {
+  return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+    Resources.getSystem().configuration.locales[0]
+  } else {
+    config.locale
+  }
```
StylianosGakis, line 256: Could we use `ConfigurationCompat.getLocales(Resources.getSystem().configuration)` here which as I understand does the same thing internally here anyway while allowing us to avoid any API version checks?

StylianosGakis (approved):
Thanks, I now hate languages 😂

app/app/src/main/kotlin/com/hedvig/android/app/MainActivity.kt
panasetskaya: it gives us nullable Locale, if I remember correctly

PR #2379, opened 2025-01-20
Change bottom sheet info to not come from backend

StylianosGakis:
image


app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/summary/SummaryDestination.kt
```diff
@@ -247,8 +246,8 @@ private fun SummaryScreen(
             quote = addonQuote,
             canExcludeAddons = content.canExcludeAddons,
             toggleHomeAddonExclusion = {
-              if (addonQuote.exclusionDialogInfo != null && !addonQuote.isExcludedByUser) {
-                exclusionBottomSheetState.show(addonQuote.exclusionDialogInfo)
+              if (!addonQuote.isExcludedByUser) {
+                exclusionBottomSheetState.show(ExclusionDialogInfo(addonQuote.addonId, addonQuote.exposureName))
```
panasetskaya, line 250: the thing that confuses me a bit about addons is their display name. We have display name of addon quote and display name i addon variant. Are they the same or different?

panasetskaya (approved):

PR #2392, opened 2025-01-24
Fix/UI fixes addons

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/AddonBanner.kt
```diff
@@ -26,7 +26,7 @@ fun FeatureAddonBanner(
   onButtonClick: () -> Unit,
   modifier: Modifier = Modifier,
 ) {
-  val containerColor = HedvigTheme.colorScheme.fillNegative
+  val containerColor = HedvigTheme.colorScheme.borderPrimary
```
StylianosGakis, line 29: I know this is up to spec, but using the border color as a container color is a bit sus.
I'll bring this up with design just to see if this is temporary or not.

StylianosGakis: https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1737721490406599

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/yourinfo/YourInfoTab.kt
```diff
@@ -156,8 +156,6 @@ internal fun YourInfoTab(
         )
       }
     }
-
-    Spacer(Modifier.height(8.dp))
```
StylianosGakis, line 160: Was this part of the UI review too? From looking at the code surrounding this it feels like the coverage rows  below would then be too close to the optional notification card  above no?

StylianosGakis (approved):

PR #2393, opened 2025-01-24
Add image viewer for images in chat

StylianosGakis:
Clicking on any image should take one to the full screen version of it The image is also zoomable


panasetskaya (approved):
Can I ask why did you prefer to make it a separate destination and not just a UI thing (like we have for cross sells)?

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/ChatLoadedScreen.kt
```diff
@@ -658,8 +675,8 @@ private fun ChatAsyncImage(
         } else {
           Modifier
         },
-      )
-      .clip(HedvigTheme.shapes.cornerLarge),
+      ),
```
panasetskaya, line 678: should we delete the commented out clipping?

PR #2401, opened 2025-01-30
Feature/chat uploading indication

panasetskaya:


StylianosGakis (approved):
`var showUploading by remember { mutableStateOf(false) }`
should probably instead be
`val numberOfOngoingUploads by remember { MutableStateFlow(0) }`
and I will try to explain why.

Right now think of the following two possible scenarios which break this functionality:
1. Member hits send for a media file which is going to take 5 seconds to finish. `showUploading` is now true
2. Member now also sends a text message soon after, `showUploading` is set to true again
3. Text message finishes quickly, and it sets `showUploading` to false
4. The mp4 is still uploading, yet we have stopped showing the loading indicator.
This scenario results in us hiding the indicator despite there still being uploading works ongoing.

---

Let's go over this scenario again now if instead of doing `showUploading = true` we did `numberOfOngoingUploads.update { it + 1 }` and when it was over we did `numberOfOngoingUploads.update { it - 1 }`.
And in line 297, we changed `showUploading = showUploading,` to (I think) `showUploading = numberOfOngoingUploads.collectAsState() > 0`

1. Member hits send for a media file which is going to take 5 seconds to finish. `numberOfOngoingUploads` is now set to 1
2. Member now also sends a text message soon after, `numberOfOngoingUploads` is set to 2
3. Text message finishes quickly, and it decrements `numberOfOngoingUploads` to 1
4. The mp4 eventually ends, and decrements `numberOfOngoingUploads` to 0
5. Now that it's 0, the `showUploading` is derived back to `false` since the value is 0

StylianosGakis (approved):
`var showUploading by remember { mutableStateOf(false) }`
should probably instead be
`val numberOfOngoingUploads by remember { MutableStateFlow(0) }`
and I will try to explain why.

Right now think of the following scenario which I think breaks this functionality:
1. Member hits send for a media file which is going to take 5 seconds to finish. `showUploading` is now true
2. Member now also sends a text message soon after, `showUploading` is set to true again
3. Text message finishes quickly, and it sets `showUploading` to false
4. The mp4 is still uploading, yet we have stopped showing the loading indicator.
This scenario results in us hiding the indicator despite there still being uploading works ongoing.

---

Let's go over this scenario again now if instead of doing `showUploading = true` we did `numberOfOngoingUploads.update { it + 1 }` and when it was over we did `numberOfOngoingUploads.update { it - 1 }`.
And in line 297, we changed `showUploading = showUploading,` to (I think) `showUploading = numberOfOngoingUploads.collectAsState() > 0`

1. Member hits send for a media file which is going to take 5 seconds to finish. `numberOfOngoingUploads` is now set to 1
2. Member now also sends a text message soon after, `numberOfOngoingUploads` is set to 2
3. Text message finishes quickly, and it decrements `numberOfOngoingUploads` to 1
4. The mp4 eventually ends, and decrements `numberOfOngoingUploads` to 0
5. Now that it's 0, the `showUploading` is derived back to `false` since the value is 0

PR #2407, opened 2025-02-03
add pdf decoder

panasetskaya:
I first started to implement a bigger solution where we actually rendered a pdf file inside the app on a separate screen as a lazy list of pages, but then I thought that we maybe need just a preview of the pdf file in the chat itself for better readability, and the actual pdf opening is probably better in the browser for the same reasons you mentioned when we talked about opening image in browser. Then a decoder for coil (first page of the pdf) would already suffice for this purpose. 
Pls let me know if you agree, if not - I'll continue with separate screen whole file rendering :)


StylianosGakis (approved):
Yeah looks good to me, the full PDF solution might come down the line when https://developer.android.com/jetpack/androidx/releases/pdf gets a stable release and we can tell if it does what we want or not.

PR #2411, opened 2025-02-04
Fix/chat persist scroll position when coming back to it

StylianosGakis:
### Fix chat resetting the scroll position when navigating back to it

By moving the flow creation and caching (through cachedIn) to the
ViewModel instance, the data is persisted in a way that it is available
on the first frame. This, in combination with the LazyListState
restoring its scroll position results in everything working as expected.

---

### Allow overriding the CoroutineScope passed into MoleculeViewModel

This makes use of the new(ish) constructor of ViewModel which accepts a
a CoroutineScope which can then later be accessed through the
`ViewModel.viewModelScope` extension function

---

### Stop scrolling to the bottom of the chat on the first frame

We simply drop the first ever "latest message" as it is the one which
appears on the first time any messages are rendered on the screen.
In the cases where we just came to this screen, we will have been at the
beginning of the list regardless. And in the scenarios where we come
back to this screen which has existed before we instead want to stay at
the scroll position we had in the past


panasetskaya (approved):
seems cool!

PR #2412, opened 2025-02-05
add HedvigHttpLoggingInterceptor

panasetskaya:
could you please take a look here? it's the exact copy of the Okhttp interceptor but with buffer compared to MAX_LOG_SIZE. In my case (big files uploads) the app stops crashing with these changes (but the logger.level is still .BODY), but there is maybe smth I miss here

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
@@ -1,0 +1,302 @@
+package com.hedvig.android.app.logginginterceptor
+
+import java.io.EOFException
+import java.io.IOException
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.TreeSet
+import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
```diff
@@ -0,0 +1,302 @@
...
+      } else {
+        val buffer = Buffer()
+        requestBody.writeTo(buffer)
+
+        val contentType = requestBody.contentType()
+        val charset: Charset = contentType?.charset(UTF_8) ?: UTF_8
+
+        logger.log("")
+        if (buffer.size > MAX_LOG_SIZE) {
+          logger.log("--> END ${requestBody.contentLength()}-byte body")
```
StylianosGakis, line 192: So that's interesting, it is able to figure out what the size of the buffer is at this point, but it fails to do the `readString` on it since it is only at that point that it tries to put everything in memory. I don't quite understand how this works internally but if it can know about its size then we are all good here. I doubt we'd ever want to log these huge contents in any case, so simply omitting them entirely sounds like a good workaround for our needs!

Great job looking into this!

StylianosGakis (approved):

PR #2414, opened 2025-02-06
GEN-3226 Feature/travel addon deeplink

panasetskaya:
- [x]  need to add a new key [slack_link](https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1738751099665359?thread_ts=1738678338.784249&cid=C03U9C6Q7TP) at TravelAddonTriageDestination.kt.118

not sure about navigateUp vs popBackStack here

### DIFF
--- a/app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/triage/TravelAddonTriageDestination.kt
+++ b/app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/triage/TravelAddonTriageDestination.kt
@@ -21,13 +21,11 @@ import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.datasource.CollectionPreviewParameterProvider
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.compose.dropUnlessResumed
 import com.hedvig.android.design.system.hedvig.ButtonDefaults.ButtonSize.Large
-import com.hedvig.android.design.system.hedvig.EmptyState
-import com.hedvig.android.design.system.hedvig.EmptyStateDefaults.EmptyStateButtonStyle.NoButton
-import com.hedvig.android.design.system.hedvig.EmptyStateDefaults.EmptyStateIconStyle.INFO
 import com.hedvig.android.design.system.hedvig.HedvigErrorSection
 import com.hedvig.android.design.system.hedvig.HedvigFullScreenCenterAlignedProgress
-import com.hedvig.android.design.system.hedvig.HedvigMultiScreenPreview
+import com.hedvig.android.design.system.hedvig.HedvigPreview
 import com.hedvig.android.design.system.hedvig.HedvigTextButton
 import com.hedvig.android.design.system.hedvig.HedvigTheme
 import com.hedvig.android.design.system.hedvig.Surface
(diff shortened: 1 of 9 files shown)


app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/triage/TravelAddonTriageDestination.kt
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    Loading -> HedvigFullScreenCenterAlignedProgress()
+
+    is Success -> {
+      LaunchedEffect(uiState.insuranceIds) {
+        val params = uiState.insuranceIds
+        launchFlow(params)
+      }
+    }
```
StylianosGakis, line 81: In the success case here we'd be emitting no UI at all.
Without knowing how this looks like, often this has led to animation glitches in navigation since we go from taking up 0x0 size to taking up the full size, and this size change is done with an animation by expanding the content from the center or top-left which is not what we want.

Can we either add an empty Box() in here, or even better wrap the entire `when` statement here in a `Surface(color = ...background, modifier = Modifier.fillMaxSize()) { when(uiState)... }` so that we always do emit at least the background properly?

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/triage/TravelAddonTriageDestination.kt
```diff
@@ -0,0 +1,165 @@
...
+              ),
+            ),
+        ) {
+          Spacer(Modifier.weight(1f))
+          val buttonText = stringResource(R.string.open_chat)
+
+          HedvigErrorSection(
+            onButtonClick = onNavigateToNewConversation,
+            subTitle = stringResource(R.string.GENERAL_ERROR_BODY),
+            // todo: another key here! ask Richard
```
StylianosGakis, line 119: Is this decided by now?

app/feature/feature-addon-purchase/src/main/kotlin/com/hedvig/android/feature/addon/purchase/ui/triage/TravelAddonTriageDestination.kt
```diff
@@ -0,0 +1,165 @@
...
+            modifier = Modifier.fillMaxSize(),
+            buttonText = buttonText,
+          )
+          Spacer(Modifier.weight(1f))
+          HedvigTextButton(
+            stringResource(R.string.general_close_button),
+            onClick = dropUnlessResumed { popBackStack() },
```
StylianosGakis, line 126: So from a button which normally takes you back one item in the backstack, but it is *not* the "up" arrow on the top-left of the screen, then what we must do is call this function here https://github.com/HedvigInsurance/android/blob/1f6e80cbcdea079155a8b83593b4815971989d25/app/app/src/main/kotlin/com/hedvig/android/app/navigation/RememberNavigator.kt#L33-L37
Easiest to pass down the `popBackStack` function from this navigator https://github.com/HedvigInsurance/android/blob/761a99440cba8211540210bf447f3ce729bf5282/app/app/src/main/kotlin/com/hedvig/android/app/navigation/HedvigNavHost.kt#L95 down to this screen, and call that one on back from this button.
Since this screen does not seem to have a top app bar which we'd call navigateUp() from the top-left arrow, we probably just don't need to pass it down here at all in the first place.

StylianosGakis (approved):
A note about `navigateUp` and `popBackStack` here where we need to call a different function.
And about always emitting some Ui from the screen.
All good otherwise 😊

PR #2416, opened 2025-02-07
Feature/chat multiple doc and videoplayer 48mb limit

panasetskaya:
Filter out files that are too big before we start the upload with hardcoded limit. Later remove when BE will change their implementation to fail immediately when a big file is sent.
[link to slack discussion](https://hedviginsurance.slack.com/archives/C9K7XRQ0G/p1738766959269399)

- [ ] translate key: CHAT_FILE_SIZE_TOO_BIG_ERROR


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/CbmChatRepository.kt
```diff
@@ -351,8 +362,17 @@ internal class CbmChatRepositoryImpl(
   }
 
   private suspend fun Raise<String>.uploadMediaToBotService(uri: Uri): String {
+    val fileSize = contentResolver.openFileDescriptor(uri, "r")?.use {
+      it.statSize
+    } ?: raise("Could not determine file size for uri: $uri")
```
StylianosGakis, line 367: Do we want to exit out eagerly if we can't figure out the size, even though it might've been just fine to upload the file anyway if it was not that big?
Do we ever expect `statSize` to fail here for some reason, or are we just being careful just in case?

StylianosGakis (approved):

PR #2417, opened 2025-02-07
try open file picker on failed media resend

panasetskaya:


StylianosGakis (changes requested):
I think there is a solution after all. It's this https://developer.android.com/training/data-storage/shared/documents-files#persist-permissions right here which seems to be made for our use case of persisting the access to the file across app restarts or phone restarts.

Can we give it a try before we see if we need this PR as a replacement?

StylianosGakis:
Superseded by https://github.com/HedvigInsurance/android/pull/2442

PR #2418, opened 2025-02-10
Feature/videoplayer oom

panasetskaya:
As far as I understand, what happens here is that if we try to upload a too big file that exceeds the whole app JVM memory (maybe we even should use `getRuntime().totalMemory()` instead of `.maxMemory()` here - not sure), 
then (despite that I tried as you see to buffer writing in the fileService - it does not help, I'm not sure if I should revert it as it was or is it still better than writing it all in the sink in one go?) the system starts urgently shutdown everything, and we can't even catch this OOM error bc there's nothing left that can catch it, so to speak :) 
So here I'm making this file size check before we even start.

Combined with this [user-friendly file size limit PR](https://github.com/HedvigInsurance/android/pull/2416) I think we're probably good. That linked PR has temporary hard-coded limitation according to our current BE requirements + user-friendly error handling.

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
@@ -182,20 +182,24 @@ class HedvigHttpLoggingInterceptor @JvmOverloads constructor(
         logger.log("--> END ${request.method} (one-shot body omitted)")
       } else {
         val buffer = Buffer()
-        requestBody.writeTo(buffer)
+        try {
+          requestBody.writeTo(buffer)
+          val contentType = requestBody.contentType()
+          val charset: Charset = contentType?.charset(UTF_8) ?: UTF_8
 
-        val contentType = requestBody.contentType()
-        val charset: Charset = contentType?.charset(UTF_8) ?: UTF_8
-
-        logger.log("")
-        if (buffer.size > MAX_LOG_SIZE) {
-          logger.log("--> END ${requestBody.contentLength()}-byte body")
-        } else if (buffer.isProbablyUtf8()) {
-          logger.log(buffer.readString(charset))
-          logger.log("--> END ${request.method} (${requestBody.contentLength()}-byte body)")
-        } else {
+          logger.log("")
+          if (buffer.size > MAX_LOG_SIZE) {
+            logger.log("--> END ${requestBody.contentLength()}-byte body")
+          } else if (buffer.isProbablyUtf8()) {
+            logger.log(buffer.readString(charset))
+            logger.log("--> END ${request.method} (${requestBody.contentLength()}-byte body)")
+          } else {
+            logger.log(
+              "--> END ${request.method} (binary ${requestBody.contentLength()}-byte body omitted)")
+          }
+        } catch (e: OutOfMemoryError) {
           logger.log(
-            "--> END ${request.method} (binary ${requestBody.contentLength()}-byte body omitted)")
+          "--> END ${request.method} (binary ${requestBody.contentLength()}-byte body omitted) with OutOfMemoryError: $e")
         }
       }
     }
--- a/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
@@ -20,13 +20,14 @@ import okio.GzipSource
 private const val MAX_LOG_SIZE = 100 * 1024
 
 class HedvigHttpLoggingInterceptor @JvmOverloads constructor(
-  private val logger: Logger = Logger.DEFAULT
+  private val logger: Logger = Logger.DEFAULT,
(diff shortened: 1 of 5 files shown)


app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
```diff
@@ -24,9 +24,24 @@ class FileService(
...
       override fun writeTo(sink: BufferedSink) {
-        contentResolver.openInputStream(uri)?.use { inputStream ->
-          sink.writeAll(inputStream.source())
-        } ?: error("Could not open input stream for uri:$uri")
+        if (fileDoesNotExceedMemory(uri)) {
+          contentResolver.openInputStream(uri)?.use { inputStream ->
+            val buffer = ByteArray(4 * 1024)
+            var bytesRead: Int
+            val outputStream = sink.outputStream()
+            try {
```
StylianosGakis, line 32: Since before we were doing `writeAll` which has the implementation here https://github.com/square/okio/blob/86043a419d328742889b1d55ab159cc7701ccddb/okio/src/commonMain/kotlin/okio/internal/Buffer.kt#L1084-L1092

The impl here looks a bit different, with an extra buffer which is used to put the data temporarily as far as I can tell. And the buffer size looks to be half of what `writeAll` seems to do inside okio.
Perhaps we can copy-paste their impl here?

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
```diff
@@ -24,9 +24,24 @@ class FileService(
...
+              while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                outputStream.write(buffer, 0, bytesRead)
+                outputStream.flush()
+              }
+            } catch (e: OutOfMemoryError) {
+              System.gc()
```
StylianosGakis, line 38: From reading about this online, it seems like calling this gives very few guarantees on what it will do or not.
I really feel like there must be a better approach here rather than consuming all of our memory, catching the exception and trying to run GC to fix it.
I suggest we try to run through some more examples on how to do this properly before merging this because I don't think this will work at least in most scenarios.

StylianosGakis (changes requested):
Let's wait a bit with merging this one, I'd like to run some more experiments around this because I feel like there is a happy path we can take here where everything is chunked properly and we are able to try and upload things without exhausting all available memory and only then catching the problem.

Let's revisit this asap after we've done the other necessary merges + release for next week's release.

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
```diff
@@ -73,4 +88,21 @@ class FileService(
   }
 
   private fun getFileExtension(path: String): String = MimeTypeMap.getFileExtensionFromUrl(path)
+
+  fun fileDoesNotExceedMemory(uri: Uri): Boolean {
```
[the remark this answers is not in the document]
panasetskaya: Yeah, but I think the one with 48 is meant as just a temporary solution.

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
panasetskaya: That really doesn't work tbh, it's just a leftover of the things I've tried. Should've removed

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
panasetskaya: I think the idea here was to chunk it to smaller bits throughout smaller buffers. But again, didn't help with this particular problem, just smth I tried (but didn't worsen anything either). If we do earlier check with getRuntime().totalMemory() we might as well go back to writeAll tbh

app/app/src/main/kotlin/com/hedvig/android/app/logginginterceptor/HedvigLoggingInterceptor.kt
```diff
@@ -182,20 +184,26 @@ class HedvigHttpLoggingInterceptor @JvmOverloads constructor(
         logger.log("--> END ${request.method} (one-shot body omitted)")
       } else {
         val buffer = Buffer()
-        requestBody.writeTo(buffer)
-
-        val contentType = requestBody.contentType()
-        val charset: Charset = contentType?.charset(UTF_8) ?: UTF_8
+        try {
+          requestBody.writeTo(buffer)
```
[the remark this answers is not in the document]
panasetskaya: honestly it don't even come to this point with earlier checks

app/core/core-file-upload/src/main/kotlin/com/hedvig/android/core/fileupload/FileService.kt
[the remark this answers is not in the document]
panasetskaya: This check is the main thing, and the other can be reverted here to what it was tbh with no harm done

PR #2419, opened 2025-02-10
GEN-3228 Feature/coinsured deeplink

panasetskaya:
Deeplink to edit coInsured, with and without arguments.

Not sure what to do with navigation here, again. if I move triage into the graph and make it graph's startDestination, scaffold back Icon drom missinginfo/addOrRemove destination recreates the startDestination, so I have to press button twice; if I move outside the graph, set startDestination back to AddMissingInfo and get there by deeplink with no arguments, it crashes bc it tries to recreate graph with no argument in the start destination.

### DIFF
--- a/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/GetInsurancesForEditCoInsuredUseCase.kt
+++ b/app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/GetInsurancesForEditCoInsuredUseCase.kt
@@ -58,5 +58,5 @@ internal data class InsuranceForEditOrAddCoInsured(
 
 internal enum class EditCoInsuredDestination {
   MISSING_INFO,
-  ADD_OR_REMOVE
+  ADD_OR_REMOVE,
 }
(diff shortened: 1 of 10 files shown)


app/feature/feature-edit-coinsured/src/main/kotlin/com/hedvig/android/feature/editcoinsured/data/GetInsurancesForEditCoInsuredUseCase.kt
```diff
@@ -0,0 +1,62 @@
...
+          add(
+            InsuranceForEditOrAddCoInsured(
+              destination = destination,
+              displayName = contract.currentAgreement.productVariant.displayName,
+              exposureName = contract.exposureDisplayName,
+              id = contract.id,
+            ),
+          )
+        }
+      }
```
StylianosGakis, line 47:
```suggestion
      filtered.map { contract ->
        val destination = if (contract.coInsured?.any { it.hasMissingInfo } == true) {
          EditCoInsuredDestination.MISSING_INFO
        } else {
          EditCoInsuredDestination.ADD_OR_REMOVE
        }
        InsuranceForEditOrAddCoInsured(
          destination = destination,
          displayName = contract.currentAgreement.productVariant.displayName,
          exposureName = contract.exposureDisplayName,
          id = contract.id,
        )
      }
```
Just a bit less typing by using the map directly instead of building a list where all items are included anyway

StylianosGakis (commented):
So a question here.
Only `EditCoInsuredTriage` seems to have deep links attached to it, the others can only be reached through normal app navigation.
Is there a possibility that the graph could look like this instead:
```
destination: EditCoInsuredTriage
graph: EditCoInsuredGraphDestination
 \
  \
   -> destination: CoInsuredAddInfo
   |
   -> destination: CoInsuredAddOrRemove
```
So that if you deep link to `EditCoInsuredTriage` you are not yet inside the graph. And if you navigate to the graph, you do so only by providing an ID. And navigating to it from the deep linkable destination would only be done by also doing a popUpTo itself inclusive.

In fact, if the graph `EditCoInsuredGraphDestination` can not exist without an ID anyway, could we flatten all 3 destinations here so that they are not inside some sort of sub-graph of their own?
That way if someone does navigateUp() on them, they will be redirected to the Root graph, aka Home screen instead.

PR #2427, opened 2025-02-13
base premium instead of premium for change tier

panasetskaya:
base premium for comparison with previous price (instead of total price with addons). [link to slack](https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1739439606409599)


StylianosGakis (approved):
🔥

PR #2432, opened 2025-02-14
RND-1139: Automatically refresh after updating

hugokallstrom:
Refresh insurance detail view after modifying co-insured. Already works in travel certificate flow so no changes needed there, however backend takes a couple of seconds to update so if you are too fast in completing the flow no updates will be seen.

### DIFF
--- a/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/GetContractForContractIdUseCase.kt
+++ b/app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/GetContractForContractIdUseCase.kt
@@ -10,38 +10,34 @@ import com.hedvig.android.feature.insurances.data.InsuranceContract
 import com.hedvig.android.logger.LogPriority
 import com.hedvig.android.logger.logcat
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 
 internal interface GetContractForContractIdUseCase {
-  fun invoke(contractId: String): Flow<Either<GetContractForContractIdError, InsuranceContract>>
+  suspend fun invoke(contractId: String): Flow<Either<GetContractForContractIdError, InsuranceContract>>
 }
 
 internal class GetContractForContractIdUseCaseImpl(
   private val getInsuranceContractsUseCaseProvider: Provider<GetInsuranceContractsUseCase>,
 ) : GetContractForContractIdUseCase {
-  override fun invoke(contractId: String): Flow<Either<GetContractForContractIdError, InsuranceContract>> {
-    return flow {
-      getInsuranceContractsUseCaseProvider
-        .provide()
-        .invoke(forceNetworkFetch = false)
-        .map { insuranceContractResult ->
-          either {
-            val contract = insuranceContractResult
-              .mapLeft { GetContractForContractIdError.GenericError(it) }
-              .bind()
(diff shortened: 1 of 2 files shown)


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/GetContractForContractIdUseCase.kt
```diff
@@ -10,38 +10,34 @@ import com.hedvig.android.feature.insurances.data.InsuranceContract
...
-            ensureNotNull(contract) {
-              GetContractForContractIdError.NoContractFound(
-                ErrorMessage("No contract found with id: $contractId").also {
-                  logcat(LogPriority.ERROR) { it.message.toString() }
-                },
-              )
-            }
+  override suspend fun invoke(contractId: String): Flow<Either<GetContractForContractIdError, InsuranceContract>> {
+    return getInsuranceContractsUseCaseProvider
+      .provide()
```
StylianosGakis, line 24: I'd suggest we keep the old API with the function returning a `Flow` but without it itself being `suspend`
I know it looked like it was a leftover from some older code, but it was like that on purpose.

So the difference is that:

---

* With a `suspend` fun that returns flow 

When we first call `invoke` here, we run `getInsuranceContractsUseCaseProvider.provide()` at call time, and then get back a `Flow` which is run on this instance we have gotten from this `provide()` method.
If the following sequence happens:
* We run this code, and we get a flow back
* We do `collect()` on this flow, and we get the data that we need
* At this point, we stop collecting for whatever reason.
* We switch to/from demo mode, so the `provide()` function would now return a different instance of this use case
* We start collecting again from this flow. But since the `provide()` was done only once on the first time we called this function, we are now going to be collecting from the old use case which is not the one that the provider would give us now.

---

* With a non `suspend` fun that returns flow as we had it before

In contrast, with the `provide()` being done *inside* the flow itself, it means that whenever there is a new caller who is going to be collecting this flow, they will first things first call the provide() and then continue with the rest of the work.
So any new collector will for sure get the most up to date class provided from the provider.

---

This is not something which will realistically happen here, however any function which is both suspend and returns a Flow is a smell that you are setting up yourself for very very subtle potential bugs like this, so we've been trying to keep our eyes open to fix any cases where we do it.

StylianosGakis (approved):
A little (not so little) note, which as a tl;dr says that it would be nice to keep the old `fun invoke(contractId: String): Flow<Either<GetContractForContractIdError, InsuranceContract>>` signature and not add the `suspend` in there since it theoretically leaves the door open for subtle bugs.

app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/GetContractForContractIdUseCase.kt
hugokallstrom: I see, that is very subtle indeed! Thanks for explaining 🙏

panasetskaya: it was my suggestion to simplify this:) but regarding this, Stelios: can you maybe take a look at this problem? it doesn't work as a flow (neither with old impl nor with this new one - with forceFetch), fetching gets broken somewhere in the chain. It does the refresh only if we exceed the viewModel timeout, it seems, so the loadIteration gets triggered again.

PR #2438, opened 2025-02-17
Fix/tier deeplink

panasetskaya:
change-tier deeplink

### DIFF
--- a/app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
+++ b/app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
@@ -157,4 +157,5 @@ val HedvigDeepLinkContainer.allDeepLinkUriPatterns: List<String>
     travelAddon.first(),
     travelCertificate.first(),
     changeTierWithoutContractId.first(),
+    changeTierWithContractId.first(),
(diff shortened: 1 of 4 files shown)


app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -150,4 +156,5 @@ val HedvigDeepLinkContainer.allDeepLinkUriPatterns: List<String>
     conversation.first(),
     travelAddon.first(),
     travelCertificate.first(),
+    changeTierWithoutContractId.first(),
```
StylianosGakis, line 159:
```suggestion
    changeTierWithoutContractId.first(),
    changeTierWithContractId.first(),
```
I think it'd then be able to match both, but I'd need to test to be sure. In any case I think it should be safe to commit this change anyway

StylianosGakis (approved):

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
panasetskaya: yes, absolutely, thanks for catching

PR #2441, opened 2025-02-18
terminate-insurance deeplink with no arg 

panasetskaya:


app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
```diff
@@ -83,7 +83,7 @@ private fun ChooseInsuranceToTerminateScreen(
   when (uiState) {
     ChooseInsuranceToTerminateStepUiState.NotAllowed -> {
       HedvigScaffold(
-        navigateUp = navigateUp,
+        navigateUp = popBackStack,
```
StylianosGakis, line 86: Why did we change that here in this pr? 👀

StylianosGakis (approved):

StylianosGakis:
Actually a question. Would this work if we only added 1 deep link, the one with the ID and the `= null` change that we did?
Could we test this out perhaps? My understanding of it is that since it's nullable and has a default it should also match the deep link properly.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/choose/ChooseInsuranceToTerminateDestination.kt
panasetskaya: you're right about the link itself being redundant, just default null argument for destination works.

StylianosGakis: > bc with navigateUp we didn't go back to the previous app if we come from the deeplink on topAppBar icon click, so system back button and topAppBar back icon had different behaviour.

Yeah that is what we want here as per https://developer.android.com/guide/navigation/principles#the_up_button_never_exits_your_app

panasetskaya: Right! What about the "close" textButton etc? In one of the previous pr we put popBackStack there

PR #2445, opened 2025-02-19
Feature/termination move text to bottom

StylianosGakis:
This also fixes the merge conflicts etc and thus supersedes https://github.com/HedvigInsurance/android/pull/2439. When this is merged that one will automatically be considered merged too

Had to change the model a bunch to to bunch up the termination reasons and the feedback text linked to them in a map since now the feedback is singular for the entire screen.
The rest of the changes in the tests etc were to account for this shift in how this screen works


app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/survey/TerminationSurveyViewModel.kt
```diff
@@ -227,32 +200,44 @@ internal sealed interface TerminationSurveyEvent {
...
+  val nextNavigationStep: SurveyNavigationStep?,
+  val navigationStepLoading: Boolean,
+  val errorWhileLoadingNextStep: Boolean,
+  val showEmptyQuotesDialog: Boolean,
+  val intentAndIdToRedirectToChangeTierFlow: Pair<String, ChangeTierDeductibleIntent>?,
+  val actionButtonLoading: Boolean,
 ) {
+  val selectedOption: TerminationSurveyOption? = reasons.firstOrNull { it.id == selectedOptionId }
   val continueAllowed: Boolean = selectedOption != null && selectedOption.suggestion == null
+
```
panasetskaya, line 228: There is a question of disabling the "continue" button. As it is now it doesn't really work well in cases where we just show info - but disable the continue button. Maybe something like:

```val continueAllowed: Boolean = selectedOption != null &&
    (selectedOption.suggestion == null ||
      selectedOption.suggestion is  SurveyOptionSuggestion.Info || selectedOption.suggestion.buttonTitle == null)
```

panasetskaya (approved):

PR #2455, opened 2025-02-24
Try catch taking persistent URI permissions

StylianosGakis:


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/CbmChatRepository.kt
```diff
@@ -278,14 +272,23 @@ internal class CbmChatRepositoryImpl(
...
-        CbmChatMessage.FailedToBeSent.ChatMessagePhoto(
-          messageId?.toString() ?: Uuid.randomUUID().toString(),
-          clock.now(),
-          uri,
-        )
+      handleFailedToSendMedia(messageId, uri, conversationId)
+    }
+  }
+
+  private suspend fun handleFailedToSendMedia(messageId: Uuid?, uri: Uri, conversationId: Uuid) {
```
panasetskaya, line 279: could we also make handleFailedToSendMedia return Either with some specific class like ErrorMessageFailedPersistentPermission(): ErrorMessage so we could distinguish later what we want to show in UI?

panasetskaya (approved):

PR #2458, opened 2025-02-25
Expose more info on message failures, and use it to show errors like file too big or generic failures to send messages

StylianosGakis:


app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/CbmChatRepository.kt
```diff
@@ -400,6 +406,22 @@ internal class CbmChatRepositoryImpl(
   }
 }
 
+sealed class MessageSendError(val originalError: ErrorMessage) : ErrorMessage by originalError {
+  class GenericError(originalError: ErrorMessage) : MessageSendError(originalError)
```
panasetskaya, line 410: why this delegate? just curious

panasetskaya (approved):

app/feature/feature-chat/src/main/kotlin/com/hedvig/android/feature/chat/data/CbmChatRepository.kt
StylianosGakis: It's so that if someone is not interested in the more detailed error type, they can just treat it as `ErrorMessage` and be fine with it.
It helped me a lot when I was writing this as I otherwise had to go and change all the callers of any of these functions to us `MessageSendError` instead, and then do `.originalError` on it if they were using the `ErrorMessage` in some way before.
With this addition, it was much simpler to just go and change only in the places that were interested in those details of which kind of `MessageSendError` it was.

particularly, this commit https://github.com/HedvigInsurance/android/pull/2458/commits/f9a9aeebfcfd7c980f44668a3c6f6617d91abe81 still had all code green without touching the presenter which I did in a follow-up step here https://github.com/HedvigInsurance/android/pull/2458/commits/a99b5aec7146975cf4095a52ca6a18fcf1362e02

And since it all worked in the end I just left it like that.

PR #2462, opened 2025-02-27
Feature/chat videoplayer scroll

panasetskaya:
fix a bit the scroll as you suggested Stelios, and made banner closeable if the conversation is still open


StylianosGakis (approved):
Looking great!

PR #2470, opened 2025-03-05
Notifications/marketing channel title

StylianosGakis:
If this is approved here https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1741170122734959 we can just merge it.

It moves the old title to the description where it probably was meant to be in the first place. And adds a new title, being "Campaigns"/"Kampanjer".

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/CrossSellNotificationSender.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/CrossSellNotificationSender.kt
@@ -27,8 +27,6 @@ class CrossSellNotificationSender(
     // TODO drop support for this notification type if we see we never get it
     logcat(LogPriority.ASSERT) { "We never expect to get a notification of type $type" }
     return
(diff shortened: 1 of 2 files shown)


app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/CrossSellNotificationSender.kt
```diff
@@ -29,6 +29,8 @@ class CrossSellNotificationSender(
     return
     val title = remoteMessage.data[DATA_MESSAGE_TITLE]
     val body = remoteMessage.data[DATA_MESSAGE_BODY]
+    val title = remoteMessage.titleFromCustomerIoData() ?: remoteMessage.data[DATA_MESSAGE_TITLE]
```
panasetskaya, line 32: duplicates?

StylianosGakis: what the hell happened here 😂

PR #2474, opened 2025-03-07
GEN-3399 Select insurance step in the moving flow

panasetskaya:
Select insurance step in the beginning of the moving flow.

- [x] wait for schema update and change accordingly (see todos in the code): [link to slack](https://hedviginsurance.slack.com/archives/C08DFK77170/p1741269816122969)
- [x]  Contract.supportsMoving (insurances) and memberActionOptions.isMovingEnabled (HC) are turned off, when Sladan changes that for staging, test the flow properly with 2 homes
- [x] still not able to test with the real flow, we return error for moving
- [x] nav stuck in loop!


app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowState.kt
```diff
@@ -202,24 +207,26 @@ internal fun MovingFlowState.Companion.fromFragments(
...
-    housingType = housingType,
+    moveFromAddressId = moveFromAddressId,
+    housingType = null,
     addressInfo = AddressInfo(null, null),
     movingDateState = MovingFlowState.MovingDateState(
       selectedMovingDate = null,
       allowedMovingDateRange = moveIntentFragment.minMovingDate..moveIntentFragment.maxMovingDate,
     ),
-    propertyState = propertyState,
+    propertyState = null,
```
StylianosGakis, line 224: Probably missing the big picture here, but how come housingType and propertyState now start as null now that we have the map above? Is it that we do not want anything to be preselected or something like that?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/addhouseinformation/AddHouseInformationViewModel.kt
```diff
@@ -225,7 +225,8 @@ private fun MovingFlowState.toInputForSubmission(validContent: ValidAddressInput
       ),
       moveFromAddressId = moveFromAddressId,
       movingDate = this.movingDateState.selectedMovingDate!!,
-      numberCoInsured = this.propertyState.numberCoInsuredState.selectedNumberCoInsured,
+      // property state should be set before
+      numberCoInsured = this.propertyState?.numberCoInsuredState?.selectedNumberCoInsured!!,
```
StylianosGakis, line 229: Is the `!!` in reality something that should go on the `propertyState!!` instead, if that is what we assert that will not be null here?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/enternewaddress/EnterNewAddressViewModel.kt
```diff
@@ -421,7 +423,8 @@ private fun MovingFlowState.toContent(): EnterNewAddressUiState.Content {
       },
     ),
     numberCoInsured = ConstrainedNumberInput(
-      initialValue = propertyState.numberCoInsuredState.selectedNumberCoInsured,
+      // property state should be set before (HousingType step)
+      initialValue = propertyState?.numberCoInsuredState?.selectedNumberCoInsured!!,
```
StylianosGakis, line 427: Same comment as above ^^

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/selectcontract/SelectContractViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+          } else {
+            SelectContractState.Error.UserPresentable(userError)
+          }
+        }
+        moveIntent
+      }.fold(
+        ifLeft = { error ->
+          Snapshot.withMutableSnapshot {
+            currentState = error
+          }
```
StylianosGakis, line 114: Since we only edit 1 mutableState here, we do not risk there being inconsistent states where one thing is edited before the other thing also has a chance to be edited.
```suggestion
          currentState = error
```

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/selectcontract/SelectContractViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+        },
+        ifRight = { intent ->
+          Snapshot.withMutableSnapshot {
```
StylianosGakis, line 117: Same idea here, we seem to just edit the `currentState` once in all branches of the when statement here, so we can do it without the `withMutableSnapshot`

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/selectcontract/SelectContractViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+                  intent = intent,
+                  selectedAddress = null,
+                  navigateToHousingType = false,
+                  buttonLoading = false,
+                )
+              }
+
+              currentAddressesSize == 1 -> {
+                submittingAddressId = intent.currentHomeAddresses[0].id
+                currentState = Redirecting(
```
StylianosGakis, line 131: ^^ That was a lie.
It seems like here we do set two variables together. I'd suggest we just wrap these two with the `Snapshot.withMutableSnapshot` here so that we make it more obvious why we use it here 😊

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/selectcontract/SelectContractViewModel.kt
StylianosGakis: vv Further clarified below

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/HousingTypeDestination.kt
```diff
@@ -204,68 +202,68 @@ private fun PreviewInsuranceDestinationAnimation(
...
-        maxHouseNumberCoInsured = null,
-        maxHouseSquareMeters = null,
-        maxApartmentNumberCoInsured = null,
-        maxApartmentSquareMeters = null,
-        isApartmentAvailableforStudent = null,
-        extraBuildingTypes = listOf(),
-        suggestedNumberCoInsured = 2,
-        currentHomeAddresses = listOf(),
-      ),
+//      initiatedMovingIntent = MoveIntentV2CreateMutation.Data.MoveIntentCreate.MoveIntent(
```
StylianosGakis, line 212: Are these temporarily commented or should we remove them completely?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/HousingTypeViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+          currentState = state.copy(navigateToNextStep = false)
+        }
+
+        DismissStartError -> {
+          loadIteration++
+        }
+      }
+    }
+
+    LaunchedEffect(Unit) {
```
StylianosGakis, line 64: Looks like this is not keyed on loadIteration. 
Was this on purpose, or just omitted by accident?

StylianosGakis (approved):
So this will require some more changes after the schema changes, right?

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/data/MovingFlowState.kt
panasetskaya: yes, it's because select current home address should be the first step now, and we don't know yet what type of house will the new home be.

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/start/HousingTypeViewModel.kt
panasetskaya: tbh just left it as it was, but i guess it's bc we don't really load anything here, so it can just be set initially without giving opportunity to reload?

StylianosGakis (approved):

StylianosGakis (approved):
SelectContractForMoving outside of the graph looks good too 👍

app/feature/feature-movingflow/src/main/kotlin/com/hedvig/android/feature/movingflow/ui/selectcontract/SelectContractDestination.kt
```diff
@@ -45,14 +45,15 @@ internal fun SelectContractDestination(
   viewModel: SelectContractViewModel,
   navigateUp: () -> Unit,
   exitFlow: () -> Unit,
-  onNavigateToNextStep: (moveIntentId: String) -> Unit,
+  onNavigateToNextStep: (moveIntentId: String, popUpDestination: Boolean) -> Unit,
```
StylianosGakis, line 48: Can we make this name a bit more descriptive regarding what we're popping here which is extra?

StylianosGakis (approved):

PR #2473, opened 2025-03-06
GEN-3356: Redesigned comparison table

StylianosGakis:
| Medium | Huge | Tiny |
| --- | --- | --- |
| image | image | image |

PR #2475, opened 2025-03-07
Design: Tab row fit in one line

StylianosGakis:
The core change is that we can use textMeasurer to do this part
```
val (minItemWidth, fixedItemHeight) = remember(textMeasurer, tabTitles, density) {
  var width = 0
  var height = 0
  for (title in tabTitles) {
    val textLayoutResult = textMeasurer.measure(title, textStyle, maxLines = 1)
    width = maxOf(width, textLayoutResult.size.width)
    height = maxOf(height, textLayoutResult.size.height)
  }
  with(density) {
    val extraHorizontalSpace = tabInternalPadding.calculateStartPadding(layoutDirection)
      .plus(tabInternalPadding.calculateEndPadding(layoutDirection))
    val extraVerticalSpace = tabInternalPadding.calculateTopPadding() + tabInternalPadding.calculateBottomPadding()
    DpSize(width.toDp() + extraHorizontalSpace, height.toDp() + extraVerticalSpace)
  }
}
```
Which lets us get the real size immediately without waiting for a frame, so it makes the rest of the calculations much easier as a consequence.

This was done on top of https://github.com/HedvigInsurance/android/pull/2473 because I noticed that we got a fourth line of rows more often than we really needed to. This PR does not change the fact that it's possible to have a second or an nth row, it only makes sure that it does not go to the next row unless there really is no space for the content.

New screenshots in the same scenario after this PR:
| 1/7 | 3/7 | 7/7 |
| --- | --- | --- |
| image | image | image |

### DIFF
--- a/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
+++ b/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tab.kt
@@ -48,6 +48,7 @@ import com.hedvig.android.design.system.hedvig.tokens.LargeTabTokens
 import com.hedvig.android.design.system.hedvig.tokens.MediumTabTokens
 import com.hedvig.android.design.system.hedvig.tokens.MiniTabTokens
 import com.hedvig.android.design.system.hedvig.tokens.SmallTabTokens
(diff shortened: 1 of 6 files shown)


panasetskaya:
amazing!

panasetskaya (approved):

PR #2477, opened 2025-03-10
GEN-3413 Feature/text fallback claim flow

panasetskaya:
- added free text fallback it is allowed by BE on audio recording step
- added free text if there is any i the summary step
- redid summary a lil bit so the empty entries are not shown


StylianosGakis (approved):
👍

PR #2479, opened 2025-03-14
A11y: Double title component

StylianosGakis:
Just a simple component to wrap this repeated behavior in all relevant places.

Probably missed some spots, but we can do those as we see them in the near future


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/a11y/DoubleTitleHeading.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.style.LineBreak
+import com.hedvig.android.design.system.hedvig.HedvigText
+import com.hedvig.android.design.system.hedvig.HedvigTheme
+
+@Composable
+fun DoubleTitleHeading(
+  title: String,
+  description: String,
```
panasetskaya, line 16: if description would be nullable, we could also use it in Travel certificate flow

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/terminationreview/TerminationConfirmationDestination.kt
```diff
@@ -141,16 +142,10 @@ private fun AreYouSureScreen(
...
-    )
-    HedvigText(
-      text = stringResource(id = R.string.TERMINATION_FLOW_SUMMARY_SUBTITLE),
-      style = HedvigTheme.typography.headlineMedium,
-      color = HedvigTheme.colorScheme.textSecondary,
-      modifier = modifier.padding(horizontal = 16.dp),
+    DoubleTitleHeading(
+      stringResource(id = R.string.TERMINATION_FLOW_CANCELLATION_TITLE),
+      stringResource(id = R.string.TERMINATION_FLOW_SUMMARY_SUBTITLE),
+      modifier.padding(horizontal = 16.dp),
```
panasetskaya, line 148: I think we have the same structure in TerminationScaffold, just divided into two parts (title in Scaffold itself, subtitle in the column)

panasetskaya (approved):

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/a11y/DoubleTitleHeading.kt
StylianosGakis: That would be great, we should do it then yes.

app/feature/feature-terminate-insurance/src/main/kotlin/com/hedvig/android/feature/terminateinsurance/step/terminationreview/TerminationConfirmationDestination.kt
StylianosGakis: Yeah we have a bit of an interesting setup there, since we always have the same itle so we've moved it into the scaffold itself.

We might just have to make a slight inconvenient change there and move the title on each screen, which means we need to remember to call it from all places, but it will allow us to use this component

PR #2481, opened 2025-03-17
Fix/pending contracts

panasetskaya:
we were not showing pending insurances (that are in the process of switching from another insurance company) at all, so added them 
- to insurances
- to termination.
Also removed some unused fields (in CancelInsuranceData)


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/insurancedetail/ContractDetailDestination.kt
```diff
@@ -243,60 +253,52 @@ private fun ContractDetailScreen(
...
                     YourInfoTab(
-                      coverageItems = state.insuranceContract.currentInsuranceAgreement.displayItems
+                      coverageItems = displayItems
                         .map { it.title to it.value },
-                      coInsured = state.insuranceContract.currentInsuranceAgreement.coInsured,
-                      allowEditCoInsured = state.insuranceContract.supportsEditCoInsured,
-                      contractHolderDisplayName = state.insuranceContract.contractHolderDisplayName,
-                      contractHolderSSN = state.insuranceContract.contractHolderSSN,
-                      allowChangeAddress = state.insuranceContract.supportsAddressChange,
+                      coInsured = if (isNotPending) contract.currentInsuranceAgreement.coInsured else listOf(),
```
StylianosGakis, line 258: For all of these, it also looks like we just want a sane default for the non-active active, and whatever it already was for the active ones.

Could we make some extension functions at the bottom of this file that instead do

```
fun AbstractInsuranceContract.coInsured(): CoInsured {
 when(this) {
  Active -> currentInsuranceAgreement.coInsured
  Pending -> emptyList()
 }
}
```

And we do this for all the fields here where you had to do an `if(isNotPending)` check.

This would also mean that if we ever introduced a new entry in the sealed hierarchy it'd be easier to support those too.

And I say perhaps it can be extension functions at the bottom of this file incase we don't want to make that other file have too many things. But perhaps it'd be nicer for all of these to be there.

Let me know what you think about these ideas

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCaseImplTest.kt
```diff
@@ -137,7 +144,12 @@ class GetInsuranceContractsUseCaseImplTest {
       )
       val result = subjectUseCase.invoke().first()
       assertk.assertThat(result).isRight().transform {
-        it.first().supportsTierChange
+        val first = it.first()
+        if (first is AbstractInsuranceContract.InsuranceContract) {
```
StylianosGakis, line 148: Would be a tiny bit nicer with a `when` here, so that if we get a new entry in the hierarchy the IDE will give us a compile-time error here instead of just defaulting to `else` even though perhaps we'd want to do something else there.

app/feature/feature-insurances/src/test/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCaseImplTest.kt
```diff
@@ -116,7 +118,12 @@ class GetInsuranceContractsUseCaseImplTest {
...
```
StylianosGakis, line 122: Would be a tiny bit nicer with a when here, so that if we get a new entry in the hierarchy the IDE will give us a compile-time error here instead of just defaulting to else even though perhaps we'd want to do something else there.

StylianosGakis (approved):
Very good job 🔥
Had some thoughts/opinions on some things, left comments accordingly

PR #2483, opened 2025-03-18
GEN-3388 Feature/claim closed flow

panasetskaya:
Handling CLAIM_CLOSED notification type.

### DIFF
--- a/app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
@@ -29,6 +29,7 @@ import com.hedvig.android.app.apollo.LoggingInterceptor
 import com.hedvig.android.app.apollo.LogoutOnUnauthenticatedInterceptor
 import com.hedvig.android.app.logginginterceptor.HedvigHttpLoggingInterceptor
 import com.hedvig.android.app.notification.senders.ChatNotificationSender
+import com.hedvig.android.app.notification.senders.ClaimClosedNotificationSender
 import com.hedvig.android.app.notification.senders.ContactInfoSender
 import com.hedvig.android.app.notification.senders.CrossSellNotificationSender
 import com.hedvig.android.app.notification.senders.GenericNotificationSender
@@ -264,6 +265,14 @@ private val notificationModule = module {
       get<HedvigBuildConstants>(),
       get<HedvigDeepLinkContainer>(),
       HedvigNotificationChannel.Chat,
+    )
+  } bind NotificationSender::class
+  single<ClaimClosedNotificationSender> {
+    ClaimClosedNotificationSender(
+      get<Context>(),
+      get<HedvigBuildConstants>(),
+      get<HedvigDeepLinkContainer>(),
+      HedvigNotificationChannel.Payments, // todo: it is related to pay-outs
     )
   } bind NotificationSender::class
   single<ContactInfoSender> {
--- a/app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
+++ b/app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
@@ -272,7 +272,7 @@ private val notificationModule = module {
       get<Context>(),
       get<HedvigBuildConstants>(),
       get<HedvigDeepLinkContainer>(),
-      HedvigNotificationChannel.Payments, // todo: it is related to pay-outs
+      HedvigNotificationChannel.Payments,
     )
   } bind NotificationSender::class
   single<ContactInfoSender> {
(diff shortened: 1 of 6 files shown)


app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
```diff
@@ -266,6 +267,14 @@ private val notificationModule = module {
       HedvigNotificationChannel.Chat,
     )
   } bind NotificationSender::class
+  single<ClaimClosedNotificationSender> {
+    ClaimClosedNotificationSender(
+      get<Context>(),
+      get<HedvigBuildConstants>(),
+      get<HedvigDeepLinkContainer>(),
+      HedvigNotificationChannel.Payments, // todo: it is related to pay-outs
```
StylianosGakis, line 275: The description of that channel is:
title:    Payment information
subtitle: Noticiations regarding your payments to and from Hedvig

I think it might be correct anyway, and we can drop this todo, wdyt?

app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/ClaimClosedNotificationSender.kt
```diff
@@ -0,0 +1,81 @@
...
+  private val buildConstants: HedvigBuildConstants,
+  private val hedvigDeepLinkContainer: HedvigDeepLinkContainer,
+  private val notificationChannel: HedvigNotificationChannel,
+) : NotificationSender {
+  override suspend fun sendNotification(type: String, remoteMessage: RemoteMessage) {
+    val claimId = remoteMessage.data.claimIdFromData()
+    val intentUri = if (claimId != null) {
+      hedvigDeepLinkContainer.claimDetails.first().replace("{claimId}", claimId)
+    } else {
+      hedvigDeepLinkContainer.home.first()
```
StylianosGakis, line 35: Reasonable fallback here.
Perhaps we also want to do a log as Error whenever this happens, and if it happens often to see why

app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/ClaimClosedNotificationSender.kt
```diff
@@ -0,0 +1,81 @@
...
+    val claimClosedPendingIntent: PendingIntent? = PendingIntentCompat.getActivity(
+      context,
+      0,
+      buildConstants.intentForNotification(intentUri),
+      FLAG_UPDATE_CURRENT,
+      true,
+    )
+    val title = remoteMessage.titleFromCustomerIoData() ?: remoteMessage.data[DATA_MESSAGE_TITLE]
+    val body = remoteMessage.bodyFromCustomerIoData() ?: remoteMessage.data[DATA_MESSAGE_BODY]
+    logcat { "Mariia: remoteMessage ${remoteMessage.data}" }
```
StylianosGakis, line 47: Remember to remove this line
```suggestion
```

app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/ClaimClosedNotificationSender.kt
```diff
@@ -0,0 +1,81 @@
...
+    logcat { "ChatNotificationSender sending notification with deeplink uri:$intentUri" }
+    val claimClosedPendingIntent: PendingIntent? = PendingIntentCompat.getActivity(
+      context,
+      0,
+      buildConstants.intentForNotification(intentUri),
+      FLAG_UPDATE_CURRENT,
+      true,
+    )
+    val title = remoteMessage.titleFromCustomerIoData() ?: remoteMessage.data[DATA_MESSAGE_TITLE]
+    val body = remoteMessage.bodyFromCustomerIoData() ?: remoteMessage.data[DATA_MESSAGE_BODY]
```
StylianosGakis, line 46: I think we don't need the fallback here if it's not present in CIO data. Those notifications that have that have it for historical reasons, as they used to be sent that way before and I'm not sure if it's still the case. For new notifications like this one it should always come from CIO as far as I understand

app/app/src/main/kotlin/com/hedvig/android/app/notification/senders/ClaimClosedNotificationSender.kt
```diff
@@ -0,0 +1,81 @@
...
+    logcat { "Mariia: remoteMessage ${remoteMessage.data}" }
+    val notification = NotificationCompat
+      .Builder(context, notificationChannel.channelId)
+      .setSmallIcon(ic_hedvig_h)
+      .setContentTitle(title)
+      .setContentText(body)
+      .setPriority(PRIORITY_MAX)
+      .setAutoCancel(true)
+      .setCategory(CATEGORY_EVENT)
```
StylianosGakis, line 55: This seems to imply that it's a: "Notification category: calendar event." perhaps it's not the right categorisation.
In fact looking at all of the categories none actually match what this is. Perhaps we can do without any category here at all?

StylianosGakis (approved):

app/app/src/main/kotlin/com/hedvig/android/app/di/ApplicationModule.kt
panasetskaya: yeah, I was just explaining why I chose that:)

PR #2484, opened 2025-03-18
GEN-3438 Force-show cross-sells after a closed claim was seen

StylianosGakis:
Use the backend response to conditionally force-show cross sells in the home screen.
Purposefully not stored in persistent storage so that if the app is killed after this was set, we do not still show the cross sells in the home screen.


app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
```diff
@@ -271,3 +276,28 @@ internal data class HomeData(
 
   companion object
 }
+
+/**
+ * The reason this exists is because the standard combine function only allows up to 5 generic flows.
+ */
```
panasetskaya, line 282: fire

PR #2487, opened 2025-03-19
Handle null or empty pn and emails from backend and do not allow one to submit new input if that would mean deleting one of the two fields

StylianosGakis:
Relax the constraints a bit to allow for such cases, since our backend is allowed to give us such responses.

Make a clearer distinction now between null Email or PhoneNumber states, which indicate that there is an empty input versus actually invalid inputs.

Do not allow submitting the new data if it would mean it'd delete an existing email/phone number from backend

Some discussion around the business requirements of this screen https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1742407752612449

### DIFF
--- a/app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/contactinfo/ContactInfoViewModel.kt
+++ b/app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/contactinfo/ContactInfoViewModel.kt
@@ -70,8 +70,21 @@ internal sealed interface ContactInfoUiState {
     private val phoneNumberIsDifferentFromUploadedPhoneNumber: Boolean
       get() = uploadedPhoneNumber.valueForTextField != phoneNumberState.text
 
+    private val emailIsDeletingKnownInfo: Boolean
+      get() = when (uploadedEmail) {
+        null -> false
+        else -> emailState.text.isBlank()
+      }
+    private val phoneNumberIsDeletingKnownInfo: Boolean
+      get() = when (uploadedPhoneNumber) {
+        null -> false
+        else -> phoneNumberState.text.isBlank()
+      }
+
     val canSubmit: Boolean
       get() = (emailIsDifferentFromUploadedEmail || phoneNumberIsDifferentFromUploadedPhoneNumber) &&
+        !emailIsDeletingKnownInfo &&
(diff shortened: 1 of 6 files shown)


app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/contactinfo/ContactInfoViewModel.kt
```diff
@@ -46,22 +51,40 @@ internal sealed interface ContactInfoUiState {
...
+      get() = when (uploadedEmail) {
+        null -> false
+        else -> emailState.text.isBlank()
+      }
+    private val phoneNumberIsDeletingKnownInfo: Boolean
+      get() = when (uploadedPhoneNumber) {
+        null -> false
+        else -> phoneNumberState.text.isBlank()
+      }
+
```
panasetskaya, line 83: I see it's getting complicated)

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/data/ContactInfoRepository.kt
```diff
@@ -46,11 +46,16 @@ internal data class ContactInformation(
...
+       */
+      fun fromString(input: String?): Either<ErrorMessage, Email?> {
         return when {
-          input.isNullOrBlank() -> ErrorMessage(invalidInputErrorMessage(input)).left()
-          !isValidEmail(input) -> ErrorMessage(invalidInputErrorMessage(input)).left()
-          else -> Email(input).right()
+          input.isNullOrBlank() -> null.right()
+          isValidEmail(input) -> Email(input).right()
+          else -> ErrorMessage(invalidInputErrorMessage(input)).left()
         }
```
panasetskaya, line 59: not fully understand what would be the invalid input:)

panasetskaya (approved):

app/feature/feature-profile/src/main/kotlin/com/hedvig/android/feature/profile/data/ContactInfoRepository.kt
StylianosGakis: Input like "email @ ..@" would fall down to this else statement for example.
Anything that does not match the regex that is used within `isValidEmail`

PR #2498, opened 2025-03-24
Feature/new claims deatils UI

StylianosGakis:
So this is what I had in mind regarding not doing the conversion in compose right before displaying it

### DIFF
--- a/app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
+++ b/app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
@@ -26,7 +26,6 @@ import androidx.compose.foundation.lazy.grid.GridCells
 import androidx.compose.foundation.lazy.grid.LazyVerticalGrid
 import androidx.compose.foundation.lazy.grid.items
 import androidx.compose.foundation.rememberScrollState
-import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.verticalScroll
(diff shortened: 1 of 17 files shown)


app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
```diff
@@ -706,21 +714,23 @@ private fun ClaimDetailHedvigAudioPlayerItem(signedAudioUrl: SignedAudioUrl, mod
...
           endSlot = {
-            val valueAsValidDate = item.second.toValidLocalDate(getLocale())
+            val formatter = rememberHedvigDateTimeFormatter()
             HedvigText(
-              text = valueAsValidDate ?: item.second,
+              text = when (val item = displayItem.value) {
+                is Date -> formatter.format(item.date.toJavaLocalDate())
+                is DateTime -> formatter.format(item.localDateTime.toJavaLocalDateTime())
+                is Text -> item.text
+              },
```
panasetskaya, line 733: Could we go all the way in and format them all also in the useCase, and just use pure already formatted Strings here?

panasetskaya (approved):

app/feature/feature-claim-details/src/main/kotlin/com/hedvig/android/feature/claim/details/ui/ClaimDetailsDestination.kt
StylianosGakis: Not quite, because we need the locale in order to format them, and the locale needs to come from compose because it may change at any point from the composition local which it comes from

PR #2502, opened 2025-03-26
GEN-3519 add subtitle and details to claim summary

panasetskaya:
In odyssey I want smth like this: [odyssey pr](https://github.com/HedvigInsurance/odyssey/pull/686), bc currently for pet we don't have any subtitle (bc subtitle is option in entrypoint, and car/dog entrypoint don't have any options)

- [x] logic on odyssey
- [x] schema published
- [x] keys translated

before -> after


app/feature/feature-odyssey/src/main/kotlin/com/hedvig/android/feature/odyssey/step/summary/ClaimSummaryViewModel.kt
```diff
@@ -143,6 +145,14 @@ internal data class ClaimSummaryInfoUiState(
       if (claimTypeTitle != null) {
         add(resources.getString(R.string.CLAIMS_CASE) to claimTypeTitle)
       }
+      //Subtitle
+      if (claimSubtitle != null) {
+        add(resources.getString(R.string.CLAIMS_SUMMARY_WHAT_IS_ABOUT) to claimSubtitle)
+      }
+//      //TODO: details, when we'll get SelectContractStep with summary
```
StylianosGakis, line 152: Future field coming from the schema?

StylianosGakis (approved):

PR #2508, opened 2025-04-02
Feature/a11y bottom navigation focus

panasetskaya:
nav bar/rail and home screen a11y

- [x] keys translated
```
TALKBACK_CLAIM_STATUS_CARD
TALKBACK_UNKNOWN
TALKBACK_CLAIM_STATUS
```

### DIFF
--- a/app/ui/claim-status/src/main/kotlin/com/hedvig/android/ui/claimstatus/ClaimStatusCard.kt
+++ b/app/ui/claim-status/src/main/kotlin/com/hedvig/android/ui/claimstatus/ClaimStatusCard.kt
@@ -10,6 +10,8 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.hedvig.android.design.system.hedvig.ButtonDefaults.ButtonSize.Medium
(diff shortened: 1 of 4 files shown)


app/ui/claim-status/src/main/kotlin/com/hedvig/android/ui/claimstatus/ClaimStatusCard.kt
```diff
@@ -45,9 +47,13 @@ fun ClaimStatusCard(
   onClick: (claimId: String) -> Unit,
   modifier: Modifier = Modifier,
 ) {
+  val description = stringResource(R.string.TALKBACK_CLAIM_STATUS_CARD)
```
StylianosGakis, line 50: Could this be "Information about your claim" perhaps?

app/ui/claim-status/src/main/kotlin/com/hedvig/android/ui/claimstatus/internal/ClaimProgressRow.kt
```diff
@@ -32,15 +35,22 @@ import hedvig.resources.R
 
 @Composable
 internal fun ClaimProgressRow(claimProgressItemsUiState: List<ClaimProgressSegment>, modifier: Modifier = Modifier) {
+  val currentStatus =
```
StylianosGakis, line 38: Just saying "Unknown" might be a bit confusing here. Should we give a more descriptive description, or otherwise no description at all?

StylianosGakis (approved):

app/ui/claim-status/src/main/kotlin/com/hedvig/android/ui/claimstatus/internal/ClaimProgressRow.kt
panasetskaya: well it would be "Claim status: unknown" actually, and I think it rarely happens actually (it's usually at least "submitted")

PR #2510, opened 2025-04-02
GEN-3531 Feature/x-sell sheet after finishing various flows

StylianosGakis:
In the same spirit as the functionality we had for showing the play store review.

When a feature wants to "opt-in" to getting a x-sell sheet after it's done, it needs to simply call `CrossSellAfterFlowRepository.completedCrossSellTriggeringSelfServiceSuccessfully` and that marks the sheet ready to be shown.

Then, the only other piece is that it needs to know which screen it will end up in after finishing, and make sure it's part of the screens specified `HedvigAppState.isInScreenEligibleForCrossSells`.

With these two in place, since the x-sell bottom sheet is now globally available (outside of the NavHost), it can at will be presented, and it fetches the available cross-sells for the member itself.
https://github.com/HedvigInsurance/android/blob/a4497663e365d442379b611ddddfad0a03bfa9b7/app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigApp.kt#L113-L124
That is now these


app/data/data-addons/src/main/kotlin/com/hedvig/android/data/addons/data/GetTravelAddonBannerInfoUseCase.kt
```diff
@@ -31,8 +31,13 @@ internal class GetTravelAddonBannerInfoUseCaseImpl(
...
     val mappedSource = when (source) {
-      TravelAddonBannerSource.TRAVEL_CERTIFICATES, TravelAddonBannerSource.DEEPLINK -> UpsellTravelAddonFlow.APP_UPSELL_UPGRADE
-      TravelAddonBannerSource.INSURANCES_TAB -> UpsellTravelAddonFlow.APP_ONLY_UPSALE
+      TravelAddonBannerSource.TRAVEL_CERTIFICATES,
+      TravelAddonBannerSource.DEEPLINK,
+      -> UpsellTravelAddonFlow.APP_UPSELL_UPGRADE
+
+      TravelAddonBannerSource.INSURANCES_TAB,
+      TravelAddonBannerSource.AFTER_FINISHING_SUCCESSFUL_FLOW,
+      -> UpsellTravelAddonFlow.APP_ONLY_UPSALE
```
panasetskaya, line 40: these UPSELL and UPSALE are very confusing tbh, but that was same as from the be, so)

app/data/data-addons/src/main/kotlin/com/hedvig/android/data/addons/data/GetTravelAddonBannerInfoUseCase.kt
```diff
@@ -72,7 +77,7 @@ internal class GetTravelAddonBannerInfoUseCaseImpl(
           description = bannerData.descriptionDisplayName,
           labels = bannerData.badges,
           eligibleInsurancesIds = nonEmptyContracts,
-          bannerSource = mappedSource,
+          bannerSource = source,
```
panasetskaya, line 80: why this change?

StylianosGakis: Yeah it's a good question.
This was done in this commit https://github.com/HedvigInsurance/android/pull/2510/commits/6b90f4668c1198aa7498364ed1bed0dd20855de6
Because as I was working with this I realized that the graphQL type itself was being propagated to all other callers, and I think it made it so that a module which needed to use it did not have access to this type at all.
So I changed it so that the public API exposes our own type instead, and it can be mapped to the GQL type in here, at the moment it needs to be sent with gql.

Is there something that looks odd now, or something that looks broken perhaps? We can adjust it further if so

app/data/data-addons/src/main/kotlin/com/hedvig/android/data/addons/data/GetTravelAddonBannerInfoUseCase.kt
StylianosGakis: I am not gonna lie, I also got confused, so I just did a little chat here https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1743605412280519 to make sure I got it right.

app/data/data-addons/src/main/kotlin/com/hedvig/android/data/addons/data/GetTravelAddonBannerInfoUseCase.kt
panasetskaya: do we actually use this property anywhere? I can only see it in previews. Maybe it's a leftover of the logic that don't exist anymore or am I missing smth obvious?

StylianosGakis: You are 100% correct, just deleted the field in https://github.com/HedvigInsurance/android/pull/2510/commits/e5c7f79bc783be0aa569e58746a2c17316b917b8

panasetskaya (approved):

PR #2516, opened 2025-04-08
GEN-3571 Bs/material3

StylianosGakis:
With the new changes on top it makes it so that the sheet both animates in when being shown, and animates out when it's hidden.

Need to work a bit more around the contentPadding not being there and if we need to add it back in again or if the callers need to handle this themselves instead. Was a bit tricky to make it work along with the window insets that are used by default in the sheet.


panasetskaya (approved):
as far as i see, now we don't have any contentPadding in any of the BS? can't we apply standard 16.dp everywhere?

StylianosGakis:
Yep, that's why I mention this in the comment. I will look into that today.

The reason I removed it for now is that the m3 sheet only takes in WindowInsets instead of any contentPadding, and combining the incoming ContentPadding with the insets in there was non-obvious. I will figure out how to do this today

PR #2519, opened 2025-04-11
GEN-3451 Fix/a11y more little fixes

panasetskaya:
I want to check it all again later tbh. The travel addon flow is not done here, that will be a separate pr


app/feature/feature-choose-tier/src/main/kotlin/com/hedvig/android/feature/change/tier/ui/stepcustomize/SelectTierDestination.kt
```diff
@@ -319,19 +333,31 @@ private fun CustomizationCard(
...
           ExpandedRadioOptionData(
             chosenState = if (chosenTierInDialog == pair.first) Chosen else NotChosen,
             title = pair.first.tierDisplayName ?: "-",
-            premium = stringResource(R.string.TIER_FLOW_PRICE_LABEL, pair.second.amount.toInt()),
+            premiumString = stringResource(R.string.TIER_FLOW_PRICE_LABEL, pair.second.amount.toInt()),
             tierDescription = pair.first.tierDescription,
             onRadioOptionClick = {
               onChooseTierInDialogClick(pair.first)
             },
+            voiceoverDescription = voiceDescription,
```
StylianosGakis, line 360: In all the places we say "voiceover" perhaps we can say "contentDescription" instead to keep it the same as what is used by compose anyway?

StylianosGakis (approved):

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/icon/Refresh.kt
```diff
@@ -217,7 +217,7 @@ private fun IconPreview() {
     ) {
       Image(
         imageVector = HedvigIcons.Refresh,
-        contentDescription = null,
+        contentDescription = com.hedvig.android.compose.ui.EmptyContentDescription,
```
panasetskaya, line 220: I have a question here Stelios, didn't we want to put EmptyContentDescription only to the places where we are 100% sure we need a null description? Is it the case for all the icons? Refresh, Info?

StylianosGakis: Yeah you are right that doing that would be a big mistake here.
This one in particular, if I see correctly, is set only on the @Preview function here right? 👀

panasetskaya: No, you're absolutely right, it's all preview for all these icons, I didn't look properly

PR #2521, opened 2025-04-23
GEN-3542 Feature/insurance certificate

panasetskaya:
- generate insurance evidence flow (data and ui: email confirmation + download screens)
- deeplink support
- push notification support

(no navigation to the feature from the Profile tab in this PR, only through deeplink/push)

- [x] INSURANCE_EVIDENCE.READ_MORE_DESCRIPTION not translated yet


hugokallstrom (approved):
Looks great! Just some general feedback—try to reduce the PR size to around 200-300 lines of code when possible. 

In this case, you could perhaps have the notification logic as a separate PR. And even have the separate views (email input and overview) as separate PRs.

app/feature/feature-insurance-certificate/src/main/kotlin/com/hedvig/android/feature/insurance/certificate/data/GetInsuranceEvidenceInitialDataUseCase.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import arrow.core.Either
+import arrow.core.raise.either
+import com.apollographql.apollo.ApolloClient
+import com.hedvig.android.apollo.ErrorMessage
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.core.common.ErrorMessage
+import octopus.InsuranceEvidenceInitialDataQuery
+
+internal interface GetInsuranceEvidenceInitialDataUseCase {
```
StylianosGakis, line 11: Hmm, do you think we can mention the fact that the data is `email` specifically, and not just "data"?

StylianosGakis (approved):

PR #2526, opened 2025-04-28
GEN-3612 Feature/discounts UI update

panasetskaya:
[figma](https://www.figma.com/design/zItETkT4mu5ANVDYED5gEm/UX-Improvements--Small-Projects-?node-id=362-4954)
I originally wanted to wait for a couple of things (now in ReferredBy we use name instead of code, since we don't yet get it from the BE; also the new short exposure name is coming for the contracts), but it seems to be still in process and iOs released it as it is, so we can merge it like this as well.
I'll tear down the redeem campaign flag etc. in another PR.


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/ui/discounts/DiscountRow.kt
```diff
@@ -38,111 +44,135 @@ internal fun DiscountRows(discounts: List<Discount>, modifier: Modifier = Modifi
...
-            HedvigText(
-              text = it,
-              color = if (discountIsExpired) {
-                HedvigTheme.colorScheme.textDisabled
-              } else {
-                HedvigTheme.colorScheme.textSecondary
-              },
-              style = HedvigTheme.typography.bodySmall,
+        Column {
+          Row {
```
StylianosGakis, line 70: Is this row wrapper necessary here?

StylianosGakis (approved):

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/ui/discounts/DiscountRow.kt
panasetskaya: nope, that's a leftover. Removing, thanks)

PR #2532, opened 2025-05-06
GEN-2464 Cleanup no/dk

StylianosGakis:


app/market/market-set/src/main/kotlin/com/hedvig/android/market/set/SetMarketUseCase.kt
```diff
@@ -1,69 +0,0 @@
...
-internal class SetMarketUseCaseImpl(
-  private val internalSetMarketUseCase: InternalSetMarketUseCase,
-  private val languageService: LanguageService,
-) : SetMarketUseCase {
-  override suspend fun setMarket(market: Market, preferSystemDefaultLocale: Locale?) {
-    val selectedLanguage = languageService.getSelectedLanguage()
-    val existingLanguage = if (preferSystemDefaultLocale == null) {
-      selectedLanguage
-    } else {
-      selectedLanguage ?: Language.from(preferSystemDefaultLocale.toLanguageTag())
```
panasetskaya, line 30: could you pls tell me more about where this language-related logic in this useCase is moved or if it is deleted completely? Just a bit lost since there are a lot of deletions here. Do we check for the system language anywhere if we don't have anything set?

app/language/language-migration/src/main/kotlin/com/hedvig/android/language/LanguageLaunchCheckUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+internal class AndroidLanguageLaunchCheckUseCase(
+  private val languageService: LanguageService,
+) : LanguageLaunchCheckUseCase {
+  override suspend fun invoke(defLocale: Locale) {
+    val currentSelectedLanguage = languageService.getSelectedLanguage()
+    logcat { "LanguageAndMarketLaunchCheckUseCase: currentLanguage: $currentSelectedLanguage" }
+    if (currentSelectedLanguage == null) {
+      val defLanguage = Language.from(defLocale.toLanguageTag())
+      languageService.setLanguage(defLanguage).also {
+        logcat { "SetMarketUseCase setting language to $it" }
```
panasetskaya, line 24: change logs here?

panasetskaya (approved):
should we also wait for right before 12th to merge this?

app/language/language-migration/src/main/kotlin/com/hedvig/android/language/LanguageLaunchCheckUseCase.kt
StylianosGakis: Fixed

app/market/market-set/src/main/kotlin/com/hedvig/android/market/set/SetMarketUseCase.kt
StylianosGakis: Now what happens is that on every app startup this is called
https://github.com/HedvigInsurance/android/blob/5764c6b1788c495348261003ad5a0487f3bb38b7/app/app/src/main/kotlin/com/hedvig/android/app/MainActivity.kt#L108-L109
Which calls
https://github.com/HedvigInsurance/android/blob/5764c6b1788c495348261003ad5a0487f3bb38b7/app/language/language-migration/src/main/kotlin/com/hedvig/android/language/LanguageLaunchCheckUseCase.kt#L19-L23
Which calls
https://github.com/HedvigInsurance/android/blob/0235b84f5b465e82ee3256f569e48b849d689f2d/app/language/language-android/src/main/kotlin/com/hedvig/android/language/AndroidLanguageService.kt#L15-L19

So on every startup, the language is set.
If it was a language which is not supported, it will default to en_SE
If it was already en_SE nothing happens
If it was already sv_SE also nothing happens as we'd again try to set the locale to what it was already.

This should cover our cases here, for old clients and new alike from the tests I've done so far

panasetskaya: wait, I don't fully understand here. How do we use currentSelectedLanguage? What if the system is Swedish, but member has previously set the app language to English, do we change it back to Swedish on every start?

PR #2533, opened 2025-05-08
a11y: unscrollable screens

panasetskaya:
Fixed the scroll in the places where I noticed that, applying scroll actively in the dialogs (bc we have a few dialogs with lazyList - I think it's better to explicitly set it every time). AlertDialog and ErrorDialog will have vertical scroll by default.

Also added HedvigVerySmallScreenPreview to make it easier not to forget about these cases (it does not entirely make up for huge fonts scenarios, but it's close), and HedvigShortMultiScreenPreview (bc HedvigMultiScreenPreview is a bit excessive I think in most cases).

### DIFF
--- a/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Dialog.kt
+++ b/app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Dialog.kt
@@ -243,9 +243,13 @@ fun HedvigDialog(
     onDismissRequest = onDismissRequest,
     properties = dialogProperties,
   ) {
-    HedvigDialogContent(dialogProperties.usePlatformDefaultWidth, applyDefaultPadding, style,
+    HedvigDialogContent(
+      dialogProperties.usePlatformDefaultWidth,
+      applyDefaultPadding,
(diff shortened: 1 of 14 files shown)


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Dialog.kt
```diff
@@ -231,6 +235,7 @@ fun MultiSelectDialog(
 @Composable
 fun HedvigDialog(
   onDismissRequest: () -> Unit,
+  applyVerticalScroll: Boolean,
```
StylianosGakis, line 238: So in which scenarios would we want this to be `applyVerticalScroll = false`? Would  this not make it as easy as before to make the wrong decision regarding if this  will be scrollable or not?

panasetskaya: We have a few scenarios where we pass content with LazyList there, in which case applying scroll would break it. That why I did not apply default there, since we'd need to think about what goes there.

StylianosGakis (approved):

PR #2537, opened 2025-05-13
Staging package name same as production

StylianosGakis:
Change com.hedvig.test.app to com.hedvig.app for staging app

This allows it to have the same package name as the production app,
allowing us to link the firebase project with the real app, which in
turn allows us to provide .aab files to firebase app distribution

Will run CI to see if this behaves as it should. And adjust accordingly until it works


panasetskaya (approved):
pls explain where we specify using staging backend, I'm lost :)

PR #2539, opened 2025-05-19
GEN-3713 Feature/contact UI update

panasetskaya:
[ticket](https://www.notion.so/hedviginsurance/Phone-number-action-for-nulled-phone-numbers-in-app-1edc3f71cb0a8070b1ffeb3dc5265532?pvs=4)
[figma](https://www.figma.com/design/zItETkT4mu5ANVDYED5gEm/UX-Improvements--Small-Projects-?node-id=853-19288&m=dev)

Made [updates in member-service](https://github.com/HedvigInsurance/member-service/pull/1034), so we now have a new mutation `memberUpdateContactInfo` and a new property in MemberActions `isContactInfoUpdateNeeded` (true if phone number is missing or was added/reviewed more than one year). 

In this PR, if `MemberActions.isContactInfoUpdateNeeded` is true, we show a reminder on Home and Profile screens, that leads to Contact screen. The UI of Contact screen is also updated.

Will update the mutation used for contact screen to the new `memberUpdateContactInfo` in a separate PR, to keep this one shorter :)

### DIFF
--- a/app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetMemberRemindersUseCase.kt
+++ b/app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetMemberRemindersUseCase.kt
@@ -2,6 +2,7 @@ package com.hedvig.android.memberreminders
 
 import arrow.core.Either
 import arrow.core.NonEmptyList
+import com.hedvig.android.memberreminders.MemberReminder.ContactInfoUpdateNeeded
 import java.util.UUID
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.combine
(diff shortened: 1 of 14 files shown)


app/member-reminders/member-reminders-public/src/main/kotlin/com/hedvig/android/memberreminders/GetMemberRemindersUseCase.kt
```diff
@@ -123,4 +132,8 @@ sealed interface MemberReminder {
     val contractId: String,
     override val id: String = UUID.randomUUID().toString(),
   ) : MemberReminder
+
+  data class ContactInfoUpdateNeeded(
+    override val id: String = UUID.randomUUID().toString(),
+  ) : MemberReminder
```
StylianosGakis, line 138:
```suggestion
  data object ContactInfoUpdateNeeded : MemberReminder {
    override val id: String = UUID.randomUUID().toString()
  }
```
I wonder, would a syntax like this work here? Since we never pass in a custom ID, and we just need it to be unique but we won't ever really have 2 of these reminders in a list of reminders.

StylianosGakis (approved):

PR #2542, opened 2025-05-20
Add ".../submit-claim" deep link to claim-flow

StylianosGakis:


app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -9,6 +9,7 @@ interface HedvigDeepLinkContainer {
   val helpCenterQuestion: List<String> // A specific question inside the help center
 
   val insurances: List<String> // The insurances destination, which also shows cross sells
+  val claimFlow: List<String> // The claim flow starting pledge destination
```
panasetskaya, line 12: do we need it in the allDeepLinkUriPatterns too?

panasetskaya (approved):

app/navigation/navigation-core/src/main/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
StylianosGakis: Yes would be good!

PR #2549, opened 2025-06-09
GEN-3793 x-sell v2 

panasetskaya:


app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -101,21 +118,69 @@ private fun <T> InternalHedvigBottomSheet(
...
+      },
   ) {
     Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
       content()
     }
   }
 }
 
+@Composable
+fun CrossSellDrugHandle(
```
StylianosGakis, line 138:
```suggestion
fun CrossSellDragHandle(
```

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -101,21 +118,69 @@ private fun <T> InternalHedvigBottomSheet(
...
+  modifier: Modifier = Modifier,
+  text: String? = stringResource(R.string.CROSS_SELL_BANNER_TEXT),
+) {
+  val direction = LocalLayoutDirection.current
+  Box(
+    modifier
+      .fillMaxWidth()
+      .height(40.dp)
+      .layout { measurable, constraints ->
+        val paddingStart = contentPadding.calculateStartPadding(direction).roundToPx()
```
StylianosGakis, line 149: Since this is quite an odd thing to see here without context, perhaps a comment might be useful
```suggestion
        // M3 sheet does not allow us to "break out" of the content padding so we do it through a layout
        val paddingStart = contentPadding.calculateStartPadding(direction).roundToPx()
```

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/HedvigBottomSheet.kt
```diff
@@ -101,21 +118,69 @@ private fun <T> InternalHedvigBottomSheet(
...
+      verticalAlignment = Alignment.CenterVertically,
+    ) {
+      Icon(
+        HedvigIcons.Campaign,
+        contentDescription = EmptyContentDescription,
+        tint = HedvigTheme.colorScheme.signalGreenElement,
+        modifier = Modifier.size(20.dp),
+      )
+      Spacer(Modifier.width(8.dp))
+      if (text != null) {
```
StylianosGakis, line 173:
```suggestion
      if (text != null) {
        Spacer(Modifier.width(8.dp))
```
So that if the text is null the icon is not shifted to the left by 4.dp

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
```diff
@@ -88,30 +85,23 @@ fun CrossSellSheet(
...
-    crossSells = crossSells,
-    onCrossSellClick = onCrossSellClick,
-    withSubHeader = false,
-  )
-  if (travelAddonBannerInfo != null) {
+  if (recommendedCrossSell != null) {
+    RecommendationSection(recommendedCrossSell, onCrossSellClick)
+  }
+  if (otherCrossSells.isNotEmpty()) {
+    Spacer(Modifier.height(64.dp))
```
StylianosGakis, line 97: If the recommended cross sell is in fact null, do we want the 64.dp empty space on top of the rest of the other cross sells content?

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
```diff
@@ -127,41 +117,47 @@ private fun ColumnScope.CrossSellsSheetContent(
...
+      painter = painterResource(id = recommendedCrossSell.type.iconRes()),
+      contentDescription = EmptyContentDescription,
+      modifier = Modifier
+        .size(140.dp),
+    )
+    Spacer(Modifier.height(24.dp))
+    HedvigText(
+      text = recommendedCrossSell.title,
+      style = HedvigTheme.typography.bodySmall,
+    )
```
StylianosGakis, line 140: Do these happen to be a11y headers by any chance? Not sure tbh

StylianosGakis (approved):

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
panasetskaya: I don't think so. but "Discover more coverage" down there could in fact be

PR #2552, opened 2025-06-12
Feature/x sell v2 tooltip

panasetskaya:


StylianosGakis (commented):
I know this is gonna be extra work, but I would really like it if we did not used shared prefs here

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -315,9 +320,16 @@ private fun HomeScreen(
...
+              .windowInsetsPadding(
+                WindowInsets.safeDrawing.only(WindowInsetsSides.Horizontal),
+              )
               .padding(horizontal = 16.dp),
           )
+        } else {
+          CrossSellsTooltip(uiState)
         }
+      } else if (uiState is Success) {
+        CrossSellsTooltip(uiState)
```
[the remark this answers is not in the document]
panasetskaya: I think the idea is that we want to sort out the inbox tooltip first. If there is inbox icon, we check it first -> then if no tooltip is needed there, we handle the cross sell tooltip. But we also want to to handle crossSells if there is no inbox icon at all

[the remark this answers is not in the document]
panasetskaya: There probably is a better way of writing it, but I can't think of one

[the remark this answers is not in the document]
StylianosGakis: Would it be possible at all to contain all of this on one if statement, both for if we're in a success state and if we've got no inbox tooltip?
And have it be as a completely separate if statement outside of that other if statement.

But tbh don't worry about this too much, it's not a huge deal, but if we ever want to make changes to this logic we might miss the fact that we need to do these changes in two spots.

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomePresenter.kt
```diff
@@ -49,19 +55,43 @@ internal class HomePresenter(
...
+      combine(
+        getHomeDataUseCaseProvider.provide().invoke(forceNetworkFetch),
+        crossSellHomeNotificationServiceProvider.provide().showRedDotNotification(),
+        crossSellHomeNotificationServiceProvider.provide().getLastEpochDayNewRecommendationNotificationWasShown(),
+      ) { homeResult: Either<ApolloOperationError, HomeData>, showRedDot: Boolean, epochDay: Long? ->
+        homeResult to (showRedDot to epochDay)
+      }.collectLatest { result: Pair<Either<ApolloOperationError, HomeData>, Pair<Boolean, Long?>> ->
+        val homeResult = result.first
+        val showCrossSellRedDot = result.second.first
+        val lastDayCrossSellToolTipShown = result.second.second
```
StylianosGakis, line 90: Creative 😅

```suggestion
        homeResult to (showRedDot to epochDay)
      }.collectLatest { (homeResult: Either<ApolloOperationError, HomeData>, (showCrossSellRedDot: Boolean, lastDayCrossSellToolTipShown: Long?)) ->
```

Could you check if this code also works?
Not sure if it's any better, but it's just a bit daunting reading the `result.second.first` and `result.second.second` down there :D

app/notification-badge-data/notification-badge-data-public/src/main/kotlin/com/hedvig/android/notification/badge/data/crosssell/CrossSellNotificationBadgeService.kt
```diff
@@ -3,51 +3,43 @@ package com.hedvig.android.notification.badge.data.crosssell
...
-      .filter(CrossSellIdentifier::isKnownCrossSell)
-      .map(CrossSellIdentifier::rawValue)
-      .toSet()
-
-    val alreadySeenCrossSells = notificationBadgeStorage.getValue(notificationBadge).first()
+      ?.rawValue
     notificationBadgeStorage.setValue(
       notificationBadge = notificationBadge,
-      newStatus = potentialCrossSells + alreadySeenCrossSells,
+      newStatus = potentialCrossSellRecommendation?.let { setOf(it) } ?: emptySet(),
```
StylianosGakis, line 42:
```suggestion
      newStatus = setOfNotNull(potentialCrossSellRecommendation),
```
Does this compile here perhaps? I think it should do the same thing

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/ToolbarIcons.kt
```diff
@@ -52,14 +60,25 @@ fun ToolbarFirstVetIcon(onClick: () -> Unit, modifier: Modifier = Modifier) {
...
 @Composable
 fun ToolbarCrossSellsIcon(onClick: () -> Unit, modifier: Modifier = Modifier) {
+  var isRotated by remember { mutableStateOf(false) }
+  val fullRotation by animateFloatAsState(
+    targetValue = if (isRotated) 360f else 0f,
+    animationSpec = tween(1500, 50),
+  )
+  LaunchedEffect(Unit) {
+    isRotated = true
+  }
```
StylianosGakis, line 70:
```suggestion
  val isRotated by produceState(false) { value = true }
  val fullRotation by animateFloatAsState(
    targetValue = if (isRotated) 360f else 0f,
    animationSpec = tween(1500, 50),
  )
```
I wonder, does this behave the same way here?
And if yes, do you prefer it or find it less nice to look at?

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/ToolbarIcons.kt
```diff
@@ -52,14 +60,25 @@ fun ToolbarFirstVetIcon(onClick: () -> Unit, modifier: Modifier = Modifier) {
 
 @Composable
 fun ToolbarCrossSellsIcon(onClick: () -> Unit, modifier: Modifier = Modifier) {
+  var isRotated by remember { mutableStateOf(false) }
+  val fullRotation by animateFloatAsState(
+    targetValue = if (isRotated) 360f else 0f,
+    animationSpec = tween(1500, 50),
```
StylianosGakis, line 66: Do we have specs from design about the easing that we want this rotation to have?
If not, I suppose the default `FastOutSlowInEasing` looks good enough?

app/design-system/design-system-hedvig/src/main/kotlin/com/hedvig/android/design/system/hedvig/Tooltip.kt
```diff
@@ -72,8 +72,8 @@ fun HedvigTooltip(
     if (!showTooltip) return@LaunchedEffect
     delay(0.5.seconds)
     transientShowTooltip = showTooltip
-    tooltipShown()
     delay(5.seconds)
+    tooltipShown()
```
StylianosGakis, line 76: So this was the way it was on purpose. It considered the tooltip as "shown" the moment it was in fact shown.

With this new change we are going to have this situation:
1. Member sees the tooltip as we show it.
2. 3 seconds later, they navigate away from this screen.
3. Then when they come back to this screen 
4. We show the tooltip again, since we never marked it as "seen"

That's why we marked it as seen asap, but then kept a "transient" state of if we show it or not, so that despite the fact that we have it marked as "seen" in the datastore or wherever we store this information

app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -329,6 +352,50 @@ private fun HomeScreen(
...
+@Composable
+private fun ColumnScope.CrossSellsTooltip(uiState: Success, setEpochDayWhenLastToolTipShown: (Long) -> Unit) {
+  if (uiState.crossSellsAction != null) {
+    // setEpochDayWhenLastToolTipShown(1L) //todo: remove testing
+    val shouldShowCrossSellsTooltip = uiState.crossSellsAction.crossSellRecommendationNotification.showToolTip
+    logcat {
+      "Mariia: ColumnScope.CrossSellsTooltip shouldShowCrossSellsTooltip: $shouldShowCrossSellsTooltip today: ${
+        java.time.LocalDate.now().toEpochDay()
+      }"
+    }
```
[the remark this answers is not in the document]
panasetskaya: nah, just left it for now, will remove before merge

PR #2649, opened 2025-09-12
Allow removing addon from coverage choosing screen

StylianosGakis:
Hardcoded parts that assume this is in fact a travel insurance
May or may not require additions from backend to avoid such hardcoding

(Video recorded before commit 8062bbf5ccbcc577cc99ebafe78016f78acc3335)
https://github.com/user-attachments/assets/1d9a9501-9539-4f8d-8a58-4ebd3ab4d055


panasetskaya (commented):
I think the addon should not have its separate card in the summary, should be included as a line in price breakdown for the main insurance - is it something you want to add in a separate pr?

panasetskaya (approved):

PR #2658, opened 2025-09-25
Fix/moving flow changes after ios discussion

StylianosGakis:
Update which fields are used from the backend to get the correct breakdown for all possible cases


panasetskaya (commented):
left comments [here in slack](https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1758809039778559?thread_ts=1758726318.795939&cid=C03HT2JRDPG)

PR #2910, opened 2026-03-30
Feature: pet chip (RND-1824)

panasetskaya:
- [x] check for accessibility
- [x] translations
- [x] publ schema to prod
- [x] BE not merged yet


app/feature/feature-chip-id/src/main/kotlin/com/hedvig/android/feature/chip/id/data/GetPetContractsForChipIdUseCase.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.apollographql.apollo.ApolloClient
+import com.apollographql.apollo.cache.normalized.FetchPolicy
+import com.apollographql.apollo.cache.normalized.fetchPolicy
+import com.hedvig.android.apollo.ApolloOperationError
+import com.hedvig.android.apollo.safeExecute
+import com.hedvig.android.data.contract.toContractGroup
+import com.hedvig.android.logger.logcat
+import octopus.GetPetContractsForChipIdQuery
+
+internal interface GetPetContractsForChipIdUseCase {
```
StylianosGakis: So this returns all contracts that are missing chipID right?
If yes, could we rename it to smth like `GetPetContractsWithMissingChipId` because `for chip id` is a bit vague about what it means imo

app/feature/feature-chip-id/src/main/kotlin/com/hedvig/android/feature/chip/id/ui/AddChipIdScreen.kt
```diff
@@ -0,0 +1,343 @@
...
+@Composable
+private fun ChipIdTextField(
+  textFieldState: TextFieldState,
+  labelText: String,
+  keyboardActionHandler: KeyboardActionHandler?,
+) {
+  val interactionSource = remember { MutableInteractionSource() }
+  val digitsOnlyTransformation = InputTransformation.byValue { _, proposed ->
+    proposed.filter { it.isDigit() }
+  }
```
StylianosGakis: Do you think it would be worth it for us to create a DigitsOnly InputTransformation just like some exist in the library itself https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/text/input/InputTransformation.kt;l=213-227?q=MaxLengthFilter which would also determine the KeyboardType.Number inside there so wherever we use it it's already properly preconfigured?
And have it live in design-system perhaps

app/feature/feature-chip-id/src/main/kotlin/com/hedvig/android/feature/chip/id/ui/AddChipIdViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+      if (submitIteration == 0) return@LaunchedEffect
+
+      submittingData = true
+      errorType = null
+
+      updateChipIdUseCase.invoke(insuranceId = contractId, petId = chipIdState.text.toString()).fold(
+        ifLeft = { error ->
+          Snapshot.withMutableSnapshot {
+            submittingData = false
+            errorType = ChipIdErrorType.GeneralError
```
StylianosGakis: Do we want to distinguish here between the generic errors and the errors that come from a UserError response? If those are properly localized etc we might want to surface them to the UI too?

app/feature/feature-chip-id/src/main/kotlin/com/hedvig/android/feature/chip/id/ui/AddChipIdScreen.kt
```diff
@@ -0,0 +1,343 @@
...
+  )
+
+  AnimatedContent(
+    targetState = uiState.errorType,
+    transitionSpec = { fadeIn() + expandVertically() togetherWith fadeOut() + shrinkVertically() },
+    modifier = Modifier.padding(top = 4.dp),
+  ) { errorType ->
+    if (errorType != null) {
+      val errorMessage = when (errorType) {
+        ChipIdErrorType.WrongInput -> stringResource(Res.string.CHIP_ID_WRONG_INPUT)
```
StylianosGakis, line 194: Does the wronginput error message stay on the screen until there is an attempt to re-submit again? Unless I am missing it somewhere here, could we dismiss this error when the member starts typing something? Or perhaps when they reach 15 digits, not sure which one is preferrable

hedvig-lint/lint-baseline/lint-baseline-feature-chip-id.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 9.0.0" type="baseline" client="gradle" dependencies="false" name="AGP (9.0.0)" variant="all" version="9.0.0">
+
+    <issue
+        id="UnknownIssueId"
+        message="Unknown issue id &quot;androidx.media3.common.util.UnstableApi&quot;">
+        <location
+            file="build.gradle.kts"/>
+    </issue>
```
StylianosGakis, line 9: I really wonder why this is messed up as it is here 🙈😭

StylianosGakis (approved):

app/feature/feature-chip-id/src/main/kotlin/com/hedvig/android/feature/chip/id/ui/AddChipIdScreen.kt
panasetskaya: I went with another transformation in the end, the visual mask like we do for ssn input:)

hedvig-lint/lint-baseline/lint-baseline-feature-chip-id.xml
panasetskaya: :elmofire:

StylianosGakis (approved):

PR #2885, opened 2026-03-12
Remove Odyssey from termination flow

hugokallstrom:
## Summary
- Replace 7 Odyssey flow-based GraphQL mutations with 2 direct operations: `terminationSurvey` query and `terminateContract` mutation (both served by Underwriter, which internally delegates to Product-Pricing for contract state changes)
- Client-driven navigation instead of backend state machine — survey options and termination action are fetched upfront, navigation computed locally
- Unified `DeflectSuggestion` destination replacing separate `DeflectAutoCancel` and `DeflectAutoDecom` screens
- Net deletion of ~1,100 lines (675 added, 1,804 removed)

## What changed
- **GraphQL**: 3 new operations, 7 old ones deleted
- **Data models**: Flat `TerminationSurveyData`/`TerminationAction`/`TerminationResult` replacing deep `TerminateInsuranceStep` sealed hierarchy
- **Repository**: 2-method interface (`getTerminationSurvey`, `terminateContract`) replacing 6 flow-based methods
- **ViewModels**: Survey VM navigates locally; Confirmation VM calls `terminateContract` directly with reason/comment
- **DI**: Removed `TerminationFlowContextStorage`, `FeatureManager`, `DataStore` dependencies
- **Tests**: Survey presenter tests rewritten; repo tests stubbed pending schema deployment

## Blocked on
- New GraphQL schema not yet deployed — `terminationSurvey` query and `terminateContract` mutation types don't exist in the schema yet. Build will fail on Apollo codegen until schema is updated.

## Test plan
- [ ] Deploy new GraphQL schema to staging
- [ ] Run `./gradlew downloadOctopusApolloSchemaFromIntrospection` and fix any codegen mismatches
- [ ] Verify full termination flow: choose insurance → survey → date picker → confirmation → success
- [ ] Verify deletion flow (no date picker)
- [ ] Verify deflection suggestions (AUTO_DECOMMISSION, AUTO_CANCEL)
- [ ] Verify change tier redirect from survey
- [ ] Run `./gradlew :feature-terminate-insurance:test`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2911, opened 2026-03-31
Termination flow UI fixes

hugokallstrom:
## Summary
- Changed survey subtitle to "Before you cancel"
- Fixed padding on info box (4.dp → 16.dp)
- Removed "Continue termination" for sold/scrapped car deflect screens
- Added "Contact us" button for those screens

Builds on top of #2885 addressing Slack UI feedback.

https://claude.ai/code/session_017Hvwsj3qNyt6VC4GfRC3H6

### DIFF
--- a/app/apollo/apollo-octopus-public/src/commonMain/graphql/com/hedvig/android/apollo/octopus/schema.graphqls
+++ b/app/apollo/apollo-octopus-public/src/commonMain/graphql/com/hedvig/android/apollo/octopus/schema.graphqls
@@ -1308,6 +1308,10 @@ type Contract {
   """
   coOwners: [ContractCoOwner!]
   """
+  Checks if petId is missing on the current or upcoming agreement.
+  """
+  missingPetId: Boolean!
+  """
   User messages to show due to self change is blocked.
   """
   selfChangeBlockers: ContractSelfChangeBlockers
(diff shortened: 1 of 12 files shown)


StylianosGakis (approved):
I think these changes look good, but also I believe it has missed some of the comments we made. Very likely because we made them by writing a very vague comment and added a screenshot to carry the rest of the context.
Perhaps we can merge this but then also have one of us manually go through the flow again to make sure we cover the other things too, like the date picking card not taking full width, making sure that “Cancel without a quote" is always the last option and not shuffled etc.

Note that I haven't ran this code to manually check myself yet, just from looking at the diff

hugokallstrom:
> I think these changes look good, but also I believe it has missed some of the comments we made. Very likely because we made them by writing a very vague comment and added a screenshot to carry the rest of the context. Perhaps we can merge this but then also have one of us manually go through the flow again to make sure we cover the other things too, like the date picking card not taking full width, making sure that “Cancel without a quote" is always the last option and not shuffled etc.
> 
> Note that I haven't ran this code to manually check myself yet, just from looking at the diff

You're absolutely right! ;) Seems like claude couldnt "see" the screenshots from Slack. I added a commit just now fixing the rest of the comments.

panasetskaya:
- i think Claude forgot about ordering of suboptions ("Cancel without a quote" shouldn't be on top)
- checked with iOs too, "Your car is back on the road" should be "We'll update your insurance automatically" 😄 and all the other strings in the car deflect strings need to be checked (Transportstyrelsen not mentioned etc.)

hugokallstrom:
Fixed the car deflect strings — now using the correct Lokalise keys:
- `AUTO_CANCEL_SOLD`, `AUTO_CANCEL_SCRAPPED`, `AUTO_CANCEL_DECOM` (all mention Transportstyrelsen)
- `AUTO_RECOMMISSION_TITLE` → "We'll update your insurance automatically"
- `AUTO_CANCEL_RECOMMISSION` → "Once your car is registered as active with Transportstyrelsen..."

Re: "Cancel without a quote" ordering — that's controlled by the backend, so will be handled separately there.

panasetskaya (approved):

PR #2932, opened 2026-04-29
Feature: manual charge (GEN-5217)

panasetskaya:
## a11y
- [x] if these are UI changes - check for their accessibility
- [x] translations
- [x] _merge backend corresponding branch first:_ [link](https://github.com/HedvigInsurance/account-service/pull/955)


app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/GetManualChargeInfoUseCase.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+internal class GetManualChargeInfoUseCaseImpl(
+  private val apolloClient: ApolloClient,
+) : GetManualChargeInfoUseCase {
+  override suspend fun invoke(): Either<ErrorMessage, ManualChargeInfo> = either {
+
+    val currentMember = apolloClient.query(ManualChargeInfoQuery())
+      .fetchPolicy(NetworkFirst)
+      .safeExecute(::ErrorMessage)
```
StylianosGakis: If we run a one-shot thing like `safeExecute` then the second response is never going to be read anyway. So we can just use NetworkOnly instead
```suggestion
      .fetchPolicy(NetworkOnly)
      .safeExecute(::ErrorMessage)
```

app/navigation/navigation-core/src/commonMain/kotlin/com/hedvig/android/navigation/core/HedvigDeepLinkContainer.kt
```diff
@@ -198,6 +200,10 @@ internal class HedvigDeepLinkContainerImpl(
   override val petIdWithContractId: List<String> = baseDeepLinkDomains.map { baseDeepLinkDomain ->
     "$baseDeepLinkDomain/pet-id?contractId={contractId}"
   }
+
+  override val manualCharge: List<String> = baseDeepLinkDomains.map { baseDeepLinkDomain ->
+    "$baseDeepLinkDomain/manual-charge" //todo: check with other platforms
```
StylianosGakis: Is this final now?
```suggestion
    "$baseDeepLinkDomain/manual-charge"
```

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigAppState.kt
```diff
@@ -162,6 +167,15 @@ internal class HedvigAppState(
     ),
   )
 
+  val showPaymentsBadge: StateFlow<Boolean> = missedPaymentNotificationServiceProvider
+    .prodImpl
```
StylianosGakis: Could we here just use `provide` to get the appropriate impl for if we're in demo mode or not?

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/ui/manualcharge/ManualChargeViewModel.kt
```diff
@@ -0,0 +1,105 @@
...
+          },
+          ifRight = {
+            screenState = ManualChargeUiState.Success(
+              manualChargeInfo = currentState.manualChargeInfo,
+              navigateToSuccess = Unit
+            )
+          }
+        )
+      }
+    }
```
StylianosGakis, line 65: This looks like if someone double-taps the action and triggers this twice then it will be sent twice to the backend, do you believe this is the case? If yes we could try to guard against it either by doing nothing if there's an ongoing request on the fly.

edit:
Thinking about this a bit more the old iteration should be cancelled as the LaunchedEffect key will cancel the previous scope aka cancel the request. This is probably 99% of the time fine, unless it is cancelled right after the request was sent, yet the response is still in the process of coming back.
If we know that the backend can handle these requests as idempotent then this shouldn't be a problem at all. If the backend will get confused by this being called twice then we'd need to solve it either on our end or on the backend at least.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/TriggerManualChargeUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+}
+
+internal class TriggerManualChargeUseCaseImpl(
+  private val apolloClient: ApolloClient
+): TriggerManualChargeUseCase {
+  override suspend fun invoke(): Either<ErrorMessage, Unit>  = either {
+    val result = apolloClient
+      .mutation(ManuallyChargeMemberMutation())
+      .safeExecute()
+      .mapLeft { raise(ErrorMessage()) }
```
StylianosGakis: A stylistic change I suppose, but since we `bind()` right after, just mapping it to an empty ErrorMessage should be enough, and the error will be raised on the next line anyway
```suggestion
      .mapLeft { ErrorMessage() }
```
I suppose the error message is kept empty because this message string is surfaced on the UI since we expect the userError.message to be user-presentable otherwise. It'd maybe be a convenient thing to at least log something here in that case so that we don't ever have to chase after an empty ErrorMessage

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/MemberCharge.kt
```diff
@@ -167,7 +169,8 @@ internal fun MemberChargeFragment.toMemberCharge(
 
 internal fun String?.toChargeMethod(): MemberPaymentChargeMethod {
   return when {
-    this?.startsWith("kivra", ignoreCase = true) == true -> MemberPaymentChargeMethod.KIVRA
+    this?.startsWith("kivra", ignoreCase = true) == true ||
+      this?.startsWith("invoice", ignoreCase = true) == true -> MemberPaymentChargeMethod.KIVRA
```
StylianosGakis: Kivra and invoice are both just clumped to Kivra, this is because they are both going through Kivra anyway? Is this a safe assumption to make here? I am asking because I am just missing context, not that I believe it's wrong.

app/notification-badge-data/notification-badge-data-public/src/main/kotlin/com/hedvig/android/notification/badge/data/payment/GetIfMissedPaymentUseCase.kt
```diff
@@ -0,0 +1,53 @@
...
+                  logcat { "GetIfMissedPaymentUseCaseImpl: error when loading missed payment: $it" }
+                  false
+                },
+                { data ->
+                  data.currentMember.missedChargeIdToChargeManually != null
+                },
+              )
+            },
+        )
+        delay(5.seconds)
```
StylianosGakis: So this use case ends up being a hot flow on HedvigAppState which is read by the HedvigAppUI. This means that this will in fact run 12 times a minute for every single person with the app open at any given moment, whereas the people interested seeing this dot will be a comparatively a small number of people.

I think we need to avoid doing this, have we discussed this with iOS too, perhaps worth making a little thread, I will start one

StylianosGakis (approved):

app/notification-badge-data/notification-badge-data-public/src/main/kotlin/com/hedvig/android/notification/badge/data/payment/GetIfMissedPaymentUseCase.kt
StylianosGakis: https://hedviginsurance.slack.com/archives/C0AUJD7PRJ6/p1778249665165339

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/data/MemberCharge.kt
panasetskaya: we probably should name MemberPaymentChargeMethod.Kivra as MemberPaymentChargeMethod.Invoice instead too, since Kivra is just one of the Invoice subtypes. 
We only have Trustly, Kivra, Unknown now.

app/feature/feature-payments/src/main/kotlin/com/hedvig/android/feature/payments/ui/manualcharge/ManualChargeViewModel.kt
panasetskaya: added loading state for the button to block

PR #2935, opened 2026-05-11
Fix: if exception during setDataSource fail gracefully

panasetskaya:
show error state instead of crashing on this type of recording errors: [crashlytics](https://console.firebase.google.com/u/0/project/hedvig-app/crashlytics/app/android:com.hedvig.app/issues/0a055ca8cbb72888bf577de8b6d7d918?time=7d&types=crash&sessionEventKey=6A0182E9015400011634258979F67FDC_2216976214432843549)


StylianosGakis (approved):
Since we only got the file not found exception now, could we do without the try/catch in AudioRecordingManager so that it does not swallow any future exceptions which we might miss? And keep this part
```
val file = File(filePath)
    if (!file.exists()) {
```

That way we cover the crash we got now but don't silently miss any future problems

PR #2939, opened 2026-05-15
Chore: deprecated unleash

panasetskaya:
Related to [this unleash bug](https://github.com/Unleash/unleash-android-sdk/issues/141). 

Unleash Frontend API only returns enabled feature toggles, so any default true features will not be able to turn off. They deprecated the overload with default value, and instead recommend to use new kill-switch flags for such features.

Only 2 flags affected for us, HELP_CENTER and TERMINATION_FLOW. Kill-switches for both added to unleash.


StylianosGakis (approved):
Thank you for this! And did we deprecate the old flags in Unleash as well?
Also what should we do for iOS, could we make a PR there as well which updates calling these new flags as well for these two features?

panasetskaya (commented):
well I asked Sladan and he was fine with how it works now for them, so dunno about ios actually. I didn't deprecate the current ones in unleash

StylianosGakis:
> well I asked Sladan and he was fine with how it works now for them, so dunno about ios actually. I didn't deprecate the current ones in unleash

https://github.com/HedvigInsurance/ugglan/pull/2444 I created a PR there too

This will allow us to eventually delete the duplicated flags

PR #2955, opened 2026-05-26
Chore: migrate to one markdown library

panasetskaya:
Migrate to one markdown in the whole project (remove com.halilibo.compose-richtext and leave only KMP mikepenz)
With help of AI, can't add either initial prompt or further tweaks though, lost when cleared history without thinking:)


StylianosGakis (approved):
🌟

PR #2963, opened 2026-06-02
Add OSS license gate to Android (RND-1996)

hugokallstrom:
What
An action that scans licenses in a repo and blocks for certain licenses that we can not include because of compliance reasons.

This action will run on every PR and will scan licenses for dependencies

See the full action impl here: https://github.com/HedvigInsurance/prod-env/pull/757

Why
For compliance reasons we need to make sure we dont merge certain licenses. For example, adding a dependency with strong copy-left license would mean we have to open source our source code.

### DIFF
--- a/.github/workflows/license-inventory.yml
+++ b/.github/workflows/license-inventory.yml
@@ -1,0 +1,11 @@
+name: license-inventory
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


StylianosGakis (approved):
Let's try this!

PR #2973, opened 2026-06-05
Simplify providing all the NotificationSenders to Metro

StylianosGakis:
Each sender's HedvigNotificationChannel was fixed but was provided from the outside. Moving it inside the class itself makes it possible to simply annotate them with @ContributesIntoSet and get the same result


panasetskaya (approved):

StylianosGakis:
Merged by
https://github.com/HedvigInsurance/android/pull/2975

PR #3004, opened 2026-06-23
Feature/firebase analytics [RND-2072]

StylianosGakis:
Add firebase tracking

Add screen tracking as the first example usage. Track the screen's name and parameters by default, with `TrackedScreen` being a way to be more deliberate about what to track exactly


panasetskaya (approved):
baseline file probably not pushed?

app/app/src/main/kotlin/com/hedvig/android/app/ui/HedvigApp.kt
```diff
@@ -287,6 +293,33 @@ private fun ReportCurrentDestinationEffect(
...
+  eventTrackingClient: EventTrackingClient,
+  screenParameterExtractor: ScreenParameterExtractor,
+) {
+  LaunchedEffect(backstackController, eventTrackingClient, screenParameterExtractor) {
+    snapshotFlow { backstackController.currentDestination }
+      .filterNotNull()
+      .collect { destination ->
+        val screenClass = destination::class.simpleName ?: destination.toString()
+        val screenName = screenClass.removeSuffix("Key")
+        eventTrackingClient.trackScreen(
```
panasetskaya, line 314: 🔥

PR #3006, opened 2026-06-24
Feature: xsell addon recommendation [GEN-5439]

panasetskaya:
## What

Adds support for the addon recommendation in the cross-sell sheet shown after a successful self-service flow.

- The cross-sell sheet and Home queries now fetch `recommendedAddon`, and the sheet renders it (pillow with plus badge, title, description, button) with priority over the recommended cross-sell.
- Flows that change a contract now pass its contract id in `CrossSellInput`, which is what makes the backend return an addon recommendation: change tier passes the changed contract id, the moving flow passes `newContractId` from the commit mutation, closed claims pass the claim's `contractId`.
- Addon-only sheets show the green banner with the default `CROSS_SELL_BANNER_TEXT` copy ("Popular choice"), since `RecommendedAddonCrossSell` has no banner text field. Same fallback as iOS.
- A successful response with nothing to show (no recommendation, no addon, no other cross-sells) no longer opens an empty sheet.
- Home's top bar cross-sells action now also shows if only an addon recommendation is returned (defensive; Home sends no contract id today).

## Remaining scope (waiting on storefront)

The design also shows a benefits checkmark list and per-type banner text ("Add extra safety when traveling" / "Add extra coverage"). Per the sync on Jul 3, storefront is getting two new properties for this (an optional banner and a list of strings) and the plan is to consume them here and get a design preview before merging. Until the schema publishes, the banner falls back to the default copy and the list is not rendered.

## Testing

- Unit tests: addon mapping in `GetHomeDataUseCase`, the Home top bar gate, `CrossSellInput` contract id present/absent, and the sheet presenter's empty/non-empty handling.
- Manually verified against staging: addon card after a change tier flow (button deep links in-app), recommended cross-sell fallback for a flow without a contract id, and the empty response case.

## a11y

[pr in storefront](https://github.com/HedvigInsurance/storefront/pull/1300)

todo:
- [x] pass new contractId from moving flow
- [x] (optionally) pass contract Id from claim closed [pr in claim service](https://github.com/HedvigInsurance/claims-service/pull/1250/changes)
- [x] wait for storefront pr schema published to prod


app/data/data-cross-sell-after-flow/src/main/kotlin/com/hedvig/android/data/cross/sell/after/flow/CrossSellAfterFlowRepository.kt
```diff
@@ -33,6 +35,7 @@ sealed class CrossSellInfoType() {
     val info: ClaimInfo,
   ) : CrossSellInfoType() {
     override val source: String = "closedClaim"
+    override val contractId: String? = null //todo!
```
StylianosGakis: Is this a real todo or do we just want it to be `null` on purpose?

app/data/data-cross-sell-after-flow/src/main/kotlin/com/hedvig/android/data/cross/sell/after/flow/CrossSellAfterFlowRepository.kt
```diff
@@ -52,22 +55,28 @@ sealed class CrossSellInfoType() {
...
+    override val contractId: String?
+  ) : CrossSellInfoType() {
     override val source: String = "changeTier"
     override val extraInfo: Map<String, Any?>? = null
   }
 
   data object Addon : CrossSellInfoType() {
     override val source: String = "addon"
     override val extraInfo: Map<String, Any?>? = null
+    override val contractId: String? = null //todo!
```
StylianosGakis: Same as above ^^

app/data/data-cross-sell-after-flow/src/main/kotlin/com/hedvig/android/data/cross/sell/after/flow/CrossSellAfterFlowRepository.kt
```diff
@@ -52,22 +55,28 @@ sealed class CrossSellInfoType() {
...
   }
 
   data object EditCoInsured : CrossSellInfoType() {
     override val source: String = "editCoInsured"
     override val extraInfo: Map<String, Any?>? = null
+    override val contractId: String? = null //todo!
```
StylianosGakis: Same as above ^^

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
```diff
@@ -205,7 +208,15 @@ private fun CrossSellsSheetContent(
...
-      if (recommendedCrossSell != null) {
+      if (recommendedAddon != null) {
+        Spacer(Modifier.height(16.dp))
+        AddonRecommendationSection(
+          recommendedAddon,
+          onButtonClick = onCrossSellClick,
+          dismissSheet = dismissSheet,
+          imageLoader = imageLoader,
+        )
+      } else if (recommendedCrossSell != null) {
```
StylianosGakis, line 228: Do we ever want to show both at the same time? Just curious because now it's an either or situation, can never show both

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
```diff
@@ -265,7 +277,14 @@ private fun CrossSellsFloatingSheetContent(
           .padding(bottom = 24.dp),
         verticalArrangement = Arrangement.spacedBy(40.dp),
       ) {
-        if (recommendedCrossSell != null) {
+        if (recommendedAddon != null) {
```
StylianosGakis, line 289: Same as above

StylianosGakis (approved):

app/data/data-cross-sell-after-flow/src/main/kotlin/com/hedvig/android/data/cross/sell/after/flow/CrossSellAfterFlowRepository.kt
panasetskaya: removed those, closed Claim now passes a contractdId

app/ui/cross-sells/src/main/kotlin/com/hedvig/android/crosssells/CrossSells.kt
panasetskaya: I don't think so. Right now we only show addon rec in smart-crossSell (where there is no other cross sells except for that). 
And backend also returns either addon + nothing else, or no addon - and something else.

StylianosGakis (approved):

PR #3007, opened 2026-06-26
Feature: new pillows [GEN-5452]

panasetskaya:
add new pillows, fix contract type mapping


app/feature/feature-insurances/src/main/kotlin/com/hedvig/android/feature/insurances/data/GetInsuranceContractsUseCase.kt
```diff
@@ -145,6 +146,7 @@ private fun ContractFragment.toContract(
   contractHolderSSN: String?,
   isMovingFlowEnabled: Boolean,
 ): EstablishedInsuranceContract {
+  logcat { "Mariia: contract type: ${currentAgreement.productVariant.typeOfContract}" }
```
StylianosGakis: log

StylianosGakis (changes requested):
I think before merging we can try to do this too https://developer.android.com/studio/write/convert-webp

Should cut down the size of each one of these like to make them like 5-10x smaller than what they are even with the lossless option without any visual differences.

StylianosGakis (approved):

PR #3019, opened 2026-07-14
Render structured claim summary answers (RND-2030)

hugokallstrom:
## Why

The claims-service summary step now exposes `keyDetails` (facts for a collapsed
"Claim details" card) and `answers` (ordered question/answer pairs whose value is
a Text / Audio / Files union), for the new "claims-flow-new" summary UI. Today the
Android summary renders one flat, always-expanded card, so it can't show the new
collapsed card + per-question answers (with inline audio playback).

## What

- `SummaryFragment` selects the new `keyDetails` and `answers` fields; `StepContent.Summary`
  gains `keyDetails` + `answers` (with an `Answer.Value` = Text/Audio/Files model), mapped
  in `ClaimIntentExt`.
- `ChatClaimSummaryTopContent` renders a collapsed "Claim details" card from `keyDetails`
  (falling back to `items` when empty) plus a "Show all answers" button — shown only when
  `answers` is non-empty — that opens a `HedvigBottomSheet` listing each question with its
  answer inline (text, a `HedvigAudioPlayer` for recorded answers, or files). The existing
  Recording and Uploaded-files sections are unchanged.
- Adds the `claim_status_show_all_answers` string.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


StylianosGakis (approved):
Regarding the
```
else -> {
 logcat { "SummaryFragment.Answer: Unknown answer value type" }
 raise(ClaimChatErrorMessage.NeedsUpdate)
}
```
When the answer is not of a known value, this I think would error out the entire screen, which if we do make such a change might be a bit too much.
How would we feel if we  just ignored such unknown values, logged it, and did not render anything there, just rendering whatever else is available anyway to prevent blocking the rest of the content from showing?

I think the val freeTexts: List<String>, part of data class Summary is now completely unused too, perhaps we can clean that part up if it’s not going to be used.

PR #3023, opened 2026-07-20
Feature: New home screen

StylianosGakis:
New Home Screen layout from Figma: https://www.figma.com/design/SogcacjzOxkCC46XcZP8lQ/App-P2-2026?node-id=429-7853&m=dev

image

Omit the parts that we do not yet have, like the news section, to be follow-up PRs


app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/ui/HomeDestination.kt
```diff
@@ -761,6 +1373,7 @@ private fun WelcomeMessage(homeText: HomeText, modifier: Modifier = Modifier) {
         fontSize = 28.0.sp,
         lineBreak = LineBreak.Heading,
         textAlign = TextAlign.Center,
+        color = HedvigTheme.colorScheme.textSecondary,
```
StylianosGakis:
```suggestion
```

PR #3035, opened 2026-08-04
Feature: send claimId for closed-claim cross-sell flow

StylianosGakis:
Thread the claim id into CrossSellInput.claimId for the CLOSED_CLAIM flow source. The backend added the optional claimId field to CrossSellInput (only allowed when flowSource is CLOSED_CLAIM) and asked clients to populate it for the closed-claim x-sell flow.


victor-reyes-by (approved):
🚀

PR #3056, opened 2026-08-14
chore/dedebug

StylianosGakis:
## a11y


panasetskaya (approved):

panasetskaya:
Nice! So we'll build half as much for tests, if I understood correctly, without "debug" for a lot of modules?

StylianosGakis:
Yeah I think that's part of the idea, just getting rid of things we never needed in the first place. Here is more info https://autonomousapps.com/blog/simplify-your-android-library-development/post/

PR #3058, opened 2026-08-14
Bug/onboarding flag issue

panasetskaya:
Fix for the bug where onboarding showed on every cold start regardless of the disable_onboarding F flag.

Seems like it was an existing bug that never surfaced since we don't have any other flag that is needed only once on start and is never re-fetched again.

The cause description from Claude:

>     Root cause was our own bootstrap list. DefaultUnleash.start(bootstrap = …) writes the seeded                                                                      
>      toggles into the cache like any other state; readyOnFeaturesReceived() flips ready on the first                                                                   
>      non-empty cache update, so bootstrap alone made isReady() true instantly; and                                                                                     
>      initializeLocalBackup() only loads the on-disk backup while !ready, so seeding bootstrap also                                                                     
>      meant the backup could never load. Net effect: awaitReady() returned immediately with only the                                                                    
>      three bootstrapped toggles in cache, and every other flag read false until the first poll landed                                                                  
>      ~1s later. OnboardingGate is the app's only one-shot (.first()) flag read that runs at startup,                                                                   
>      so it was the only place the defect was visible.                                                                                                                  

So the fix mainly removed bootstrap from isReady(), and did adjustments: added 3 existing defaults for 3 flags to neverFetchedDefaults map consulted only while !client.isReady(), added valueOf(feature), switched featureUpdatedFlow to UnleashStateListener.onStateChanged().

### DIFF
--- a/app/featureflags/feature-flags/src/androidMain/kotlin/com/hedvig/android/featureflags/HedvigUnleashClient.kt
+++ b/app/featureflags/feature-flags/src/androidMain/kotlin/com/hedvig/android/featureflags/HedvigUnleashClient.kt
@@ -7,11 +7,9 @@ import com.hedvig.android.featureflags.flags.unleashKey
 import com.hedvig.android.logger.logcat
 import io.getunleash.android.DefaultUnleash
 import io.getunleash.android.UnleashConfig
-import io.getunleash.android.data.Toggle
 import io.getunleash.android.data.UnleashContext
-import io.getunleash.android.events.HeartbeatEvent
-import io.getunleash.android.events.UnleashFetcherHeartbeatListener
 import io.getunleash.android.events.UnleashReadyListener
+import io.getunleash.android.events.UnleashStateListener
 import kotlin.coroutines.resume
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.channels.awaitClose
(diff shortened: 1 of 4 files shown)


app/featureflags/feature-flags/src/androidMain/kotlin/com/hedvig/android/featureflags/HedvigUnleashClient.kt
```diff
@@ -26,31 +24,43 @@ private const val DEVELOPMENT_CLIENT_KEY = "*:development.f2455340ac9d599b5816fa
...
-
-      override fun togglesChecked() {
-      }
-
-      override fun togglesUpdated() {
+    val listener = object : UnleashStateListener {
+      override fun onStateChanged() {
         trySend(Unit)
       }
     }
```
StylianosGakis: I think I am mostly concerned with this, if it's equivalent and fits all of our needs here.

I think we can make this a bit more complete with this change
```suggestion
    val listener = object : UnleashStateListener, UnleashReadyListener {
      override fun onStateChanged() { trySend(Unit) }
      override fun onReady() { trySend(Unit) }
    }
```

StylianosGakis (approved):

PR #3065, opened 2026-08-18
fix for crossSell pics rough edges

panasetskaya:
one of the things we talked about with Nils - the xsell and quotes pillow edges were rough on Android (pixelized): Home, Insurances, onboarding and the cross-sell sheet. The image was larger than the box it was drawn into, and the on-device downscale lost detail.
<img width="219" height="165" alt="rough_edges" src="https://github.com/user-attachments/assets/e007627f-14eb-4f78-9217-78625bcae155" />
                                                                                                              
Solution in 3 parts:                                                                                     
                                                                                                                                                                       
- variant selection, where there is one.`CrossSell` exposes both`pillowImageSmall` and `pillowImageLarge`, but we only ever selected large and used it at every size. 
- CDN resize, where it doesn't. `ShopSessionDisplay.pillowImage` has no variants and is published at 832x832, which we were collapsing into a 126px box. Use `storyblokResized()` which appends Storyblok's `m/126x126`
- asset resize (where we don't use BE for pillows - for addons). The 12 `data-contract` pillows were 400x400 but are drawn at 40-48.dp everywhere; they're now 192x192 (48.dp at xxxhdpi).These also back QuoteCard, the moving flow, chip-id and onboarding contract cards.


StylianosGakis (approved):

StylianosGakis:
Really interesting! This is a good TIL

PR #3069, opened 2026-08-19
Never show the payments quick action on home

StylianosGakis:
## a11y


panasetskaya:
was there a decision to hide it from Home?

panasetskaya (approved):

PR #3070, opened 2026-08-19
Close onboarding payment flow on user cancellation instead of offering retry

StylianosGakis:
## a11y


panasetskaya:
do we have a payment flow in the onboarding? it's just the connect trustly flow, from anywhere, right?

panasetskaya (approved):

PR #3082, opened 2026-08-21
Fix: ppi subtitles

panasetskaya:
PPI needs different wording for edit tier. Sister to this [ios PR](https://github.com/HedvigInsurance/ugglan/pull/2584)


StylianosGakis (approved):
Regarding the the minHeight change, how does it now look like in very small font size situations? Usually these elements don’t have a min size so that when the font size is smaller the entire thing shrinks along with it to keep it more proportional.

https://hedviginsurance.slack.com/archives/C03HT2JRDPG/p1787324247960809?thread_ts=1787323597.713849&cid=C03HT2JRDPG

PR #3092, opened 2026-08-25
Share one phone number field across onboarding, profile and Swish

StylianosGakis:
## Problem

Three screens take a phone number and agreed on nothing. Onboarding and Profile write the **same** `memberUpdateContactInfo` field yet disagreed on what it may hold. Swish writes its own field (`paymentMethodSetupSwishPayout`) and had no rules at all.

| | Onboarding | Profile | Swish |
|---|---|---|---|
| Character filter | none | whitespace only, letters pass | none |
| Line limits | multiline | multiline | multiline |
| Minimum | 6 characters | 6 digits | 10 characters |
| Malformed value on submit | accepted | rejected | n/a |

Live consequences: a number could be typed across several lines and saved that way, Swish's Save enabled on `abcdefghij`, and onboarding would re-save a malformed stored number that Profile rejected. Numbers with line breaks in them **are already in the backend** as a result.

## The rules, in one place

`PhoneNumberRules` lives in `core-common-public/commonMain/.../validation`, next to the existing `ValidateEmail` that Profile's `Email` type already uses. It is plain Kotlin. `ui-phone-number` holds only what needs Compose: `HedvigPhoneNumberField` and the `InputTransformation`.

| | `MemberPhoneNumber` | `SwishPhoneNumber` |
|---|---|---|
| Leading `+` | allowed | not allowed |
| Minimum | 6 digits | 10 digits |
| Maximum | 15 digits (E.164) | 15 digits |
| One line, no return key | yes | yes |

Landlines run to eight or nine digits and foreign numbers are equally valid for a member's own number, so that minimum is a sanity floor. Swish pays out to a Swedish mobile, so ten digits is the real length and a country code is never part of the input.

**Counts are in digits, and ASCII digits.** Onboarding counted characters, so with `+` allowed `+12345` would have passed there while Profile rejected it. Swish counted characters, which is what let ten letters enable Save. `Char.isDigit()` is true for Arabic-Indic digits that Profile's own regex rejects, so digits are `'0'..'9'`.

**A `+` is accepted but never interpreted, and never silently dropped.** Nothing prepends a country code or drops a leading zero: treating `+46` and `07` as the same number needs a country-code selector plus a normalising step, and half of that is worse than none. A `+` is kept when it leads the number, and otherwise the edit is refused rather than mangled, because dropping it turns an international number into a domestic one that does not exist.

**Profile validates through these rules too**, rather than keeping its own regex and its own copy of the minimum. Its tests pin down thirteen shapes that must pass or fail, and all thirteen behave identically through the shared rules, which is what makes this a consolidation rather than a rewrite.

## Why a stored value is never reshaped

`ContactInformation.phoneNumber` is documented as being shown "exactly as the backend holds it, which is not necessarily a valid `PhoneNumber`… so that the member can correct it". So while a value is malformed, only edits that would add *further* disallowed characters are refused; deleting a stray separator or typing the rest of the number still works. Once a value is well formed, every disallowed character in an edit must be one that edit introduced, so filtering the whole value is the same as filtering only the insertion, which is what lets a number pasted with spaces or dashes come out clean rather than being rejected whole.

Onboarding drops whitespace before validating and submitting, as Profile always did, so the common stored shape `070 123 45 67` submits rather than being called invalid.

## Verification

21 unit tests on the rules, run as `commonTest` (CI covers these via `jvmTest`). Full `test jvmTest lint` green, ktlint clean on every module touched.

Driven on a device against a real test member:

- `+46701234567abc-x` → `+46701234567`; the IME's return key is a checkmark, so there is no newline key to press.
- The onboarding keypad, which bypasses the input transformation entirely (see below), stops at exactly 15 digits: 12 taps past the cap added 7.
- That member's stored number was literally `01234\n99\n3`. It renders on one line, Save was refused while it was malformed, and a clean `0701234567` then submitted successfully, `memberUpdateContactInfo` returning it with no error.
- Profile: the same stored value renders on one line where it previously wrapped to three; typing `abc5+x` appends only the `5`; `ImeAction.Next` and focus-hides-error unchanged.

## Things a review of this branch already caught and fixed

Worth reading, because two were subtle:

- **`TextFieldState.edit` skips `InputTransformation`.** It is a developer edit; only `editAsUser` applies the transformation. Onboarding's in-app keypad appended through `edit`, so it walked straight past the digit cap. Keypad input now goes through `appendPhoneNumberDigit`, which holds the cap and documents why it has to.
- **An over-long value trapped the member.** Every edit from a value past the cap proposes something still past it, so backspace was refused along with everything else. Shortening is now always allowed.
- A `+` was dropped when anything preceded it, `(+46) 70 …` included.
- Onboarding rejected the space-formatted stored numbers Profile accepts.
- Swish silently dropped a pasted `+` and enabled Save on the remains, which was a regression against the old length check.

## Worth a reviewer's attention

- **Profile is stricter while typing now.** It used to accept letters and complain only at submit. An improvement, but a behaviour change on a screen outside the ticket that prompted this. It also now rejects a stored number longer than fifteen digits, which its regex never checked.
- **Swish keeps `maxDigits = 15`** even though ten is a Swedish mobile's real length, so `07012345678901` still enables Save. Tightening a cap on a payout input wants a decision rather than an inference from a KDoc, and it is no looser than the length check it replaced.
- **The digit minimum is 6, not 7.** Per the last word in the Slack thread; one constant if we want 7.
- **Corrupt data already in the backend** is not cleaned by this. Both screens now refuse to re-save it, and fixing it in the field works, but nothing prompts a member who never opens either screen.


app/core/core-common-public/src/commonMain/kotlin/com/hedvig/android/core/common/validation/PhoneNumberRules.kt
```diff
@@ -0,0 +1,109 @@
...
+  fun digitsIn(text: CharSequence): Int = text.count { it.isAsciiDigit() }
+
+  fun hasEnoughDigits(text: CharSequence): Boolean = digitsIn(text) >= minDigits
+
+  /**
+   * Whether [text] holds nothing but what these rules allow. A field built on these rules keeps typed
+   * input this way on its own, so the case worth checking before submitting is a stored value that
+   * was never valid.
+   */
+  fun isWellFormed(text: CharSequence): Boolean {
```
panasetskaya, line 26: what if the member inserts a number with dashes? will we clean it earlier?

app/core/core-common-public/src/commonMain/kotlin/com/hedvig/android/core/common/validation/PhoneNumberRules.kt
```diff
@@ -0,0 +1,109 @@
...
+     * as eight digits and foreign numbers are equally valid, so the minimum only has to rule out
+     * input too short to be a number at all.
+     */
+    val MemberPhoneNumber = PhoneNumberRules(minDigits = 6, allowLeadingPlus = true)
+
+    /**
+     * Swish pays out to a Swedish mobile number, so the country code is never part of the input and
+     * ten digits is the real length rather than a sanity floor.
+     */
+    val SwishPhoneNumber = PhoneNumberRules(minDigits = 10, allowLeadingPlus = false)
```
panasetskaya, line 86: When we'll start Swish project, I'll check the swish requirements better, I think they‘ll have the upper limit too

panasetskaya (approved):

app/core/core-common-public/src/commonMain/kotlin/com/hedvig/android/core/common/validation/PhoneNumberRules.kt
StylianosGakis: 🤖 🤖 🤖 
Good catch, it was half-handled. Two paths, and only one was clean:

- **Typed or pasted dashes** were already stripped by the field, so `070-123-45-67` became `0701234567` on the way in.
- **A dash already in the stored value** was not. The field shows a stored number exactly as the backend holds it so the member can fix it, and `isWellFormed` then called it invalid on submit, so Save showed "That does not appear to be a valid phone number" against a number that reads perfectly fine, and they had to delete the dashes by hand.

Reachable too: until this branch, onboarding's Save only checked `length >= 6` with no character validation, which is exactly where the `01234\n99\n3` on my test member came from. Dashes would have gone in the same way.

Fixed in e879b1224d: submission now drops separators, so a stored `070-123 45 67` submits as `0701234567` and repairs itself. A letter still makes the whole value unreadable rather than being silently dropped, since that is a mistake rather than formatting, and so does a `+` that does not lead the number. Profile drops its whitespace-only trimming for the same cleaning, so both screens agree.

Profile's own thirteen valid/invalid fixtures all still behave identically, which is what made me comfortable changing its factory.

app/core/core-common-public/src/commonMain/kotlin/com/hedvig/android/core/common/validation/PhoneNumberRules.kt
StylianosGakis: 🤖 🤖 🤖 
Agreed, and that matches why I left it. A reviewer flagged that `SwishPhoneNumber` inherits `maxDigits = 15` while the KDoc claims ten is the real length, so `07012345678901` still enables Save. I did not tighten it: that is a payout input and "at least 10 digits" was what we landed on in the thread, so it wanted a real requirement rather than an inference from a comment I wrote. It is also no looser than the `text.length >= 10` check it replaced.

So when you get to the Swish requirements, the change should be a one-liner: add `maxDigits = <n>` to `SwishPhoneNumber`. Everything else, the single line, the digit counting, the refusal to silently drop a `+`, already applies there.

PR #3093, opened 2026-08-25
Payment protection: qualification period info on change amount summary

StylianosGakis:
Android side of [ugglan#2592](https://github.com/HedvigInsurance/ugglan/pull/2592). [GEN-5624](https://app.notion.com/p/hedviginsurance/App-If-increased-amount-in-change-amount-flow-Information-box-about-90-day-qualification-period-3c2c3f71cb0a80729756c7847c13419a)

When a member **increases** their payment protection amount, the change amount summary screen now shows an info card about the qualification period before the higher amount takes effect.

## Changes

- `SummaryState.Success` gains `showQualificationPeriodInfo`, computed in `SummaryPresenter` where both the selected and the current quote are already loaded. True only when the contract group is `PAYMENT_PROTECTION` **and** the selected quote's tier level is above the current one, so lowering or keeping the amount, and every other contract group, show nothing. Same predicate as iOS.
- `SummaryDestination` renders it as a `HedvigNotificationCard(..., Info)` directly above the `Total` row, matching how the moving flow places its "other insurances" notice (`feature-movingflow/.../summary/SummaryDestination.kt`), which is the treatment Richard asked for.
- The preview's `Success` fixture is extracted to a `previewSuccessState` val so the provider can add a second entry with the card visible without duplicating the 60-line quote.
- New `SummaryPresenterTest` covers the three cases from the iOS tests: payment protection + increase, payment protection + no increase, non-payment-protection + increase.

## Copy

Uses `CHANGE_INSURANCE_AMOUNT_PAYMENT_PROTECTION_INFO`, the renamed key, with both the English and Swedish copy in place:

> Your updated amount will take effect after your qualification period. Your current amount remains active during this period.

> Det nya försäkringsbeloppet börjar gälla efter att kvalificeringstiden passerat. Ditt nuvarande belopp gäller som vanligt under tiden.

Per Daniel's point that qualifying periods differ (sick leave, self-employed), the copy states no fixed number of days, so nothing in the app hardcodes 90.

**Note for iOS:** ugglan#2592 still references the old `CHANGE_TIER_PAYMENT_PROTECTION_INFO`, which no longer exists in Lokalise. It needs the same rename before it builds against current strings.

## Notes

- The first commit is a plain `./gradlew downloadStrings` run. Besides the key above it incidentally picks up the `CLAIM_CHAT_STROLLER_THEFT_*` keys that landed in Lokalise in the meantime.

### DIFF
--- a/app/core/core-resources/src/androidMain/res/values-sv-rSE/strings.xml
+++ b/app/core/core-resources/src/androidMain/res/values-sv-rSE/strings.xml
@@ -152,6 +152,7 @@
   <string name="CHANGE_ADDRESS_YEAR_OF_CONSTRUCTION_ERROR">Ange byggår</string>
   <string name="CHANGE_ADDRESS_YEAR_OF_CONSTRUCTION_LABEL">Byggår</string>
   <string name="CHANGE_ADDRESS_YOU_PLUS">Du + %1$d</string>
+  <string name="CHANGE_INSURANCE_AMOUNT_PAYMENT_PROTECTION_INFO">Det nya försäkringsbeloppet börjar gälla efter att kvalificeringstiden passerat. Ditt nuvarande belopp gäller som vanligt under tiden.</string>
   <string name="CHANGE_PAYOUT_METHOD_BUTTON_LABEL">Ändra utbetalningskonto</string>
   <string name="CHANGE_TIER_BUTTON_TITLE">Ändra skyddsnivå</string>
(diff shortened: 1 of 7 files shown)


panasetskaya (approved):

StylianosGakis:
https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1787733972793869?thread_ts=1787644402.111539&cid=C03U9C6Q7TP
More info here. Might need adjustment before merge

panasetskaya (changes requested):
will be done from BE

PR #3094, opened 2026-08-25
Hide expired quote cards on home

StylianosGakis:
Bottom of a two PR stack.

## What

The Your quotes section rendered whatever `ongoingShopSessions` the backend returned, in whatever order it returned them. Two consequences:

1. **Expired quotes stayed on screen.** `ShopSessionDisplay.validTo` documents itself as the client's tool for this:

   > Mirrors the underlying session's `validTo`. Clients can use this to hide cards that are expired.

   We never selected the field, so an expired session sat on home until the backend stopped returning it.

2. **Card order was arbitrary.** The mapping was a bare `.map {}`, so ordering was whatever the response happened to contain.

## How

Select `validTo` and `lastActivityAt`, drop sessions that are no longer valid, and put the most recently touched session first. Neither field is added to the `OngoingShopSession` model, since nothing reads them past the filter and sort.

## Testing

Two new tests in `GetHomeUseCaseTest`, both watched failing before the fix:

* the sorting test returned backend order (`middle, oldest, newest`) against the expected `newest, middle, oldest`
* the expiry test returned 2 sessions including `expired-session` where 1 was expected

The two pre-existing shop session tests still pass. `OctopusFakeResolver` resolves `DateTime` to `Instant.DISTANT_FUTURE`, so their unset `validTo` reads as not expired.

## Note for the reviewer

One assumption worth a second opinion: that the backend does return expired sessions. That follows from the schema telling clients to hide them (a server that filtered would not need to), but I did not confirm it against a live response.

## a11y

### DIFF
--- a/app/feature/feature-home/src/main/graphql/QueryHome.graphql
+++ b/app/feature/feature-home/src/main/graphql/QueryHome.graphql
@@ -89,7 +89,9 @@ query Home($claimsHistoryFlag: Boolean!, $resumeClaimEnabled: Boolean!, $disable
           amount
           currencyCode
         }
+        lastActivityAt
         resumeUrl
+        validTo
         pillowImage {
           src
         }
--- a/app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
+++ b/app/feature/feature-home/src/main/kotlin/com/hedvig/android/feature/home/home/data/GetHomeDataUseCase.kt
@@ -156,18 +156,22 @@ internal class GetHomeDataUseCaseImpl(
             val otherCrossSellsData = crossSellsData.otherCrossSells.map {
               it.toCrossSell()
             }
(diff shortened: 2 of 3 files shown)


panasetskaya:
I'm pretty sure `lastActivityAt` from storefront is actually whenever it was created btw (and ordered with most recent first: `ORDER BY s.createdAt DESC`), and storefront filters out `AND s.validTo > CURRENT_TIMESTAMP`, so should be only the valid ones. 
so maybe for both points we can rely on BE here.

StylianosGakis:
Closing this in favour of relying on the backend, per @panasetskaya's point above: storefront already filters `AND s.validTo > CURRENT_TIMESTAMP` and orders `ORDER BY s.createdAt DESC`, so both the client side filter and the sort here were redundant. That also answers the assumption I flagged in the description, and it went the other way from my guess.

Dropping this also drops `validTo` and `lastActivityAt` from the Home query, which is a small payload win.

#3095 now sits directly on `develop` with just the dismissal work.

PR #3104, opened 2026-08-27
fix(tracking): restore Datadog screen-level view names after Nav3 migration

StylianosGakis:
Restores Datadog RUM screen-level view names, which the Nav2 → Nav3 migration silently removed, and fixes a Firebase screen-name collision found on the way.

## Why

`129d7c34c7` ("refactor(nav): migrate feature graphs and app shell to Nav3") deleted `NavigationViewTrackingEffect` from `HedvigAppState`. That effect hooks `NavController`, so it is Nav2-only and could not be carried over, and nothing replaced it. Nothing in `:datadog-android` changed, which is why this was invisible from the Datadog side: the config file genuinely is untouched, and `ActivityViewTrackingStrategy` has been there since 2023.

With only the Activity layer left, and the app having one Activity, **every RUM view from 14.3.6 onward reported as `com.hedvig.android.app.MainActivity`.**

Evidence, distinct `@view.name` values per app version over 30 days:

| Version | Views | Distinct view names |
|---|---|---|
| 14.4.5 | 22,305 | **4** |
| 14.4.1 | 42,469 | **4** |
| 14.3.6 | 577 | **3** |
| 14.3.2 | 1,564 | 64 |
| 14.2.4 | 7,508 | 79 |
| 14.0.10 | 3,432 | 67 |

That broke **all 18** custom `android.*` RUM metrics, which every one of them filter on `@view.name` or `@view.url`, plus the three Android SLOs built on those metrics. Control: `trace.android.request.hits`, which is APM and not view-scoped, was flat over the same window (1,392,972 → 1,408,961), so usage never changed. The 13 `ios.*` metrics were unaffected.

## What changed in the app

**Commit 1** wires `Navigation3TrackingEffect` (from `dd-sdk-android-compose` 3.12.1, already pinned and already on `:app`'s classpath) off `Backstack.entries`.

It sits behind a small private wrapper for a load-bearing reason: `Navigation3TrackingEffect` is `@NonRestartableComposable`, so calling it directly from `HedvigApp` would attribute the back-stack snapshot read to `HedvigApp`'s restart scope and recompose the whole app shell on every navigation.

View names become the nav key canonical class names, with no `/{arg}` placeholder suffix, because Nav3 derives them from the class rather than a route pattern:

```
- com.hedvig.feature.claim.chat.ClaimOutcomeNewClaimDestination/{outcome}
+ com.hedvig.feature.claim.chat.navigation.ClaimOutcomeNewClaimKey
```

**Commit 2** fixes a pre-existing Firebase defect this surfaced. The screen name was `simpleName.removeSuffix("Key")`, which is not unique across features. Four pairs collide today and were being reported as one screen each:

| Key | Modules |
|---|---|
| `SubmitSuccessKey` | feature-choose-tier, feature-addon-purchase |
| `SubmitFailureKey` | feature-choose-tier, feature-addon-purchase |
| `FirstVetKey` | feature-home, feature-help-center |
| `ForeverKey` | feature-forever, feature-payments |

Names are now feature-qualified (`ChangeTier.SubmitSuccess`), the qualifier is dropped where it would only repeat the key name, and `screenClass` carries the fully qualified class name. `ScreenNameTest` scans the classpath and fails the build if any two nav keys ever resolve to the same name.

⚠️ This changes existing Firebase screen names, so historical Firebase data splits at this release boundary.

## Datadog side, already done outside this PR

The 10 repairable metric filters were rewritten to accept **both** the old route-pattern names and the new key names, so each series stays continuous across the release. Every rewrite was verified to match the identical event count before and after, so none of them moved a current number.

All consumers reference the metrics by name, so the dashboard, the claims monitor and all three Android SLOs needed no edits of their own.

## Follow-up, tracked in this PR

`docs/plans/2026-08-27-datadog-android-metric-recovery.md` records what still needs doing once this ships, because nothing else would remind anyone:

- **Remove the legacy `OR` branches** from 8 metrics once the pre-14.3.6 install base is gone. Includes the trigger condition, the verification queries, and the exact target filter per metric.
- Two traps written down: the terminate-insurance metrics need no cleanup, and chat/login must **not** be simplified to a package wildcard, since that pulls in `Inbox` and the other login screens (measured: +824 on 8,089 for chat) and inflates both the metric and its SLO denominator.
- 7 metrics held for deletion, with the reason they are held.
- The missing claim-failure signal, and the measured +1.4% scope change to `android.claimflow.network.count`.

## Verification

- `:app:compileDebugKotlin` and `:app:ktlintCheck` pass
- `ScreenNameTest` passes, 4 tests, including the exhaustive classpath collision scan
- Datadog metric filters re-read back from the API after writing, and a whole-org diff confirmed 31 metrics before and after with exactly the 10 intended changes and all 13 iOS metrics byte-identical

Worth a reviewer's eye on the restart-scope reasoning for the wrapper composable, and on the feature-qualified screen-name scheme, since that one changes Firebase event names.

## a11y
- [x] if these are UI changes - check for their accessibility

No UI changes. Both commits touch analytics side effects only, with no change to rendered output.


panasetskaya (approved):
🔥

PR #3108, opened 2026-08-28
Cover KMP source sets with a ktlint namespace-import rule

StylianosGakis:
Stacked on #3110. Review that one first.

### Why a second mechanism

The `NamespaceImport` Android Lint check from #3107 cannot reach KMP modules, so the 19 design system violations it just cleaned up are ones it could never have caught.

I tried to fix that properly before writing this:

| Attempt | Result |
|---|---|
| Apply standalone `com.android.lint` alongside the KMP plugin | Works, catches `commonMain`. But every `generateAndroidMainLintModel` then depends on **itself**, and the build dies with a circular dependency as soon as two such modules depend on each other |
| `checkDependencies = false` | Same cycle |
| `android.experimental.lint.analysisPerComponent=false` | Same cycle |
| `com.android.internal.lint` | Runnable lint, but JVM components only. A full run with a planted violation still in `commonMain` came back green |

The root cause is upstream. AGP's `com.android.kotlin.multiplatform.library` contributes a `lint {}` DSL but registers no task that runs it, and it sits outside `com.android.base`, so plugin logic keying off the Android plugins does nothing there. [Google issue 246751841](https://issuetracker.google.com/issues/246751841) has tracked this since 2022, AGP release notes through 9.3.2 never mention KMP lint, and the [KMP plugin docs](https://developer.android.com/kotlin/multiplatform/plugin) don't mention lint in either the supported or the unsupported section. The consistent community answer is to use ktlint or detekt for common code.

### What this does

kotlinter already runs ktlint over every source set, so `hedvig-ktlint` implements the same rule as a `RuleSetProviderV3` custom ruleset against ktlint 1.8.0, the version kotlinter 5.6.0 bundles. It is wired into every module through kotlinter's `ktlint` dependency configuration.

### The tradeoff

ktlint has no type resolution, so unlike the Lint check it cannot ask whether an owner is a class or a package. `CAPITALIZED_PACKAGES` carries that cost: Kotlin/Native interop packages are named after the framework they bind, so `platform.Foundation.systemLocale` is shaped exactly like a member import and has to be excluded by prefix.

Both checks are kept deliberately. On Android modules Lint remains the better signal, since it resolves types and reports inline in the IDE as you type. A violation there is reported twice, which only happens when someone is already wrong.

### Verification

- Fires on KMP `commonMain`: a planted import in `Tooltip.kt` reports `[hedvig:namespace-import]`.
- **Zero false positives.** A full `lintKotlin` sweep across all 1,374 files and every source set returns no `namespace-import` hits.
- Edge cases pass: `platform.Foundation.systemLocale` in `core-locale`'s `nativeMain` and every `Duration.Companion.seconds` import are clean.

### Unrelated pre-existing failure worth knowing about

`:feature-insurances` and `:feature-terminate-insurance` currently fail `lintKotlin` on `standard:indent`, `standard:wrapping` and similar. I confirmed with a control run, changes stashed, that these are pre-existing and unrelated to this PR (38 identical errors). It does mean `ktlintCheck` is red on develop today.

### DIFF
--- a/build-logic/convention/src/main/kotlin/HedvigGradlePlugin.kt
+++ b/build-logic/convention/src/main/kotlin/HedvigGradlePlugin.kt
@@ -38,6 +38,14 @@ private fun Project.configureKtlint(libs: LibrariesForLibs) {
     reporters = arrayOf(ReporterType.checkstyle.name)
   }
 
+  // Our own rules run on every source set, which is what gives KMP modules the coverage that
+  // Android Lint cannot reach.
+  if (name != "hedvig-ktlint") {
+    dependencies {
+      add("ktlint", project(":hedvig-ktlint"))
+    }
+  }
+
   tasks.withType<org.jmailen.gradle.kotlinter.tasks.LintTask>().configureEach {
     exclude { it.file.path.contains("generated/") }
     reports.set(
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -27,6 +27,8 @@ kmpNativeCoroutines = "1.0.5"
 kotlin = "2.4.10"
 kotlinpoet = "2.3.0"
 kotlinter = "5.6.0"
+# Must match the ktlint that kotlinter bundles, so the custom ruleset links against the same API
+ktlintCore = "1.8.0"
 ksp = "2.3.10"
 ktor = "3.5.1"
 license = "0.9.9"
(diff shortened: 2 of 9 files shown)


StylianosGakis:
Superseded, no content lost. See the reasoning in the comment on #3110.

The ktlint rule from this PR now leads the stack in #3122, since it is the mechanism with full coverage. The lint.xml change that was sitting in here unmentioned is extracted into #3121, together with a convention plugin fix so a new module no longer fails `lint` by existing.

- #3121 (base `develop`) → #3122 (base #3121) → #3123 (base #3122)

All three are drafts, rebased on current develop, and green.

PR #3110, opened 2026-08-28
Import the type, never the namespace

StylianosGakis:
Adds an import rule, enforces it, and fixes every existing violation.

### Why

PR #3100 was one screen refactor in one file, +250/-139. Most of that was not the refactor: 29 gratuitous new imports and ~60 rewritten call sites turning `Res.string.FOO` into `string.FOO`, `Clock.System.now()` into `System.now()`, and `HomeEvent.RefreshData` into `RefreshData`. The real change was buried under the churn, and the shortened forms read worse than what they replaced. `string.FOO` names nothing, and `System.now()` reads as `java.lang.System`.

### The rule

**Import the type, never the namespace.** An import may shorten a qualified reference only when the short name still says what it is to someone reading that line cold.

Types pass, and this stays the house style (~1,200 such imports already exist):

```kotlin
import ...HomeUiState.Success                 // `is Success ->` reads fine
import kotlin.time.Duration.Companion.seconds // enables the `5.seconds` idiom
```

Members reached through a receiver that carries the meaning do not:

```kotlin
import hedvig.resources.Res.string   // ❌ use Res.string.FOO
import kotlin.time.Clock.System      // ❌ use Clock.System.now()
```

The rule also bans import-only changes to lines you are not otherwise editing.

### Enforcement

A `NamespaceImport` check in `hedvig-lint`, alongside the existing `ComposeM2Api`. It **resolves** the owner rather than matching path text, which is what keeps `platform.Foundation.systemLocale` from being flagged, since `Foundation` is a capitalized Kotlin/Native interop package and not a class. Deny and allow lists are configurable through the shared `lint.xml`.

`hedvig-lint` had no test source set, so this adds one plus the `lint-tests` dependency. Seven cases cover both directions of the rule.

### The fixes

All 24 violations, across 11 files. Two of them were latent bugs rather than style:

- `TopAppBarLayoutForActions` has no `windowInsets` parameter and was silently reading the import. Only the compiler caught it, after the import was removed.
- `TopAppBar.kt`'s import was otherwise entirely unused, the same class of miss as an unused `kotlin.time.Clock` import elsewhere. ktlint's `no-unused-imports` catches neither.

For the 19 design system `Defaults` members, the owning object lives in the same file, so they are now qualified as `TooltipDefaults.defaultStyle`, matching what `TopAppBar.kt` already did. References from inside an owning object's own body are untouched, since they resolve without an import.

### Verification

- 7/7 detector tests pass.
- `:design-system-hedvig:compileKotlinMetadata` and release compiles of the three affected feature modules succeed.
- Positive control: re-adding `import kotlin.time.Clock.System` to `HomeDestination.kt` fails `:feature-home:lint` with the expected message. Reverted, lint clean.
- Repo-wide rescan: zero violations remain.

### Known gap

The Lint check does not run on KMP modules, so 19 of the 24 fixes are in a module it cannot police. That is handled in the stacked PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -480,6 +480,47 @@ Configuration in `.editorconfig`:
 - **Entry functions:** `{feature}Entries`
 - **Use cases:** `{Action}{Domain}UseCase` (e.g., `GetHomeDataUseCase`)
 
+### Imports
+
+**Import the type, never the namespace.** An import may shorten a qualified reference only when the
+short name still says what it is to someone reading that line cold, without scrolling to the import
+list. Sealed subclasses, enum entries and other types pass that test. A member reached through a
+receiver that carries the meaning does not.
+
+Always allowed (this is the house style, ~1200 such imports exist):
+
+```kotlin
+import com.hedvig.android.feature.home.home.ui.HomeUiState.Success   // `is Success ->` reads fine
+import com.hedvig.android.design.system.hedvig.TooltipDefaults.BeakDirection.TopEnd
+import kotlin.time.Duration.Companion.seconds                        // enables the `5.seconds` idiom
+```
+
+Never allowed, because the receiver is the meaning:
+
+```kotlin
+import hedvig.resources.Res.string      // ❌ `stringResource(string.FOO)`  → use `Res.string.FOO`
+import hedvig.resources.Res.drawable    // ❌ `painterResource(drawable.x)` → use `Res.drawable.x`
+import kotlin.time.Clock.System         // ❌ `System.now()`               → use `Clock.System.now()`
+import ...hedvig.TooltipDefaults.defaultStyle  // ❌ `defaultStyle` alone names nothing
+```
+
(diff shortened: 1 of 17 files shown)


StylianosGakis:
Superseded by a properly ordered stack, no content lost.

The problem with this pair was that this PR introduced an Android Lint check that cannot police 19 of the 24 fixes it made, and #3108 existed to plug that hole. The ktlint rule is the mechanism with full coverage (Android, JVM and KMP), so it now leads and the Lint check is an optional layer on top.

- #3121 Fix two lint configuration traps (base `develop`). The lint.xml and baseline changes that were buried in #3108, extracted, plus a convention plugin fix so adding any new module no longer fails `lint`.
- #3122 Import the type, never the namespace (base #3121). Replaces this PR: the rule, all fixes, and ktlint enforcement across every source set.
- #3123 Add a type-resolving NamespaceImport lint check (base #3122). The Lint check from this PR, rescoped as precision plus inline IDE feedback.

All three are drafts, rebased on current develop, and green.

PR #3138, opened 2026-09-07
Lift the liquid glass buttons off the ButtonSize axis

StylianosGakis:
Extract the glass button outside of the entire rest of the button system because it does not respect the multiple-size scenario anyway and we had to do if-checks for the glass scenario to adjust the size of the button due to this fact. We had code like
```
@Composable
private fun ButtonSize.sizeIn(style: Style): Size = when {
  style.glassMaterial != null && this == ButtonSize.Large -> Size.Glass
  else -> size
}
```
which felt wrong so I figured might as well take that special case out since we only have one size in the app for now. If we get more such buttons we can consider bringing it back into the same size system

🤖 AI description:

📄 [**The button seam**](https://claude.ai/code/artifact/0b474493-2b02-468d-a680-64fe4ec2490e) — visual walkthrough of the final API, the size scale, and the Figma mapping.

`sizeIn(style)` let a `ButtonStyle` reach across and change the metrics a `ButtonSize` asked for, gated on `this == ButtonSize.Large`. So `RoundedPrimary` at `Medium` silently rendered with standard Medium padding: no error, no warning. Nothing hit it today (all three call sites are `Large`), but `design-showcase` was rendering the bogus combinations.

- `RoundedPrimary`/`RoundedLiquidGlass` leave `ButtonStyle` for `LiquidGlassButtonStyle { Tinted, Regular }`.
- New `HedvigLiquidGlassButton`, which takes no `ButtonSize`.
- Both funnel into a private `ButtonImpl` receiving `Style` and `Size` already resolved, so the axes can't cross. `sizeIn()` is deleted.
- Showcase gets a dedicated Glass row instead of a cell per size.

**The misuse is now a compile error**, not a silent fallback. Three probes against the new API:

```
Argument type mismatch: actual type is 'ButtonDefaults.LiquidGlassButtonStyle',
  but 'ButtonDefaults.ButtonStyle' was expected.
No parameter with name 'buttonSize' found.
Unresolved reference 'RoundedPrimary'.
```

The two enums are disjoint, `HedvigLiquidGlassButton` has no size parameter, and `Size` is a private sealed interface so `Size.LiquidGlass` isn't nameable outside `Button.kt`. (`Modifier.liquidGlass` stays public for `ToolbarIcons`, so a glass material can still be hand-applied to any button — but that's explicit, and nothing gets silently substituted.)

**Naming**, following @panasetskaya's review comment. She was right that `Primary` isn't glass: it resolves to the same nine color tokens as `ButtonStyle.Primary` at every state. But the Figma file settles what these actually are. [Make a claim](https://www.figma.com/design/SogcacjzOxkCC46XcZP8lQ/App-P2-2026?node-id=429-8162) and [Help & support](https://www.figma.com/design/SogcacjzOxkCC46XcZP8lQ/App-P2-2026?node-id=429-8163) are two instances of one component with a boolean `Tinted` variant, and **both** BG children carry the same drop shadow, `0px 8px 40px rgba(0,0,0,0.12)`, which is exactly our `glassDropShadow`. So they are a pair, design calls the pair liquid glass, and the code now says the same.

Metrics check out against Figma: fill `#121212`, 20dp horizontal padding, 48dp height (12 + 24 + 12), full corner radius.

Verified on a Pixel 6 Pro: home action row renders all three buttons correctly, and the showcase shows the six remaining styles per size plus the Glass row.

### DIFF
--- a/app/design-system/design-system-hedvig/src/commonMain/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
+++ b/app/design-system/design-system-hedvig/src/commonMain/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
@@ -90,31 +90,71 @@ fun HedvigButton(
     onClickLabel = onClickLabel,
     isLoading = isLoading,
   ) {
-    val buttonColors = buttonStyle.style.buttonColors
-    val loadingTransition = updateTransition(isLoading, label = "loading transition")
-    loadingTransition.AnimatedContent(
-      transitionSpec = {
-        fadeIn(tween(durationMillis = 220, delayMillis = 90)) togetherWith fadeOut(tween(90))
-      },
-      contentAlignment = Alignment.Center,
-    ) { loading ->
-      if (loading) {
-        LayoutWithoutPlacement(
-          sizeAdjustingContent = { HedvigText(text = text, modifier = Modifier.withoutPlacement()) },
-        ) {
-          val desc = stringResource(Res.string.TALKBACK_LOADING_STATE_BUTTON)
-          ThreeDotsLoading(
-            stableColor = buttonColors.activeLoadingIndicatorColor,
-            temporaryColor = buttonColors.inactiveLoadingIndicatorColor,
-            modifier = Modifier.wrapContentSize(Alignment.Center)
-              .semantics {
-                contentDescription = desc
-              },
-          )
-        }
-      } else {
-        HedvigText(text = text, textAlign = TextAlign.Center)
+    ButtonLabel(text = text, isLoading = isLoading, buttonColors = buttonStyle.style.buttonColors)
+  }
+}
+
+/**
+ * A pill of the iOS liquid glass material, lifted off the content behind it by a drop shadow.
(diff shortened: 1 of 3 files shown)


app/design-system/design-system-hedvig/src/commonMain/kotlin/com/hedvig/android/design/system/hedvig/Button.kt
```diff
@@ -310,12 +376,19 @@ object ButtonDefaults {
...
 
+  /**
+   * The glass styles, which [HedvigGlassButton] renders. They are deliberately not part of
+   * [ButtonStyle]: glass is drawn at a single size, so it does not cross the [ButtonSize] axis and
+   * there is no such thing as a small glass button.
+   */
+  enum class GlassButtonStyle {
     /** A filled pill with the sheen and drop shadow of the iOS glass material. */
-    RoundedPrimary,
+    Primary,
```
panasetskaya: the only comment now it that this button now that we have all buttons rounded is just a Primary. It's not glass in itself.

panasetskaya (approved):

PR #3148, opened 2026-09-11
Record post-release monitor work, the action-based precedent, and the auth SLOs that already exist

StylianosGakis:
🤖 AI description:

Doc only. Records the post-release follow-up, the action-based precedent on iOS, and the auth SLOs that already exist. Every claim in it was checked against live Datadog on 2026-09-11, and several earlier ones did not survive that check, so the diff is part new material and part correction.

**After the release: take the SLO window back to 7 days.** `error_budget(...).over("30d")` is a trailing window, so once the budget is burned the monitor stays red until the events age out, up to a month. Seven days recovers four times faster. The arithmetic is recorded with it: at ~703 login attempts a week, 7d at 99% is about 7 failures of budget against 30 at 30d, so if that turns out to be normal variance the lever is the target, not the window. The figure is flagged as a floor, since login-view impressions are themselves still recovering from the Nav3 breakage and the same query read half as much a few days earlier.

**Accepted, not fixed: this SLO cannot see an unreachable auth service.** A request that never reaches the server produces a RUM error and no resource, so it lands in neither side of the ratio and an auth outage reads 100% or no-data with nothing firing. The original plan was to flip `notify_no_data` on monitor `93408872`. That does not work: the monitor is `type: "slo alert"`, and the flag is not honoured there. Org-wide, all 32 SLO alert monitors have it false and the only 2 setting it true are ordinary metric monitors. Covering it properly needs a second absence monitor, which is deliberately not being built, because the gap is as old as the SLO and a genuinely unreachable auth service is already caught server-side on APM traces. What the doc records instead is the reading rule: a green `Auth: login (Android)` means nothing without a non-zero denominator.

**The iOS action-based precedent, with its counter-example.** The doc used to say no iOS metric broke because they are action-based. That is not the reason. The Nav3 migration was Android-only, and only 3 of 13 `ios.*` metrics are action-based. Of those three only two are genuinely rename-proof, because they take their names from a Swift enum the app owns. The third, `ios.claims.end.count`, filters on a GraphQL `__typename` owned by the backend schema, so it dies silently on a schema rename. The proposed `CLAIM_SUBMISSION_FAILED` action is therefore catching up to a working pattern, with one recorded example of how to get it wrong. `ios.login.network.error` filtering on Lokalise-managed translated strings is recorded as the worse coupling it is.

**What already measures auth.** Two server-side SLOs cover auth availability on APM traces with the full population and no client sampling. `Auth: login (iOS)` also exists, contrary to an earlier claim in this doc that it did not, and it reads 99.95% OK while dividing an error-type metric by a resource-type metric, the same mismatch that made the Android SLO report 534% of its budget. It looks healthy for the same reason the Android one did right up until it did not. Also notes `Purchase: completed signs` (`completed / started`, 70% target) as existing precedent for a funnel SLO on explicit events.

Other corrections: the install-base trigger for the `OR`-branch cleanup is split into 30d/7d/1d, because the 12% figure the doc leaned on is a 30-day average and the 7-day number is 1.1%, so the trigger is much closer than it read; and the burn-rate note now points at the 9 burn-rate monitors the org already runs rather than claiming the thresholds have to be sized from scratch.

### DIFF
--- a/docs/plans/2026-08-27-datadog-android-metric-recovery.md
+++ b/docs/plans/2026-08-27-datadog-android-metric-recovery.md
@@ -345,11 +345,22 @@ Android cannot fix. Worth knowing before anyone treats it as the primary defence
 rebuilds backend auth availability inside RUM.
 
 `Auth: login (iOS)` does exist (`9028984bed7351dc92b7e58e5c7db82f`, 7d at 99%, monitor `93102231`)
-and reads 99.95%, state OK. That is not reassurance. It divides `ios.login.network.error`
-(`event_type: error`) by `ios.login.network.count` (`event_type: resource`), the same numerator and
-denominator mismatch that made the Android SLO report 534% of its budget. It looks healthy for the
-same reason the Android one looked healthy right up until it did not. Android got paged first
-because its view names broke, not because iOS is better instrumented.
+and reads 99.95%, state OK. **That is not reassurance.** Checked 2026-09-11, it is
+
+```
+(sum:ios.login.network.count - sum:ios.login.network.error) / sum:ios.login.network.count
+```
+
+which is the Android defect exactly. `ios.login.network.count` is `event_type: resource`,
+`ios.login.network.error` is `event_type: error`, so the numerator subtracts one kind of event from
+a count of a different kind and can go negative. That is what made the Android SLO report 534% of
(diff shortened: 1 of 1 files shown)


StylianosGakis:
Superseded by #3159, which is the same commits on a branch whose name matches what the change actually concluded. Closing this one; no review had happened here yet.

PR #3150, opened 2026-09-11
Add a scripted demo-mode claim flow

StylianosGakis:
Scripted path for Demo mode, to ease development but also show the flow to demo mode users

🤖 AI description:

Lets the claim chat run without a backend, so the P2-2026 design work can exercise states that are awkward to get the real backend to produce on demand: a burst of task descriptions, a contract picker with and without a prefill, and the audio recording step.

Eight of the use cases the flow runs through were concrete `internal` classes with no interface, so interfaces are extracted here first. Each then gets a `Demo` implementation and a `Switching` binding, following the `DemoSwitcher` pattern already used by home, insurances and payments. The `Switching` class carries the only `@ContributesBinding` for its type, which is why the existing `@ContributesBinding` on `SkipStepUseCaseImpl` and `RegretStepUseCaseImpl` moves onto the switchers rather than staying put.

Two things worth a reviewer's attention:

This ships. These KMP modules use AGP's `androidLibrary {}` target, which has no build types, so there is no debug source set to hide a fake in, and the use cases are `internal` so it cannot live in `:app`. Demo mode is the sanctioned alternative, which means real demo-mode users now get the scripted claim flow. That is a product change, not just test scaffolding.

The switchers sit in `androidMain` rather than `commonMain` because `core-demo-mode` is a JVM library and `commonMain` has to stay platform-agnostic. `nativeMain` builds no graph, so nothing breaks there.

`DemoClaimIntentScript`'s task schedule is deliberately uneven and delivers two descriptions in a single emission. That is the case where rendering only the most recent description silently drops messages, and a test asserts the burst stays in the schedule so it cannot be flattened away by accident.

Verified: `:app:assembleDebug` succeeds, which is what proves the graph still resolves; `:feature-claim-chat:jvmTest` passes 6 tests; `ktlintCheck` is clean. Note `allTests` fails locally on `linkDebugTestIosSimulatorArm64` with a missing `DatadogCore` framework, unrelated to this change, and CI runs `test jvmTest` rather than `allTests`.

### DIFF
--- a/app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/data/SkipStepUseCase.kt
+++ b/app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/data/SkipStepUseCase.kt
@@ -9,7 +9,6 @@ import com.hedvig.android.core.common.ErrorMessage
 import com.hedvig.android.core.common.di.AppScope
 import com.hedvig.android.language.LanguageService
 import com.hedvig.android.logger.logcat
-import dev.zacsweers.metro.ContributesBinding
 import dev.zacsweers.metro.Inject
 import dev.zacsweers.metro.SingleIn
 import octopus.ClaimIntentSkipStepMutation
@@ -18,7 +17,6 @@ internal interface SkipStepUseCase {
   suspend fun invoke(id: StepId): Either<ClaimChatErrorMessage, ClaimIntent>
 }
 
-@ContributesBinding(AppScope::class)
 @SingleIn(AppScope::class)
 @Inject
 internal class SkipStepUseCaseImpl(
--- a/app/feature/feature-claim-chat/build.gradle.kts
+++ b/app/feature/feature-claim-chat/build.gradle.kts
@@ -55,12 +55,19 @@ kotlin {
       implementation(projects.uiForceUpgrade)
       implementation(projects.partnersDeflect)
     }
+    commonTest.dependencies {
+      implementation(libs.assertK)
+      implementation(libs.coroutines.test)
+      implementation(libs.kotlin.test)
+      implementation(libs.turbine)
+    }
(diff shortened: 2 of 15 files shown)


app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/data/SkipStepUseCase.kt
```diff
@@ -18,7 +17,6 @@ internal interface SkipStepUseCase {
   suspend fun invoke(id: StepId): Either<ClaimChatErrorMessage, ClaimIntent>
 }
 
-@ContributesBinding(AppScope::class)
```
panasetskaya, line 21: could you explain here pls? :)

panasetskaya (commented):
This is only for the Demo purpose, right?

app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/data/SkipStepUseCase.kt
StylianosGakis: Since we now have this
```
@ContributesBinding(AppScope::class, binding = binding<SkipStepUseCase>())
internal class SwitchingSkipStepUseCase(
  override val demoManager: DemoManager,
  override val prodImpl: SkipStepUseCaseImpl,
  override val demoImpl: SkipStepUseCaseDemo,
) : SkipStepUseCase, DemoSwitcher<SkipStepUseCase>()
```

Now that one is the one which is automatically contributing an implementation of `SkipStepUseCase`. So whenever someone asks for `SkipStepUseCase` it gets the demo-switched one instead.

StylianosGakis:
Yes demo, but I used it extensively during dev too, so it doubles down as useful for dev work but also if someone stumbles upon the demo mode

PR #3151, opened 2026-09-11
Spec and plan for the claim chat P2-2026 design changes

StylianosGakis:
🤖 AI description:

Docs only. Scopes the claim submission changes from the P2-2026 Figma review into a spec and a task-by-task implementation plan.

In scope: animated and paced task descriptions, the summary answer hierarchy, the insurance picker as a bottom sheet, the input mode row with glass overlays, a genuinely sticky input, and ghost skip buttons. Items 3 and 4 were dropped in the meeting.

Three findings changed the shape of the work compared to what the meeting assumed:

Draft preservation across input mode switches is new work, not something to preserve. `SwitchToFreeText` explicitly nulls the text and `SwitchToAudioRecording` resets the recording, and the state models the two modes as alternatives so it cannot hold both. It is deferred, with the reasoning kept for whoever picks it up.

The input is not currently bottom-anchored at all. It is the last list item stretched to viewport height, so making it sticky is a structural change to the screen's scroll model rather than a styling fix. That item is deliberately isolated on its own branch so it can be reverted alone.

Item 5 is smaller than it looked. `SINGLE_SELECT` already renders through a picker, `RadioGroup` already shows option subtitles, `HedvigBigCard` already has a content-slot overload, and the sheet title is the backend's own field title, so no new Lokalise key is needed.

The plan sequences the work in waves because separate branches remove git conflicts but not file conflicts: three items are genuinely independent, the two input items share `ClaimChatDestination.kt`, and the skip-button change touches five files the others are rewriting.


StylianosGakis:
Spec complete, deleting

PR #3153, opened 2026-09-11
Animate and pace claim chat task descriptions

StylianosGakis:
Improve transition etc when we get new task steps back to back

🤖 AI description:

Stacked on #3150, so the diff here is only the two commits on top. Rebase onto develop once that merges.

Two changes to the task step's loading text.

The first is a real behaviour fix rather than styling. `TaskStepTopContent` rendered `descriptions.lastOrNull()`, and the presenter accumulates descriptions across polls, so whenever one poll carried more than one new description the intermediate ones were never shown. The design treats these as a sequence the user reads. `pacedDescriptions` now walks the backlog in order, holding each for a minimum dwell.

The second is the transition. The `AnimatedContent` had no `transitionSpec`, so it inherited the default fade plus `scaleIn` plus an animated `SizeTransform`. Since the descriptions differ a lot in length ("Analyzing..." against "Going through the details..."), animating the bounds shifted the neighbouring Rive animation on every swap. It now slides and fades vertically with a snapped size transform, so the indicator holds still.

Verified: 4 new tests pass, the 6 existing demo-flow tests still pass, `:feature-claim-chat:assemble` and `ktlintCheck` are clean.

Not yet verified by hand in demo mode. A device is connected but reaching demo mode goes through the login screen, which would log out the session on it. Worth a look in the running app before this leaves draft, particularly that the indicator really does stay put.

### DIFF
--- a/app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/ui/step/TaskDescriptionQueue.kt
+++ b/app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/ui/step/TaskDescriptionQueue.kt
@@ -1,0 +1,50 @@
+package com.hedvig.android.feature.claim.chat.ui.step
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.produceState
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.snapshotFlow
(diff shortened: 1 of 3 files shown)


panasetskaya (approved):
do we have a real case to test on in staging? wanna see how it looks like

StylianosGakis:
> do we have a real case to test on in staging? wanna see how it looks like

I believe the demo mode flow should be a good way to test this, try the integration/claim-chat-p2-2026 branch perhaps?

PR #3156, opened 2026-09-11
Present both claim chat input modes over the conversation

StylianosGakis:
Present both claim chat input modes over the conversation

🤖 AI description:

Stacked on #3150. Independent of #3153, #3154 and #3155, which touch different files.

Item 6a. The audio step now offers the two input modes as one row of equal-width secondary buttons (Write / Record) with a ghost skip below, so neither mode reads as the preferred route. Picking one swaps that row for an inline card in the same slot: the voice recorder, or a text field that raises the keyboard and rides above it. No sheet and no scrim, so the question stays readable while answering, which is what the Figma frames show.

The voice card is also resized. Its waveform never draws taller than `WAVE_MAX_HEIGHT` (30dp), so the 158dp band it lived in was mostly empty in every state, and in the idle state it was a row of dots in 158dp of nothing. The band is now sized to the waves with symmetric insets, which takes the card from ~326dp to ~195dp. That is what makes it fit in landscape, without locking orientation anywhere.

Three leftovers from when this was a bottom sheet are gone: a `safeDrawing` bottom inset that added the nav bar height inside the card, a doubled horizontal padding on the button row, and a `verticalScroll` that crashed on open. The card sits in the Column's unweighted slot, which Compose measures with an infinite max height, and `verticalScroll` throws on that.

Deliberately not done: draft preservation across mode switches, as the spec records, so cancelling still discards.

The two labels use `claims_write` and `claims_record`. They were hardcoded with TODOs to add them to Lokalise; Mariia pointed out in review that `claims_record` already exists, and so does `claims_write`, both carrying exactly those values and sitting in the same `claims_` family as the skip button in this row. The `strings.xml` files come along in this branch because they had drifted, and the scheduled download would have produced the same.

Verified on a Pixel in demo mode, portrait and landscape, through the full record / listen / start over / send cycle. `ktlintCheck` and `jvmTest` pass.


app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/ui/step/audiorecording/AudioRecordingStepSections.kt
```diff
@@ -221,111 +230,162 @@ internal fun AudioRecorderBubble(
...
-                      SignedAudioUrl.fromSignedAudioUrlString(recordingState.audioPath.remoteUrl),
-                    ),
-                  )
+                  buttonSize = ButtonDefaults.ButtonSize.Large,
+                  modifier = Modifier.weight(1f),
+                ) {
+                  Icon(HedvigIcons.Mic, null, Modifier.size(24.dp))
+                  Spacer(Modifier.width(8.dp))
+                  // TODO: Add "Record" / "Spela in" to Lokalise
+                  HedvigText("Record")
```
panasetskaya: a hardcode (i think claims_record can work here?)

panasetskaya (commented):
A bit of UI artifacts that are off:
active input background color here
Image

horizontal alignment here (maybe pre-existed)
Image

also I think we are not using minimal amount of symbols anymore in the text input, is it on purpose?

StylianosGakis:
Will fix the indicator artifact and the lokalization problem.

The horizontal alignment bug does seem a bit unrelated, will try to check that too.

PR #3157, opened 2026-09-11
Make the claim chat input sticky above the keyboard

StylianosGakis:
Make the claim chat input sticky above the keyboard

🤖 AI description:

Stacked on #3156, which is stacked on #3150. This is the structural item and the spec deliberately keeps it on its own branch so it can be reverted without taking the other work with it. Nothing from items 1, 2, 5 or 6a is folded in here, and that property is worth preserving in review.

The input was never actually bottom-anchored. It was the last item of the transcript list, stretched to at least viewport height, which only looked pinned while the list sat at its end. Scrolling up to reread earlier answers carried it away.

The screen is now a column: the transcript scrolls in a weighted LazyColumn holding each step's question plus the answers of past steps, and the current step's input sits in a pinned container below it carrying imePadding, so it rises with the keyboard and settles back to the bottom edge on dismissal. The list takes bottom content padding equal to the measured input height so content still scrolls clear of it.

Three mechanisms existed only to sustain the old illusion and are removed with it: `LastItemHeightAdjustingState` (now unreferenced, so the file is deleted), the `requiredHeightIn` on the final item, and the fixed 3000px scroll nudge, which becomes a plain scroll to the newest step.

The split between pinned and inline uses the `isCurrentStep` flag that was already threaded through, and the pinned container appears on the same animation state the inline one used, so the intro sequence still gates it.

**The arrow condition is fixed.** It was reading `lazyListState.canScrollForward`, and the docked input it controls changes that measurement: standing the input down gives the list its height back, so reaching the end brings the input back, which makes the list scrollable again, which stands the input down. Mariia saw that loop as the arrow flickering. It also settles in a state where scrolling forward never brings the input back at all, leaving the step unanswerable until the member scrolls *up* far enough for the arrow to reappear and taps it.

The input now stands down only on a deliberate drag backwards past 24dp, and comes back only on reaching the end of the list, so nothing the input does to the list's height can flip either.

Verified on device across four window sizes, walking to the describe step and scrolling back then forward:

| Window | Height | Before | After |
| --- | --- | --- | --- |
| 1080x2400 portrait | 914dp | ok | ok |
| 2400x1600 landscape | 610dp | input never returns | returns at the end of the list |
| 2400x1080 landscape | 411dp | input never returns | returns at the end of the list |
| 1080x1080 square | 411dp | input never returns | returns at the end of the list |

It reproduced in any window short enough for the transcript to overflow, which is why portrait phone never showed it. Six samples after settling show no flicker.

One thing still wants eyes: the handoff, which is the known risk in the spec. When a step is answered its input leaves the pinned container and reappears as a transcript entry, and I cannot tell from a build whether that transition reads cleanly.

`:app:assembleDebug`, `ktlintCheck` and `jvmTest` all pass.


panasetskaya (changes requested):
this seems to break UI:

Image

Image

Image

StylianosGakis:
Is it the problem regarding the arrow + the card showing at the same time? 
I am trying to fix that now if yes, but if it's anything else too please lmk

PR #3166, opened 2026-09-16
Answer a claim chat select step by tapping a pill

StylianosGakis:
Make buttons that aren't skippable just be buttons rather than a selectable group according to latest design requirements. Context: https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1789477561460749?thread_ts=1789379513.512749&cid=C03U9C6Q7TP

🤖 AI description:

Stacked on #3150, which carries the demo script this is exercised through.

From the Slack thread: no prefilled value means no confirm button and a tap on a pill answers the step; a prefilled value keeps the button, because there is then something to accept or change. That covers Richard's ask for the initial triage step without special casing it, since nothing is ever prefilled there.

Richard's "if we have skip, we can only have skip there" applies to the no-prefill case, where a tap already answers the step. Mariia caught that reading it more broadly stranded the member: on a step that was both prefilled and skippable the pills only selected, the confirm button was gone, and skip was the only way forward. Confirm now follows the prefill alone, so such a step shows both buttons. The demo script gains a prefilled skippable step and a test pins all four combinations of the two flags.

The pills also carry a button role instead of sitting in a `selectableGroup` that announces each one as selected or not selected. That framing described a selection the member never makes once a tap submits.

`RoundCornersPill` is split in two while we are here. It used to take a nullable `onClick`, so an answer already given and a control the member taps shared a body built on the plain `Surface`, which opts out of merging its descendants and marks itself a traversal group. That is right for a bubble to be read past and wrong for something to act on, and the click was bolted on top of it. The overload with `onClick` now uses the clickable `Surface`, so the label, the click and the role sit on one node; the one without keeps the selectable text and the bubble semantics. Call sites that passed `onClick = null` just drop the argument.

The demo script gains two pill steps to exercise both halves, one with an answer already on it and one without.

Verified in demo mode: a step with no prefill shows pills and no confirm, and tapping one submits and moves on; the prefilled step still shows its confirm button with the prefilled pill highlighted. `ktlintCheck` and `jvmTest` pass.


app/feature/feature-claim-chat/src/commonMain/kotlin/com/hedvig/android/feature/claim/chat/ui/step/ContentSelectStep.kt
```diff
@@ -60,38 +61,44 @@ internal fun ContentSelectStep(
...
-                      itemId,
-                    ),
-                  )
-                }
-              },
-              isLoading = currentContinueButtonLoading,
-              enabled = !isSubmitting && selectedOptionId != null,
-              modifier = Modifier.fillMaxWidth(),
-            )
+            if (hasPrefilledAnswer && !canSkip) {
```
panasetskaya: shouldn't it just be if (hasPrefilledAnswer) here?

StylianosGakis: From the discussion https://hedviginsurance.slack.com/archives/C03U9C6Q7TP/p1789478542727219?thread_ts=1789379513.512749&cid=C03U9C6Q7TP this is what I understood, perhaps I misunderstood though

panasetskaya: but if we have prefilled answers, and the step is skippable, by this logic we will not show Confirm button, but the pills will also not work as buttons (bc there are prefilled answers). So no way to submit answers. Am I wrong?

_If we have skip, we can only have skip there i would say_
this seems about no prefilled values scenario?

panasetskaya (approved):

PR #3182, opened 2026-09-18
Eng: a11y automated checks in design system

panasetskaya:
Adds `./gradlew accessibilityChecks` (also an Android Studio run configuration), which runs Accessibility Test Framework checks over the design system on a connected device. 

It never fails the build — findings land in the HTML report instead.                                                                                       

84 cases currently, 12 failing, every one a touch target under 48dp. No contrast or labelling problems in either theme; fixes come in a separate branch.

Wdyt?


StylianosGakis (approved):
Wow yeah this is very interesting, I don't see why we shouldn't try it out

